PR #2, opened 2023-09-03
Fix LoopData not being a public type

xrtxn:
The project fails to build.


YaLTeR:
Thanks; wonder why it builds fine here

PR #8, opened 2023-09-16
Add keyboard.repeat-rate and keyboard.repeat-delay

kchibisov:
Add option to let the users configure the keyboard repeat delay and rate. Also change the default values to mimic the Xorg, which is also the values used by default by the majority of the compositors.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -3,6 +3,14 @@
 
 input {
     keyboard {
+        // Control the repeat rate of the keyboard.
+
+        // Repeat delay in milliseconds.
+        repeat-delay 660
+
+        // Repeat rate in characters per second.
+        repeat-rate 25
+
         xkb {
             // You can set rules, model, layout, variant and options.
             // For more information, see xkeyboard-config(7).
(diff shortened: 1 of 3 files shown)


resources/default-config.kdl
```diff
@@ -3,6 +3,14 @@
 
 input {
     keyboard {
+        // Control the repeat rate of the keyboard.
+
+        // Repeat delay in milliseconds.
+        repeat-delay 200
```
YaLTeR, line 9: Are you sure about this 200? [Sway](https://github.com/swaywm/sway/blob/b3519c2d2f269f34ff3712ebf1903905a56541a6/sway/input/keyboard.c#L992) and [wlroots](https://github.com/swaywm/sway/blob/b3519c2d2f269f34ff3712ebf1903905a56541a6/sway/input/keyboard.c#L992) default to 600. My GNOME install has 500 (I haven't changed it).

YaLTeR (commented):
Thanks, looks good, just a question about the defaults.

resources/default-config.kdl
kchibisov: Hm, yeah, it could be actually 600. I just remember that we picked the 200 for sctk/anvil at some point, but I don't remember why 200 exactly, maybe I've pushed all of that in the end.

It's 660 in the Xorg actually, but the 25 value is accurate.

PR #29, opened 2023-09-30
protocols: add IME protocols

### LINKED ISSUE #22: IME support
Those protocols are implemented in smithay, so it's a matter of wiring them up. Catacomb
also has them implemented.

I can work on this as well.

kchibisov:
This commit adds support for the `input_method_v2`, `text_input_v3`, and `virtual_keyboard`. The implementation follows the one in the anvil and catacomb, but those protocols are mostly enabled and forget type of things.

Fixes #22.

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -4,22 +4,26 @@ mod xdg_shell;
 
 use smithay::backend::allocator::dmabuf::Dmabuf;
 use smithay::backend::renderer::ImportDma;
+use smithay::desktop::PopupKind;
 use smithay::input::pointer::CursorImageStatus;
 use smithay::input::{Seat, SeatHandler, SeatState};
 use smithay::reexports::wayland_server::protocol::wl_data_source::WlDataSource;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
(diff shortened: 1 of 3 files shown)


kchibisov:
I haven't tested `popup` but the code is the same as in anvil, the regular ime was tested with `anthywl`.

src/handlers/mod.rs
```diff
@@ -48,6 +53,31 @@ impl SeatHandler for State {
...
+    fn new_popup(&mut self, surface: PopupSurface) {
+        if let Err(err) = self.niri.popups.track_popup(PopupKind::from(surface)) {
+            warn!("error tracking ime popup {err:?}");
+        }
+    }
+    fn parent_geometry(&self, parent: &WlSurface) -> Rectangle<i32, Logical> {
+        self.niri
+            .global_space
+            .elements()
+            .find_map(|window| {
```
YaLTeR, line 68: So, the global space doesn't actually have any windows, it's just for the outputs. It should be something like:
```
self.niri.monitor_set.find_window_and_output(parent).map(|window, _| window.geometry())
```

YaLTeR:
Thanks!

PR #31, opened 2023-10-01
config/input: add cursor section

kchibisov:
This should allow users to configure theme and size for the cursor, as well as automatically set `XCURSOR_THEME` and `XCURSOR_SIZE` env variables.


YaLTeR:
Thanks! I moved the cursor {} config section to the top level.

PR #52, opened 2023-10-17
Try to pick user login shell instead of assuming bash

kchibisov:
The trick is inspired by `gnome-session`.

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -1,4 +1,15 @@
-#!/bin/bash --login
+#!/bin/sh
+
+if [ -n "$SHELL" ] &&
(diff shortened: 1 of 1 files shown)


resources/niri-session
```diff
@@ -1,9 +1,14 @@
-#!/bin/bash --login
-
-# Make sure there's no already running session.
-if systemctl --user -q is-active niri.service; then
```
YaLTeR, line 4: Why delete this check though?

kchibisov: Good catch.

YaLTeR:
Thanks

PR #55, opened 2023-10-25
Add support for wl_compositor@v6

kchibisov:


src/niri.rs
```diff
@@ -1194,12 +1194,17 @@ impl Niri {
                     .as_ref()
                     .map(|surface| (surface, bbox_from_surface_tree(surface, surface_pos)));
 
+                // FIXME we basically need to pick the largest scale factor across the overlapping
+                // outputs, this is how it's usually done in clients as well.
+                let mut cursor_scale = 1;
+                let mut dnd_scale = 1;
```
kchibisov, line 1200: This doesn't really work, because you run this on every frame meaning that the scale is being send in a busy loop. There're two approaches I can see

1) Cache on per-surface basis the current scale factor and don't resent it.
2) Send scale factor for the pointer once it `enters` surface, given that surface
can't span multiple outputs.

Which approach should we pick?

Though, I wonder whether we should send this in reply to `surface::commit` when the surface is a cursor surface? And then we need to resend scale when output actually got changed for the cursor, but I'm not sure how to identify that with the smithay's API.

YaLTeR: I think caching the current scale factor per-surface and not re-sending is what I'd prefer here.

> Send scale factor for the pointer once it enters surface, given that surface
can't span multiple outputs.

This doesn't quite work in spirit, because the cursor (or, better example, the DnD surface) can be large and significantly overlap another output with a higher scale.

> Though, I wonder whether we should send this in reply to surface::commit when the surface is a cursor surface? And then we need to resend scale when output actually got changed for the cursor, but I'm not sure how to identify that with the smithay's API.

Yeah, I'm not sure Smithay has API for this atm. I think if we go with the caching approach it should work fine as is.

src/niri.rs
```diff
@@ -1209,12 +1214,22 @@ impl Niri {
...
                         }
                     }
                 }
+
+                if surface.version() >= 6 {
+                    surface.preferred_buffer_scale(cursor_scale);
+                }
+                if let Some((surface, _)) = dnd {
+                    if surface.version() >= 6 {
+                        surface.preferred_buffer_scale(dnd_scale);
```
YaLTeR, line 1230: There's also dnd surface handling in the branch above, also needs to be handled.

kchibisov: Is there some other `dnd` surface other the one in `dnd` struct?

YaLTeR: I mean the `CursorImageStatus::Hidden | CursorImageStatus::Named(_)` branch above, which may still have a DnD surface. It does not get the preferred buffer scale yet as far as I can see.

src/niri.rs
kchibisov: The new smithay API should at least cache events we're sending. So caching is now done globally.

kchibisov:
What should we do with subsurfaces and other surfaces not having configure events? For cursor and its related surface we're using the output refresh stuff. I'd guess that for subsurfaces we should just handle them from the `commit` hook?

YaLTeR:
> What should we do with subsurfaces and other surfaces not having configure events?

You mean like, window appears and gets its preferred scale, then it creates a subsurface which doesn't until the window moves to another output? Yeah, I'm not sure. We could send them from `Layout::refresh()` which is called in loop callback and already handles states like activated I guess.

kchibisov:
I think it should work now.

YaLTeR (approved):

src/handlers/mod.rs
```diff
@@ -71,6 +72,17 @@ delegate_text_input_manager!(State);
...
     fn new_popup(&mut self, surface: PopupSurface) {
+        if let Some((_, output)) = surface
+            .get_parent()
+            .and_then(|parent| self.niri.layout.find_window_and_output(&parent.surface))
+        {
+            let scale = output.current_scale().integer_scale();
+            let transform = output.current_transform();
+            let wl_surface = surface.wl_surface();
+            with_states(wl_surface, |data| {
+                send_surface_state(wl_surface, data, scale, transform);
```
YaLTeR, line 83: Why do IME popups get the state in new_popup, but Xdg popups get it on first commit, wouldn't it make sense to unify those? Also what happens when an IME popup changes its parent?

src/niri.rs
```diff
@@ -1168,10 +1169,15 @@ impl Niri {
...
                     if let Some(mut overlap) = geo.intersection(bbox) {
                         overlap.loc -= surface_pos;
+                        dnd_scale = dnd_scale.max(output.current_scale().integer_scale());
                         output_update(output, Some(overlap), surface);
                     } else {
                         output_update(output, None, surface);
                     }
+
+                    with_states(surface, |data| {
+                        send_surface_state(surface, data, dnd_scale, Transform::Normal);
```
YaLTeR, line 1179:
```suggestion
                        send_surface_state(surface, data, dnd_scale, output.current_transform());
```

src/niri.rs
```diff
@@ -1209,12 +1220,22 @@ impl Niri {
...
+                            dnd_scale = dnd_scale.max(output.current_scale().integer_scale());
                             output_update(output, Some(overlap), surface);
                         } else {
                             output_update(output, None, surface);
                         }
                     }
                 }
+
+                with_states(surface, |data| {
+                    send_surface_state(surface, data, cursor_scale, Transform::Normal);
```
YaLTeR, line 1232:
```suggestion
                    send_surface_state(surface, data, cursor_scale, output.current_transform());
```

src/niri.rs
```diff
@@ -1209,12 +1220,22 @@ impl Niri {
...
+                });
+                if let Some((surface, _)) = dnd {
+                    with_states(surface, |data| {
+                        send_surface_state(surface, data, dnd_scale, Transform::Normal);
```
YaLTeR, line 1236:
```suggestion
                        send_surface_state(surface, data, dnd_scale, output.current_transform());
```

YaLTeR (changes requested):

src/handlers/mod.rs
kchibisov: ime popup changes its parent during` new_popup`, that's the reason the logic is here.

kchibisov: xdg popup on the other side has a fixed parent which must be specified on `commit` and then it follows the parent window, so it gets events due to `with_surfaces` calls.

src/niri.rs
kchibisov: There's no `output` to call `current_transform`, since the surface can span multiple outputs and deciding which transform to use is not as trivial as `scale` where you can pick max.

So you need some heuristics to pick transform the way we pick scale and then apply, what you've suggested doesn't work.

src/handlers/mod.rs
YaLTeR: I see, thanks.

src/niri.rs
YaLTeR: Right, I see.

YaLTeR (approved):

YaLTeR:
Thanks.

PR #57, opened 2023-10-27
Drop `GbmDevice` in the end

kchibisov:
The `GbmDevice` must be destroyed after destroying all the objects associated with it.


YaLTeR:
Thanks

PR #58, opened 2023-10-27
Don't send key on release from niri actions

### LINKED ISSUE #28: Don't set `Key::Release` on hotkey end
Some applications may run their bindings on a `Release` of the key, like firefox.

You can test it muting itself with `Mod+m` binding.

kchibisov:
Some clients run logic on `Release`, thus don't send the key originally used for running `niri` actions.

Fixes #28.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use smithay::backend::input::{
     AbsolutePositionEvent, Axis, AxisSource, ButtonState, Device, DeviceCapability, Event,
     GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _, GestureSwipeUpdateEvent as _,
@@ -6,7 +8,7 @@ use smithay::backend::input::{
     TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState,
 };
 use smithay::backend::libinput::LibinputInputBackend;
-use smithay::input::keyboard::{keysyms, FilterResult, KeysymHandle, ModifiersState};
+use smithay::input::keyboard::{keysyms, FilterResult, Keysym, ModifiersState};
 use smithay::input::pointer::{
     AxisFrame, ButtonEvent, GestureHoldBeginEvent, GestureHoldEndEvent, GesturePinchBeginEvent,
     GesturePinchEndEvent, GesturePinchUpdateEvent, GestureSwipeBeginEvent, GestureSwipeEndEvent,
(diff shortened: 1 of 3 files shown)


src/input.rs
```diff
@@ -136,232 +127,229 @@ impl State {
...
+                // Filter actions when the session is locked or we released the previous
+                // action hotkey.
+                let action = match action {
+                    _ if event.state() == KeyState::Released => return,
+                    Some(
+                        Action::Quit
+                        | Action::ChangeVt(_)
+                        | Action::Suspend
+                        | Action::PowerOffMonitors,
+                    ) if self.niri.is_locked() => return,
```
YaLTeR, line 154: This is quite a weird-looking conditional now, also I believe the logic here got reversed: these four actions are the only ones allowed when niri is locked.

kchibisov: Hm, indeed.

YaLTeR:
What happens on a sequence like:
1. Press Mod
2. Press Shift
3. Press h
4. Release Shift
5. Release h
6. Release Mod
if Mod+h (but not Mod+Shift+h) is a compositor action? Is the app going to get its "h" release?

kchibisov:
>if Mod+r (but not Mod+Shift+r) is a compositor action? Is the app going to get its "r" release?

Not anymore.

src/input.rs
```diff
@@ -135,233 +128,257 @@ impl State {
...
+                            (_, Entry::Vacant(_)) => FilterResult::Forward,
                         }
                     },
-                );
+                ) else {
+                    return;
+                };
 
-                // Filter actions when the session is locked.
-                if self.niri.is_locked() {
```
YaLTeR, line 156: I think if you change the `let Some(Some(` above to `let Some(mut action) =` then you can leave this part of the code completely unchanged, save for the `if !pressed { return; }`.

YaLTeR (commented):
I think after merging, I will need to extract this filtering code to be testable and add tests for several of these input scenarios. This has certainly reached the complexity level where I'm no longer sure the code is correct.

src/input.rs
```diff
@@ -135,233 +128,257 @@ impl State {
...
+                            action(&config, comp_mod, keysym, *mods),
+                            this.niri.suppress_keys.entry(key_code),
+                        ) {
+                            (action, Entry::Occupied(mut entry)) => {
+                                let key = entry.get_mut();
+                                if pressed {
+                                    *key += 1;
+                                } else {
+                                    *key -= 1;
+                                }
```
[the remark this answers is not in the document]
kchibisov: >Can we legitimately go above 1 here? I'm thinking it might be better to use a simple bool instead of a counter.

yes, with keyboard having duplicated keys.

>Also, for the VT switch keys like Ctrl-Alt-F1, I'm not sure we ever see key release events. A counter will keep counting up unnecessarily, while a bool will just stay true.

I think the only solution for that is to `clear` the `HashMap`.

src/input.rs
kchibisov: Yeah, but why have the old logic with extra `None` and increasing indention for the giant `match`.

src/input.rs
```diff
@@ -899,3 +825,243 @@ impl State {
...
+                keysym: close_keysym,
+                modifiers: Modifiers::COMPOSITOR | Modifiers::CTRL,
+            },
+            actions: vec![Action::CloseWindow],
+        }]);
+
+        let comp_mod = CompositorMod::Super;
+        let mut suppressed_keys = HashSet::new();
+
+        // The key_code we pick is arbitrary, the only thin
```
YaLTeR, line 969:
```suggestion
        // The key_code we pick is arbitrary, the only thing
```

src/input.rs
```diff
@@ -899,3 +825,243 @@ impl State {
...
+            filter,
+            FilterResult::Intercept(Some(Action::CloseWindow))
+        ));
+        assert!(suppressed_keys.contains(&close_key_code));
+
+        let filter = close_key_event(&mut suppressed_keys, mods, false);
+        assert!(matches!(filter, FilterResult::Intercept(None)));
+        assert!(suppressed_keys.is_empty());
+
+        // Remove mode to make it for a binding.
```
YaLTeR, line 1017:
```suggestion
        // Remove mod to make it for a binding.
```

YaLTeR (approved):

YaLTeR:
Thanks, seems to work fine. Is this good to merge?

PR #64, opened 2023-11-01
pointer input for layer surface

cmeissl:
This PR allows to give layer surfaces pointer focus. `waybar` for example can trigger certain actions on pointer input.


YaLTeR:
Thanks! Tested with waybar, seems to work fine.

PR #65, opened 2023-11-01
Improve layout handling

kchibisov:
Review by commit and then rebase. Draft until upstream merges changes.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -16,6 +16,11 @@ input {
         // Delay is in milliseconds before the repeat starts. Rate is in characters per second.
         // repeat-delay 600
         // repeat-rate 25
+
+        // Niri can remember the keyboard layout globally (the default) or per-window.
+        // - "global" - layout change is global for all windows.
+        // - "window" - layout is tracked for each window individually.
+        // track-layout "global"
     }
 
     // Next sections include libinput settings.
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -230,6 +230,14 @@ binds {
     Mod+Shift+Minus { set-window-height "-10%"; }
     Mod+Shift+Equal { set-window-height "+10%"; }
 
+    // Actions to switch layouts.
+    // Note: if you uncomment these, make sure you do NOT have
+    // a matching layout switch hotkey configured in xkb options above.
+    // Having both at once on the same hotkey will break the switching,
+    // since it will switch twice upon pressing the hotkey (once by xkb, once by niri).
+    // Mod+Space       { switch-layout "next"; }
+    // Mod+Shift+Space { switch-layout "prev"; }
+
     Print { screenshot; }
     Ctrl+Print { screenshot-screen; }
     Alt+Print { screenshot-window; }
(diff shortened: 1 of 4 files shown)


resources/default-config.kdl
```diff
@@ -210,6 +217,10 @@ binds {
     Mod+F { maximize-column; }
     Mod+Shift+F { fullscreen-window; }
 
+    // Actions to switch layouts.
+    //Mod+Space       { switch-layout "next"; }
+    //Mod+Shift+Space { switch-layout "prev"; }
```
YaLTeR, line 222:
```suggestion
    // Keyboard layout switching.
    Mod+Space       { switch-layout "next"; }
    Mod+Shift+Space { switch-layout "prev"; }
```
This makes sense to have by default I think.

resources/default-config.kdl
```diff
@@ -16,6 +16,13 @@ input {
         // Delay is in milliseconds before the repeat starts. Rate is in characters per second.
         // repeat-delay 600
         // repeat-rate 25
+
+        // How compositor should track the layout.
+        //
+        // - "global" - layout change is global for all windows.
+        // - "window" - layout is tracked for each window individually.
+        //
+        //track-layout "global"
```
YaLTeR, line 25: I think this would make more sense as a boolean toggle:
```suggestion
        // Uncomment this line to make the layout change per-window, rather than globally.
        // per-window-layout
```

src/niri.rs
```diff
@@ -382,8 +382,43 @@ impl State {
...
         let keyboard = self.niri.seat.get_keyboard().unwrap();
-        if keyboard.current_focus() != focus {
+        let current_focus = keyboard.current_focus();
+        if current_focus != focus {
+            if self.niri.config.borrow().input.keyboard.track_layout == TrackLayout::Window {
+                let current_layout =
+                    keyboard.with_kkb_state(self, |context| context.active_layout());
+
+                let mut new_layout = current_layout;
+                // Store the currently active layout for the surface.
```
YaLTeR, line 392: I guess only toplevels ever get keyboard focus, not subsurfaces and whatnot? What about popups? Those should probably follow the parent window.

resources/default-config.kdl
kchibisov: Not really, Most folks switch in xkb options.

resources/default-config.kdl
kchibisov: I'm not sure about boolean, the enum is at lest clear what it does by default.

src/niri.rs
kchibisov: I think they grab, but not sure they receive focus? If you can test what happens it would be fine, we could resolve parents here, if it matters.

resources/default-config.kdl
YaLTeR: It shouldn't conflict with xkb switching right?

kchibisov: They don't conflict, it's just not everyone use multiple layouts, it's normally just one.

src/niri.rs
kchibisov: I tested and it works with popups just fine, the layout is changed per window, so parent inheritcs.

YaLTeR:
Did some minor changes. Should be good to merge now?

YaLTeR (approved):

PR #66, opened 2023-11-02
Check alternative cursor names when loading cursor

kchibisov:
Some themes don't have all the cursors specified as in w3c specification, thus try to check for alternative names as well.


YaLTeR:
Thanks

PR #77, opened 2023-11-04
Add Nix Flake

billksun:
This PR adds a basic Nix Flake that adds a Nix build and development environment.

The flake is based on the flake file used by System76's [`pop-os/cosmic-comp`](https://github.com/pop-os/cosmic-comp/blob/master_jammy/flake.nix) project.

Please note I am neither a Rust programmer nor someone familiar with Wayland development, so please take a look at the dependencies declared under `buildInputs` and `nativeBuildInputs` and let me know if there anything that needs to be removed or added.

If you have Nix installed and Flake enabled, you should be able to just run `nix build` to build the `niri` binary. The resulting executable will be in `./result/bin/niri`.

This flake also sets up a development environment that brings in `rust-analyzer` so you can use any LSP aware editor and get IDE features out-of-the-box.


YaLTeR:
Hey, thanks for the PR.

> Please note I am neither a Rust programmer nor someone familiar with Wayland development, so please take a look at the dependencies declared under buildInputs and nativeBuildInputs and let me know if there anything that needs to be removed or added.

Uh, well, I am not familiar with nix and how it may split dependencies into packages so I'm not the best person to assess this either. Looks reasonable at a glance.

In general, to merge this, I would like there to be a person I could ping should any questions or issues arise with the flake.

billksun:
> In general, to merge this, I would like there to be a person I could ping should any questions or issues arise with the flake.

I'd be more than happy to be the point of contact. This will motivate me to dive deeper into Nix as well.

YaLTeR:
I gave it a try in an official nixos container and then in a Fedora 39 toolbox, and hit this EGL issue in both:
```
└─ ./result/bin/niri
2023-11-07T13:35:01.793605Z DEBUG niri::config: loaded config from "/var/home/yalter/.config/niri/config.kdl"
2023-11-07T13:35:01.793723Z  INFO backend_winit: smithay::backend::winit: Initializing a winit backend
2023-11-07T13:35:01.807698Z  INFO backend_winit: smithay::backend::egl::display: Skipping EGL platform because one or more required extensions are not supported. Missing extensions: ["EGL_KHR_platform_wayland"] window=94212173716464
2023-11-07T13:35:01.807717Z  INFO backend_winit: smithay::backend::egl::display: Skipping EGL platform because one or more required extensions are not supported. Missing extensions: ["EGL_EXT_platform_wayland"] window=94212173716464
2023-11-07T13:35:01.807723Z  INFO backend_winit: smithay::backend::egl::display: Skipping EGL platform because one or more required extensions are not supported. Missing extensions: ["EGL_ANGLE_platform_angle", "EGL_ANGLE_platform_angle_vulkan", "EGL_EXT_platform_wayland"] window=94212173716464
2023-11-07T13:35:01.807731Z ERROR backend_winit: smithay::backend::egl::display: Unable to find suitable EGL platform window=94212173716464
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Egl(DisplayNotSupported)', src/backend/winit.rs:39:66
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: niri::backend::winit::Winit::new
   4: niri::niri::State::new
   5: niri::main
```
Also, apparently, nix flakes are experimental? This makes me think that perhaps a better place for this would be a wiki page on the repository? This way the flake can also be edited by others without going through the pull requests. I could then link it from the README.

billksun:
@YaLTeR, can you give me the steps/commands you used?

I'm not that familiar with accessing the GPU in containers. That said, I used the following command:
```
sudo docker run --device=/dev/dri -it nixos/nix
```
But running `niri` gave me an "Operation not permitted" error.

YaLTeR:
You can try running it in a Fedora toolbox. That one automatically forwards everything needed to run compositors (I do all of my niri development and testing in a toolbox). The commands are something along
```
toolbox create
toolbox enter
```

YaLTeR:
Hm, I am not sure actually. Maybe you can just `sudo umount /nix` from within toolbox?

YaLTeR:
Still seems to have the same issue after the last commit for me.

YaLTeR:
If it works on NixOS then perhaps we could add it as is to the wiki, then if someone else figures out non-NixOS they could edit it?

flake.nix
```diff
@@ -0,0 +1,84 @@
...
+          checks.niri = niri;
+          packages.default = niri;
+
+          devShells.default = pkgs.mkShell {
+            buildInputs = [
+              pkgs.nil
+              pkgs.cargo
+              pkgs.rust-analyzer
+            ];
+          };
```
ErrorNoInternet, line 81:
```suggestion
          devShells.default = pkgs.mkShell {
            nativeBuildInputs = with pkgs; [
              pkg-config
              clang
            ];

            buildInputs = with pkgs; [
              nil
              cargo
              rust-analyzer
              libinput
              libxkbcommon
              mesa
              pipewire
              seatd
              stdenv.cc.cc.lib
              systemd
            ];

            LIBCLANG_PATH="${pkgs.llvmPackages.libclang.lib}/lib";
          };
```

ErrorNoInternet:
Works for me on NixOS (`nix run .`)! It would definitely work on non-NixOS as well, just probably not in containers.

Only a few missing dependencies in the devShell

YaLTeR:
> It would definitely work on non-NixOS as well, just probably not in containers.

Well, it's pretty strange that it doesn't work in a toolbox. Toolbox exposes pretty much everything by now; I can run niri both as a window and on a TTY just fine from a toolbox, yet the nix build doesn't run for some reason.

YaLTeR:
I tried `nix run .` (rather than `./result/niri`) and it gives me the same error (on a toolbox).

ErrorNoInternet:
What if you try `cargo r` in `nix develop`?

YaLTeR:
It doesn't build, complains about missing libraries (libudev).

ErrorNoInternet:
Try this: https://github.com/YaLTeR/niri/pull/77#discussion_r1395500427

YaLTeR:
When I try to `cargo build` after `nix develop` now it proceeds further, but then it seems like it tries using a different compiler version for the final step somehow, breaking everything with errors like:
```
           crate `tracy_client` compiled by rustc 1.74.0 (79e9716c9 2023-11-13): /var/home/yalter/source/rs/niri/target/debug/deps/libtracy_client-16fcf4a8512604f3.rlib
   = help: please recompile that crate using this compiler (rustc 1.76.0-nightly (6b771f6b5 2023-11-15)) (consider running `cargo clean` first)
```
I tried running `cargo clean` and restarting the process several times.

YaLTeR:
Hm, I had a rustup override set on the folder and it must've been picking that up (just for the final step for some reason?) which was breaking it. I removed the override and now it built and gave me this error:
```
2023-11-17T04:42:42.873775Z  INFO backend_winit: smithay::backend::winit: Initializing a winit backend
thread 'main' panicked at src/backend/winit.rs:39:66:
called `Result::unwrap()` on an `Err` value: EventLoopCreation(Os(OsError { line: 81, file: "/var/home/yalter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.2/src/platform_impl/linux/wayland/event_loop/mod.rs", error: WaylandError(Connection(NoWaylandLib)) }))
stack backtrace:
   0: rust_begin_unwind
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/result.rs:1652:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/result.rs:1077:23
   4: niri::backend::winit::Winit::new
             at ./src/backend/winit.rs:39:32
   5: niri::niri::State::new
             at ./src/niri.rs:263:28
   6: niri::main
             at ./src/main.rs:87:21
   7: core::ops::function::FnOnce::call_once
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library/core/src/ops/function.rs:250:5
```

billksun:
I apologize for the lack of communication. I am wrapping up a conference trip, so should be able to take a look more closely soon.

I was able to reproduce the error in the official Nix container. From the series of warnings before the error, it looked like EGL is expecting some of these wayland extensions to be available but was unable to locate any?

I could not get fedora toolbox to work. It apparently had exposed the local libc from the host to `/nix` in the container, and thus prevents me from installing Nix within the container itself, since the nix installer also tries to install into `/nix`.

ErrorNoInternet:
I'm currently on a trip as well, so my responses are a bit delayed.

Seems like I'm missing a few packages in the devShell, try copying all the `nix build` buildInputs section to the devShell `buildInputs` section.

ErrorNoInternet:
I'm currently on a trip as well, so my responses are a bit delayed.

Seems like I'm missing a few packages in the devShell, try copying all the `nix build` buildInputs section to the devShell `buildInputs` section.

I'll spin up a Fedora 39 VM and see if I can debug that EGL DisplayNotSupported error later.

twoneis:
I have been working on trying to get this https://github.com/NixOS/nixpkgs/pull/264517#issue-1969621196 to work and while it is very different, I think the lack of the RUSTFLAGS like mentioned here https://github.com/NixOS/nixpkgs/pull/264517#issuecomment-1817817060 may also help in this case. (To be clear, this should fix the EGL issue first reported, not EventLoopCreation error you had later, so be sure to base it on the original request). Please let me know if this helps with the error or creates a new one. I am still testing things related to the nixpkgs request and don't have the time to test both right now, but I might give it a try after that, but I thought this may be helpful in the meantime.

I also think there are other things that could be taken from that pull request, like using rustPlatform.bindgenHook instead of setting LIBCLANG_PATH manually, but I'm by no means an expert at this and don't know for sure if that would actually be preferred.

billksun:
@twoneis, I added the RUSTFLAGS and I still get the same EGL error when starting `niri` in a container.

I found a post with similar error message, mainly some EGL extension is not supported, and the solution is adding `hardware.opengl.enable = true` to the NixOS configuration: https://discourse.nixos.org/t/wlr-errors-when-trying-to-start-sway/28698/3

If the issue is the same, I think this reinforces the idea that the container was not setup correctly for hardware opengl acceleration.

twoneis:
@billksun Yes, I didn't try the flake but only the nixpkgs version which is quite different and saw the same issue was mentioned. After trying the flake on my own system, I can confirm it works without any changes on my system. Sorry for the confusion, I should have done that simple test first.

billksun:
> I removed the override and now it built and gave me this error:

@YaLTeR, it appears this error is due to the binary compiled by `cargo` unable to find the dependency installed via Nix. This makes sense as Nix libraries are installed in the Nix store and most binaries won't be able to find them.

By using @twoneis' suggestion using RUSTFLAGS to force link to Nix's `libEGL` and `libwayland-client`, the binary compiled by `cargo` will be able to find the libraries, and you should get the same error as running the binaries produced by `nix build` in your previous [comment](https://github.com/YaLTeR/niri/pull/77#issuecomment-1798540387) (at least I did when running from the container, but you might get further since the container worked for you before).

Before adding the RUSTFLAGS, the binary built by `cargo build` on `NixOS` will result in the same error you got. But after adding it (and adding additional `buildInputs` to the `devShell` like @ErrorNoInternet  suggested), I can successfully launch Niri in a window, just like the binary built by `nix build`. This is consistent with the hypothesis that the `cargo` built executable was not able to find the libraries without help. 

Unfortunately, I still haven't figure out why Niri won't launch inside the container, but I will continue to investigate.

bryceberger:
I use nix and rust in some personal projects, and took a stab at this flake. My changes are at https://github.com/bryceberger/niri/tree/nixify, unsure if you want to / the best way to merge them in here.

I am able to run with both `cargo run` and `nix run`. (NixOS, no system rust dependencies, `hardware.opengl.enable = true;`).

billksun:
@bryceberger, can you see if you can get the built binaries (both via `cargo` and `nix`) to run inside a container? That's the blocker right now, as non-NixOS users won't be able to take advantage of the flake.

I'm not too familiar with container development that enables direct access to the host hardware, and all the articles that I am able to find are on making Nvidia GPUs available to the container, while I only have access to AMD GPUs.

bryceberger:
I've had... many many issues getting this to work before. In the meantime, would https://github.com/nix-community/nixGL/ work?

bryceberger:
See https://github.com/bryceberger/niri/commit/756e8e9e046d2ad301d957dd8555f032d269e310 for a docker image that _should_ run on non-nixos.

bryceberger:
After installing and building in a fresh Ubuntu 22.04 VM, I am able to run the binaries built by `cargo build` and `nix build` by just doing `nix run --impure github:guibou/nixGL -- ...`. I was able to run the docker image as well (after adding the user to the `video` group and rebooting), but don't personally see the point of a docker image when the normal `nix build` works just fine.

billksun:
@bryceberger, `nixGL` indeed does the trick!

I am in favor of dropping the container support for a pure Nix approach. We just need to include the instructions for using `nixGL` for non-NixOS users.

I can add you as contributor to my branch, so you can push your changes to this PR as well. Alternatively, we can just abandon this PR and you can create a new PR.

bryceberger:
In favor of adding me to your branch to keep the discussion in one place. I'll add some wording to the readme and do a bit of rebasing to clean up the history.

YaLTeR:
May I once again suggest that making and editing a wiki page could make all of this easier? 😄 I think you don't need permissions to edit wiki pages, right?

YaLTeR:
> But if the wiki is preferred we can make references in the README and direct users to the wiki page, would that be okay?

Yeah, certainly. I can do it myself even. Something along "You can also use nix to build and run niri, check this wiki page for instructions."

YaLTeR:
Hm, for some reason I was sure that wikis can be edited by anyone, but it turns out that by default that's not the case (it can be allowed though). I guess it's not possible to allow that for individual pages, is it?

YaLTeR:
Hey, so @sodiboo made a flake here: https://github.com/sodiboo/niri-flake

I think it's better this way if it lives in a repo controlled by someone actually using nix. I linked that from the README.

Do y'all agree with this?

billksun:
> Hey, so @sodiboo made a flake here: https://github.com/sodiboo/niri-flake
> 
> I think it's better this way if it lives in a repo controlled by someone actually using nix. I linked that from the README.
> 
> Do y'all agree with this?

That flake serves a different purpose. Our flake is for development on any Linux distro, and @sodiboo's is for packaging and installation on NixOS.

sodiboo:
Hey, @YaLTeR, the one I made is specifically for easily *installing niri*, as in providing a NixOS module and the final package, useful for an end-user who isn't actually editing the rust code. This PR does *not* do that, it creates a development environment useful for actually editing niri code, in a reproducible environment (= identical versions of all dependencies including rust toolchains for all nix-based contributors).

I guess this can be confusing from an outside perspective, but a flake is really just a unit of code (with some special "purity" properties that make desirable to use for all kinds of package management stuff). It can have multiple outputs, and as mentioned on matrix, the flake in this PR serves a distinctly different purpose than the one I made. I don't really know how to explain this that well.

Really, these two should be merged into a single unit of nix, as flakes can and do and are "supposed" to do all of these things at once, and the development environment "should" be in the niri repository along with the rest of the source code (this way, getting a full development environment after cloning is just `nix develop`, even if your system never had Rust installed). For packaging as a NixOS module, separating the repositories is fine, but at least for now, this PR is not made obsolete by my flake.

They're nontrivial to merge due to using different cargo-to-nix "adapters". I picked crate2nix to solve one problem: build times. Without going too much into details, the built-in rustPlatform.buildRustPackage (not used by this PR either) makes all of niri one derivation. A derivation is the smallest unit that can be rebuilt, and since everything is "pure", this means cargo cannot cache anything and the whole dependency graph is rebuilt every commit. crate2nix solves this by not building with cargo. it only uses cargo to resolve dependencies, and then on subsequent commits only the dependencies that are updated are rebuilt, because each one is its own derivation and can be rebuilt individually.

I suppose my flake "could" be adapted easily to allow you to build/run using `nix` instead of `cargo`, but that feels unnatural to me and I don't think it's how rust programmers on nix want to work? I certainly wouldn't. maybe I could get used to it.

YaLTeR:
Hmm, thanks for the explanation, I understand it a bit better now.

> these two should be merged into a single unit of nix, as flakes can and do and are "supposed" to do all of these things at once

I guess that would be the better approach indeed?

> I don't think it's how rust programmers on nix want to work?

Well, I can't really answer that either, what is the expected workflow there.

> the development environment "should" be in the niri repository along with the rest of the source code

Yeah well I'm just still a bit uneasy about having files in the repo which I can't fix myself if they break (and looking at the various attempts at nixing niri in different places does not inspire me with confidence that I'd be able to easily figure out what to do if something breaks).

What do people do with other projects where nix files live out of tree? Maybe there's some sort of "established" way of communicating where to find the nix files?

kchibisov:
>What do people do with other projects where nix files live out of tree? Maybe there's some sort of "established" way of communicating where to find the nix files?

You usually write them yourself, which is relatively easy if you know nix, and a lot do `FHS` because nix flakes are not great for contiguous dev cycle, especially when you need to fixup binaries.

However a lot of people have no clue how to write nix flakes, because none of that is really documented, because _experimental_, and a lot of users use nix-shell instead (they are not compat, flake can use shell, but not the other way around iirc).

flakes just bring _complexity_ to normal workflow and doesn't solve anything. `nix-shell` be more suitable, however there's a catch, `nix-shell` is really slow because the evaluation is not cached, so both tools suck in their own way and solve really nothing, probably at this point you just go into `FHS` for all your `src` and call it a day, since users which actually do non-trivial dev (compilers, etc) do exactly that (imagine resolving headers with all of that mess in your non-wrapped compiler).

billksun:
> What do people do with other projects where nix files live out of tree? Maybe there's some sort of "established" way of communicating where to find the nix files?

In my experience, with out-of-tree Nix support, it basically means users are on their own. And you end up with a lot of people writing their own individual Nix flakes, which is okay for someone that wants to fork your code or just want to play with it in their own sandbox. But this doesn't help collaboration, because my build can be different from another person's, because we could be working on different versions of dependencies, using different build tools, 

> they are not compat, flake can use shell, but not the other way around iirc

They actually can be made compatible: https://nixos.wiki/wiki/Flakes#Using_flakes_with_stable_Nix

kchibisov:
>In my experience, with out-of-tree Nix support, it basically means users are on their own. And you end up with a lot of people writing their own individual Nix flakes, which is okay for someone that wants to fork your code or just want to play with it in their own sandbox. But this doesn't help collaboration, because my build can be different from another person's, because we could be working on different versions of dependencies, or using different build tools.

Should I submit my gentoo's ebuilds then and dev scripts? What about guix, etc? It's all opinionated, so unless project requires to build itself with nix I don't see a point. And niri is clearly not harder to build than any other project gui project in rust (e.g. alacritty).

billksun:
> Should I submit my gentoo's ebuilds then and dev scripts? What about guix, etc? It's all opinionated [...]

It's opinionated, sure. But it's also not without practical benefits. Do you expect your contributors to be mostly Gentoo users? If yes, then bundling ebuilds and dev scripts makes it easier to get your contributors onboard. Do you expect your contributors to be Linux users? If yes, then Guix gives the general Linux audience that choose not to fuss with manual development environment setup an option to get started quickly.

Why Nix over Guix then? Well, if you look at [repology](https://repology.org/repositories/graphs) as a gauge of project size (activity, contributors, and output), Guix has _26k_ packages and _52%_ of them are up-to-date, whereas Nix is in it's own class above all tracked distros when it comes to both number of packages available (_88k_) and percentage of up-to-date packages (_79%_). You can deduce that there are a _lot_ more Nix users than Guix users. So this can be a practical justification for supporting Nix over Guix.

To illustrate the problem that Nix solves, I had a hard time trying to replicate @YaLTeR's development environment, in fact, I never got my container environment to work because of extra configuration that may be needed on my system. A development environment specified via Nix has a much higher chance of working than one defined by a container image that still requires additional manual system configuration. 

Also as someone that is still deciding whether to incorporate Rust into my main toolbox, having a Nix development environment specified in the project allows me to quickly contribute to the project without committing my system setup to the Rust tool-chain, so I don't have to worry about polluting my system with packages I no longer need after I am finished with my contribution.

On the other hand, if I were a seasoned Rust developer, Nix allows me to work on multiple projects that may require different versions of the Rust compiler, libraries, and native dependencies; all at the same time, while having the same hermetic build guarantees, giving me a consistent developer experience across projects.

I've also on-boarded many student interns at my workplace using Nix flakes. Most students have no experience using our tool-chain, but having a flake allows them to easily reproduce the development environment without needing any knowledge on how to create Nix flakes. To them, it's just a simple command and they have everything they need.

A note on out-of-tree Nix flakes, I neglected to mention that it is much more common for packaging/installation flakes and derivations to be out-of-tree. After all, the vast majority of the 88k packages in [nixpkgs](https://github.com/NixOS/nixpkgs) are projects that don't have built-in Nix support.

But for development environments, it makes sense to bundle it with the project, because it saves a contributor from needing to setup their own environment or write their own Nix flake, and lets them take advantage of reproducible development environments even if they themselves are not experienced Nix users (like my example with the student interns above).

In regards to `FHS`, I never had to mess with it unless I am dealing with precompiled binaries (from non-Nix packages). Furthermore, most of complexities of Nix the package manager and Nix the language and its libraries are only materialized when you are packaging software or creating configuration modules for Nix.

If using `nix build` to build Niri is something that the project owners don't feel comfortable with, we can reduce the flake to only providing the development environment, and forgo the hermetic builds feature (along with the Nix library used here, `crane`, which promises faster CI build times while still using `cargo` underneath). I do this with my Python projects as some legacy projects are not well supported via the Nix tool-chain.

YaLTeR:
Giving this a try again and looks like it needs changes to account for the newly split crate:

```
⬢[yalter@toolbox niri]$ nix --experimental-features 'nix-command flakes' run .
error: builder for '/nix/store/v7f7yvvv5gkq770cp9i0kr8msrjxxj8q-niri-deps-73db4795203e7c27fc1970ae87c31ed836f572c9.drv' failed with exit code 101;
       last 10 log lines:
       > ++ command cargo --version
       > cargo 1.76.0-nightly (26333c732 2023-11-28)
       > ++ command cargo check --release --locked --all-targets
       > error: failed to load manifest for dependency `niri-config`
       >
       > Caused by:
       >   failed to read `/build/source/niri-config/Cargo.toml`
       >
       > Caused by:
       >   No such file or directory (os error 2)
       For full logs, run 'nix log /nix/store/v7f7yvvv5gkq770cp9i0kr8msrjxxj8q-niri-deps-73db4795203e7c27fc1970ae87c31ed836f572c9.drv'.
error: 1 dependencies of derivation '/nix/store/cjmn1hiqz1yn4rc4gg8lg7qvmb3bs4g8-niri-73db4795203e7c27fc1970ae87c31ed836f572c9.drv' failed to build
```

I tried adding the `niri-config` folder into what looks to be the source file array:

```nix
          src = nix-filter.lib.filter {
            root = ./.;
            include = [
              ./src
              ./niri-config
              ./Cargo.toml
              ./Cargo.lock
              ./resources
            ];
          };
```

I'm not sure if this was the right change, it seemed to make it build.

I then ran `build .` and tried running the result with nixGL as follows:

```
nix --experimental-features 'nix-command flakes' run github:guibou/nixGL#nixGLIntel result/bin/niri
```

It did seem to run fine as a winit window, just missing xcursors:

```
2024-01-07T17:13:21.684804Z  WARN niri::cursor: error loading xcursor default@24: no default icon
2024-01-07T17:13:30.081875Z  WARN niri::cursor: error loading xcursor text@24: no default icon
```

I then tried running niri on the TTY but it failed to draw anything on screen with this output in the log:

```
⬢[yalter@toolbox niri]$ nix --experimental-features 'nix-command flakes' run github:guibou/nixGL#nixGLIntel result/bin/niri
2024-01-07T17:15:32.649643Z  INFO niri: starting version 0.1.0-alpha.3 (unknown commit)
2024-01-07T17:15:32.658299Z DEBUG niri_config: loaded config from "/var/home/yalter/.config/niri/config.kdl"
2024-01-07T17:15:33.106182Z  INFO niri::backend::tty: using as the render node: "/dev/dri/renderD128"
2024-01-07T17:15:33.115647Z DEBUG niri::backend::tty: device added: 57857 "/dev/dri/card1"
did not find extension DRI_Mesa version 1
failed to bind extensions
MESA-LOADER: failed to open zink: /nix/store/pns85bw3vgcdjpcr9in30sysnnyki869-mesa-22.1.4-drivers/lib/dri/zink_dri.so: cannot open shared object file: No such file or directory (search paths /nix/store/2l4j4zp09047g61c0hw5ndw8k7lakd0c-mesa-22.1.4-drivers/lib/dri:/nix/store/pns85bw3vgcdjpcr9in30sysnnyki869-mesa-22.1.4-drivers/lib/dri, suffix _dri)
did not find extension DRI_Mesa version 1
failed to bind extensions
did not find extension DRI_Mesa version 1
failed to bind extensions
2024-01-07T17:15:33.536310Z  WARN niri::backend::tty: error adding device: No such file or directory (os error 2)
2024-01-07T17:15:33.536342Z DEBUG niri::backend::tty: device added: 57856 "/dev/dri/card0"
did not find extension DRI_Mesa version 1
failed to bind extensions
MESA-LOADER: failed to open zink: /nix/store/pns85bw3vgcdjpcr9in30sysnnyki869-mesa-22.1.4-drivers/lib/dri/zink_dri.so: cannot open shared object file: No such file or directory (search paths /nix/store/2l4j4zp09047g61c0hw5ndw8k7lakd0c-mesa-22.1.4-drivers/lib/dri:/nix/store/pns85bw3vgcdjpcr9in30sysnnyki869-mesa-22.1.4-drivers/lib/dri, suffix _dri)
did not find extension DRI_Mesa version 1
failed to bind extensions
did not find extension DRI_Mesa version 1
failed to bind extensions
2024-01-07T17:15:33.640028Z  WARN niri::backend::tty: error adding device: No such file or directory (os error 2)
2024-01-07T17:15:33.640050Z  INFO niri: listening on Wayland socket: wayland-2
```

I'm not sure what this means; the zink error is likely a red herring, it shouldn't try to load it in the first place.

> Do you expect your contributors to be Linux users? If yes, then Guix gives the general Linux audience that choose not to fuss with manual development environment setup an option to get started quickly.

Well, I expect contributors to be Linux users, so I provide a list of packages to install (for both Ubuntu as the "default linux" and Fedora which I use myself) and try to avoid weird build steps (it's just `cargo build`). Then it's up to the user to figure out and install the corresponding packages for their distro. Not ideal I know, but also this is a system component; for my apps I can and do provide a Flatpak manifest that gets you started in one click. I could provide a Dockerfile I guess but I don't feel like figuring out the right incantation to forward the TTY stuff (like toolbox does), and running as a winit window only is kinda meh.

I completely understand the appeal of nix here and there seems to be a lot of interest, which is why I'm still considering it, but it's quite a bit different from any traditional building, and I haven't had the time to learn nix yet to be freely able to adjust it myself when changes require it. If we can live with the nix file in-tree being occasionally broken until someone comes around to fix it, then I guess I can merge it.

> we can reduce the flake to only providing the development environment

I suppose if that means simplifying the nix file and reducing the changes needed in case of breakage, then it is indeed a good idea.

sodiboo:
> > we can reduce the flake to only providing the development environment
> 
> I suppose if that means simplifying the nix file and reducing the changes needed in case of breakage, then it is indeed a good idea.

If you're worried about not being able to fix things if they break, then this is a good compromise, and it would still be quite useful to me. Currently, if i need to make a change to the niri source tree, i usually tend to do something like `LD_LIBRARY_PATH="$(nix eval --raw nixpkgs#wayland)/lib:$(nix eval --raw nixpkgs#libglvnd)/lib:$LD_LIBRARY_PATH" nix-shell -p pkg-config udev libseat libxkbcommon libinput mesa pipewire rustPlatform.bindgenHook --run "cargo run --release"` because i don't want the dependencies to persist.

foo-dogsquared:
I'm just going to make an interjection here and say I don't like this change. You can pull statistics on how Nix is widely used or tout its practical benefits but at the end of the day, Nix is just a developer tool and it is on the developer's opinion to include Nix support including providing a way to build the package and a provisioning a devshell.

Anyways, since the developer don't use Nix and [already expressed concern about supporting a tool that they're not prepared to deal with](https://github.com/YaLTeR/niri/pull/77#issuecomment-1879569309), can I suggest a change that this flake should be put in the `contrib/` folder instead? This is [what Neovim project did](https://github.com/neovim/neovim/tree/367e52cc79a786bbee4456b30f9ec5db7e28d6a5/contrib) and it is a good way to indicate that this is a community-led contribution and not explicitly supported by the developer themselves.

YaLTeR:
That sounds like a nice compromise, would that work for nix users?

sodiboo:
There's certainly no real issue with it, as it could then be imported something like `github:YaLTeR/niri?path=contrib`, which is slightly less aesthetically pleasing but has no functional downside.

But, I personally don't really see much benefit in doing so? Like, it will still break just as often, and it's no "clearer" that it's community-maintained than if it were just mentioned in the README?

The only real change I see compared to keeping it in the root is that it becomes slightly more cumbersome to access (albeit that's definitely nitpicking)

I could maybe see a maintainability benefit if it were a separate repository, perhaps included as a submodule (not sure if that would actually be any better in practice though), but just moving it into a subdirectory seems kinda pointless to me.

billksun:
> That sounds like a nice compromise, would that work for nix users?

One of the main issue with `flake.nix` not existing in project root is losing automatic provisioning of the development shell as soon as you enter the project directory that you get by using [direnv](https://direnv.net/) with [nix-direnv](https://github.com/nix-community/nix-direnv) . It's not the end of the world, you just always have to type `nix develop` to get into the development shell. But having your tools ready as soon as you enter the project directory does make the development experience next-level. Technically, you could still automatically load the development shell by specifying the path of the `flake.nix` file in the configuration file of `direnv`, but this is not the default and will likely throw people off.

What if we add a comment at the top of the `flake.nix` file, specifying that the file is a community supported feature and list out the maintainers (me and anyone else interested) and their email addresses (or GitHub usernames)?

YaLTeR:
Alright I guess, let's do that. Do you mind doing the niri-config crate change then, possibly doing something about missing xcursors, and also anything else you might have left to do?

kchibisov:
>One of the main issue with flake.nix not existing in project root is losing automatic provisioning of the development shell as soon as you enter the project directory that you get by using [direnv](https://direnv.net/) with [nix-direnv](https://github.com/nix-community/nix-direnv) . It's not the end of the world, you just always have to type nix develop to get into the development shell. But having your tools ready as soon as you enter the project directory does make the development experience next-level. Technically, you could still automatically load the development shell by specifying the path of the flake.nix file in the configuration file of direnv, but this is not the default and will likely throw people off.

You can point to flake in literally any path in the `.envrc`, it can reside in arbitrary location.

billksun:
> You can point to flake in literally any path in the `.envrc`, it can reside in arbitrary location.

Indeed, and I did mentioned it, along with my concern:

>Technically, you could still automatically load the development shell by specifying the path of the flake.nix file in the configuration file of direnv, but this is not the default and will likely throw people off.

billksun:
> Alright I guess, let's do that. Do you mind doing the niri-config crate change then, possibly doing something about missing xcursors, and also anything else you might have left to do?

Sure, sounds like a plan.

README.md
```diff
@@ -57,6 +55,26 @@ First, install the dependencies for your distribution.
...
+nix build
+```
+
+To execute the resulting binary:
+
+```
+./results/bin/niri
+```
+    
+For Nix users not on NixOS, you may need [NixGL](https://github.com/nix-community/nixGL) to run the resulting binary:
```
YaLTeR, line 72:
```suggestion
We have a community-maintained flake which provides a devshell with required dependencies. Use `nix build` to build niri, and then run `./results/bin/niri`.
    
If you're not on NixOS, you may need [NixGL](https://github.com/nix-community/nixGL) to run the resulting binary:
```

README.md
```diff
@@ -73,6 +91,10 @@ To do that, put files into the correct directories according to this table.
 
 Doing this will make niri appear in GDM and, presumably, other display managers.
 
+### NixOS
+
+For NixOS users, check out https://github.com/sodiboo/niri-flake
```
YaLTeR, line 96: Could you keep this inside the tip above for now? I think it makes more sense, because people are likely first interested in "normal" installation, which is why I put that closer to the top.

YaLTeR:
Gave it a go one more time, seems to work. Thanks!

PR #84, opened 2023-11-12
Toggle full width when maximize button is clicked

SED4906:
Handling maximize_request and unmaximize_request is simple enough.

### DIFF
--- a/src/handlers/xdg_shell.rs
+++ b/src/handlers/xdg_shell.rs
@@ -91,15 +91,14 @@ impl XdgShellHandler for State {
     }
 
     fn maximize_request(&mut self, surface: ToplevelSurface) {
-        // FIXME
-
         // The protocol demands us to always reply with a configure,
         // regardless of we fulfilled the request or not
+        self.niri.layout.toggle_full_width();
         surface.send_configure();
     }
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
So, proper xdg-shell maximize and unmaximize are slightly more involved. Particularly, they may communicate that the surface had been maximized with a Maximized surface state. Though, that state also requires the client to match the requested geometry exactly, which is not necessarily desired, and can change the client to look different (e.g. GTK windows will lose rounded corners). Then, repeated (un-)maximize requests should do nothing, rather than toggle the state. The maximize capability should also be indicated upon construction of XdgShellState (it currently isn't).

A potentially interesting way to present maximize is like PaperWM, so remove gaps from the window similar to fullscreen, but leave layer-shell struts intact. The only problem I see is that it is quite similar to the already-existing full-width.

All that said, the behavior that this PR implements is an ok fallback and better than just doing nothing imo.

YaLTeR:
Thanks

YaLTeR:
I had to revert this PR unfortunately. It seems that some GTK windows send un/maximize requests upon window creation, which causes the wrong (currently active) window to maximize. I reproduce it by opening a xdp-gnome file chooser, or opening a GTK 4 app via Fuzzel.

Issue #89, opened 2023-12-08
Feature: option for [focus|move]-window-[up|down] to fallback to {focus|move-window-to]-workspace-[up|down]


rabidpug:
Hi,

I migrated to Niri recently and I've been really enjoying the experience. For such a young project I've found it incredibly stable and really had no issues that weren't due to other programs.

I do have a couple of proposals that I think would improve the flexibility of some of the commands and reduce the number of keybinds required to navigate.

The proposal for this issue - I'd like to be able to have the option for the `focus-window` and `move-window` commands to fall back to the `focus-workspace` and `move-window-to-workspace` commands respectively when the former command doesn't do anything.
IE if I have 2 windows in the same column and I'm focused on the top window, with `Mod+Down` set to `focus-window-down` I'd like to have it go:
`Mod+Down` - focus the bottom window
`Mod+Down` - no more windows down to focus, so focus next workspace
etc
This way the vast majority of my navigation can be done with Mod+UDLR alone.

Thanks!

YaLTeR:
Hello,

> I migrated to Niri recently and I've been really enjoying the experience. For such a young project I've found it incredibly stable and really had no issues that weren't due to other programs.

Thanks! That's great to hear :)

> I'd like to be able to have the option for the focus-window and move-window commands to fall back to the focus-workspace and move-window-to-workspace commands respectively when the former command doesn't do anything.

That's an interesting idea, I haven't thought of that before. Shouldn't be too difficult to implement either, I think.

YaLTeR:
Added in https://github.com/YaLTeR/niri/pull/93.

PR #93, opened 2023-12-18
Add optional fallback to workspace focus/move for window focus/move

rabidpug:
Took a crack at implementing #89 as a learning exercise for myself. Seems to be working but open to guidance on my approach as this is unfamiliar territory for me.

This change maintains the existing functionality for focus-window-up/down and move-window-up/down by adding an optional bool parameter that, when true, has the commands fall back to focus-workspace-up/down and move-window-to-workspace-up/down respectively when there is no window to focus/swap with.


src/layout.rs
```diff
@@ -3335,12 +3395,12 @@ mod tests {
         FullscreenWindow(#[proptest(strategy = "1..=5usize")] usize),
         FocusColumnLeft,
         FocusColumnRight,
-        FocusWindowDown,
-        FocusWindowUp,
+        FocusWindowDown(#[proptest(strategy = "any::<bool>()")] bool),
```
YaLTeR, line 3398: I don't think this is needed; the others specify a strategy just because they need something other than the default (i.e. numbers 1 through 5 instead of the whole usize range). For a bool it's not an issue.

YaLTeR (commented):
Thanks! The implementation looks good.

However, "UI" wise, I was thinking about this, and I think I would prefer four new commands instead of a boolean argument, to make it clearer what's going on. I haven't quite settled on the names though, there's the obvious `focus-window-or-workspace-up` and the less clear `move-window-up-or-to-workspace-up`, but I'm open for suggestions, especially for that second one. Lengthiness is less of a concern compared to clarity I'd say, it's not like you're typing these over and over.

src/layout.rs
rabidpug: Ahhk, yea I definitely wasn't sure what I was doing there haha

rabidpug:
I was definitely on the fence about the unclear boolean argument, figured you'd have a stance on the best approach. I've refactored to separate commands with the names you mentioned, can update if we think of better ones.

YaLTeR:
Thanks!

PR #96, opened 2024-01-05
Add `center-column-on-focus` option

tversteeg:
Idea expanded upon in #95.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -143,6 +143,14 @@ layout {
         // top 64
         // bottom 64
     }
+
+    // When to center a column when changing focus, options are:
+    // - "never", default behavior, focusing an off-screen column will keep at the left
+    //   or right edge of the screen.
+    // - "on-overflow", focusing a column will center it if it doesn't fit
+    //   together with the previously focused column.
+    // - "always", the focused column will always be centered.
+    center-focused-column "never"
 }
 
 // Add lines like this to spawn processes at startup.
--- a/src/layout/workspace.rs
+++ b/src/layout/workspace.rs
@@ -3,7 +3,7 @@ use std::iter::zip;
 use std::rc::Rc;
 use std::time::Duration;
 
-use niri_config::{PresetWidth, SizeChange, Struts};
+use niri_config::{CenterFocusedColumn, PresetWidth, SizeChange, Struts};
 use smithay::backend::renderer::element::surface::WaylandSurfaceRenderElement;
 use smithay::backend::renderer::element::utils::RelocateRenderElement;
 use smithay::backend::renderer::{ImportAll, Renderer};
@@ -397,9 +397,7 @@ impl<W: LayoutElement> Workspace<W> {
         new_offset - self.working_area.loc.x
     }
 
-    fn animate_view_offset_to_column(&mut self, current_x: i32, idx: usize) {
-        let new_view_offset = self.compute_new_view_offset_for_column(current_x, idx);
-
+    fn animate_view_offset(&mut self, current_x: i32, idx: usize, new_view_offset: i32) {
         let new_col_x = self.column_x(idx);
         let from_view_offset = current_x - new_col_x;
         self.view_offset = from_view_offset;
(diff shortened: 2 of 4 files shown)


src/layout/workspace.rs
```diff
@@ -426,20 +424,97 @@ impl<W: LayoutElement> Workspace<W> {
...
+            CenterColumnOnMove::NoFit => {
+                let width = self.columns[idx].width();
+
+                // Calculate whether the selected column together with the left column fit on the view
+                let fit_left = self
+                    .column_left_idx(idx)
+                    .map(|left_idx| {
+                        let left_width = self.columns[left_idx].width();
+
+                        width + left_width + self.options.gaps < self.view_size.w
```
YaLTeR, line 477: Non-fullscreen columns always operate on the working area (which includes struts and layer-shell). The view size is only for fullscreen columns.

src/layout/workspace.rs
```diff
@@ -426,20 +424,97 @@ impl<W: LayoutElement> Workspace<W> {
...
+                let fit_right = self
+                    .column_right_idx(idx)
+                    .map(|right_idx| {
+                        let right_width = self.columns[right_idx].width();
+
+                        width + right_width + self.options.gaps < self.view_size.w
+                    })
+                    .unwrap_or_default();
+
+                if fit_left || fit_right {
```
YaLTeR, line 491: This is kind of weird, like if you have a maximized window, then two narrow columns, and you go from the maximized window to the first narrow column, it will flush against the right side of the screen, because the column to the right of it (which is off screen) would fit together with it. In this case it would make more sense to, for instance, make the rightmost column visible too. Or to just center the current column.

src/layout/workspace.rs
```diff
@@ -426,20 +424,97 @@ impl<W: LayoutElement> Workspace<W> {
...
     }
 
+    fn animate_view_offset_to_column(&mut self, current_x: i32, idx: usize) {
+        let new_view_offset = self.compute_new_view_offset_for_column(current_x, idx);
+        self.animate_view_offset(current_x, idx, new_view_offset);
+    }
+
+    fn animate_view_offset_to_column_centered(&mut self, current_x: i32, idx: usize) {
+        if self.columns.is_empty() {
+            self.animate_view_offset_to_column(current_x, idx);
```
YaLTeR, line 434:
```suggestion
```

src/layout/workspace.rs
```diff
@@ -426,20 +424,97 @@ impl<W: LayoutElement> Workspace<W> {
...
+        let col = &self.columns[idx];
+        if col.is_fullscreen {
+            self.animate_view_offset_to_column(current_x, idx);
+            return;
+        }
+
+        let width = col.width();
+
+        // If the column is wider than the working area, then on commit it will be shifted to left
+        // edge alignment by the usual positioning code, so there's no use in doing anything here.
```
YaLTeR, line 447:
```suggestion
        // edge alignment by the usual positioning code, so there's no use in trying to center it here.
```

src/config.rs
```diff
@@ -253,6 +255,17 @@ pub enum PresetWidth {
...
+pub enum CenterColumnOnMove {
+    /// Moving to a column will keep not center the column.
+    #[default]
+    Never,
+    /// Moving to a column will always center the column.
+    Always,
+    /// Moving to a column will center the column if the column doesn't fit on the screen together with one of its neighbors.
+    NoFit,
+}
+
```
YaLTeR, line 268: I suspect if you run `cargo test` you will find out that you need to add this to the test below too, so please do that with a non-default value.

src/layout/workspace.rs
tversteeg: I see the problem here, but I'm struggling to find a nice, clean solution without having to refactor a lot of code. Do you have any suggestions?

YaLTeR: What about your original idea where you check if it fits together with the previously focused window?

resources/default-config.kdl
```diff
@@ -136,6 +136,14 @@ default-column-width { proportion 0.5; }
 // If you leave the brackets empty, the windows themselves will decide their initial width.
 // default-column-width {}
 
+// When to center a column when moving, options are:
```
YaLTeR, line 139:
```suggestion
// When to center a column when changing focus, options are:
```

resources/default-config.kdl
```diff
@@ -136,6 +136,14 @@ default-column-width { proportion 0.5; }
...
+// - "never", default behavior, focusing an off-screen column will keep at the left
+//   or right edge of the screen.
+// - "no-fit", focusing to a column will center the column if the column doesn't fit
```
YaLTeR, line 142:
```suggestion
// - "no-fit", focusing a column will center the column if the column doesn't fit
```

src/config.rs
```diff
@@ -253,6 +255,17 @@ pub enum PresetWidth {
 #[derive(knuffel::Decode, Debug, Clone, PartialEq)]
 pub struct DefaultColumnWidth(#[knuffel(children)] pub Vec<PresetWidth>);
 
+#[derive(knuffel::DecodeScalar, Debug, Default, PartialEq, Eq, Clone, Copy)]
+pub enum CenterColumnOnFocus {
+    /// Moving to a column will keep not center the column.
```
YaLTeR, line 260:
```suggestion
    /// Focusing a column will keep not center the column.
```

src/config.rs
```diff
@@ -253,6 +255,17 @@ pub enum PresetWidth {
...
+    #[default]
+    Never,
+    /// Moving to a column will always center the column.
```
YaLTeR, line 263:
```suggestion
    /// Focusing a column will always center the column.
```

src/config.rs
```diff
@@ -253,6 +255,17 @@ pub enum PresetWidth {
...
+    Always,
+    /// Moving to a column will center the column if the column doesn't fit on the screen together with one of its neighbors.
```
YaLTeR, line 265:
```suggestion
    /// Focusing a column will center the column if the column doesn't fit on the screen together with one of its neighbors.
```

YaLTeR (commented):
The merge conflict is because I moved some settings into a new `layout { }` scope, just move this one there too I think.

src/layout/workspace.rs
tversteeg: Yeah, that's great, don't know why I forgot that haha. Thanks!

src/layout/workspace.rs
```diff
@@ -426,13 +424,69 @@ impl<W: LayoutElement> Workspace<W> {
...
 
         let current_x = self.view_pos();
-        self.animate_view_offset_to_column(current_x, idx);
+        match self.options.center_column_on_focus {
+            CenterColumnOnFocus::Always => {
+                self.animate_view_offset_to_column_centered(current_x, idx)
+            }
+            CenterColumnOnFocus::NoFit => {
+                let source_x = self.column_x(self.active_column_idx);
+                let source_width = self.columns[self.active_column_idx].width();
```
YaLTeR, line 469: In case the focus jumps over multiple columns, do we perhaps want to compare against the neighbor column in the direction, rather than the source column? I.e. `idx + 1` if `active_column_idx > idx` and `idx - 1` otherwise?

YaLTeR (approved):

YaLTeR:
Thanks!

PR #99, opened 2024-01-11
Use the notify library for the config file watcher

tversteeg:
The [notify library](https://github.com/notify-rs/notify) uses an OS specific filesystem notification under the hood, on Linux this is [inotify](https://en.wikipedia.org/wiki/Inotify). This improves the performance of niri because it doesn't have to poll the metadata of the file every 500 ms. The debounce time is set to 500 ms, which ensures that only 1 event every 500 ms will trigger a configuration reload.

### DIFF
--- a/src/watcher.rs
+++ b/src/watcher.rs
@@ -1,6 +1,6 @@
 //! File modification watcher.
 
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::mpsc::{self, RecvTimeoutError};
 use std::thread;
 use std::time::{Duration, Instant};
@@ -14,6 +14,37 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(100);
 const FALLBACK_POLLING_TIMEOUT: Duration = Duration::from_secs(1);
 
 pub fn watch(path: PathBuf, changed: SyncSender<()>) {
+    let Ok(path_metadata) = path.metadata() else {
+        warn!("config file {path:?} is not valid");
+        return;
+    };
+    if !path_metadata.file_type().is_file() {
+        warn!("config file {path:?} is not a file");
+        return;
+    }
+    let Ok(canonical_path) = path.canonicalize() else {
+        warn!("config file {path:?} could not be canonicalized");
+        return;
+    };
+    // When the file is a symlink check both the linked directory and the original one.
+    let paths = match canonical_path.symlink_metadata() {
+        Ok(symlink_metadata) if symlink_metadata.file_type().is_symlink() => {
+            vec![path.clone(), canonical_path]
+        }
+        _ => vec![canonical_path],
+    };
+
+    let mut parents = paths
+        .iter()
+        .map(|path| {
+            let mut path = path.clone();
+            path.pop();
+            path
+        })
+        .collect::<Vec<PathBuf>>();
+    parents.sort_unstable();
+    parents.dedup();
+
     let (tx, rx) = mpsc::channel();
     let mut watcher = match RecommendedWatcher::new(
         tx,
@@ -21,7 +52,7 @@ pub fn watch(path: PathBuf, changed: SyncSender<()>) {
     ) {
         Ok(watcher) => watcher,
         Err(err) => {
-            error!("Unable to watch config file {path:?}: {err}");
+            error!("unable to watch config file {path:?}: {err:?}");
             return;
         }
     };
@@ -29,14 +60,16 @@ pub fn watch(path: PathBuf, changed: SyncSender<()>) {
     thread::Builder::new()
         .name(format!("Filesystem Watcher for {}", path.to_string_lossy()))
         .spawn(move || {
-            // Watch the configuration file directory.
-            let parent = path.parent().map(Path::to_path_buf).unwrap_or_default();
-            if let Err(err) = watcher.watch(&parent, RecursiveMode::NonRecursive) {
-                error!("Unable to watch config directory {parent:?}: {err}");
-            }
-
             let mut debouncing_deadline: Option<Instant> = None;
             let mut events_received_during_debounce = Vec::new();
+
+            for parent in &parents {
+                // Watch the configuration file directory.
+                if let Err(err) = watcher.watch(parent, RecursiveMode::NonRecursive) {
+                    error!("unable to watch config directory {parent:?}: {err:?}");
+                    return;
+                }
+            }
 
             loop {
                 let event = match debouncing_deadline.as_ref() {
@@ -67,7 +100,8 @@ pub fn watch(path: PathBuf, changed: SyncSender<()>) {
 
                         if events_received_during_debounce
                             .drain(..)
-                            .any(|event| event.paths.contains(&path))
+                            .flat_map(|event| event.paths.into_iter())
+                            .any(|path| paths.contains(&path))
                         {
                             if let Err(err) = changed.send(()) {
                                 warn!("error sending change notification: {err:?}");
@@ -76,16 +110,16 @@ pub fn watch(path: PathBuf, changed: SyncSender<()>) {
                         }
                     }
                     Ok(Err(err)) => {
-                        debug!("Config watcher errors: {err:?}");
+                        debug!("config watcher errors: {err:?}");
                     }
                     Err(err) => {
-                        debug!("Config watcher channel dropped unexpectedly: {err}");
+                        debug!("config watcher channel dropped unexpectedly: {err:?}");
                         break;
                     }
                 }
             }
 
-            debug!("Exiting watcher thread for {}", path.to_string_lossy());
+            debug!("exiting watcher thread for {}", path.to_string_lossy());
         })
         .unwrap();
 }
--- a/src/watcher.rs
+++ b/src/watcher.rs
@@ -1,74 +1,91 @@
 //! File modification watcher.
 
-use std::path::PathBuf;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::path::{Path, PathBuf};
+use std::sync::mpsc::{self, RecvTimeoutError};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
+use notify::{Config, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use smithay::reexports::calloop::channel::SyncSender;
 
-pub struct Watcher {
-    should_stop: Arc<AtomicBool>,
-}
+const DEBOUNCE_DELAY: Duration = Duration::from_millis(100);
 
-impl Drop for Watcher {
-    fn drop(&mut self) {
-        self.should_stop.store(true, Ordering::SeqCst);
-    }
-}
+/// The fallback for `RecommendedWatcher` polling.
+const FALLBACK_POLLING_TIMEOUT: Duration = Duration::from_secs(1);
 
-impl Watcher {
-    pub fn new(path: PathBuf, changed: SyncSender<()>) -> Self {
-        let should_stop = Arc::new(AtomicBool::new(false));
+pub fn watch(path: PathBuf, changed: SyncSender<()>) {
+    let (tx, rx) = mpsc::channel();
+    let mut watcher = match RecommendedWatcher::new(
+        tx,
+        Config::default().with_poll_interval(FALLBACK_POLLING_TIMEOUT),
+    ) {
+        Ok(watcher) => watcher,
+        Err(err) => {
+            error!("Unable to watch config file {path:?}: {err}");
+            return;
+        }
+    };
 
-        {
-            let should_stop = should_stop.clone();
-            thread::Builder::new()
-                .name(format!("Filesystem Watcher for {}", path.to_string_lossy()))
-                .spawn(move || {
-                    // this "should" be as simple as mtime, but it does not quite work in practice;
-                    // it doesn't work if the config is a symlink, and its target changes but the
-                    // new target and old target have identical mtimes.
-                    //
-                    // in practice, this does not occur on any systems other than nix.
-                    // because, on nix practically everything is a symlink to /nix/store
-                    // and due to reproducibility, /nix/store keeps no mtime (= 1970-01-01)
-                    // so, symlink targets change frequently when mtime doesn't.
-                    let mut last_props = path
-                        .canonicalize()
-                        .and_then(|canon| Ok((canon.metadata()?.modified()?, canon)))
-                        .ok();
+    thread::Builder::new()
+        .name(format!("Filesystem Watcher for {}", path.to_string_lossy()))
+        .spawn(move || {
+            // Watch the configuration file directory.
+            let parent = path.parent().map(Path::to_path_buf).unwrap_or_default();
+            if let Err(err) = watcher.watch(&parent, RecursiveMode::NonRecursive) {
+                error!("Unable to watch config directory {parent:?}: {err}");
+            }
 
-                    loop {
-                        thread::sleep(Duration::from_millis(500));
+            let mut debouncing_deadline: Option<Instant> = None;
+            let mut events_received_during_debounce = Vec::new();
 
-                        if should_stop.load(Ordering::SeqCst) {
-                            break;
+            loop {
+                let event = match debouncing_deadline.as_ref() {
+                    Some(debouncing_deadline) => rx.recv_timeout(
+                        debouncing_deadline.saturating_duration_since(Instant::now()),
+                    ),
+                    None => {
+                        let event = rx.recv().map_err(Into::into);
+
+                        debouncing_deadline = Some(Instant::now() + DEBOUNCE_DELAY);
+
+                        event
+                    }
+                };
+
+                match event {
+                    Ok(Ok(event)) => match event.kind {
+                        EventKind::Any
+                        | EventKind::Create(_)
+                        | EventKind::Modify(_)
+                        | EventKind::Other => {
+                            events_received_during_debounce.push(event);
                         }
+                        _ => (),
+                    },
+                    Err(RecvTimeoutError::Timeout) => {
+                        debouncing_deadline = None;
 
-                        if let Ok(new_props) = path
-                            .canonicalize()
-                            .and_then(|canon| Ok((canon.metadata()?.modified()?, canon)))
+                        if events_received_during_debounce
+                            .drain(..)
+                            .any(|event| event.paths.contains(&path))
                         {
-                            if last_props.as_ref() != Some(&new_props) {
-                                trace!("file changed: {}", path.to_string_lossy());
-
-                                if let Err(err) = changed.send(()) {
-                                    warn!("error sending change notification: {err:?}");
-                                    break;
-                                }
-
-                                last_props = Some(new_props);
+                            if let Err(err) = changed.send(()) {
+                                warn!("error sending change notification: {err:?}");
+                                break;
                             }
                         }
                     }
+                    Ok(Err(err)) => {
+                        debug!("Config watcher errors: {err:?}");
+                    }
+                    Err(err) => {
+                        debug!("Config watcher channel dropped unexpectedly: {err}");
+                        break;
+                    }
+                }
+            }
 
-                    debug!("exiting watcher thread for {}", path.to_string_lossy());
-                })
-                .unwrap();
-        }
-
-        Self { should_stop }
-    }
+            debug!("Exiting watcher thread for {}", path.to_string_lossy());
+        })
+        .unwrap();
 }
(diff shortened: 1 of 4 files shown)


YaLTeR:
The problem with inotify-based solutions compared to the current one is that they are very brittle. For example, I removed the `.config/niri` folder and replaced it with a symlink, after which on this PR it reloaded the config one more time and then stopped, no matter what I did. The host compositor (running main) kept reloading just fine.

kchibisov:
Given that niri is a `linux/bsd` only I'd suggest to use `inotify` and `kqueue` directly and plug both into the `calloop`. The `notify` crate is a bit big and spawns a separate thread to do basic things and adds a lot of deps to something you likely already have wrappers in deps.

Like for example `rustix` already has `inotify`.

>For example, I removed the .config/niri folder and replaced it with a symlink, after which on this PR it reloaded the config one more time and then stopped, no matter what I did. 

Yeah, but it's not efficient, if you do weird things you should re-run rediscovery code and hook inotify into the new location. It's true that if you have a deep level of symlinks, or weird layout it's hard, but it's users trying to do weird things and trying to mess.

In general IPC based configuration is better than live-reload, especially when you can easily do it (unix), it's not like you have anything to live-reload and you don't have to develop debouncing.

YaLTeR:
> Yeah, but it's not efficient, if you do weird things you should re-run rediscovery code and hook inotify into the new location. It's true that if you have a deep level of symlinks, or weird layout it's hard, but it's users trying to do weird things and trying to mess.

Is the current 500 ms metadata poll really that bad to warrant guessing when else inotify will break and how to fix it this time?

kchibisov:
>Is the current 500 ms metadata poll really that bad to warrant guessing when else inotify will break and how to fix it this time?

We have inotify in alacritty for a long time, no-one ever complained that it's not working besides one dude in weird setup.

YaLTeR:
I checked the same symlinking around thing with alacritty and it seemed to keep reloading throughout so there's that.

tversteeg:
What should I do with this PR?

kchibisov:
you can port alacritty code to watch file, not sure whether you should use notify library though, since it's a bit too much.

tversteeg:
I took a look at alacritty, and they also use the notify library, except they manually implement debouncing, which the additional `notify-debouncer-mini` dependency handles perfectly without adding any other extra sub-dependencies.

Only problem is that I get a very strange compilation error which seems unrelated to my changes.

kchibisov:
mini debouncer was dead broken, but maybe they've fixed it? We use notify library only because we have to work on windows and macOS, however on windows/bsd one could plug right into calloop and save resources, but notify library doesn't provide such handling.

Though, maybe it's fine to use notify given that zbus already brings all its deps.

YaLTeR:
The error is because you updated Smithay (did you run `cargo update`?), and it had a breaking change.

tversteeg:
Oh right, I see, I updated a bunch of dependencies, apparently. I was rebasing to main and due to the long list of merge conflicts I thought it was easier to just remove Cargo.lock and regenerate it again. I'll fix it.

YaLTeR:
Sorry for not commenting here; I'll need to check how notify debouncer mini works, just haven't had the time yet.

tversteeg:
> Sorry for not commenting here; I'll need to check how notify debouncer mini works, just haven't had the time yet.

No problem, take your time! It's not anything crucial anyway.

YaLTeR:
Gave this another try, seems like it still doesn't catch renaming the `~/.config/niri` folder and then creating a new `~/.config/niri` folder with a new `config.kdl`. Also indeed a bit unfortunate that notify brings in `mio`, then spawns a thread, and `notify-debouncer-mini` also spawns its own thread.

tversteeg:
Shall I close the PR in that case?

YaLTeR:
I'm ok with moving to notify, just it needs to be done in such a way that it catches moving folders and symlinks around. Alacritty catches all of that, so maybe could be worth checking what it's doing. Extra threads and mio are unfortunate but not a deal breaker I guess.

src/watcher.rs
```diff
@@ -14,30 +14,63 @@ const DEBOUNCE_DELAY: Duration = Duration::from_millis(100);
 const FALLBACK_POLLING_TIMEOUT: Duration = Duration::from_secs(1);
 
 pub fn watch(path: PathBuf, changed: SyncSender<()>) {
+    let Ok(path_metadata) = path.metadata() else {
+        error!("config file {path:?} is not valid");
```
YaLTeR, line 18: I use `error!` in niri for programming errors that are nevertheless recoverable. So these all should be `warn!` since the config path is controlled by the user.

YaLTeR:
I added a set of tests for the cases the watcher currently handles. Here's this PR rebased on main:

<details>

<summary>notify.patch</summary>

```patch
commit 1d6e0afb0fad3f383366cefb30f54bf198a30a2f
Author: Thomas Versteeg <t@versteeg.email>
Date:   Mon Feb 12 03:21:56 2024 +0700

    Use the notify library for the config file watcher
    
    The notify library uses a OS specific filesystem notification under the
    hood, on Linux this is inotify. This improves the performance of niri
    because it doesn't have to poll the metadata of the file every 500 ms.
    The debounce time is set to 500 ms which ensures that only 1 event every
    500 ms will trigger a configuration reload.

diff --git a/Cargo.lock b/Cargo.lock
index 0a80b6f..484c276 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1042,6 +1042,18 @@ dependencies = [
  "rustc_version",
 ]
 
+[[package]]
+name = "filetime"
+version = "0.2.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1ee447700ac8aa0b2f2bd7bc4462ad686ba06baa6727ac149a2d6277f0d240fd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "redox_syscall 0.4.1",
+ "windows-sys 0.52.0",
+]
+
 [[package]]
 name = "flate2"
 version = "1.0.28"
@@ -1613,6 +1625,26 @@ dependencies = [
  "hashbrown",
 ]
 
+[[package]]
+name = "inotify"
+version = "0.9.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f8069d3ec154eb856955c1c0fbffefbf5f3c40a104ec912d4797314c1801abff"
+dependencies = [
+ "bitflags 1.3.2",
+ "inotify-sys",
+ "libc",
+]
+
+[[package]]
+name = "inotify-sys"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e05c02b5e89bff3b946cedeca278abc628fe811e604f027c45a8aa3cf793d0eb"
+dependencies = [
+ "libc",
+]
+
 [[package]]
 name = "input"
 version = "0.9.0"
@@ -1756,6 +1788,26 @@ dependencies = [
  "syn 1.0.109",
 ]
 
+[[package]]
+name = "kqueue"
+version = "1.0.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7447f1ca1b7b563588a205fe93dea8df60fd981423a768bc1c0ded35ed147d0c"
+dependencies = [
+ "kqueue-sys",
+ "libc",
+]
+
+[[package]]
+name = "kqueue-sys"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ed9625ffda8729b85e45cf04090035ac368927b8cebc34898e7c120f52e4838b"
+dependencies = [
+ "bitflags 1.3.2",
+ "libc",
+]
+
 [[package]]
 name = "lazy_static"
 version = "1.4.0"
@@ -2051,6 +2103,18 @@ dependencies = [
  "simd-adler32",
 ]
 
+[[package]]
+name = "mio"
+version = "0.8.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f3d0b296e374a4e6f3c7b0a1f5a51d748a0d34c85e7dc48fc3fa9a87657fe09"
+dependencies = [
+ "libc",
+ "log",
+ "wasi",
+ "windows-sys 0.48.0",
+]
+
 [[package]]
 name = "ndk"
 version = "0.8.0"
@@ -2101,6 +2165,7 @@ dependencies = [
  "log",
  "niri-config",
  "niri-ipc",
+ "notify",
  "notify-rust",
  "pangocairo",
  "pipewire",
@@ -2192,6 +2257,23 @@ dependencies = [
  "minimal-lexical",
 ]
 
+[[package]]
+name = "notify"
+version = "6.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6205bd8bb1e454ad2e27422015fb5e4f2bcc7e08fa8f27058670d208324a4d2d"
+dependencies = [
+ "bitflags 2.4.2",
+ "filetime",
+ "inotify",
+ "kqueue",
+ "libc",
+ "log",
+ "mio",
+ "walkdir",
+ "windows-sys 0.48.0",
+]
+
 [[package]]
 name = "notify-rust"
 version = "4.10.0"
diff --git a/Cargo.toml b/Cargo.toml
index 1101136..f949ce9 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ input = { version = "0.9.0", features = ["libinput_1_21"] }
 keyframe = { version = "1.1.1", default-features = false }
 libc = "0.2.153"
 log = { version = "0.4.20", features = ["max_level_trace", "release_max_level_debug"] }
+notify = { version = "6.1.1", default-features = false }
 niri-config = { version = "0.1.1", path = "niri-config" }
 niri-ipc = { version = "0.1.1", path = "niri-ipc", features = ["clap"] }
 notify-rust = { version = "4.10.0", optional = true }
diff --git a/src/main.rs b/src/main.rs
index bcf0fea..3df518b 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,6 @@ use std::{env, mem};
 
 use clap::Parser;
 use directories::ProjectDirs;
-use niri::animation;
 use niri::cli::{Cli, Sub};
 #[cfg(feature = "dbus")]
 use niri::dbus;
@@ -18,7 +17,7 @@ use niri::niri::State;
 use niri::utils::{
     cause_panic, spawn, version, REMOVE_ENV_RUST_BACKTRACE, REMOVE_ENV_RUST_LIB_BACKTRACE,
 };
-use niri::watcher::Watcher;
+use niri::{animation, watcher};
 use niri_config::Config;
 use portable_atomic::Ordering;
 use sd_notify::NotifyState;
@@ -197,9 +196,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     };
 
     // Set up config file watcher.
-    let _watcher = if let Some(path) = path.clone() {
+    if let Some(path) = path.clone() {
         let (tx, rx) = calloop::channel::sync_channel(1);
-        let watcher = Watcher::new(path.clone(), tx);
+        watcher::watch(path.clone(), tx);
         event_loop
             .handle()
             .insert_source(rx, move |event, _, state| match event {
@@ -207,10 +206,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 calloop::channel::Event::Closed => (),
             })
             .unwrap();
-        Some(watcher)
-    } else {
-        None
-    };
+    }
 
     // Spawn commands from cli and auto-start.
     spawn(cli.command);
diff --git a/src/watcher.rs b/src/watcher.rs
index e5b9458..e42ed59 100644
--- a/src/watcher.rs
+++ b/src/watcher.rs
@@ -1,88 +1,139 @@
 //! File modification watcher.
 
 use std::path::PathBuf;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{mpsc, Arc};
+use std::sync::mpsc::{self, RecvTimeoutError};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
+use notify::{Config, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
 use smithay::reexports::calloop::channel::SyncSender;
 
-pub struct Watcher {
-    should_stop: Arc<AtomicBool>,
-}
+const DEBOUNCE_DELAY: Duration = Duration::from_millis(100);
 
-impl Drop for Watcher {
-    fn drop(&mut self) {
-        self.should_stop.store(true, Ordering::SeqCst);
-    }
+/// The fallback for `RecommendedWatcher` polling.
+const FALLBACK_POLLING_TIMEOUT: Duration = Duration::from_secs(1);
+
+pub fn watch(path: PathBuf, changed: SyncSender<()>) {
+    watch_with_start_notification(path, changed, None)
 }
 
-impl Watcher {
-    pub fn new(path: PathBuf, changed: SyncSender<()>) -> Self {
-        Self::with_start_notification(path, changed, None)
+pub fn watch_with_start_notification(
+    path: PathBuf,
+    changed: SyncSender<()>,
+    started: Option<mpsc::SyncSender<()>>,
+) {
+    let Ok(path_metadata) = path.metadata() else {
+        warn!("config file {path:?} is not valid");
+        return;
+    };
+    if !path_metadata.file_type().is_file() {
+        warn!("config file {path:?} is not a file");
+        return;
     }
-
-    pub fn with_start_notification(
-        path: PathBuf,
-        changed: SyncSender<()>,
-        started: Option<mpsc::SyncSender<()>>,
-    ) -> Self {
-        let should_stop = Arc::new(AtomicBool::new(false));
-
-        {
-            let should_stop = should_stop.clone();
-            thread::Builder::new()
-                .name(format!("Filesystem Watcher for {}", path.to_string_lossy()))
-                .spawn(move || {
-                    // this "should" be as simple as mtime, but it does not quite work in practice;
-                    // it doesn't work if the config is a symlink, and its target changes but the
-                    // new target and old target have identical mtimes.
-                    //
-                    // in practice, this does not occur on any systems other than nix.
-                    // because, on nix practically everything is a symlink to /nix/store
-                    // and due to reproducibility, /nix/store keeps no mtime (= 1970-01-01)
-                    // so, symlink targets change frequently when mtime doesn't.
-                    let mut last_props = path
-                        .canonicalize()
-                        .and_then(|canon| Ok((canon.metadata()?.modified()?, canon)))
-                        .ok();
-
-                    if let Some(started) = started {
-                        let _ = started.send(());
+    let Ok(canonical_path) = path.canonicalize() else {
+        warn!("config file {path:?} could not be canonicalized");
+        return;
+    };
+    // When the file is a symlink check both the linked directory and the original one.
+    let paths = match canonical_path.symlink_metadata() {
+        Ok(symlink_metadata) if symlink_metadata.file_type().is_symlink() => {
+            vec![path.clone(), canonical_path]
+        }
+        _ => vec![canonical_path],
+    };
+
+    let mut parents = paths
+        .iter()
+        .map(|path| {
+            let mut path = path.clone();
+            path.pop();
+            path
+        })
+        .collect::<Vec<PathBuf>>();
+    parents.sort_unstable();
+    parents.dedup();
+
+    let (tx, rx) = mpsc::channel();
+    let mut watcher = match RecommendedWatcher::new(
+        tx,
+        Config::default().with_poll_interval(FALLBACK_POLLING_TIMEOUT),
+    ) {
+        Ok(watcher) => watcher,
+        Err(err) => {
+            error!("unable to watch config file {path:?}: {err:?}");
+            return;
+        }
+    };
+
+    thread::Builder::new()
+        .name(format!("Filesystem Watcher for {}", path.to_string_lossy()))
+        .spawn(move || {
+            let mut debouncing_deadline: Option<Instant> = None;
+            let mut events_received_during_debounce = Vec::new();
+
+            for parent in &parents {
+                // Watch the configuration file directory.
+                if let Err(err) = watcher.watch(parent, RecursiveMode::NonRecursive) {
+                    error!("unable to watch config directory {parent:?}: {err:?}");
+                    return;
+                }
+            }
+
+            if let Some(started) = started {
+                let _ = started.send(());
+            }
+
+            loop {
+                let event = match debouncing_deadline.as_ref() {
+                    Some(debouncing_deadline) => rx.recv_timeout(
+                        debouncing_deadline.saturating_duration_since(Instant::now()),
+                    ),
+                    None => {
+                        let event = rx.recv().map_err(Into::into);
+
+                        debouncing_deadline = Some(Instant::now() + DEBOUNCE_DELAY);
+
+                        event
                     }
-
-                    loop {
-                        thread::sleep(Duration::from_millis(500));
-
-                        if should_stop.load(Ordering::SeqCst) {
-                            break;
+                };
+
+                match event {
+                    Ok(Ok(event)) => match event.kind {
+                        EventKind::Any
+                        | EventKind::Create(_)
+                        | EventKind::Modify(_)
+                        | EventKind::Other => {
+                            events_received_during_debounce.push(event);
                         }
-
-                        if let Ok(new_props) = path
-                            .canonicalize()
-                            .and_then(|canon| Ok((canon.metadata()?.modified()?, canon)))
+                        _ => (),
+                    },
+                    Err(RecvTimeoutError::Timeout) => {
+                        debouncing_deadline = None;
+
+                        if events_received_during_debounce
+                            .drain(..)
+                            .flat_map(|event| event.paths.into_iter())
+                            .any(|path| paths.contains(&path))
                         {
-                            if last_props.as_ref() != Some(&new_props) {
-                                trace!("file changed: {}", path.to_string_lossy());
-
-                                if let Err(err) = changed.send(()) {
-                                    warn!("error sending change notification: {err:?}");
-                                    break;
-                                }
-
-                                last_props = Some(new_props);
+                            if let Err(err) = changed.send(()) {
+                                warn!("error sending change notification: {err:?}");
+                                break;
                             }
                         }
                     }
+                    Ok(Err(err)) => {
+                        debug!("config watcher errors: {err:?}");
+                    }
+                    Err(err) => {
+                        debug!("config watcher channel dropped unexpectedly: {err:?}");
+                        break;
+                    }
+                }
+            }
 
-                    debug!("exiting watcher thread for {}", path.to_string_lossy());
-                })
-                .unwrap();
-        }
-
-        Self { should_stop }
-    }
+            debug!("exiting watcher thread for {}", path.to_string_lossy());
+        })
+        .unwrap();
 }
 
 #[cfg(test)]
@@ -90,7 +141,7 @@ mod tests {
     use std::error::Error;
     use std::fs::File;
     use std::io::Write;
-    use std::sync::atomic::AtomicU8;
+    use std::sync::atomic::{AtomicU8, Ordering};
     use std::time::SystemTime;
 
     use calloop::channel::sync_channel;
@@ -122,7 +173,7 @@ mod tests {
 
         let (tx, rx) = sync_channel(1);
         let (started_tx, started_rx) = mpsc::sync_channel(1);
-        let _watcher = Watcher::with_start_notification(config_path.clone(), tx, Some(started_tx));
+        watch_with_start_notification(config_path.clone(), tx, Some(started_tx));
         loop_handle
             .insert_source(rx, |_, _, _| {
                 changed.fetch_add(1, Ordering::SeqCst);
```

</details>

It fails to handle the following cases:
```
test watcher::tests::create_dir_and_file ... FAILED
test watcher::tests::create_file ... FAILED
test watcher::tests::recreate_file ... ok
test watcher::tests::change_file ... ok
test watcher::tests::change_linked_file ... ok
test watcher::tests::change_file_in_linked_dir ... ok
test watcher::tests::recreate_dir ... FAILED
test watcher::tests::swap_dir ... FAILED
test watcher::tests::swap_just_link ... FAILED
test watcher::tests::swap_dir_link ... FAILED
```

Most of these have to do with the fact that the directory itself either does not exist at the start, or gets recreated at runtime. Tbh it would be good to keep this working, but I guess not absolutely critical. (Alacritty fails most of these too..)

The important one is swap_just_link because it's an actual case on NixOS. I'm not sure why it isn't caught but it must keep working.

YaLTeR:
Btw one reason I want live reload to keep working in as many cases as possible is because if it breaks, you have to restart the compositor, and that's *very annoying* (compared to, say, restarting a terminal). Sure it can be worked around by adding a `niri msg` to restart live reload, but you need to *know* about it and know that this problem can even happen, which especially new users won't.

YaLTeR:
I wonder if it's even possible to handle these cases well without massive complexity... And like, the cases are pretty ordinary, too. Installed niri, liked it, want to put the config into your dotfiles repo? One reasonable way is to move niri folder into the repo, then symlink it back into .config. Now inotify broke unless you were also watching .config itself..

tversteeg:
> I added a set of tests for the cases the watcher currently handles.

This is great, thanks a lot! I'll fix as much as possible.

> One reasonable way is to move niri folder into the repo, then symlink it back into .config. Now inotify broke unless you were also watching .config itself..

This should still work because if a symlink is detected both the folder where the symlink is and the folder the symlink points to should be watched.

tversteeg:
I've taken a look at the new tests, and I think that it doesn't make sense to use the notify library with these added requirements.

YaLTeR:
Let's close for now, just don't delete the branch in case we want to pick it up later. I have some future ideas on how to use notify together with the current polling: e.g. keep polling less frequently, like once a second, as the fallback that always works, but also try to set up a *simple* notify just to get much faster reloads in the simple case. For this idea however, 2 extra threads for notify already starts to sound pretty bad. Maybe we can do it on top of calloop somehow.

PR #120, opened 2024-01-21
tty: reset device to recover from certain errors

cmeissl:
after a session resume (tty switch) we might
end up with a state mismatch. in this case
we can try to reset the device to recover
from the mismatch.

depends on https://github.com/Smithay/smithay/pull/1295

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1406,25 +1406,22 @@ dependencies = [
 
 [[package]]
 name = "input"
-version = "0.8.3"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e6e74cd82cedcd66db78742a8337bdc48f188c4d2c12742cbc5cd85113f0b059"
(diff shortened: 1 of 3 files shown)


YaLTeR:
I'm giving this some testing. Some comments so far:
1. We need to queue a redraw after resetting the connectors, because otherwise the monitors will stay off until the next time there's a redraw.
    * In my local checkout I added a new `RenderResult::Retry` for this, and made it not break the in-progress lock state, since we're retrying.
    * Is there a possibility of a retry busy-loop here?
2. I guess the same check needs to happen in `use_mode()` [here](https://github.com/YaLTeR/niri/blob/df48337d83f78bbd2a923863e73941b6bd236a28/src/backend/tty.rs#L1325). Since on main I made the output config change wait until session resume, and with this PR on top of main the `use_mode()` call in this case will fail with `DrmError(TestFailed(crtc::Handle(95)))` so the mode will not apply until the next time the config is changed.

cmeissl:
superseded by [cbbb7a26fc0bff426d67bb2804167f01ade5fbe4](https://github.com/YaLTeR/niri/commit/cbbb7a26fc0bff426d67bb2804167f01ade5fbe4)

PR #121, opened 2024-01-22
handle symlinks for hot-reloading (nix)

sodiboo:
on nix, using home-manager, we typically have config files as symlinks to nix store. this breaks the current implementation, as it will not detect anything has changed. nix store does not store mtime (because the result should be identical no matter when you build it).

this PR adds *correct* handling of the watcher, not just for nix, but in general:

- if it is a symlink and the mtime changes, but it still points to the same target file and that file hasn't updated mtime, then we can ignore this because the config file hasn't changed, only the symlink.
- likewise, if the symlink path changes (from a previous symlink in a chain), but the ultimate target is unchanged, we can ignore this.

- we cannot ignore if the symlinks end up pointing to a *completely different canonical path*, then it's a different file entirely and we reload it.
- we cannot ignore if the target file's mtime is modified. the config changed and we reload it.
- if the path changes *and* the mtime, then we need to reload it even more than before.
- if neither the canonical path, nor the mtime of config changed, then the config didn't change. doesn't matter what the symlinks do, it still points to the same target.

### DIFF
--- a/src/watcher.rs
+++ b/src/watcher.rs
@@ -27,7 +27,18 @@ impl Watcher {
             thread::Builder::new()
                 .name(format!("Filesystem Watcher for {}", path.to_string_lossy()))
                 .spawn(move || {
-                    let mut last_mtime = path.metadata().and_then(|meta| meta.modified()).ok();
+                    // this "should" be as simple as mtime, but it does not quite work in practice;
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks, seems to work well.

PR #137, opened 2024-01-28
Add option to rotate outputs

### LINKED ISSUE #106: Add monitor rotation setting
Needs to thread down to everything and make sure direct scanout still works (Wayland clients need to know about the rotation, submit correctly rotated buffers). Check that screencast and screenshot still works.

axtloss:
Adds support for output Transformation using a config option `transform` in the `output` section to allow setting it per output
closes #106

Attempting to do a screenshot on a rotated output (tested with rotation 270) currently cuts the available output to screenshot by half, any part of the selected area that goes into the "invisible" area becomes transparent

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -187,6 +187,8 @@ pub struct Output {
     pub name: String,
     #[knuffel(child, unwrap(argument), default = 1.)]
     pub scale: f64,
+    #[knuffel(child, unwrap(argument, str), default = Transform::Normal)]
+    pub transform: Transform,
     #[knuffel(child)]
     pub position: Option<Position>,
     #[knuffel(child, unwrap(argument, str))]
@@ -199,8 +201,58 @@ impl Default for Output {
             off: false,
             name: String::new(),
             scale: 1.,
+            transform: Transform::Normal,
             position: None,
             mode: None,
+        }
+    }
+}
+
+/// Output transform, which goes counter-clockwise.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Transform {
+    Normal,
+    _90,
+    _180,
+    _270,
+    Flipped,
+    Flipped90,
+    Flipped180,
+    Flipped270,
+}
+
+impl FromStr for Transform {
+    type Err = miette::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "normal" => Ok(Self::Normal),
+            "90" => Ok(Self::_90),
+            "180" => Ok(Self::_180),
+            "270" => Ok(Self::_270),
+            "flipped" => Ok(Self::Flipped),
+            "flipped-90" => Ok(Self::Flipped90),
+            "flipped-180" => Ok(Self::Flipped180),
+            "flipped-270" => Ok(Self::Flipped270),
+            _ => Err(miette!(concat!(
+                r#"invalid transform, can be "90", "180", "270", "#,
+                r#""flipped", "flipped-90", "flipped-180" or "flipped-270""#
+            ))),
+        }
+    }
+}
+
+impl From<Transform> for smithay::utils::Transform {
+    fn from(value: Transform) -> Self {
+        match value {
+            Transform::Normal => Self::Normal,
+            Transform::_90 => Self::_90,
+            Transform::_180 => Self::_180,
+            Transform::_270 => Self::_270,
+            Transform::Flipped => Self::Flipped,
+            Transform::Flipped90 => Self::Flipped90,
+            Transform::Flipped180 => Self::Flipped180,
+            Transform::Flipped270 => Self::Flipped270,
         }
     }
 }
@@ -726,6 +778,7 @@ mod tests {
 
             output "eDP-1" {
                 scale 2.0
+                transform "flipped-90"
                 position x=10 y=20
                 mode "1920x1080@144"
             }
@@ -826,6 +879,7 @@ mod tests {
                     off: false,
                     name: "eDP-1".to_owned(),
                     scale: 2.,
+                    transform: Transform::Flipped90,
                     position: Some(Position { x: 10, y: 20 }),
                     mode: Some(Mode {
                         width: 1920,
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -65,6 +65,10 @@ input {
     // Scale is a floating-point number, but at the moment only integer values work.
     scale 2.0
 
+    // Transform allows to rotate the output counter-clockwise, valid values are:
+    // normal, 90, 180, 270, flipped, flipped-90, flipped-180 and flipped-270.
+    transform "normal"
+
     // Resolution and, optionally, refresh rate of the output.
     // The format is "<width>x<height>" or "<width>x<height>@<refresh rate>".
     // If the refresh rate is omitted, niri will pick the highest refresh rate
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -648,20 +648,22 @@ impl State {
             let mut resized_outputs = vec![];
             for output in self.niri.global_space.outputs() {
                 let name = output.name();
-                let scale = self
-                    .niri
-                    .config
-                    .borrow()
-                    .outputs
-                    .iter()
-                    .find(|o| o.name == name)
-                    .map(|c| c.scale)
-                    .unwrap_or(1.);
+                let config = self.niri.config.borrow_mut();
+                let config = config.outputs.iter().find(|o| o.name == name);
+
+                let scale = config.map(|c| c.scale).unwrap_or(1.);
                 let scale = scale.clamp(1., 10.).ceil() as i32;
-                if output.current_scale().integer_scale() != scale {
+
+                let transform = config
+                    .map(|c| c.transform.into())
+                    .unwrap_or(Transform::Normal);
+
+                if output.current_scale().integer_scale() != scale
+                    || output.current_transform() != transform
+                {
                     output.change_current_state(
                         None,
-                        None,
+                        Some(transform),
                         Some(output::Scale::Integer(scale)),
                         None,
                     );
@@ -1172,18 +1174,21 @@ impl Niri {
         let global = output.create_global::<State>(&self.display_handle);
 
         let name = output.name();
-        let scale = self
-            .config
-            .borrow()
-            .outputs
-            .iter()
-            .find(|o| o.name == name)
-            .map(|c| c.scale)
-            .unwrap_or(1.);
-        let scale = scale.clamp(1., 10.).ceil() as i32;
 
-        // Set scale before adding to the layout since that will read the output size.
-        output.change_current_state(None, None, Some(output::Scale::Integer(scale)), None);
+        let config = self.config.borrow();
+        let c = config.outputs.iter().find(|o| o.name == name);
+        let scale = c.map(|c| c.scale).unwrap_or(1.);
+        let scale = scale.clamp(1., 10.).ceil() as i32;
+        let transform = c.map(|c| c.transform.into()).unwrap_or(Transform::Normal);
+        drop(config);
+
+        // Set scale and transform before adding to the layout since that will read the output size.
+        output.change_current_state(
+            None,
+            Some(transform),
+            Some(output::Scale::Integer(scale)),
+            None,
+        );
 
         self.layout.add_output(output.clone());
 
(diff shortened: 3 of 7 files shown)


resources/default-config.kdl
```diff
@@ -65,6 +65,10 @@ input {
     // Scale is a floating-point number, but at the moment only integer values work.
     scale 2.0
 
+    // Transform allows rotating the output, values are:
+    // Normal, 90, 180, 270, Flipped, Flipped90, Flipped180 and Flipped270
```
YaLTeR, line 69: To be consistent with the rest of the config, make these kebab case, i.e. "normal", "flipped-90".

src/niri.rs
```diff
@@ -655,10 +655,31 @@ impl State {
...
                     .unwrap_or(1.);
                 let scale = scale.clamp(1., 10.).ceil() as i32;
+
+                let transform = self
+                    .niri
+                    .config
+                    .borrow()
+                    .outputs
+                    .iter()
+                    .find(|o| o.name == name)
```
YaLTeR, line 665: Extract this part into a variable since `scale` above also uses it.

niri-config/src/lib.rs
```diff
@@ -187,6 +187,8 @@ pub struct Output {
     pub name: String,
     #[knuffel(child, unwrap(argument), default = 1.)]
     pub scale: f64,
+    #[knuffel(child, unwrap(argument), default = String::from("Normal"))]
+    pub transform: String,
```
YaLTeR, line 191: Make this an `enum Transform` with the same variants as `smithay::utils::Transform`, implement `FromStr` for it, and implement `From<Transform> for smithay::utils::Transform`. This way invalid values won't be accepted, and you won't have to convert manually elsewhere in the code. See `TapButtonMap` and `AccelProfile` for an example.

src/niri.rs
```diff
@@ -1173,8 +1194,28 @@ impl Niri {
             .unwrap_or(1.);
         let scale = scale.clamp(1., 10.).ceil() as i32;
 
+        let transform = self
+            .config
+            .borrow()
+            .outputs
+            .iter()
+            .find(|o| o.name == name)
```
YaLTeR, line 1202: Same, extract to variable.

YaLTeR (changes requested):
Thanks! Some more things:
1. Screenshot should work fine. Changing transform when the screenshot UI is open shouldn't break it. Currently if the scale changes the screenshot UI will close automatically, just do the same with transform.
2. Screencast should also work fine. Things should not break when the monitor is rotated while it's being screencasted. For now I suggest just stopping the screencast if that happens, because if the monitor is rotated 90° then the buffer dimensions would change, which there's no code to handle atm.
3. `Niri::refresh_pointer_outputs()` seems to hardcode `Transform::Normal` for the clients right now, it should use the current output transform.

axtloss:
(i will apply the code suggestions in a bit)

> Screenshot should work fine. Changing transform when the screenshot UI is open shouldn't break it. Currently if the scale changes the screenshot UI will close automatically, just do the same with transform.

Weird, for me part of the transformed display just disappears in some way, the resulting screenshot turns transparent:
image

YaLTeR:
You probably need to change some hardcoded Transform values inside screenshot UI.

axtloss:
I tested with obs, and I think I see what happens, despite the output being transformed what screen captures see are an untransformed screen
image

YaLTeR:
Yeah, I think also might be caused by some hardcoded Transform values.

axtloss:
> Yeah, I think also might be caused by some hardcoded Transform values.

Yes that appears to be the issue, changing some hardcoded values around changes makes it see the output correctly.

cmeissl:
This should at least document that the transform is counter clock wise as defined by the wayland protocol spec.
Otherwise this could create some confusion especially with automated changes based on hardware sensors which
might put out clock wise rotation. An alternative would be to name it `orientation`, define it as CW and invert before applying to the wayland outpu.

YaLTeR:
Hmm. Sway calls it "transform" and performs it clockwise. I think we should do it clockwise too?

cmeissl:
> Hmm. Sway calls it "transform" and performs it clockwise. I think we should do it clockwise too?

Also fine, as long as it is documented/specified somewhere. 
This will still require to invert the transform before setting it on the `Output`, because this transform is
[defined as CCW](https://wayland.app/protocols/wayland#wl_output:enum:transform:entry:90).

axtloss:
The latest commit fixes the screenshot issue somewhat, the preview ignores the transform and is rotated normally, causing part of the desktop to be cut off, however the actual screenshot functions fine.
I am not sure how to fix the rendering issue, I would greatly appreciate some help/indicators where the issue could be

YaLTeR:
Maybe due to the hardcoded transform in `render_to_texture()`? Might need to set it there and hardcode Normal back in screenshot UI; if you don't manage to get it working I'll take a look myself.

axtloss:
> This should at least document that the transform is counter clock wise as defined by the wayland protocol spec. Otherwise this could create some confusion especially with automated changes based on hardware sensors which might put out clock wise rotation. An alternative would be to name it `orientation`, define it as CW and invert before applying to the wayland outpu.

I made it rotate clockwise and also mentioned it in the default config

axtloss:
> Maybe due to the hardcoded transform in `render_to_texture()`? Might need to set it there and hardcode Normal back in screenshot UI; if you don't manage to get it working I'll take a look myself.

I made `render_to_texture()` use the current transform, but it still does not work, it now is upside down and still cut off. I am not really sure what else to try

YaLTeR:
I'll take a look myself then (but not today).

YaLTeR:
Figured it out. Gonna push some commits into this PR.

axtloss:
that's nice! should I invite you as a collaborator on my fork to make it easier to add the changes?

YaLTeR:
It says "Maintainers are allowed to edit this pull request." which means I should be able to push to this branch.

YaLTeR:
Pushed some fixes, please try if everything works. I changed transform to mean counter-clockwise (think it's more important to avoid confusion between the types).

There are still a few bugs left, notably with direct scanout and with changing transform on the fly, but I suspect those need Smithay fixes.

axtloss:
What bugs did you experience with changing transform on the fly? When I tested with a nested niri session it worked fine

YaLTeR:
- if you get direct scanout with 90 or 270 transform, it will show up as the inverse transform instead
- if you change transform on the fly in such a way that width and height aren't swapped (e.g. normal to 180 or normal to flipped or 90 to 270) then the monitor isn't damaged, so you'll see artifacts until the monitor is eventually fully redamaged again

axtloss:
Is it possible to damage an output or force a redraw on it? I tried using `niri.redraw()` but that just crashes niri so I probably did something wrong

YaLTeR:
I think we can fix this in Smithay instead

YaLTeR:
Hmm? I haven't seen any crashes. Could you post the backtrace please?

axtloss:
The crashes stopped happening and no matter what I do I cant seem to get them back, maybe it was fixed by me rebooting the computer

YaLTeR:
Another win for the turning it off and on again computer fixing strategy

Have you by chance been running niri as a session when you got the crashes? Then they might still be present in `journalctl -b -1 /usr/bin/niri`

YaLTeR:
I updated to the latest Smithay, and something I just noticed that might be new after the update: "flipped-90" and "flipped-270" are both vertically and horizontally flipped relative to their normal counterparts, whereas "flipped" and "flipped-180" are only horizontally flipped. In sway each flip is only horizontally flipped. @cmeissl could this be something that has regressed?

cmeissl:
> I updated to the latest Smithay, and something I just noticed that might be new after the update: "flipped-90" and "flipped-270" are both vertically and horizontally flipped relative to their normal counterparts, whereas "flipped" and "flipped-180" are only horizontally flipped. In sway each flip is only horizontally flipped. @cmeissl could this be something that has regressed?

Unlikely that this is a regression from the latest changes, can you give me simple reproduction steps?

YaLTeR:
Set monitor transform "90", then change to "flipped-90", observe that it's flipped in both ways (so it should be "flipped-270" instead).

cmeissl:
> Set monitor transform "90", then change to "flipped-90", observe that it's flipped in both ways (so it should be "flipped-270" instead).

Yeah, can reproduce this  and no, this can not be caused by this changes. 
I will take a look a this later this evening.

axtloss:
> Another win for the turning it off and on again computer fixing strategy
> 
> Have you by chance been running niri as a session when you got the crashes? Then they might still be present in `journalctl -b -1 /usr/bin/niri`

Yep theres a crash log in journalctl, though I still cannot reproduce it
```
thread 'main' panicked at src/niri.rs:1994:9:
assertion failed: matches!(state.redraw_state, RedrawState :: Queued(_) | RedrawState ::
    WaitingForEstimatedVBlankAndQueued(_))
stack backtrace:
   0: rust_begin_unwind
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/panicking.rs:127:5
   3: niri::niri::Niri::redraw
             at ./Projects/niri/src/niri.rs:1994:9
   4: niri::niri::Niri::queue_redraw::{{closure}}
             at ./Projects/niri/src/niri.rs:1631:13
   5: calloop::loop_logic::LoopHandle<Data>::insert_idle::{{closure}}
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/calloop-0.12.4/src/loop_logic.rs:156:17
   6: <core::option::Option<F> as calloop::sources::IdleDispatcher<Data>>::dispatch
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/calloop-0.12.4/src/sources/mod.rs:597:13
   7: calloop::loop_logic::EventLoop<Data>::dispatch_idles
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/calloop-0.12.4/src/loop_logic.rs:542:13
   8: calloop::loop_logic::EventLoop<Data>::dispatch
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/calloop-0.12.4/src/loop_logic.rs:560:9
   9: calloop::loop_logic::EventLoop<Data>::run
             at ./.cargo/registry/src/index.crates.io-6f17d22bba15001f/calloop-0.12.4/src/loop_logic.rs:596:13
  10: niri::main
             at ./Projects/niri/src/main.rs:249:5
  11: core::ops::function::FnOnce::call_once
             at /rustc/82e1608dfa6e0b5569232559e3d385fea5a93112/library/core/src/ops/function.rs:250:5
2024-01-29T18:10:15.297033Z DEBUG niri::watcher: exiting watcher thread for /home/xen/.config/niri/config.kdl
```

axtloss:
> I updated to the latest Smithay, and something I just noticed that might be new after the update: "flipped-90" and "flipped-270" are both vertically and horizontally flipped relative to their normal counterparts, whereas "flipped" and "flipped-180" are only horizontally flipped. In sway each flip is only horizontally flipped. @ cmeissl could this be something that has regressed?

I am able to reproduce this one a build that does not have the updated smithay yet

YaLTeR:
Oh wow, thanks for that log. That is a part that is definitely not supposed to crash like that.

YaLTeR:
Do you happen to remember what you did right before you got that crash?

axtloss:
I edited the configuration to change the transform value for a screen, if it matters, the open application were firefox, foot and emacs

YaLTeR:
I'm really not sure how that could've happened but I doubt the new transform code has anything to do with it, so I won't block this PR on that crash. However, I'll wait a bit more so that we can investigate the flipped-90/270 thing, and I will also once again check other compositors to settle on the clockwise vs. counter-clockwise.

cmeissl:
> I'm really not sure how that could've happened but I doubt the new transform code has anything to do with it, so I won't block this PR on that crash. However, I'll wait a bit more so that we can investigate the flipped-90/270 thing, and I will also once again check other compositors to settle on the clockwise vs. counter-clockwise.

Can you give https://github.com/Smithay/smithay/pull/1311 a try?

YaLTeR:
Alright, now with the extra fix from @cmeissl everything seems to work fine.

YaLTeR:
Rotation direction wise:
* sway is clockwise by default
* hyprland is counter-clockwise by default

So I guess if there's no "standard" we can stick with counter-clockwise, as that makes the types match.

YaLTeR:
Thanks @axtloss and @cmeissl!

PR #138, opened 2024-01-28
fix plane artifacts

cmeissl:
depends on https://github.com/Smithay/smithay/pull/1297

@kchibisov does this fix the issues you mentioned?


kchibisov:
>@kchibisov does this fix the issues you mentioned?

No. My issue is likely something else, since once I exit niri I can't actually start my greater and I fail to set crtc. I think it's just a driver bug.

cmeissl:
Okay, just to add something that surprised me. sway does not seem to clear planes on session resume, so when switching from a compositor that used overlay planes, but did not clear it on session suspend it will be left over and still show up in sway.

YaLTeR:
Thanks

PR #143, opened 2024-01-29
fix output transform

cmeissl:
~~based on https://github.com/YaLTeR/niri/pull/138~~
~~depends on https://github.com/Smithay/smithay/pull/1307~~

includes an additional fix for a race in popup surface re-use I encountered in niri and anvil


YaLTeR:
Thanks

PR #146, opened 2024-01-30
Add actions to move the active workspace to another monitor

Trundle:
I didn't add the new actions to the default config because I don't have good default bindings. I personally use Mod+Ctrl+Alt, but that clashes with `Alt` as mod for nested Niri and also it's three modifiers at the same time. Unfortunately, this also makes the new actions a bit hard to discover, as the default config also seems to serve as documentation about available features.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -1335,6 +1335,42 @@ impl<W: LayoutElement> Layout<W> {
         }
     }
 
+    pub fn move_workspace_to_output(&mut self, output: &Output) {
+        let MonitorSet::Normal {
+            monitors,
+            active_monitor_idx,
+            ..
+        } = &mut self.monitor_set
+        else {
+            return;
+        };
+
+        let current = &mut monitors[*active_monitor_idx];
+        if current.active_workspace_idx == current.workspaces.len() - 1 {
+            // Insert a new empty workspace.
+            let ws = Workspace::new(current.output.clone(), current.options.clone());
+            current.workspaces.push(ws);
+        }
+        let mut ws = current.workspaces.remove(current.active_workspace_idx);
+        current.active_workspace_idx = current.active_workspace_idx.saturating_sub(1);
+        current.workspace_switch = None;
+        current.clean_up_workspaces();
+
+        ws.set_output(Some(output.clone()));
+        ws.original_output = OutputId::new(output);
+
+        let target_idx = monitors
+            .iter()
+            .position(|mon| &mon.output == output)
+            .unwrap();
+        let target = &mut monitors[target_idx];
+        target.workspaces.insert(target.active_workspace_idx, ws);
+        target.workspace_switch = None;
+        target.clean_up_workspaces();
+
+        *active_monitor_idx = target_idx;
+    }
+
     pub fn set_fullscreen(&mut self, window: &W, is_fullscreen: bool) {
         match &mut self.monitor_set {
             MonitorSet::Normal { monitors, .. } => {
@@ -1737,6 +1773,7 @@ mod tests {
         SetColumnWidth(#[proptest(strategy = "arbitrary_size_change()")] SizeChange),
         SetWindowHeight(#[proptest(strategy = "arbitrary_size_change()")] SizeChange),
         Communicate(#[proptest(strategy = "1..=5usize")] usize),
+        MoveWorkspaceToOutput(#[proptest(strategy = "1..=5u8")] u8),
     }
 
     impl Op {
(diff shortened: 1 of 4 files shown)


src/layout/mod.rs
```diff
@@ -1335,6 +1335,39 @@ impl<W: LayoutElement> Layout<W> {
...
+        };
+
+        let current = &mut monitors[*active_monitor_idx];
+        if current.active_workspace_idx == current.workspaces.len() - 1 {
+            // Insert a new empty workspace.
+            let ws = Workspace::new(current.output.clone(), current.options.clone());
+            current.workspaces.push(ws);
+        }
+        let mut ws = current.workspaces.remove(current.active_workspace_idx);
+        current.active_workspace_idx = current.active_workspace_idx.saturating_sub(1);
```
YaLTeR, line 1355: I suspect that we want a `current.workspace_switch = None;` here followed by a `current.clean_up_workspaces();`. In case we hit that action during a workspace switch.

src/layout/mod.rs
```diff
@@ -1335,6 +1335,39 @@ impl<W: LayoutElement> Layout<W> {
...
+        let view_size = output_size(output);
+        let working_area = compute_working_area(output, self.options.struts);
+        ws.set_view_size(view_size, working_area);
+
+        let target_idx = monitors
+            .iter()
+            .position(|mon| &mon.output == output)
+            .unwrap();
+        let target = &mut monitors[target_idx];
+        target.workspaces.insert(target.active_workspace_idx, ws);
```
YaLTeR, line 1366: This also needs a `target.workspace_switch = None;` followed by clean up workspaces.

YaLTeR (commented):
Thanks, this looks good apart from needing to disable the workspace switch.

> Unfortunately, this also makes the new actions a bit hard to discover, as the default config also seems to serve as documentation about available features.

Could you add a comment into the default config about these actions? Somewhere around other monitor/workspace actions.

src/layout/mod.rs
```diff
@@ -1335,6 +1335,39 @@ impl<W: LayoutElement> Layout<W> {
...
+            // Insert a new empty workspace.
+            let ws = Workspace::new(current.output.clone(), current.options.clone());
+            current.workspaces.push(ws);
+        }
+        let mut ws = current.workspaces.remove(current.active_workspace_idx);
+        current.active_workspace_idx = current.active_workspace_idx.saturating_sub(1);
+
+        let view_size = output_size(output);
+        let working_area = compute_working_area(output, self.options.struts);
+        ws.set_view_size(view_size, working_area);
```
YaLTeR, line 1359: Oh yeah and you need to update `ws.output` and `ws.original_output` to `output`. That should fix the failing tests.

Trundle: Would `ws.set_output(Some(output.clone())` work? It would take care of computing the working area and also sends output leave and enter events.

src/layout/mod.rs
Trundle: Added, but I think the `clean_up_workspaces` is superfluous: whole workspaces are moved, so at least for `current`, it's not possible that a new empty workspace will exist. At least not without some other window/column moving actions being executed concurrently, and I don't think that's possible?

src/layout/mod.rs
YaLTeR: Yes, that's even better. I guess you could check the code that adds or removes monitors because that's the other place that moves workspaces between monitors.

You will still need to update `ws.original_output` though, since that code doesn't (and shouldn't), but this code should, since the user explicitly asked the workspace to move to a different output.

src/layout/mod.rs
YaLTeR: It's possible:
1. Start a workspace switch on one monitor
2. Some window closes on the same monitor, maybe even on a different workspace entirely. Since a workspace switch is ongoing, the workspaces aren't cleaned up
3. The user moves a workspace, either from this monitor, or to this monitor
4. We cancel the workspace switch here
5. If we don't clean up workspaces right away, that workspace from step 2 remains not cleaned up

Trundle: Thanks for the explanation! I didn't connect that the cleanup happens when the animation is done.

src/layout/mod.rs
Trundle: Added it

src/layout/mod.rs
Trundle: Changed and several test runs with `RUN_SLOW_TESTS=1` are happy locally

YaLTeR:
Thanks! Everything seems to work fine.

One thing I'm no completely decided on is whether it maybe makes more sense to put the new workspace after, rather than before, the previously active workspace. (It will add some annoying edge cases in the code.) I'll need to use this for a bit and get a feel for it to see if I have any preference.

YaLTeR:
I decided to tweak it to put the new workspace after the current one instead of before. This feels a bit better to me, and also it makes moving back and forth idempotent in more cases.

PR #147, opened 2024-01-30
Allow selecting windows in screenshot UI

tversteeg:
When the screenshot UI is opened, this PR allows you to select a window region by clicking on a window.

The region is not shown when dragging or when hovering over an already selected window.

Not tested on multiple outputs since I only have 1 screen. I also had to refactor the code somewhat since it now needs to draw in both open and closed states.

[output.webm](https://github.com/YaLTeR/niri/assets/2574730/c10ea323-92bf-4024-80ea-8b777ed259fc)

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -21,6 +21,7 @@ use smithay::utils::{Logical, Point, SERIAL_COUNTER};
 use smithay::wayland::pointer_constraints::{with_pointer_constraint, PointerConstraint};
 use smithay::wayland::tablet_manager::{TabletDescriptor, TabletSeatTrait};
 
+use crate::layout::LayoutElement;
 use crate::niri::State;
 use crate::screenshot_ui::ScreenshotUi;
 use crate::utils::{center, get_monotonic_time, spawn};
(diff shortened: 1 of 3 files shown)


YaLTeR:
Pretty cool! So I wonder, when do you want this window region screenshot, over a proper window screenshot (i.e. Alt-PrtSc)? The proper window screenshot will include just the window, with CSD and any popups if they stick out.

image

It might be better to have a separate mode where you can click on a window and it takes a proper window screenshot. Niri doesn't quite have a full-blown toolkit like GNOME Shell for this, but maybe if you press "W" then instead of the area, hovering a window will highlight it, and clicking will take a window screenshot of that window?

tversteeg:
For me, this window mode is mainly so I can have the screenshot utility under a single button, and deciding what part of the screen I want to take after taking the actual screenshot.

I think having a separate mode would be optimal, and this could be an in-between solution covering 90% of the use case.

I also prefer the current solution where you can click on a target which is highlighted or drag manually if you want a specific region. As a new user, I would never remember that there's a shortcut inside the screenshot mode for just selecting a window.

YaLTeR:
> deciding what part of the screen I want to take after pressing the screenshot button

We can store screenshots of all visible windows at the time of pressing the screenshot button, just like the screen currently.

> As a new user, I would never remember that there's a shortcut inside the screenshot mode for just selecting a window.

Well, could put a small help panel at the bottom.

> I also prefer the current solution where you can click on a target which is highlighted or drag manually if you want a specific region.

I think it's also kinda confusing that while you want to select a region, windows get highlighted under your cursor.

tversteeg:
> We can store screenshots of all visible windows at the time of pressing the screenshot button, just like the screen currently.

I don't understand this, how will this enable me to select what I want afterward?

> I think it's also kinda confusing that while you want to select a region, windows get highlighted under your cursor.

The windows don't get highlighted when you're actively selecting something.

YaLTeR:
> I don't understand this, how will this enable me to select what I want afterward?

Same as it works on your video. Just instead of cutting it out of the monitor screenshot, it'll save the stored window screenshot.

YaLTeR:
Right now screenshot UI stores a texture per output; you'd also store textures for windows visible on this output. Then when you select a window, you'll pick that window's texture and save it to file.

PR #167, opened 2024-02-03
add move-window-left/right commands

exoticorn:
These move the active window alternately into and out of columns. As such, they can be thought of as alternative options to the `consume-window-into-column` and `expel-window-from-column` commands.

In the expel case, they are strictly more flexible, as they can expel the active window both to the left and right.

In the consume case, they differ in that they move the active window into an adjacent column instead of pulling a window to the right into the active column. One example where this is more convenient is when opening two new windows (say terminals) and putting them into one column.

This PR does not add default bindings for these commands. Bindings can be discussed once these commands have been found useful in practice.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -459,6 +459,8 @@ pub enum Action {
     MoveWindowUp,
     MoveWindowDownOrToWorkspaceDown,
     MoveWindowUpOrToWorkspaceUp,
+    ConsumeOrExpelWindowLeft,
+    ConsumeOrExpelWindowRight,
     ConsumeWindowIntoColumn,
     ExpelWindowFromColumn,
     CenterColumn,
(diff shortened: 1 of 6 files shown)


YaLTeR:
Looks good, could you add the new actions to the tests at the bottom of layout/mod.rs? Add them to `Op` and to two `every_op` arrays in the tests.

exoticorn:
Done. The tests did uncover that I forgot to check for the workspace being empty. That's fixed now as well.

YaLTeR:
Seems to work fine, I'll give it a bit more testing. Do you mind if I rename the actions to `consume-or-expel-window-left/right`? I think that's clearer, since `move` actions don't consume into columns elsewhere, plus it allows adding other similarly named actions in the future (e.g. `consume-column-left/right` for the whole column).

exoticorn:
Feel free to rename. I was unsure about the naming myself. I arrived from the point of view that not jumping over columns is probably the more useful default when moving single windows. But your argument of being more consistent in the naming holds a lot of weight as well.

YaLTeR:
Thanks!

PR #184, opened 2024-02-11
Implement DRM leasing

### LINKED ISSUE #178: Implement DRM leasing
I don't have any VR hardware to do it myself unfortunately.

Ref. https://github.com/YaLTeR/niri/discussions/177

Supreeeme:
Tested with Monado and SteamVR (with some cage patches).
Closes #178

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -41,6 +41,9 @@ use smithay::reexports::wayland_protocols;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
 use smithay::utils::DeviceFd;
 use smithay::wayland::dmabuf::{DmabufFeedback, DmabufFeedbackBuilder, DmabufGlobal};
+use smithay::wayland::drm_lease::{
+    DrmLease, DrmLeaseBuilder, DrmLeaseRequest, DrmLeaseState, LeaseRejected,
+};
 use smithay_drm_extras::drm_scanner::{DrmScanEvent, DrmScanner};
 use smithay_drm_extras::edid::EdidInfo;
 use wayland_protocols::wp::linux_dmabuf::zv1::server::zwp_linux_dmabuf_feedback_v1::TrancheFlags;
@@ -105,7 +108,7 @@ type GbmDrmCompositor = DrmCompositor<
     DrmDeviceFd,
 >;
 
-struct OutputDevice {
+pub struct OutputDevice {
     token: RegistrationToken,
     render_node: DrmNode,
     drm_scanner: DrmScanner,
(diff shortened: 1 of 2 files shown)


src/backend/tty.rs
```diff
@@ -455,13 +494,17 @@ impl Tty {
             })
             .unwrap();
 
+        let drm_lease_state = DrmLeaseState::new::<State>(&niri.display_handle, &node).unwrap();
```
YaLTeR, line 497: This shouldn't `unwrap()`. If it's expected to "always" work then I guess `.context()?` it (and move as early as possible in the function so that nothing needs to be deinitialized should this fail). If it's not expected to "always" work then make it an `Option` in the `OutputDevice`.

src/handlers/mod.rs
```diff
@@ -374,3 +378,49 @@ impl ForeignToplevelHandler for State {
     }
 }
 delegate_foreign_toplevel!(State);
+
+impl DrmLeaseHandler for State {
+    fn drm_lease_state(&mut self, node: DrmNode) -> &mut DrmLeaseState {
```
YaLTeR, line 383: All these methods are guaranteed to receive a `node` that exists in our `OutputDevice`? Looking at those unwraps.

src/backend/tty.rs
```diff
@@ -583,6 +626,41 @@ impl Tty {
         );
         debug!("connecting connector: {output_name}");
 
+        let device = self.devices.get_mut(&node).context("missing device")?;
```
YaLTeR, line 629: Are any changes needed to `on_output_config_changed()`?

YaLTeR (commented):
This is very cool!

src/handlers/mod.rs
Supreeeme: The Smithay docs do not make this very clear but I would be inclined to believe so. Looking at the Smithay source, it seems that the only nodes that are used are created from the node used when creating a DrmLeaseState, based on [the global creation step](https://github.com/Smithay/smithay/blob/91e61f13501f21d66803efac947bfafed43080c5/src/wayland/drm_lease/mod.rs#L428), [the data initialization step](https://github.com/Smithay/smithay/blob/91e61f13501f21d66803efac947bfafed43080c5/src/wayland/drm_lease/mod.rs#L714), and the [request creation](https://github.com/Smithay/smithay/blob/91e61f13501f21d66803efac947bfafed43080c5/src/wayland/drm_lease/mod.rs#L762).

src/backend/tty.rs
Supreeeme: No, but I switched `non_desktop_connectors` to be a HashSet so that when the config is reloaded the same connector isn't added multiple times.

YaLTeR:
Thanks

RangerMauve:
Sorry to necrobump, but is there something that needs to be added to the config to enable this? I'm having trouble getting the Monado OpenXR runtime working with it.

YaLTeR:
No, should work out of the box (unless it broke somehow)

Supreeeme:
I still use it all the time, it still works. Make sure your headset is being marked as non desktop by the kernel.

RangerMauve:
Great, ty!

@Supreeeme Any tips on how to do that with Wayland? I've only seen examples for xrandr and my Google-fu hasn't worked out so far. :sweat_smile:

Supreeeme:
Check out the Monado discord, someone there just posted a script for it: https://discord.com/invite/tNqyfBHy
(it's also bridged to matrix if that's more your thing)

RangerMauve:
TY gonna try it out this Friday :bow:

PR #185, opened 2024-02-11
Basic Touch Support

cmeissl:
~~decided to give it a try despite being unfinished in smithay.
lacks testing, proper output selection, doesn't work properly with output transform,...
but hey, touch seems to work so far on my laptop :)~~

~~based on https://github.com/Smithay/smithay/pull/1326~~ merged

So, with the touch refactor merged in smithay it is time to finally finish basic touch support in niri

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -8,7 +8,7 @@ use smithay::backend::input::{
     GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _, GestureSwipeUpdateEvent as _,
     InputBackend, InputEvent, KeyState, KeyboardKeyEvent, PointerAxisEvent, PointerButtonEvent,
     PointerMotionEvent, ProximityState, TabletToolButtonEvent, TabletToolEvent,
-    TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState,
+    TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState, TouchEvent,
 };
 use smithay::backend::libinput::LibinputInputBackend;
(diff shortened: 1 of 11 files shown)


src/input.rs
```diff
@@ -171,6 +172,9 @@ impl State {
             let desc = TabletDescriptor::from(&device);
             tablet_seat.add_tablet::<Self>(&self.niri.display_handle, &desc);
         }
+        if device.has_capability(DeviceCapability::Touch) && self.niri.seat.get_touch().is_none() {
+            self.niri.seat.add_touch();
```
YaLTeR, line 176: Do we need to remove it in `on_device_removed()`?

src/input.rs
```diff
@@ -1354,6 +1362,94 @@ impl State {
...
+            return;
+        };
+        let Some(touch_location) = self.compute_touch_location(&evt) else {
+            return;
+        };
+        let serial = SERIAL_COUNTER.next_serial();
+        let under = self
+            .niri
+            .surface_under_and_global_space(touch_location)
+            .map(|under| under.surface);
```
YaLTeR, line 1394: We probably want to activate window / output under touch here, just like we do for mouse and tablet clicks?

cmeissl: I initially intended on adding that through bindings on global gestures. But I guess for now we can just active the window similar to pointer button. One important difference is that I added a check if there is a touch grab and in this case not activate a window. I added that to prevent stealing the window on multiple touches.

YaLTeR (approved):
Don't have a touch device to test, but code wise looks good.

YaLTeR:
Thanks!

PR #189, opened 2024-02-13
Use nearest scaling for integer-upscaled surfaces

YaLTeR:
Doesn't currently work for the cursor plane, as a similar logic would need to somehow be threaded inside there.

For the primary plane I'm not even sure what sets the scaling filter, seems to use linear for me.

@kchibisov please give this a try.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3005,7 +3005,7 @@ checksum = "e6ecd384b10a64542d77071bd64bd7b231f4ed5940fba55e98c3de13824cf3d7"
 [[package]]
 name = "smithay"
 version = "0.3.0"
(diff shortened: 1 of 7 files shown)


YaLTeR:
I've got confirmation from mutter folks that it's currently impossible to control the scaling for scanout. They're planning to add this as part of the new kms color transformation API.

24fpsDaVinci:
> I've got confirmation from mutter folks that it's currently impossible to control the scaling for scanout. They're planning to add this as part of the new kms color transformation API.

so we have to wait for mutter? this PR does not work at all?

YaLTeR:
We have to wait for the kernel to add the necessary API to do this properly. In the meantime I suppose I could add this as a debug flag. I'll need to redo the PR from scratch though.

PR #191, opened 2024-02-13
Add the beginnings of window rules

YaLTeR:
For now just `default-column-width` and `open-on-output`. Other interesting rules will need some refactoring, and rules that can apply dynamically will also need some Smithay additions.

@zkat I think this should partially cover your use-case of opening some applications on specific outputs. Mind giving it a try?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2130,6 +2130,7 @@ dependencies = [
  "knuffel",
  "miette",
  "niri-ipc",
+ "regex",
  "smithay",
  "tracing",
  "tracy-client",
--- a/niri-config/Cargo.toml
+++ b/niri-config/Cargo.toml
@@ -12,6 +12,7 @@ bitflags.workspace = true
 knuffel = "3.2.0"
 miette = "5.10.0"
 niri-ipc = { version = "0.1.1", path = "../niri-ipc" }
+regex = "1.10.3"
 smithay = { workspace = true, features = ["backend_libinput"] }
 tracing.workspace = true
 tracy-client.workspace = true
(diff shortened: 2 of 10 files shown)


zkat:
This looks really good! Although I only have one monitor so I can't really test different outputs. What I was looking for was different workspaces

YaLTeR:
How should it work considering niri has dynamic workspaces? I.e. if a window opens on workspace 2 while workspace 1 is still empty, workspace 1 will be removed and workspace 2 will become workspace 1.

zkat:
ohhhhh. I thought niri allowed workspace gaps if you were explicit about it. Hmmmm.

My gut feeling is it might be nice to be able to optionally create non-dynamic workspaces, for folks to be able to preconfigure things like this?

YaLTeR:
> My gut feeling is it might be nice to be able to optionally create non-dynamic workspaces, for folks to be able to preconfigure things like this?

I'll need to think more about how to do it. Dynamic workspaces are very well suited for multiple monitors, since you can move them around, and if a monitor disconnects you can just move its workspaces to another one, and then bring back later. There's no static workspace index collisions to worry about.

The case is "at startup, I want these several windows to follow this workspace and order arrangement". One way to do it is to keep a minimum number of workspaces open if there's a window rule to put a window on a specific workspace. But then there are questions like, you probably only want it on one output and not on all of them? What happens if you disconnect that one primary output and then a configured window opens? Ideally this quite common case should be easy to configure (i.e. does not require setting multiple options in different sections) and the configuration should also make sense and not break any other settings.

PR #195, opened 2024-02-14
input: enable configuring of trackpoint devices

okvik:
This adds the ability to configure trackpoint type input devices in a similar manner it can be done for regular mice, i.e. using the same set of configuration options.

I've tested this on ThinkPad X1 Extreme Gen. 5 running NixOS, and I suppose it should likely work elsewhere where there's basic trackpoint support.

Let me know if I missed anything.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -57,6 +57,8 @@ pub struct Input {
     #[knuffel(child, default)]
     pub mouse: Mouse,
     #[knuffel(child, default)]
+    pub trackpoint: Trackpoint,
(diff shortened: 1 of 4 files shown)


okvik:
> All three settings apply to trackpoints, right?

Yes. The acceleration settings work just the same. And the natural scrolling also applies because you can scroll with it by holding the middle (trackpoint) button while operating the nipple.

> Also, please add it into the default config below mouse, similarly commented out.

Will do.

YaLTeR:
Thanks!

PR #217, opened 2024-02-20
Move to --session CLI flag; add systemd feature to make it optional

YaLTeR:
Intended to help various systemd-less environments and BSD.

@jbeich I see you're maintaining the FreeBSD package, are these changes good?

@QaidVoid do these changes fix your use-case?

Ref. https://github.com/YaLTeR/niri/discussions/211


QaidVoid:
Yeah, it works. Thanks.

PR #230, opened 2024-02-24
handle buffer offset

cmeissl:
tracking of https://github.com/Smithay/smithay/pull/1341

### DIFF
--- a/src/handlers/compositor.rs
+++ b/src/handlers/compositor.rs
@@ -1,7 +1,7 @@
 use std::collections::hash_map::Entry;
 
 use smithay::backend::renderer::utils::{on_commit_buffer_handler, with_renderer_surface_state};
-use smithay::input::pointer::CursorImageStatus;
+use smithay::input::pointer::{CursorImageStatus, CursorImageSurfaceData};
 use smithay::reexports::calloop::Interest;
 use smithay::reexports::wayland_server::protocol::wl_buffer;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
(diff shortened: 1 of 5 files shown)


cmeissl:
@YaLTeR This is now just waiting for the review of https://github.com/Smithay/smithay/pull/1341

src/handlers/compositor.rs
```diff
@@ -297,13 +297,42 @@ impl CompositorHandler for State {
             &self.niri.cursor_manager.cursor_image(),
             CursorImageStatus::Surface(s) if s == &root_surface
         ) {
+            with_states(surface, |states| {
```
YaLTeR, line 300: Guess this should only happen if `surface == root_surface`

src/handlers/compositor.rs
```diff
@@ -297,13 +297,42 @@ impl CompositorHandler for State {
...
             // FIXME: granular redraws for cursors.
             self.niri.queue_redraw_all();
             return;
         }
 
         // This might be a DnD icon surface.
-        if self.niri.dnd_icon.as_ref() == Some(&root_surface) {
+        if matches!(&self.niri.dnd_icon, Some(icon) if icon.surface == root_surface) {
+            let dnd_icon = self.niri.dnd_icon.as_mut().unwrap();
+            with_states(&dnd_icon.surface, |states| {
```
YaLTeR, line 326: Guess this can only change if surface == root_surface

src/handlers/compositor.rs
cmeissl: right, while unlikely the surface could be part of a surface tree

src/input/mod.rs
```diff
@@ -2754,8 +2750,8 @@ mod tests {
         // The key_code we pick is arbitrary, the only thing
         // that matters is that they are different between cases.
 
-        let close_key_code = close_keysym.into();
-        let close_key_event = |suppr: &mut HashSet<u32>, mods: ModifiersState, pressed| {
+        let close_key_code =  Keycode::from(close_keysym.raw() + 8u32);
```
cmeissl, line 2753: Not sure about the offset here tbh

YaLTeR:
Thanks

PR #234, opened 2024-02-28
Correctly handle parsing of Binds and DefaultColumnWidth

### LINKED ISSUE #233: Putting several binds on the same key will show incorrect values in hotkey overlay
The main cause of this issue is that when creating two binds of the same key, niri doesn't complain at all. Notably, it even remembers both binds after parsing. It then handles this inconsistently. The code is not necessarily broken, but the root cause is that this code is written under the incorrect assumptions that every key corresponds to at most one `Bind` (potentially zero, if unbound of course), and that every bind contains exactly one `Action`.

The hotkey overlay informs a user of the configured hotkeys, nonexhaustively. Mainly, it shows only one keybind per action. To do this, it searches for each action separately.

The actual keyboard handler will search for the key, and perform the first action it finds. This is not necessarily the keybind that the hotkey overlay shows for this action, which is fine, but what's not fine is that it's also not necessarily the action the hotkey overlay shows for this keybind, or even necessarily any action at all.

Here's a minimal config displaying two behaviours that i don't think are very intuitive.

```kdl
binds {
    Mod+A { quit skip-confirmation=true; }
    Mod+A { spawn "alacritty"; }
    Mod+Y {}
    Mod+Y { spawn "fuzzel"; }
    Mod+Q { quit; }
}
```

Pretend you are a user who *does not* remember their config. The overlay is a quick reminder of what each button does, right?

image

Niri does elide "redundant" keybinds. It's expected that there may be more keybinds not listed.

This user, who knows how the hotkey overlay works, knows that at least the following binds exist:

- `Mod+Q` will exit niri
- `Mod+A` will spawn a terminal
- `Mod+Y` will spawn an application launcher

`Mod+Q` really does work as they expect! It will close niri. Nothing strange there.

They want to get something done, so they press `Mod+Y` and nothing happens. Niri does not perform the action it promised. They cannot launch applications.

This is an advanced user who is comfortable with the terminal, so they figure they can figure out what's happening here. So they press `Mod+A`, as the hotkey overlay told them. Niri performs the wrong action. Our user is confused as to why they have been sent back to their login manager.

A related problem stemming from the same root cause is that you can currently bind multiple actions in the same block, like so:

```kdl
binds {
    Mod+A { spawn "alacritty"; spawn "fuzzel"; }
}
```

In this case, niri just ignores the second action in the block completely. It never appears in hotkey overlays, and it is never executed. The same interactions with duplicate binds can still occur, though.

### System Information

* niri version: `niri 0.1.2 (niri-flake at 494e98c)`

sodiboo:
Previously, for `default-column-width` as well as keybinds, you could specify multiple children. Subsequent children after the first one were ignored. Now, this emits an error.

Additionally, i've also also made sure you can't register duplicate keybinds. This will emit an error, too. Previously, this resulted in some very strange behaviour, as detailed in the issue that this will fix #233

However, there is still the question of how  `Mod` and `Super`/`Alt` should interact, if they result in duplicate keybinds. Currently, this edge case just results in the old behaviour. I haven't handled it. Maybe binds that differs only by `Mod` vs `Super`/`Alt` should always count as duplicates? Should some new property be added to configure this behaviour? i.e. `Super+Alt+L nested="Ctrl+Alt+L" { spawn "swaylock"; }`, `Super+Alt+L nested=null { spawn "swaylock"; }`?

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1033,8 +1033,31 @@ where
                     if seen_keys.insert(bind.key) {
                         binds.push(bind);
                     } else {
-                        // TODO: add context of the source span for the original bind?
-                        // i couldn't figure out how to easily do this
+                        // ideally, this error should point to the previous instance of this keybind
+                        //
+                        // i (sodiboo) have tried to implement this in various ways:
+                        // miette!(), #[derive(Diagnostic)]
+                        // DecodeError::Custom, DecodeError::Conversion
+                        // nothing seems to work, and i suspect it's not possible.
+                        //
+                        // DecodeError is fairly restrictive.
+                        // even DecodeError::Custom just wraps a std::error::Error
+                        // and this erases all rich information from miette. (why???)
+                        //
+                        // why does knuffel do this?
+                        // from what i can tell, it doesn't even use DecodeError for much.
+                        // it only ever converts them to a Report anyways!
+                        // https://github.com/tailhook/knuffel/blob/c44c6b0c0f31ea6d1174d5d2ed41064922ea44ca/src/wrappers.rs#L55-L58
+                        //
+                        // besides like, allowing downstream users (such as us!)
+                        // to match on parse failure, i don't understand why
+                        // it doesn't just use a generic error type
+                        //
+                        // even the matching isn't consistent,
+                        // because errors can also be omitted as ctx.emit_error.
+                        // why does *that one* especially, require a DecodeError?
+                        //
+                        // anyways if you can make it format nicely, definitely do fix this
                         ctx.emit_error(DecodeError::unexpected(
                             &child.node_name,
                             "keybind",
@@ -1066,6 +1089,14 @@ where
 
         let mut children = node.children();
 
+        // If the action is invalid but the key is fine, we still want to return something.
+        // That way, the parent can handle the existence of duplicate keybinds,
+        // even if their contents are not valid.
+        let dummy = Self {
+            key,
+            action: Action::Spawn(vec![]),
+        };
+
         if let Some(child) = children.next() {
             for unwanted_child in children {
                 ctx.emit_error(DecodeError::unexpected(
@@ -1074,19 +1105,19 @@ where
                     "only one action is allowed per keybind",
                 ));
             }
-            let action = Action::decode_node(child, ctx)?;
-            Ok(Self { key, action })
+            match Action::decode_node(child, ctx) {
+                Ok(action) => Ok(Self { key, action }),
+                Err(e) => {
+                    ctx.emit_error(e);
+                    Ok(dummy)
+                }
+            }
         } else {
             ctx.emit_error(DecodeError::missing(
                 node,
                 "expected an action for this keybind",
             ));
-            // This value isn't used. Decoding will not succeed, because we just emitted an error.
-            // Even if it somehow were used, spawn without args is a no-op.
-            Ok(Self {
-                key,
-                action: Action::Spawn(vec![]),
-            })
+            Ok(dummy)
         }
     }
 }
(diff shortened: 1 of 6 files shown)


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
YaLTeR, line 1: Is this necessary? I don't see it in cosmic-comp for example. Is it gonna cause any warnings if nix is not installed but direnv is?

niri-config/src/lib.rs
```diff
@@ -944,6 +945,145 @@ where
...
+        for child in node.children() {
+            match Bind::decode_node(child, ctx) {
+                Err(e) => {
+                    ctx.emit_error(e);
+                }
+                Ok(bind) => {
+                    if seen_keys.insert(bind.key) {
+                        binds.push(bind);
+                    } else {
+                        // TODO: add context of the source span for the original bind?
```
YaLTeR, line 1029: If you leave this as is, then change TODO to FIXME.

niri-config/src/lib.rs
```diff
@@ -944,6 +945,145 @@ where
...
+            ctx.emit_error(DecodeError::missing(
+                node,
+                "expected an action for this keybind",
+            ));
+            // This value isn't used. Decoding will not succeed, because we just emitted an error.
+            // Even if it somehow were used, spawn without args is a no-op.
+            Ok(Self {
+                key,
+                action: Action::Spawn(vec![]),
+            })
```
YaLTeR, line 1082: Why not just return Err(DecodeError) with the error right above?

YaLTeR (commented):
Looks good.

> how Mod and Super/Alt should interact, if they result in duplicate keybinds

I think it's fine to leave them as is, i.e. they can both exist and the first one will trigger in case they conflict. Throwing a parse error conditioned on nested vs. TTY backend sounds worse than just bind working a different way on nested vs. TTY.

> Should some new property be added to configure this behaviour?

Nah, don't need new properties here.

niri-config/src/lib.rs
sodiboo: I don't know how i thought to write a comment, but did not think to write in the comment why i did it this way. This is done like so, such that `Binds` will have a happy path on return, adding dummy data to the actual binds, but crucially remembering that the `key` was parsed correctly. In other words, this ensures that "duplicate keybind" is emitted even when there is an empty block. I did miss the case where there is a block with an invalid action, though.

.envrc
sodiboo: I guess it's not necessary, but my workflow depends on it. I use an extension to vscode that loads `.envrc`, and i can't easily work without it (vscode doesn't really launch "inline" with the terminal environment, so launching from `nix develop` does not do the trick as it would with i.e. nvim). It could just as well be gitignored, though. I don't see it causing an issue without `nix-direnv`, as i tried just now making a dummy directory with and `.envrc` of `use AAAAAAAAAAAAAAAAA` and it doesn't complain at all. firstly; it asks before even trying to load anyway, but when it does load, it seems to just succeed with no additional notes.

YaLTeR: I guess you can add .direnv and .envrc to your global gitignore, then it doesn't have to be added to individual projects.

sodiboo: hm. that's... fair enough. i'll admit, i do not have extensive knowledge of git, and was genuinely unaware of this. always thought people just manually made sure to keep editor metadata and such untracked or something. should i remove that commit from my branch then, and force push, or is there some "cleaner"(?) way to do it?

niri-config/src/lib.rs
sodiboo: i think returning Err is not fine, because then we also suppress the  "duplicate keybind". my latest commit documents why we do this a lot better, though. i agree, the previous comment was unhelpful at best and did not at all communicate the intent

YaLTeR (approved):
Seems to work. Thanks!

PR #236, opened 2024-02-28
reduce memory allocations

cmeissl:
testing of https://github.com/Smithay/smithay/pull/1346


YaLTeR:
Tested running a few clients and a game, plus Tracy on another monitor.

Yellow = this PR.

image

image

image

cmeissl:
thx for testing, looks like some minimal improvement (probably from the bind happy path change). I did not expect much improvement on wall clock time on desktop anyway with the PR mostly only reducing allocations.
So good to know that there seems so be no regression on desktop :)

YaLTeR:
Updated Smithay to this in 7e0d3d31f719615a3da99325d5797a96ae1fa5d2

PR #238, opened 2024-03-03
implement tablet mode (disabling keyboard and mouse/touch pad)

exoticorn:
Can be triggered by a dedicated hardware switch, or by a toggle action.

Open question: Can we reliably distinguish internal from external devices? If yes, we might only want to disable internal devices.


src/input.rs
```diff
@@ -1572,6 +1598,13 @@ fn should_intercept_key(
         disable_power_key_handling,
     );
 
+    if tablet_mode {
+        if final_action != Some(Action::ToggleTabletMode) {
+            suppressed_keys.insert(key_code);
```
YaLTeR, line 1603: Doesn't this need to remove on key release?

src/input.rs
```diff
@@ -697,6 +703,10 @@ impl State {
     }
 
     fn on_pointer_motion<I: InputBackend>(&mut self, event: I::PointerMotionEvent) {
+        if self.niri.tablet_mode {
```
YaLTeR, line 706: I'm thinking, maybe handle all of these one level above? I.e. before dispatching in `process_input_event()`, do:
```
if self.niri.tablet_mode && matches!(event, ...) {
    return;
}
```

src/niri.rs
```diff
@@ -195,6 +195,7 @@ pub struct Niri {
     pub gesture_swipe_3f_cumulative: Option<(f64, f64)>,
 
     pub lock_state: LockState,
+    pub tablet_mode: bool,
```
YaLTeR, line 198:
```suggestion
    // Tablet mode for convertibles: disable regular pointer and keyboard input.
    pub tablet_mode: bool,
```

YaLTeR (commented):
> Open question: Can we reliably distinguish internal from external devices? If yes, we might only want to disable internal devices.

Could you check what mutter does? Its tablet support isn't perfect by any means, but there's one maintainer constantly using it with a tablet.

exoticorn:
> Could you check what mutter does? Its tablet support isn't perfect by any means, but there's one maintainer constantly using it with a tablet.

As far as I can tell, mutter just sets a property called `touch-mode` but doesn't change anything input-related in tablet/touch-mode. This now makes me question this whole PR. It could very well be that disabling keyboard & touch pad is supposed to happen at a level below the compositor.

Unfortunately, the tablet switch in my convertible is not supported on linux, so I cannot verify this behavior myself. I'm inclined to shelve this PR until testing can be done on more devices.

YaLTeR:
Alright, thanks for checking. No rush, let's wait until we figure out better what's happening.

Quackdoc:
I think it's a useful feature to have regardless. I have an old convertible laptop.... somewhere I can use for testing if it would help, but it never had consistently working "tablet mode" in any DE. so I think having niri capable of doing it at the compositor level is at least useful.

YaLTeR:
> As far as I can tell, mutter just sets a property called touch-mode but doesn't change anything input-related in tablet/touch-mode. This now makes me question this whole PR. It could very well be that disabling keyboard & touch pad is supposed to happen at a level below the compositor.

So apparently that's exactly what happens on ThinkPad convertibles at least, the keyboard is disabled at a hardware level. In Mutter / Shell it just makes the on-screen keyboard pop up, and enables auto-rotation (although I've heard feedback that the auto-rotation toggle should not really be tied to the tablet mode, as it can be useful outside also).

YaLTeR:
#747 added the ability to bind commands to switch events, so you can bind a command that toggles the on-screen keyboard when going to tablet mode. I guess that's all that's needed more or less?

salman-farooq-sh:
This might be related: https://wiki.archlinux.org/title/Tablet_PC#Tablet_mode. I recently used `tablet-mode` on a cheap convertible laptop.

Issue #42, opened 2023-10-11
Support wlr-gamma-control-unstable-v1


wesbluemarine:
Is there a way, currently, to change the color temperature like in gammastep and wlsunset?

YaLTeR:
Probably needs to be implemented.

PR #240, opened 2024-03-03
Adds support for wlr-gamma-control-unstable-v1

phuhl:
Closes https://github.com/YaLTeR/niri/issues/42

This was harder than I thought :D 

I am new to rust, wlroots development and niri, so please be patient with me and feel free to point out stupid things I might have done.

I left a few comments on where I was unsure if something was ok like that.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -1,6 +1,7 @@
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
+use std::iter::zip;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::path::Path;
 use std::rc::Rc;
@@ -702,6 +703,11 @@ impl Tty {
             Err(err) => debug!("error setting max bpc: {err:?}"),
         }
 
+        // Reset gamma to linear in case it was set before.
+        if let Err(err) = set_gamma_for_crtc(&device.drm, crtc, None) {
+            debug!("error resetting gamma: {err:?}");
+        }
+
         let surface = device
             .drm
             .create_surface(crtc, mode, &[connector.handle()])?;
@@ -1244,63 +1250,28 @@ impl Tty {
         }
     }
 
-    pub fn get_gamma_length(&self, output: &Output) -> Option<u32> {
+    pub fn get_gamma_size(&self, output: &Output) -> anyhow::Result<u32> {
         let tty_state = output.user_data().get::<TtyOutputState>().unwrap();
-        let device = self.devices.get(&tty_state.node)?;
-        let crtc_info = device.drm.get_crtc(tty_state.crtc.clone()).ok()?;
-        return Some(crtc_info.gamma_length());
+        let crtc = tty_state.crtc;
+
+        let device = self
+            .devices
+            .get(&tty_state.node)
+            .context("missing device")?;
+        let info = device
+            .drm
+            .get_crtc(crtc)
+            .context("error getting crtc info")?;
+        Ok(info.gamma_length())
     }
 
-    pub fn get_gamma(&self, output: &Output) -> Option<Vec<u16>> {
-        let tty_state = output.user_data().get::<TtyOutputState>().unwrap();
-        let device = self.devices.get(&tty_state.node)?;
-        let crtc_info = device.drm.get_crtc(tty_state.crtc.clone()).ok()?;
-        let crtc = tty_state.crtc.clone();
-        let gamma_length = crtc_info.gamma_length() as usize;
-        let mut red = vec![0; gamma_length];
-        let mut green = vec![0; gamma_length];
-        let mut blue = vec![0; gamma_length];
-        if let Err(err) = device.drm.get_gamma(crtc, &mut red, &mut green, &mut blue) {
-            warn!("error getting gamma for crtc {crtc:?}: {err:?}");
-            return None;
-        }
-        red.extend(green);
-        red.extend(blue);
-        return Some(red);
-    }
-
-    pub fn set_gamma(
-        &self,
-        niri: &mut Niri,
-        output: &Output,
-        ramp: Vec<u16>,
-    ) -> anyhow::Result<()> {
+    pub fn set_gamma(&self, output: &Output, ramp: Option<&[u16]>) -> anyhow::Result<()> {
         let tty_state = output.user_data().get::<TtyOutputState>().unwrap();
         let device = self
             .devices
             .get(&tty_state.node)
             .context("missing device")?;
-
-        let crtc_info = device.drm.get_crtc(tty_state.crtc.clone())?;
-        let crtc = tty_state.crtc.clone();
-        let gamma_length = crtc_info.gamma_length() as usize;
-
-        ensure!(ramp.len() == gamma_length * 3, "wrong gamma length");
-        let mut red = ramp.clone();
-        red.truncate(gamma_length);
-        let mut green = ramp.clone();
-        green.drain(0..gamma_length);
-        green.truncate(gamma_length);
-        let mut blue = ramp;
-        blue.drain(0..gamma_length * 2);
-        blue.truncate(gamma_length);
-
-        device
-            .drm
-            .set_gamma(crtc, &mut red, &mut green, &mut blue)
-            .context("error setting gamma")?;
-        niri.queue_redraw(output.clone());
-        Ok(())
+        set_gamma_for_crtc(&device.drm, tty_state.crtc, ramp)
     }
 
     fn refresh_ipc_outputs(&self) {
@@ -1868,6 +1839,49 @@ fn set_max_bpc(device: &DrmDevice, connector: connector::Handle, bpc: u64) -> an
     }
 
     Err(anyhow!("couldn't find max bpc property"))
+}
+
+pub fn set_gamma_for_crtc(
+    device: &DrmDevice,
+    crtc: crtc::Handle,
+    ramp: Option<&[u16]>,
+) -> anyhow::Result<()> {
+    let _span = tracy_client::span!("set_gamma_for_crtc");
+
+    let info = device.get_crtc(crtc).context("error getting crtc info")?;
+    let gamma_length = info.gamma_length() as usize;
+
+    let mut temp;
+    let ramp = if let Some(ramp) = ramp {
+        ensure!(ramp.len() == gamma_length * 3, "wrong gamma length");
+        ramp
+    } else {
+        let _span = tracy_client::span!("generate linear gamma");
+
+        // The legacy API provides no way to reset the gamma, so set a linear one manually.
+        temp = vec![0u16; gamma_length * 3];
+
+        let (red, rest) = temp.split_at_mut(gamma_length);
+        let (green, blue) = rest.split_at_mut(gamma_length);
+        let denom = gamma_length as u64 - 1;
+        for (i, ((r, g), b)) in zip(zip(red, green), blue).enumerate() {
+            let value = (0xFFFFu64 * i as u64 / denom) as u16;
+            *r = value;
+            *g = value;
+            *b = value;
+        }
+
+        &temp
+    };
+
+    let (red, ramp) = ramp.split_at(gamma_length);
+    let (green, blue) = ramp.split_at(gamma_length);
+
+    device
+        .set_gamma(crtc, red, green, blue)
+        .context("error setting gamma")?;
+
+    Ok(())
 }
 
 #[cfg(test)]
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -2,6 +2,8 @@ use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
 use std::iter::zip;
+use std::num::NonZeroU64;
+use std::os::fd::AsFd;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::path::Path;
 use std::rc::Rc;
@@ -10,6 +12,7 @@ use std::time::Duration;
 use std::{io, mem};
 
 use anyhow::{anyhow, bail, ensure, Context};
+use bytemuck::cast_slice_mut;
 use libc::dev_t;
 use niri_config::Config;
 use smithay::backend::allocator::dmabuf::Dmabuf;
@@ -165,6 +168,7 @@ struct Surface {
     name: String,
     compositor: GbmDrmCompositor,
     dmabuf_feedback: Option<SurfaceDmabufFeedback>,
+    gamma_props: Option<GammaProps>,
     /// Tracy frame that goes from vblank to vblank.
     vblank_frame: Option<tracy_client::Frame>,
     /// Frame name for the VBlank frame.
@@ -179,6 +183,13 @@ struct Surface {
 pub struct SurfaceDmabufFeedback {
     pub render: DmabufFeedback,
     pub scanout: DmabufFeedback,
+}
+
+struct GammaProps {
+    crtc: crtc::Handle,
+    gamma_lut: property::Handle,
+    gamma_lut_size: property::Handle,
+    previous_blob: Option<NonZeroU64>,
 }
 
 impl Tty {
@@ -703,8 +714,17 @@ impl Tty {
             Err(err) => debug!("error setting max bpc: {err:?}"),
         }
 
-        // Reset gamma to linear in case it was set before.
-        if let Err(err) = set_gamma_for_crtc(&device.drm, crtc, None) {
+        let mut gamma_props = GammaProps::new(&device.drm, crtc)
+            .map_err(|err| debug!("error getting gamma properties: {err:?}"))
+            .ok();
+
+        // Reset gamma in case it was set before.
+        let res = if let Some(gamma_props) = &mut gamma_props {
+            gamma_props.set_gamma(&device.drm, None)
+        } else {
+            set_gamma_for_crtc(&device.drm, crtc, None)
+        };
+        if let Err(err) = res {
             debug!("error resetting gamma: {err:?}");
         }
 
@@ -818,6 +838,7 @@ impl Tty {
             name: output_name.clone(),
             compositor,
             dmabuf_feedback,
+            gamma_props,
             vblank_frame: None,
             vblank_frame_name,
             time_since_presentation_plot_name,
@@ -1258,20 +1279,34 @@ impl Tty {
             .devices
             .get(&tty_state.node)
             .context("missing device")?;
-        let info = device
-            .drm
-            .get_crtc(crtc)
-            .context("error getting crtc info")?;
-        Ok(info.gamma_length())
+
+        let surface = device.surfaces.get(&crtc).context("missing surface")?;
+        if let Some(gamma_props) = &surface.gamma_props {
+            gamma_props.gamma_size(&device.drm)
+        } else {
+            let info = device
+                .drm
+                .get_crtc(crtc)
+                .context("error getting crtc info")?;
+            Ok(info.gamma_length())
+        }
     }
 
-    pub fn set_gamma(&self, output: &Output, ramp: Option<&[u16]>) -> anyhow::Result<()> {
+    pub fn set_gamma(&mut self, output: &Output, ramp: Option<&[u16]>) -> anyhow::Result<()> {
         let tty_state = output.user_data().get::<TtyOutputState>().unwrap();
+        let crtc = tty_state.crtc;
+
         let device = self
             .devices
-            .get(&tty_state.node)
+            .get_mut(&tty_state.node)
             .context("missing device")?;
-        set_gamma_for_crtc(&device.drm, tty_state.crtc, ramp)
+
+        let surface = device.surfaces.get_mut(&crtc).context("missing surface")?;
+        if let Some(gamma_props) = &mut surface.gamma_props {
+            gamma_props.set_gamma(&device.drm, ramp)
+        } else {
+            set_gamma_for_crtc(&device.drm, crtc, ramp)
+        }
     }
 
     fn refresh_ipc_outputs(&self) {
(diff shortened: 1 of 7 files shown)


src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+                let mut blue = ramp.clone();
+                blue.drain(0..gamma_length * 2);
+                blue.truncate(gamma_length);
+
+                device
+                    .drm
+                    .set_gamma(crtc, &mut red, &mut green, &mut blue)
+                    .unwrap_or_else(|err| warn!("error setting gamma for: {err:?}"));
+            }
+        }
```
phuhl, line 1297: In the sway implementation I saw that a new render frame was scheduled. It seems to work w/o doing that. Do we need that? How would we do that (i.e. what function needs to be called)?

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                let buf = &mut Vec::new();
+                if File::from(fd).read_to_end(buf).is_err() {
+                    error!("Failed to read gamma data");
+                    resource.failed();
+                    return;
+                }
+
+                let mut gamma = Vec::new();
+                for i in 0..buf.len() / 2 {
+                    gamma.push(((buf[i * 2 + 1] as u16) << 8) | buf[i * 2] as u16);
```
phuhl, line 173: Annoyingly, we need to convert a u8 to a u16 vector here. That means, we need to concert us with little vs big indian. Is there a better way? Is this safe?

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+use wayland_protocols_wlr::gamma_control::v1::server::{
+    zwlr_gamma_control_manager_v1, zwlr_gamma_control_v1,
+};
+use zwlr_gamma_control_manager_v1::ZwlrGammaControlManagerV1;
+use zwlr_gamma_control_v1::ZwlrGammaControlV1;
+
+const VERSION: u32 = 1;
+
+pub struct GammaControlManagerState {
+    gamma_controls: HashMap<ObjectId, ZwlrGammaControlV1>,
```
YaLTeR, line 20:
```suggestion
    gamma_controls: HashMap<WlOutput, ZwlrGammaControlV1>,
```
Wayland objects conveniently implement Hash.

src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+    pub fn get_gamma_length(&self, output: &Output) -> Option<u32> {
+        let node = output.user_data().get::<TtyOutputState>().unwrap().node;
+        let devices = self.devices.get(&node);
+        if let Some(device) = devices {
+            for (_connector, crtc) in device.drm_scanner.crtcs() {
+                let crt_info = device.drm.get_crtc(crtc).unwrap();
+                return Some(crt_info.gamma_length());
+            }
+        }
+        None
```
YaLTeR, line 1248: This will get gamma length of the first enabled output, whatever that output is. Luckily, there's `crtc` in `TtyOutputState`.

```suggestion
        let tty_state = output.user_data().get::<TtyOutputState>().unwrap();
        let device = self.devices.get(&tty_state.node).ok()?;
        let crtc_info = device.drm.get_crtc(tty_state.crtc.clone()).ok()?;
        return Some(crtc_info.gamma_length());
```

src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+    }
+
+    pub fn get_gamma(&self, output: &Output) -> Option<Vec<u16>> {
+        let node = output.user_data().get::<TtyOutputState>().unwrap().node;
+        let devices = self.devices.get(&node);
+        if let Some(device) = devices {
+            for (_connector, crtc) in device.drm_scanner.crtcs() {
+                let crt_info = device.drm.get_crtc(crtc).unwrap();
```
YaLTeR, line 1256: Same as above.

src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+                let gamma_length = crt_info.gamma_length() as usize;
+                let mut red = vec![0; gamma_length];
+                let mut green = vec![0; gamma_length];
+                let mut blue = vec![0; gamma_length];
+                device
+                    .drm
+                    .get_gamma(crtc, &mut red, &mut green, &mut blue)
+                    .unwrap_or_else(|err| warn!("error getting gamma for: {err:?}"));
```
YaLTeR, line 1264: This `unwrap_or_else()` will not return `None` as intended. What you want is this:

```suggestion
                if let Err(err) = device.drm.get_gamma(crtc, &mut red, &mut green, &mut blue) {
                    warn!("error getting gamma for crtc {crtc:?}: {err:?}");
                    return None;
                }
```

src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+    }
+
+    pub fn set_gamma(&self, output: &Output, ramp: Vec<u16>) {
+        let node = output.user_data().get::<TtyOutputState>().unwrap().node;
+
+        let devices = self.devices.get(&node);
+
+        if let Some(device) = devices {
+            for (_connector, crtc) in device.drm_scanner.crtcs() {
+                let crt_info = device.drm.get_crtc(crtc).unwrap();
```
YaLTeR, line 1280: Same as above.

src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+                let gamma_length = crt_info.gamma_length() as usize;
+
+                let mut red = ramp.clone();
+                red.truncate(gamma_length);
+                let mut green = ramp.clone();
+                green.drain(0..gamma_length);
+                green.truncate(gamma_length);
+                let mut blue = ramp.clone();
+                blue.drain(0..gamma_length * 2);
+                blue.truncate(gamma_length);
```
YaLTeR, line 1290: Make this function return `anyhow::Result<()>`, then add a check here for the correct gamma length with `ensure!(ramp.len() == gamma_length, "wrong gamma length");`.

src/backend/tty.rs
```diff
@@ -1236,6 +1236,67 @@ impl Tty {
...
+
+                device
+                    .drm
+                    .set_gamma(crtc, &mut red, &mut green, &mut blue)
+                    .unwrap_or_else(|err| warn!("error setting gamma for: {err:?}"));
```
YaLTeR, line 1295: With `anyhow::Result`, you can change this to `.context("error setting gamma")?`.

src/backend/tty.rs
YaLTeR: I don't know if it's necessary, but you can pass in `niri: &mut Niri` and call `niri.queue_redraw(output)` to schedule a render.

src/handlers/mod.rs
```diff
@@ -425,3 +428,56 @@ impl DrmLeaseHandler for State {
...
 }
 delegate_drm_lease!(State);
+
+impl GammaControlHandler for State {
+    fn gamma_control_manager_state(&mut self) -> &mut GammaControlManagerState {
+        &mut self.niri.gamma_control_manager_state
+    }
+
+    fn set_gamma(&mut self, wl_output: &WlOutput, ramp: Vec<u16>) -> Result<(), ()> {
+        if !self.backend.is_tty() {
```
YaLTeR, line 438: Instead of these checks here, just make it so the global is only available when using the tty backend.

src/handlers/mod.rs
```diff
@@ -425,3 +428,56 @@ impl DrmLeaseHandler for State {
...
+        };
+
+        self.backend.tty().get_gamma(&output)
+    }
+
+    fn destroy(&mut self, output_id: ObjectId) {
+        self.niri
+            .gamma_control_manager_state
+            .destroy_gamma_control(output_id)
+    }
```
YaLTeR, line 468: Is this really needed if it just calls a method on the gamma_control_manager_state itself?

src/handlers/mod.rs
```diff
@@ -425,3 +428,56 @@ impl DrmLeaseHandler for State {
...
+
+    fn get_gamma(&mut self, wl_output: &WlOutput) -> Option<Vec<u16>> {
+        if !self.backend.is_tty() {
+            error!("Backend is not Tty");
+            return None;
+        }
+
+        let Some(output) = Output::from_resource(wl_output) else {
+            return None;
+        };
```
YaLTeR, line 459:
```suggestion
        let output = Output::from_resource(wl_output)?;
```

src/handlers/mod.rs
```diff
@@ -425,3 +428,56 @@ impl DrmLeaseHandler for State {
...
+
+    fn get_gamma_size(&mut self, wl_output: &WlOutput) -> Option<u32> {
+        if !self.backend.is_tty() {
+            error!("Backend is not Tty");
+            return None;
+        }
+        let Some(output) = Output::from_resource(wl_output) else {
+            error!("No Output matching WlOutput");
+            return None;
+        };
```
YaLTeR, line 478:
```suggestion
        let output = Output::from_resource(wl_output)?;
```

src/handlers/mod.rs
```diff
@@ -425,3 +428,56 @@ impl DrmLeaseHandler for State {
...
+        let gamma_length = self.backend.tty().get_gamma_length(&output);
+        return gamma_length;
```
YaLTeR, line 480:
```suggestion
        self.backend.tty().get_gamma_length(&output)
```

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                    id,
+                    GammaControlState {
+                        gamma_size: gamma_size.clone(),
+                        previous_gamma_ramp,
+                        output,
+                    },
+                );
+                if !got_gamma_map {
+                    zwlr_gamma_control.failed();
+                    error!("Failed to get current gamma ramp");
```
YaLTeR, line 114: These should be `warn!` (since I imagine they don't necessarily indicate a programming bug), and I'd put them in the Handler functions on State, instead of here inside the protocol.

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                    return;
+                }
+
+                if gamma_size.is_none() {
+                    zwlr_gamma_control.failed();
+                    error!("gamma size not set");
+                    return;
+                }
+
+                zwlr_gamma_control.gamma_size(gamma_size.unwrap());
```
YaLTeR, line 134: Rather than checking is_none() + unwrap(), I'd do `if let Some(gamma_size) = gamma_size { /* success path */ } else { /* call failed() */ }`

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+        _client: &Client,
+        resource: &ZwlrGammaControlV1,
+        request: <ZwlrGammaControlV1 as Resource>::Request,
+        data: &GammaControlState,
+        _dhandle: &DisplayHandle,
+        _data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            zwlr_gamma_control_v1::Request::SetGamma { fd } => {
+                info!("Setting gamma for output {:?}", data.output);
```
YaLTeR, line 163:
```suggestion
                debug!("setting gamma for output {:?}", data.output);
```

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                let buf = &mut Vec::new();
+                if File::from(fd).read_to_end(buf).is_err() {
+                    error!("Failed to read gamma data");
```
YaLTeR, line 166:
```suggestion
                    warn!("failed to read gamma data for output {:?}", data.output);
```

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                    return;
+                }
+
+                let mut gamma = Vec::new();
+                for i in 0..buf.len() / 2 {
+                    gamma.push(((buf[i * 2 + 1] as u16) << 8) | buf[i * 2] as u16);
+                }
+                if data.gamma_size.is_none() {
+                    error!("gamma size not set");
+                    resource.failed();
```
YaLTeR, line 177: We already failed it when it was created in this case, probably can just `return` silently.

Or, even better, add a flag `data.failed`, set it when failing the resource, and just do `if data.failed { return; }` at the top.

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                    return;
+                }
+                let gamma_size = data.gamma_size.unwrap();
+                if gamma.len() != gamma_size as usize * 3 {
+                    error!(
+                        "gamma length wrong (expected {}, got {})",
+                        gamma_size,
+                        gamma.len()
+                    );
```
YaLTeR, line 186: This check will be inside `set_gamma` as I suggested above.

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                    resource.failed();
+                    return;
+                }
+
+                if state.set_gamma(&data.output, gamma).is_err() {
+                    resource.failed();
+                    error!("Failed to set gamma");
+                    return;
+                }
```
YaLTeR, line 195:
```suggestion
                if let Err(err) = state.set_gamma(&data.output, gamma) {
                    warn!("error setting gamma: {err:?}");
                    resource.failed();
                    return;
                }
```

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+                if !got_gamma_map {
+                    zwlr_gamma_control.failed();
+                    error!("Failed to get current gamma ramp");
+                    return;
+                }
+
+                if state
+                    .gamma_control_manager_state()
+                    .gamma_controls
+                    .contains_key(&output_id)
```
YaLTeR, line 121: What I'm afraid can happen is:
1. Client A sets gamma
2. Client B creates gamma control, gets `previous_gamma_ramp` equal to Client A's gamma, gets a fail here
3. Client A destroys global, resets gamma
4. Client B destroys global later, "resets" gamma to Client A's value instead of the original

So this `if` should also set `previous_gamma_ramp` to `None`.

src/protocols/gamma_control.rs
```diff
@@ -0,0 +1,237 @@
...
+        data: &GammaControlState,
+    ) {
+        if data.previous_gamma_ramp.is_none() {
+            return;
+        }
+        let ramp = data.previous_gamma_ramp.as_ref().unwrap();
+        state
+            .set_gamma(&data.output, ramp.to_vec())
+            .unwrap_or_else({
+                error!("Failed to reset gamma ramp");
```
YaLTeR, line 215: Also `if let Err(err)` and `warn!`.

src/protocols/gamma_control.rs
YaLTeR: We have [bytemuck](https://docs.rs/bytemuck/latest/bytemuck/) in our dependency tree, I think you can just use that? I assume the endianness is expected to match and "just work".

YaLTeR (commented):
Thanks, this is actually quite close to what I would write!

src/handlers/mod.rs
phuhl: How would I do that, can you point me to a place where that has been done?

src/handlers/mod.rs
phuhl: I think it is, because we go from one impl to another through the state. But I might be wrong...

src/handlers/mod.rs
YaLTeR: Several ways to do it. For example, can store an Option<GammaControlState> on Niri, and only create it if it's tty. Or add a global filter (should be done anyway for restricted clients, same as in foreign toplevel management) and always return false if backend is not tty.

src/protocols/gamma_control.rs
phuhl: After refactoring some stuff above, I'd rather keep it this way, as we do a check on gamma_size.is_none() above and need to create the GammaControlState with previous_gamma_ramp as None. We can create it all a third time of course, if you prefer, but I feel, that would be unnecessarily verbose.

You decide, keep it as is?

src/protocols/gamma_control.rs
phuhl: It seems, we don't have bytemuck, yet. I added it. Are you comfortable with that?

src/handlers/mod.rs
phuhl: Sorry, still not getting it...

What exactly are you referring to, when saying global? The GlobalDispatch? If that is the case, changing the delegate_gamma_control conditionally seems impossible to me, as that is happening before we have a state.

What is a global filter?

phuhl:
Except for the "does tty exist" thing, I think I worked all the suggestions in. Thanks for reviewing! Let me know if anything needs changing.

src/backend/tty.rs
```diff
@@ -1237,64 +1237,62 @@ impl Tty {
...
+        &self,
+        niri: &mut Niri,
+        output: &Output,
+        ramp: Vec<u16>,
+    ) -> anyhow::Result<()> {
+        let tty_state = output.user_data().get::<TtyOutputState>().unwrap();
+        let device = self
+            .devices
+            .get(&tty_state.node)
+            .ok_or(anyhow!("missing device"))?;
```
YaLTeR, line 1274:
```suggestion
            .context("missing device")?;
```
I think just this should work?

src/backend/tty.rs
```diff
@@ -1237,64 +1237,62 @@ impl Tty {
...
+
+        let mut red = ramp.clone();
+        red.truncate(gamma_length);
+        let mut green = ramp.clone();
+        green.drain(0..gamma_length);
+        green.truncate(gamma_length);
+        let mut blue = ramp.clone();
+        blue.drain(0..gamma_length * 2);
+        blue.truncate(gamma_length);
+        ensure!(ramp.len() == gamma_length, "wrong gamma length");
```
YaLTeR, line 1288: Put this above `let mut red = ramp.clone()`, this way you can avoid cloning it the third time for blue. Also shouldn't it check for `gamma_length * 3`?

src/protocols/gamma_control.rs
```diff
@@ -160,37 +161,24 @@ where
     ) {
         match request {
             zwlr_gamma_control_v1::Request::SetGamma { fd } => {
-                info!("Setting gamma for output {:?}", data.output);
-                let buf = &mut Vec::new();
-                if File::from(fd).read_to_end(buf).is_err() {
-                    error!("Failed to read gamma data");
+                if data.failed {
                     resource.failed();
```
YaLTeR, line 165: If it's failed then we don't need to send `failed()` again here, just return.

src/protocols/gamma_control.rs
```diff
@@ -160,37 +161,24 @@ where
...
-                for i in 0..buf.len() / 2 {
-                    gamma.push(((buf[i * 2 + 1] as u16) << 8) | buf[i * 2] as u16);
-                }
-                if data.gamma_size.is_none() {
-                    error!("gamma size not set");
+                debug!("setting gamma for output {:?}", data.output);
+                let buf = &mut Vec::new();
+                if File::from(fd).read_to_end(buf).is_err() {
+                    warn!("failed to read gamma data for output {:?}", data.output);
                     resource.failed();
```
YaLTeR, line 172: Also set `data.failed = true;`

src/protocols/gamma_control.rs
```diff
@@ -160,37 +161,24 @@ where
...
-                        gamma.len()
-                    );
-                    resource.failed();
-                    return;
-                }
 
-                if state.set_gamma(&data.output, gamma).is_err() {
+                if let Err(err) = state.set_gamma(&data.output, gamma, gamma_size) {
+                    warn!("error setting gamma: {err:?}");
                     resource.failed();
```
YaLTeR, line 181: Also set `data.failed = true;`

src/protocols/gamma_control.rs
```diff
@@ -205,16 +193,16 @@ where
...
-        state
-            .set_gamma(&data.output, ramp.to_vec())
-            .unwrap_or_else({
-                error!("Failed to reset gamma ramp");
-                |_| ()
-            });
+
+        if let Err(err) = state.set_gamma(&data.output, ramp.to_vec(), data.gamma_size.unwrap()) {
+            warn!("error resetting gamma: {err:?}");
+            return;
```
YaLTeR, line 203: We still need to call `state.destroy` in this case. Nothing we can really do about the fact that resetting gamma failed.

src/handlers/mod.rs
YaLTeR: This is what I'm referring to:

https://github.com/YaLTeR/niri/blob/1a784e6e66785f360da6df59ae2fb4b98370ca3c/src/niri.rs#L909-L912

This closure controls which clients can see the global. In this case, restricted clients (i.e. sandboxed with flatpak) cannot see the global. I suggest adding that and additionally always returning false from that closure if the backend is not TTY, this way on winit clients will simply not see it.

Implementing that closure is very simple:
1. https://github.com/YaLTeR/niri/blob/1a784e6e66785f360da6df59ae2fb4b98370ca3c/src/protocols/foreign_toplevel.rs#L63-L66
2. https://github.com/YaLTeR/niri/blob/1a784e6e66785f360da6df59ae2fb4b98370ca3c/src/protocols/foreign_toplevel.rs#L327-L329

YaLTeR: This is what I'm referring to:

https://github.com/YaLTeR/niri/blob/1a784e6e66785f360da6df59ae2fb4b98370ca3c/src/niri.rs#L909-L912

This closure controls which clients can see the global. In this case, restricted clients (i.e. sandboxed with flatpak) cannot see the global. I suggest adding that and additionally always returning false from that closure if the backend is not TTY, this way on winit clients will simply not see it.

Implementing that closure is very simple:
1. https://github.com/YaLTeR/niri/blob/1a784e6e66785f360da6df59ae2fb4b98370ca3c/src/protocols/foreign_toplevel.rs#L63-L66
2. https://github.com/YaLTeR/niri/blob/1a784e6e66785f360da6df59ae2fb4b98370ca3c/src/protocols/foreign_toplevel.rs#L327-L329

src/backend/tty.rs
phuhl: Yup, should check for that. Just search that bug for 15 minutes *facepalm*

src/protocols/gamma_control.rs
phuhl: data itself is read only here and the Dispatch trait doesn't allow &mut data. We could get arround it with some research on data_init maybe, but I think, setting data.failed is not vital here, what do you think?

src/protocols/gamma_control.rs
phuhl: same as above

phuhl:
Thanks again for the review, applied most of the changes

src/protocols/gamma_control.rs
YaLTeR: Ah, right, you'd need Mutexes for that... Well I guess it's not a big deal. The client is supposed to destroy upon failing anyway.

YaLTeR:
Thanks, do you mind if I do some changes myself before merging?

phuhl:
> Thanks, do you mind if I do some changes myself before merging?

Sure, go ahead

YaLTeR:
Okay, I think this should be good now. After testing it turned out that replugging a monitor with a changed gamma leaves its gamma hanging, so I decided to change the code to always reset to a linear gamma instead of the "previous" gamma. I also made it so newly connected monitors have gamma reset to linear right away.

Could you test that this still works for you?

YaLTeR:
Wow, this legacy set_gamma call is sure slow though, 10 ms per call here. Well, hopefully the new gamma properties will be better...

YaLTeR:
I have now implemented using the new GAMMA_LUT property and... it's still as slow. Turns out it's slow on sway as well, so I guess one just needs to be aware of this? I've seen times as long as 18 ms, which is a frame drop even on a 60 Hz screen, and gammastep does a gamma adjustment every few seconds in normal operation. Fun!

Well, this should be done now, so I'll be waiting for your OK and merging.

YaLTeR:
Decided to also implemented storing a pending gamma change when niri doesn't have control over the TTY, to be applied when it gets control.

phuhl:
Tested ae5887d and that worked great on my system!

phuhl:
f96ea47 also works, did not test switching TTY though

YaLTeR:
Thanks!

PR #243, opened 2024-03-04
implement wlr-screencopy

### LINKED ISSUE #33: Add wlr-screencopy
The implementation with smithay is available in https://github.com/catacombing/catacomb/tree/master/src/protocols/screencopy and could probably be reused. Though it's limited in terms of formats.

sodiboo:
Resolves #33

Significant parts of this are taken from Catacomb.

I don't fully understand everything i've done so far with this, but damnit am i proud of what i've done, and you don't even KNOW the pure amounts of dopamine rushing through my veins when i finally saw a screenshot output from my own little thing!!

As i submit this as a draft, it basically works.

- The image is upside down. This flipping happens *before* selection of the region, i.e. if you take a screenshot of the top half of the screen, the resulting image will be of the bottom half of the screen, flipped. This is likely a consequence of the order of the pixels sent to the screencopy. Maybe?

image

- After taking a screenshot, the keyboard input just gets completely fucked. I cannot type commands in the terminal. I cannot use binds to spawn processes. I cannot use binds to move windows, or even the focus. To test the screenshot functionality, i genuinely just set `WAYLAND_DISPLAY` in a terminal in the "real" compositor in order to work around this and focus on screencopy implementation. Why does this happen. I have no idea, please help.

- Does it work on multiple outputs? Probably not. Catacomb, to my knowledge, doesn't expect there to be more than 1 output(??), and my development setup does not have multiple monitors at the moment, so i can't easily test multimonitor. Therefore, no particular care has been taken towards keeping track of multiple outputs yet.

- DmaBuf/tty is currently untested. I don't see why it wouldn't work, but i have only tested it on winit so far.

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -383,13 +383,12 @@ delegate_foreign_toplevel!(State);
 
 impl ScreencopyHandler for State {
     fn frame(&mut self, screencopy: Screencopy) {
-        let output = screencopy.output().clone();
-        if let Some(state) = self.niri.output_state.get_mut(&output) {
-            state.pending_screencopy.push(screencopy);
+        if let Err(err) = self
+            .niri
+            .render_for_screencopy(&mut self.backend, screencopy)
+        {
+            warn!("error rendering for screencopy: {err:?}");
         }
-
-        // Force redraw, to prevent screencopy stalling.
-        self.niri.queue_redraw(output);
     }
 }
 delegate_screencopy!(State);
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -54,12 +54,13 @@ use smithay::{
     delegate_text_input_manager, delegate_virtual_keyboard_manager,
 };
 
-use crate::delegate_foreign_toplevel;
 use crate::niri::{ClientState, State};
 use crate::protocols::foreign_toplevel::{
     self, ForeignToplevelHandler, ForeignToplevelManagerState,
 };
+use crate::protocols::screencopy::{Screencopy, ScreencopyHandler};
 use crate::utils::output_size;
+use crate::{delegate_foreign_toplevel, delegate_screencopy};
 
 impl SeatHandler for State {
     type KeyboardFocus = WlSurface;
@@ -379,6 +380,19 @@ impl ForeignToplevelHandler for State {
     }
 }
 delegate_foreign_toplevel!(State);
+
+impl ScreencopyHandler for State {
+    fn frame(&mut self, screencopy: Screencopy) {
+        let output = screencopy.output().clone();
+        if let Some(state) = self.niri.output_state.get_mut(&output) {
+            state.pending_screencopy.push(screencopy);
+        }
+
+        // Force redraw, to prevent screencopy stalling.
+        self.niri.queue_redraw(output);
+    }
+}
+delegate_screencopy!(State);
 
 impl DrmLeaseHandler for State {
     fn drm_lease_state(&mut self, node: DrmNode) -> &mut DrmLeaseState {
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -18,7 +18,9 @@ use smithay::backend::renderer::element::solid::{SolidColorBuffer, SolidColorRen
 use smithay::backend::renderer::element::surface::{
     render_elements_from_surface_tree, WaylandSurfaceRenderElement,
 };
-use smithay::backend::renderer::element::utils::{select_dmabuf_feedback, RelocateRenderElement};
+use smithay::backend::renderer::element::utils::{
+    select_dmabuf_feedback, Relocate, RelocateRenderElement,
+};
 use smithay::backend::renderer::element::{
     default_primary_scanout_output_compare, AsRenderElements, Id, Kind, PrimaryScanoutOutput,
     RenderElementStates,
@@ -97,9 +99,10 @@ use crate::input::{apply_libinput_settings, TabletData};
 use crate::ipc::server::IpcServer;
 use crate::layout::{Layout, MonitorRenderElement};
 use crate::protocols::foreign_toplevel::{self, ForeignToplevelManagerState};
+use crate::protocols::screencopy::{Screencopy, ScreencopyManagerState};
 use crate::pw_utils::{Cast, PipeWire};
 use crate::render_helpers::renderer::NiriRenderer;
-use crate::render_helpers::{render_to_texture, render_to_vec};
+use crate::render_helpers::{render_to_shm, render_to_texture, render_to_vec};
 use crate::ui::config_error_notification::ConfigErrorNotification;
 use crate::ui::exit_confirm_dialog::ExitConfirmDialog;
 use crate::ui::hotkey_overlay::HotkeyOverlay;
@@ -154,6 +157,7 @@ pub struct Niri {
     pub layer_shell_state: WlrLayerShellState,
     pub session_lock_state: SessionLockManagerState,
     pub foreign_toplevel_state: ForeignToplevelManagerState,
+    pub screencopy_state: ScreencopyManagerState,
     pub shm_state: ShmState,
     pub output_manager_state: OutputManagerState,
     pub dmabuf_state: DmabufState,
@@ -247,6 +251,7 @@ pub struct OutputState {
     pub lock_render_state: LockRenderState,
     pub lock_surface: Option<LockSurface>,
     pub lock_color_buffer: SolidColorBuffer,
+    pub pending_screencopy: Vec<Screencopy>,
 }
 
 #[derive(Default)]
@@ -910,6 +915,9 @@ impl Niri {
             ForeignToplevelManagerState::new::<State, _>(&display_handle, |client| {
                 !client.get_data::<ClientState>().unwrap().restricted
             });
+        let screencopy_state = ScreencopyManagerState::new::<State, _>(&display_handle, |client| {
+            !client.get_data::<ClientState>().unwrap().restricted
+        });
 
         let mut seat: Seat<State> = seat_state.new_wl_seat(&display_handle, backend.seat_name());
         seat.add_keyboard(
@@ -1030,6 +1038,7 @@ impl Niri {
             layer_shell_state,
             session_lock_state,
             foreign_toplevel_state,
+            screencopy_state,
             text_input_state,
             input_method_state,
             virtual_keyboard_state,
@@ -1265,6 +1274,7 @@ impl Niri {
             lock_render_state,
             lock_surface: None,
             lock_color_buffer: SolidColorBuffer::new(size, CLEAR_COLOR_LOCKED),
+            pending_screencopy: Vec::new(),
         };
         let rv = self.output_state.insert(output.clone(), state);
         assert!(rv.is_none(), "output was already tracked");
@@ -2166,13 +2176,19 @@ impl Niri {
         // to err on the safe side.
         self.send_frame_callbacks(output);
 
-        // Render and send to PipeWire screencast streams.
-        #[cfg(feature = "xdp-gnome-screencast")]
-        {
-            backend.with_primary_renderer(|renderer| {
+        let y_invert = match backend {
+            Backend::Tty(_) => false,
+            Backend::Winit(_) => true,
+        };
+
+        backend.with_primary_renderer(|renderer| {
+            // Render and send to PipeWire screencast streams.
+            #[cfg(feature = "xdp-gnome-screencast")]
+            {
                 self.render_for_screen_cast(renderer, output, target_presentation_time);
-            });
-        }
+            }
+            self.render_for_screencopy(renderer, output, y_invert);
+        });
     }
 
     pub fn update_primary_scanout_output(
@@ -2507,6 +2523,47 @@ impl Niri {
         }
 
         feedback
+    }
+
+    fn render_for_screencopy(
+        &mut self,
+        renderer: &mut GlesRenderer,
+        output: &Output,
+        y_invert: bool,
+    ) {
+        for mut screencopy in self
+            .output_state
+            .get_mut(output)
+            .map(|s| mem::take(&mut s.pending_screencopy))
+            .unwrap_or_default()
+        {
+            let region = screencopy.region();
+            let elements = self
+                .render(renderer, output, screencopy.overlay_cursor())
+                .into_iter()
+                .rev();
+
+            let scale = output.current_scale().fractional_scale().into();
+
+            let elements = elements.map(|element| {
+                RelocateRenderElement::from_element(
+                    element,
+                    region.loc.upscale(-1),
+                    Relocate::Relative,
+                )
+            });
+
+            let damage = [Rectangle::from_loc_and_size((0, 0), region.size)];
+            screencopy.damage(&damage);
+
+            if let Err(err) =
+                render_to_shm(renderer, screencopy.buffer(), region.size, scale, elements)
+            {
+                warn!("error rendering to screencopy shm buffer: {err:?}")
+            } else {
+                screencopy.submit(y_invert);
+            }
+        }
     }
 
     #[cfg(feature = "xdp-gnome-screencast")]
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -9,7 +9,7 @@ use std::time::{Duration, Instant};
 use std::{env, mem, thread};
 
 use _server_decoration::server::org_kde_kwin_server_decoration_manager::Mode as KdeDecorationsMode;
-use anyhow::Context;
+use anyhow::{ensure, Context};
 use calloop::futures::Scheduler;
 use niri_config::{Config, TrackLayout};
 use smithay::backend::allocator::Fourcc;
@@ -251,7 +251,6 @@ pub struct OutputState {
     pub lock_render_state: LockRenderState,
     pub lock_surface: Option<LockSurface>,
     pub lock_color_buffer: SolidColorBuffer,
-    pub pending_screencopy: Vec<Screencopy>,
 }
 
 #[derive(Default)]
@@ -1274,7 +1273,6 @@ impl Niri {
             lock_render_state,
             lock_surface: None,
             lock_color_buffer: SolidColorBuffer::new(size, CLEAR_COLOR_LOCKED),
-            pending_screencopy: Vec::new(),
         };
         let rv = self.output_state.insert(output.clone(), state);
         assert!(rv.is_none(), "output was already tracked");
@@ -2176,18 +2174,10 @@ impl Niri {
         // to err on the safe side.
         self.send_frame_callbacks(output);
 
-        let y_invert = match backend {
-            Backend::Tty(_) => false,
-            Backend::Winit(_) => true,
-        };
-
         backend.with_primary_renderer(|renderer| {
             // Render and send to PipeWire screencast streams.
             #[cfg(feature = "xdp-gnome-screencast")]
-            {
-                self.render_for_screen_cast(renderer, output, target_presentation_time);
-            }
-            self.render_for_screencopy(renderer, output, y_invert);
+            self.render_for_screen_cast(renderer, output, target_presentation_time);
         });
     }
 
@@ -2525,47 +2515,6 @@ impl Niri {
         feedback
     }
 
-    fn render_for_screencopy(
-        &mut self,
-        renderer: &mut GlesRenderer,
-        output: &Output,
-        y_invert: bool,
-    ) {
-        for mut screencopy in self
-            .output_state
-            .get_mut(output)
-            .map(|s| mem::take(&mut s.pending_screencopy))
-            .unwrap_or_default()
-        {
-            let region = screencopy.region();
-            let elements = self
-                .render(renderer, output, screencopy.overlay_cursor())
-                .into_iter()
-                .rev();
-
-            let scale = output.current_scale().fractional_scale().into();
-
-            let elements = elements.map(|element| {
-                RelocateRenderElement::from_element(
-                    element,
-                    region.loc.upscale(-1),
-                    Relocate::Relative,
-                )
-            });
-
-            let damage = [Rectangle::from_loc_and_size((0, 0), region.size)];
-            screencopy.damage(&damage);
-
-            if let Err(err) =
-                render_to_shm(renderer, screencopy.buffer(), region.size, scale, elements)
-            {
-                warn!("error rendering to screencopy shm buffer: {err:?}")
-            } else {
-                screencopy.submit(y_invert);
-            }
-        }
-    }
-
     #[cfg(feature = "xdp-gnome-screencast")]
     fn render_for_screen_cast(
         &mut self,
@@ -2643,7 +2592,9 @@ impl Niri {
                     .get_or_insert_with(|| self.render::<GlesRenderer>(renderer, output, true));
                 let elements = elements.iter().rev();
 
-                if let Err(err) = render_to_dmabuf(renderer, dmabuf, size, scale, elements) {
+                if let Err(err) =
+                    render_to_dmabuf(renderer, dmabuf, size, scale, Transform::Normal, elements)
+                {
                     warn!("error rendering to dmabuf: {err:?}");
                     continue;
                 }
@@ -2661,6 +2612,42 @@ impl Niri {
         for id in casts_to_stop {
             self.stop_cast(id);
         }
+    }
+
+    pub fn render_for_screencopy(
+        &mut self,
+        backend: &mut Backend,
+        screencopy: Screencopy,
+    ) -> anyhow::Result<()> {
+        let output = screencopy.output().clone();
+        ensure!(self.output_state.contains_key(&output), "output is missing");
+
+        backend
+            .with_primary_renderer(move |renderer| {
+                let elements = self
+                    .render(renderer, &output, screencopy.overlay_cursor())
+                    .into_iter()
+                    .rev();
+
+                let region_loc = screencopy.region_loc();
+                let elements = elements.map(|element| {
+                    RelocateRenderElement::from_element(
+                        element,
+                        region_loc.upscale(-1),
+                        Relocate::Relative,
+                    )
+                });
+
+                let scale = output.current_scale().fractional_scale().into();
+                let transform = output.current_transform();
+                render_to_shm(renderer, screencopy.buffer(), scale, transform, elements)
+                    .context("error rendering to screencopy shm buffer: {err:?}")?;
+
+                screencopy.submit(false);
+
+                Ok(())
+            })
+            .context("primary renderer is missing")?
     }
 
     #[cfg(feature = "xdp-gnome-screencast")]
@@ -2722,7 +2709,14 @@ impl Niri {
                 let elements = self.render::<GlesRenderer>(renderer, &output, true);
                 let elements = elements.iter().rev();
 
-                let res = render_to_texture(renderer, size, scale, Fourcc::Abgr8888, elements);
+                let res = render_to_texture(
+                    renderer,
+                    size,
+                    scale,
+                    Transform::Normal,
+                    Fourcc::Abgr8888,
+                    elements,
+                );
                 let screenshot = match res {
                     Ok((texture, _)) => texture,
                     Err(err) => {
@@ -2752,7 +2746,14 @@ impl Niri {
         let scale = Scale::from(output.current_scale().fractional_scale());
         let elements = self.render::<GlesRenderer>(renderer, output, true);
         let elements = elements.iter().rev();
-        let pixels = render_to_vec(renderer, size, scale, Fourcc::Abgr8888, elements)?;
+        let pixels = render_to_vec(
+            renderer,
+            size,
+            scale,
+            Transform::Normal,
+            Fourcc::Abgr8888,
+            elements,
+        )?;
 
         self.save_screenshot(size, pixels)
             .context("error saving screenshot")
@@ -2778,7 +2779,14 @@ impl Niri {
             1.,
         );
         let elements = elements.iter().rev();
-        let pixels = render_to_vec(renderer, size, scale, Fourcc::Abgr8888, elements)?;
+        let pixels = render_to_vec(
+            renderer,
+            size,
+            scale,
+            Transform::Normal,
+            Fourcc::Abgr8888,
+            elements,
+        )?;
 
         self.save_screenshot(size, pixels)
             .context("error saving screenshot")
@@ -2881,6 +2889,7 @@ impl Niri {
             renderer,
             size,
             Scale::from(f64::from(output_scale)),
+            Transform::Normal,
             Fourcc::Abgr8888,
             elements,
         )?;
--- a/src/protocols/screencopy.rs
+++ b/src/protocols/screencopy.rs
@@ -1,0 +1,346 @@
+use std::os::fd::OwnedFd;
+use std::time::UNIX_EPOCH;
+
+use calloop::generic::Generic;
+use calloop::{Interest, LoopHandle, Mode, PostAction};
+use smithay::output::Output;
+use smithay::reexports::wayland_protocols_wlr::screencopy::v1::server::zwlr_screencopy_frame_v1::{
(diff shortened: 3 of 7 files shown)


sodiboo:
Cargo needs to be nightly in order for `cargo fmt` to appease the CI. this seems like a bug, given the MSRV of 1.72.0

sodiboo:
It seems that launching my branch in tty does not work. I don't know what i did to cause this, but i feel that i don't really deserve this fate. It just displays pitch black. Exiting still works, though. So, clearly, i broke something important, but not *that* important. How? What?

sodiboo:
> After taking a screenshot, the keyboard input just gets completely fucked.

Upon further inspection, this may be unrelated to my changes(??). Because, earlier, i managed to take many screenshots in a single session without the keyboard breaking. Additionally, just now, i noticed that *the input was broken without ever calling grim*. Therefore, i find it likely that this issue is not a consequence of anything i've written as part of this PR, but rather an unrelated issue that i've been running into, likely specific to the winit backend.

sodiboo:
> The winit backend does indeed have a problem that sometimes modifiers get stuck when you focus it in/out.

Oh, interesting! It is as simple as modifiers getting stuck. In my case, i believe `Super` is getting stuck, likely as some indirect result of me pressing `Super` as it gets focused. I didn't realize that's what was happening, and i just brushed is off as "oh none of the keybinds work and i can't even type in the terminal", but it feels obvious now.

sodiboo:
Either way, that's both of the major issues i encountered accounted for (flipped image; borked input). As such, there are no more major issues with the screencopying functionality per se, but it is still untested with dmabuf/tty and i also cannot easily test multi-monitor myself, though i do NOT Handle it properly (currently: "output under cursor". crashes sometimes when unwrapping this)

sodiboo:
Woah, i just had a great idea: read the error output lol. tty isn't working because of mesa driver mismatch. when i just `cargo run`, absolutely NONE of the niri-flake mitigations apply.

`nix run github:sodiboo/niri-flake#niri-unstable --override-input niri-unstable .` works fine as a bit of a hack. it takes a while to run locally.

the results? DmaBuf screencopy works fine, but this *too* is upside-down. so, what do?

sodiboo:
yeah i have no idea what to do here. workaround: `grim img.png; convert img.png -flip img.png` (and then use imagemagick also to crop, i.e. `slurp`, because `grim -g` operates on the flipped coordinate space. doing it like that would result in a correcly oriented image from a rect whose position is flipped across the middle of the screen)

sodiboo:
Reimplemented screencopy but made it run in the main render function.

- new field `pending_screencopy` keeps track of this
- image for winit is magically no longer flipped
- image for dmabuf is also no longer flipped

in keeping track of the output, i also made it easier to keep track of the rest of the attributes on `Screencopy`, instead of discarding them like in catacomb (and yeah, fair enough, phones have one screen and no cursor)

side effect:

- the cursor is now part of the screenshot because we are copying the "real" render buffer.

TODO: `overlay_cursor`

but other than the cursor now being part of screenshot, it's basically done by now, except me not having tested it on multiple outputs (again, can't easily at the moment)

sodiboo:
overlay_cursor was easy enough to implement. the cursor will flicker for a moment now when a screenshot is taken, because this is copying from the actual main render buffer, and not creating some new one somewhere. but i don't think it's too noticeable. With that impemented, i would say that this is pretty much *done*,  besides testing it on multiple outputs

sodiboo:
I fixed the cursor flickering issue in like, maybe one of the ? I just immediately redraw the frame. I originally wanted to skip the frame, but this (A) causes a frame to be skipped (undesirable) and (B) ends up not redrawing always - the display may not refresh until the cursor moves or something. I figure, that this isn't actually that bad, because the other obvious way to do it is to just render to an offscreen buffer that never ends up on the screen, and don't put the cursor on it. That approach also redraws the screen twice in one frame.

sodiboo:
I actually switched my system to this branch, along with updating my locking script (detailed as a use case for screencopy in the original issue). It works great! Though, again, don't have multiple displays to test with, but i think it would work https://github.com/sodiboo/nix-config/commit/94ccc18d10506add126af9edf31a9d27715784fd

src/niri.rs
```diff
@@ -910,6 +913,7 @@ impl Niri {
             ForeignToplevelManagerState::new::<State, _>(&display_handle, |client| {
                 !client.get_data::<ClientState>().unwrap().restricted
             });
+        let screencopy_state = ScreencopyManagerState::new::<State>(&display_handle);
```
YaLTeR, line 916: Needs a filter on `restricted` clients like above.

src/protocols/screencopy.rs
```diff
@@ -0,0 +1,279 @@
...
+                frame,
+                overlay_cursor,
+                x,
+                y,
+                width,
+                height,
+                output,
+            } => {
+                let output = Output::from_resource(&output).unwrap();
+                let rect = Rectangle::from_loc_and_size((x, y), (width, height));
```
YaLTeR, line 96: Is anything gonna break if the client sends invalid coordinates like negative width / height?

src/protocols/screencopy.rs
```diff
@@ -0,0 +1,279 @@
...
+            wl_shm::Format::Argb8888,
+            rect.size.w as u32,
+            rect.size.h as u32,
+            rect.size.w as u32 * 4,
+        );
+
+        if manager.version() >= 3 {
+            // Send desired DMA buffer parameters.
+            frame.linux_dmabuf(
+                Fourcc::Argb8888 as u32,
```
YaLTeR, line 143: I think the "default format" is actually `Abgr8888`, that's what we use for dmabufs in screenshot etc., let's use it here too.

src/protocols/screencopy.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+
+    /// Submit the copied content.
+    pub fn submit(mut self) {
+        // Notify client that buffer is ordinary.
+        self.frame.flags(Flags::empty());
+
+        // Notify client about successful copy.
+        let now = UNIX_EPOCH.elapsed().unwrap();
```
YaLTeR, line 271: We can send the real presentation time here.

src/handlers/mod.rs
```diff
@@ -8,6 +8,7 @@ use std::os::fd::OwnedFd;
 use std::sync::Arc;
 use std::thread;
 
+use log::debug;
```
YaLTeR, line 11:
```suggestion
```

src/niri.rs
```diff
@@ -247,6 +249,7 @@ pub struct OutputState {
     pub lock_render_state: LockRenderState,
     pub lock_surface: Option<LockSurface>,
     pub lock_color_buffer: SolidColorBuffer,
+    pub pending_screencopy: Option<Screencopy>,
```
YaLTeR, line 252: I guess this needs to be a `Vec<Screencopy>` if multiple clients want to copy the same frame.

src/backend/tty.rs
```diff
@@ -1102,13 +1104,114 @@ impl Tty {
...
+
         // Render the elements.
-        let elements = niri.render::<TtyRenderer>(&mut renderer, output, true);
+        let elements = niri.render::<TtyRenderer>(&mut renderer, output, overlay_cursor);
 
         // Hand them over to the DRM.
         let drm_compositor = &mut surface.compositor;
         match drm_compositor.render_frame::<_, _>(&mut renderer, &elements, [0.; 4]) {
             Ok(res) => {
+                if let Some(screencopy) = niri
```
YaLTeR, line 1120: Is there any particular reason why you put this code in both tty.rs and winit.rs? I'd put it in niri.rs in the end of `redraw()`, together with the call to `self.render_for_screen_cast()` which does a conceptually similar thing. It always renders separately, too, so you can always use the correct `overlay_cursor` setting without stepping on the real render's toes. And there's no code duplication between winit and tty.

Yes that means it won't optimize by blitting when appropriate, but I think for the first iteration that's fine (especially for taking 1 screenshot use case), and later I'll try to solve it for both PipeWire and screencopy together.

src/backend/tty.rs
```diff
@@ -1102,13 +1104,114 @@ impl Tty {
...
+                                    && buffer_data.height == region.size.h
+                                    && shm_len as i32 == buffer_data.stride * buffer_data.height)
+                                {
+                                    return Err(anyhow!("the buffer format is incorrect"));
+                                }
+
+                                renderer.with_context(|gl| {
+                                    // SAFETY: i hope this is okay
+                                    unsafe {
+                                        gl.ReadPixels(
```
YaLTeR, line 1159: I think you can use a combination of the same as what `render_helpers::render_to_vec()` does and rendering to a small texture directly with the right `RelocateRenderElement` to put the target square into it.

src/backend/tty.rs
```diff
@@ -1102,13 +1104,114 @@ impl Tty {
...
+                                            region.size.h,
+                                            gles::ffi::RGBA,
+                                            gles::ffi::UNSIGNED_BYTE,
+                                            shm_buffer.cast(),
+                                        );
+                                    }
+
+                                    // Catacomb says: Convert OpenGL's RGBA to ARGB.
+                                    // sodiboo thinks: really? this looks like RGBA -> BGRA to me?
+                                    // (is it also like,
```
YaLTeR, line 1172: Instead of converting here, you just need to advertise the right format in protocols/screencopy.rs, and then use that same format when rendering for it.

src/protocols/screencopy.rs
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        for Rectangle { loc, size } in damage {
+            self.frame
+                .damage(loc.x as u32, loc.y as u32, size.w as u32, size.h as u32);
+        }
+    }
+
+    /// Submit the copied content.
+    pub fn submit(mut self) {
```
YaLTeR, line 266: Suggestion to improve the code: since you're passing by value here, you can remove the `self.submitted` field and always send `failed()` on drop, meanwhile in this function just destructure with `let Screencopy { fields } = self;` which doesn't run `drop()`.

YaLTeR (commented):
Cool! This is quite close to good.

src/protocols/screencopy.rs
sodiboo: Hm, i hadn't even thought about that. The other implementations i looked at don't seem to validate this, but it might be an issue? I'm unsure

src/protocols/screencopy.rs
sodiboo: Isn't that what this does? i.e. "time elapsed since the UNIX epoch"

src/niri.rs
sodiboo: Something like that, yeah. I noticed some differing behaviours in this regard for tty/winit.

To "stress test" the screencopy, i tried running a loop where i call `grim` 12 times. On TTY, this causes the screen to stutter for a while (if i only call it like, say, three times, this is nearly unnoticeable to me - and since this is likely to happen after some "selection of the screen" or such, this is probably okay). However, on winit, it runs fast (???) and i previously observed that it would render the first screenshot fine, the second one sometimes, but the rest were just empty or such. I could not reproduce this behaviour now when i was trying to verify that this really did occur, so either i accidentally fixed it or it was never a real issue. But i'd imagine this kind of thing could be a result of this only storing one `Screencopy`. I didn't implement it as a Vec as i thought this edge case is unlikely, and even if it does occur, you can likely retry.

src/backend/tty.rs
sodiboo: The main reason i put it in both is because the bodies are not identical (i.e. `res` is of a different type in each, the `SyncPoint` is gotten from a different place, and of course on TTY there is the dmabuf implementation). I didn't realize there was an appropriate unified place to put this, though, and i'll see what i can do on this regard. Having a separate rendering function to hook into is most definitely what i want in reality, though, so i'll try to implement it there.

src/protocols/screencopy.rs
YaLTeR: That's the time when the function is called, however we have the precise time from the GPU when the frame is shown on screen.

Actually, come to think of it, this would require sending a screencopy on vblank rather than on redraw (since the presentation timestamp is only available on vblank), so this is fine for now.

src/backend/tty.rs
sodiboo: I don't think i can, as if i change the `wl_shm::Format` sent to the buffer, `grim` doesn't like it (which is like, one of the main consumers of this protocol). Looking at the source, it should handle a bunch of formats just fine, but on my system it only seems to work with `Argb8888`. I think it's easier to just always use the more compatible format, and we can handle converting to and from this format.

YaLTeR: Then you can just use the same format when rendering for screencopy.

src/protocols/screencopy.rs
sodiboo: This doesn't seem to be the case? i.e. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4c08007f797f3559bb36867db18f901f

Potentially, we could do `mem::forget`, which is maybe not desirable? From what i can tell, neither the `ZwlrScreencopyFrameV1` or `WlBuffer` types have a specific `Drop` impl, so it might be just fine, but i think the current way from Catacomb seems more "robust" in not relying on those details

YaLTeR: Huh, apparently you can't destructure types with Drop?

image

Then just leave as is, yeah.

src/protocols/screencopy.rs
sodiboo: ah. didn't catch what "presentation time" means. when i first glanced at this code in catacomb, i thought it was sending just unix epoch, and read your suggestion as just "send the real time".

src/backend/tty.rs
sodiboo: Interesting! Yay for eliminating all unsafe code i "wrote" except for `gl.ReadPixels`

src/backend/tty.rs
sodiboo: Do you mean, we render to a `Vec<u8>` and then copy those bytes? Is there some safe function to map a texture onto a `wl_shm` buffer? I don't understand why else we would do this, and i don't see a significant issue with the current way of doing it (besides that comment of course, but that was written in my "blindly copy and paste" stage - i feel that i understand a bit better now)

YaLTeR: We always render to a GPU texture. Then `render_to_vec()` maps that GPU texture to the main memory and copies it into a Vec; you can copy into the shm WlBuffer instead.

src/niri.rs
sodiboo: I added this functionality, but i don't know how to test it. It seems fairly simple, though, and i've basically just copied what `foreign_toplevel.rs` does

src/backend/tty.rs
sodiboo: Isn't that essentially what i'm already doing, just from a different buffer? As far as i can tell, i think what you're describing is just what `ReadPixels` does? I don't think i fully understand the suggestion.

src/backend/tty.rs
sodiboo: I've implemented it in a similar manner to `render_for_screen_cast`. Now, screencopy is handled at the end of the main render function.

src/render_helpers/mod.rs
```diff
@@ -127,7 +132,75 @@ pub fn render_to_dmabuf(
...
+
+        if let Some(mut damage) = output_rect.intersection(dst) {
+            damage.loc -= dst.loc;
+            element
+                .draw(&mut frame, src, dst, &[damage])
+                .context("error drawing element")?;
+        }
+    }
+
+    let sync_point = frame.finish().context("error finishing frame")?;
```
YaLTeR, line 179: You can replace this and the `gl.ReadPixels` call below with
```rust
    let mapping =
        render_and_download(renderer, size, scale, fourcc, elements).context("error rendering")?;
    let copy = renderer
        .map_texture(&mapping)
        .context("error mapping texture")?;
```
and then use `ptr::copy_to_nonoverlapping()` from `copy` (which is a `&[u8]`) to `shm_buffer`.

src/render_helpers/mod.rs
```diff
@@ -127,7 +132,75 @@ pub fn render_to_dmabuf(
...
+
+    Ok(sync_point)
+}
+
+pub fn render_to_shm(
+    renderer: &mut GlesRenderer,
+    buffer: &WlBuffer,
+    region: Rectangle<i32, Physical>,
+    size: Size<i32, Physical>,
```
YaLTeR, line 144: I was thinking, instead of rendering the full `size` and then copying out the `region`, to render already to a small `region.size`-sized texture, by wrapping `elements` in a `RelocateRenderElement` with relative location of `-region.loc`.

src/render_helpers/mod.rs
```diff
@@ -127,7 +132,75 @@ pub fn render_to_dmabuf(
...
+        }
+    }
+
+    let sync_point = frame.finish().context("error finishing frame")?;
+
+    shm::with_buffer_contents_mut(buffer, |shm_buffer, shm_len, buffer_data| {
+        // The buffer prefers pixels in little endian ...
+        if !(buffer_data.format == wl_shm::Format::Argb8888
+            && buffer_data.stride == region.size.w * 4
+            && buffer_data.height == region.size.h
```
YaLTeR, line 185: Can this part be validated ahead of time, at the point when the client gives us the buffer?

src/niri.rs
```diff
@@ -2509,6 +2524,46 @@ impl Niri {
...
+                screencopy.region(),
+                output.current_mode().unwrap().size,
+                output.current_scale().fractional_scale().into(),
+                transform,
+                elements.into_iter().rev(),
+            ) {
+                Ok(sync_point) => {
+                    screencopy.submit_after_sync(sync_point.export(), &self.event_loop)
+                }
+                Err(e) => warn!("{e:?}"),
```
YaLTeR, line 2562:
```suggestion
                Err(err) => warn!("error rendering to screencopy shm buffer: {err:?}"),
```

src/render_helpers/mod.rs
sodiboo: I did this in https://github.com/YaLTeR/niri/pull/243/commits/005a4626d54f8db3417a2a1bba161614e8a053f3 - but ultimately i feel that it is just worse? Because, now, this function is unsafe. It moves the concern of the buffer into the caller, and for what benefit? Performance? I think this function is much nicer when it validates the buffer internally.

I suppose we could verify it at both points (i.e. remove `debug_assert!()` macro around the condition), then the function still has the same safety guarantees but we exit early in screencopy process if the buffer is malformed.

sodiboo:
> The winit backend does indeed have a problem that sometimes modifiers get stuck when you focus it in/out.

i got so fed up with this when testing wlr-screencopy that i just went and fixed it. it was surprisingly easy to get it to work reliably. #249

sodiboo:
I implemented `RelocateRenderElement` and it works fine with grim because, well, grim doesn't take advantage of this. It just copies the whole screen, then crops on its own. I tried `wayshot`, which is [one of few consumers](https://grep.app/search?q=screencopy_manager%28%5C.%7C_v1_%29capture_output_region&regexp=true) of `capture_output_region`. and found that i was handling scaling incorrectly. I had wrongly read Catacomb's code path for this to be a convoluted "size up, then down again, by the same factor, and round", and as such simplified it to a noop. turns out this was not the case, and without those lines, tools like wayshot don't get the full buffer. So, now with these changes, not only do we render only the parts of the screen that were requested, but i even fixed the `CopyOutputRegion` path, so tools like wayshot work fine.

<details>

<summary>
side tangent: ew wayshot why do you downscale my screenshots. i legit thought i had done something wrong, but no, wayshot really just does blur the image it receives, and i guess assumes we have a scaling factor of 1
</summary>

- storing logical size: https://github.com/waycrate/wayshot/blob/cb6bd68dbbe6ab70a5d8fe3bd04cc154f0631cd8/libwayshot/src/dispatch.rs#L133-L137
- using logical size to compute regions: https://github.com/waycrate/wayshot/blob/cb6bd68dbbe6ab70a5d8fe3bd04cc154f0631cd8/libwayshot/src/lib.rs#L543-L548
- passing logical size to image fixer upper: https://github.com/waycrate/wayshot/blob/cb6bd68dbbe6ab70a5d8fe3bd04cc154f0631cd8/libwayshot/src/lib.rs#L458-L463
- downscaling the image (why!!!): https://github.com/waycrate/wayshot/blob/cb6bd68dbbe6ab70a5d8fe3bd04cc154f0631cd8/libwayshot/src/image_util.rs#L34-L39

</details>

src/render_helpers/mod.rs
sodiboo: Done.

src/render_helpers/mod.rs
sodiboo: I suppose i could. But i don't see why we'd do this. Doesn't this just make the whole operation less efficient, for no discernible gain? If the fact that most of this highlighted block is repeat several times across the file, i did fix that just now.

src/niri.rs
sodiboo: I added this field initially so that it could be accessed during the main render function, but as it stands there is not really any need to do that. Could i just render directly in the screencopy handler now (like i did in the initial implementation)? Is there any reason to prefer doing in the render loop? If we can do it there, then the field can be removed entirely and there is no need for a `Vec`.

YaLTeR: > Could i just render directly in the screencopy handler now (like i did in the initial implementation)? Is there any reason to prefer doing in the render loop?

I think the intent of screencopy is to get the same frame as will be displayed on screen. For that you need to render right after rendering for screen output. Because if you render in the handler, it might be too early: the mouse may still move until a redraw, or new client buffers may still be committed that will be included in the frame that will be shown on the output.

(To be fair, if the compositor is idling, we can still return frames that are too early. But I can worry about this myself later; this is still the right approach for the final solution.)

src/render_helpers/mod.rs
```diff
@@ -101,14 +85,69 @@ pub fn render_to_dmabuf(
...
+                size.w,
+                size.h,
+                // ... but OpenGL prefers them in big endian.
+                gles::ffi::BGRA_EXT,
+                gles::ffi::UNSIGNED_BYTE,
+                shm_buffer.cast(),
+            )
+        })?;
+
+        Ok(sync_point)
```
YaLTeR, line 136: `glReadPixels` is synchronous, it won't return until the operation completes. So the sync point is useless here (in this case, `glReadPixels` itself will wait for that sync point before copying data to the main memory).

src/render_helpers/mod.rs
```diff
@@ -101,14 +85,69 @@ pub fn render_to_dmabuf(
...
+    })?
+}
+
+fn render_elements(
```
YaLTeR, line 140: Nice refactor!

src/render_helpers/mod.rs
YaLTeR: > I suppose we could verify it at both points (i.e. remove debug_assert!() macro around the condition), then the function still has the same safety guarantees but we exit early in screencopy process if the buffer is malformed.

Yes, I think that's a good approach.

src/render_helpers/mod.rs
YaLTeR: Two things that come to mind:

1. Going through `Renderer` sets up other things in the GL context properly, that may or may not be important.
2. I wouldn't be so sure it would be less efficient. The methods used by `render_and_download()` do a `glReadPixels` into a pixel buffer object, then map it into the main memory. I've seen enough GPU weirdness that I wouldn't be surprised if that is actually faster (or at least not slower) than `glReadPixels` directly into the main memory. Ideally should be profiled though.

YaLTeR:
>  side tangent: ew wayshot why do you downscale my screenshots. i legit thought i had done something wrong, but no, wayshot really just does blur the image it receives, and i guess assumes we have a scaling factor of 1 

Gonna do a blind guess here that they did it to simplify handling of multi-monitor setups with different scaling factors. Not how I'd solve that problem though..

sodiboo:
Other consumers

- wf-recorder handles it correctly https://github.com/ammen99/wf-recorder/blob/c43e8e11ec99e72d8c137fa2ea7e4ddb934275bd/src/frame-writer.cpp#L730-L734
- wl-mirror handles it correctly https://github.com/Ferdi265/wl-mirror/blob/0dc473ac66bff697e52e5f70a7ef37c2e45bd4b0/src/mirror-screencopy.c#L323-L326

src/niri.rs
sodiboo: The previously observed (and magically disappeared) behaviour i described yesterday is probably a side effect of an unrelated bug. In that instance, grim "succeeded" 12 times with sometimes garbage data (this has since gone away). But if i change it to fork each time, then like half of them fail with a non-zero exit code.  Changing `Option` to a `Vec` here allows 12 forked grims to successfully capture the same frame.

An optimization i could imagine, is if there are multiple screencopies with different regions, it might be more efficient to render the whole frame once and only copy different regions. But, that also feels like a case of severe premature optimization. The only realistic use case for many screencopies frequently enough that this matters is screen recording, but then this doesn't apply because they're always on different frames.

YaLTeR:
Yea wayshot can fix it themselves. Especially since it only affects the winit backend.

src/niri.rs
```diff
@@ -2531,10 +2531,11 @@ impl Niri {
         output: &Output,
         y_invert: bool,
     ) {
-        if let Some(mut screencopy) = self
+        for mut screencopy in self
             .output_state
             .get_mut(output)
-            .and_then(|s| s.pending_screencopy.take())
+            .map(|s| s.pending_screencopy.drain(..).collect::<Vec<_>>())
```
YaLTeR, line 2537: Uh, no extra collect() pls. mem::take()

sodiboo: oh yeah. originally i thought mem::replace with default seems verbose. that's why i did this instead. mem::take is much more appropriate here, though

src/render_helpers/mod.rs
sodiboo: Thanks!

src/render_helpers/mod.rs
sodiboo: Hm. so it does seem. is there an equivalent asychronous call? Should the `glReadPixels` be moved into *after* the `SyncPoint`? Does it really matter?

YaLTeR: The correct thing to do here is to simply do nothing with the sync point and return None. The sync point should be used only for the dmabuf buffer type.

src/protocols/screencopy.rs
sodiboo: in debug mode: niri crashes. smithay has debug asserts to enure the size should never be negative.
in release mode: i believe the cast to `u32` results in niri telling the client that the buffer should allocate  buffer with approximately `u32::MAX - width` width (or vice versa for height, stride). niri does not crash, but the client i was testing this with (modified wayshot to allocate a negative width) does, and niri logs an error in client communication. if the client somehow detects this and allocates a better buffer, niri will reject it for being wrong (comparing to the requested region).. if the client somehow allocates the requested buffer without dying, then both the height *and* width must be negative for their product to ever be positive.

i'm gonna normalize the requested region such that negative width/height means the region extents up and left from the given point.

additionally, what about negative/offscreen coordinates, such that the resulting region is not within screen bounds? well, for these, they are simply cropped to the screen bounds, and niri will request a buffer of different specs than the given width/height that the client asked for.

src/render_helpers/mod.rs
sodiboo: From what i can tell reading various documentation, i'd imagine you might be right that this operation in itself is faster. However, there is another difference. The current implementation uses a `GlesRenderbuffer`, which from what i can tell is generally more efficient to actually render to. So, which is actually faster? Rendering to a texture, mapping the texture, and then `ptr::copy_nonoverlapping` or rendering to a Renderbuffer and then `glReadPixels`?  I don't know. Neither is excessively slow for me, so i don't think the difference is that significant, but i've pushed a commit which implements the alternative behaviour under `render_to_shm_alt`. It's a drop-in replacement for the other function, feel free to profile it if you want. If it's genuinely faster, then i agree we should use this instead.

src/protocols/screencopy.rs
YaLTeR: Doesn't it make more sense to just fail the request if width or height are <= 0?

YaLTeR:
Do you have more things planned for this PR? I think I could finish it myself tomorrow morning if you don't mind.

sodiboo:
I was thinking, potentially, we can actually uncap the width of the screenshot? i.e. allow it to go outside the monitor bounds. Because, unlike i.e. a tiling manager, in niri it actually makes sense to refer to coordinates that are offscreen. You just get the windows that are scrolled off that side of the screen.

You could interpret out of bounds coordinates as meaning "the screen in that direction" (i.e. wayshot seems to create its own coordinate system based on monitor layouts), but this is something that is just straight up not supported by wlr-screencopy. it's in the name, it copies a screen! so i don't think this alternative interpretation is applicable to wlr-screencopy in particular.

I think it's slightly harder to implement vertically (i.e. allow screenshots of "multiple workspaces"), but if we uncapped *both* in this manner, then we could totally make a `niri msg` subcommand that prints the output and region for a specific visible window (in this specific frame, of course not stable)? The scripting capabilities of that could be pretty awesome. Even without the subcommand, it's still quite useful if i want to take a wide screenshot.

Alternatively, to make less "screencopy-dependent", such a subcommand could instead include output like the currently selected workspace, the scroll position in logical pixels in the workspace, and the position of that window in its own workspace, and which workspace it is in.

That's not necessarily part of this PR. But the uncapped bounds totally are.

YaLTeR:
I think this uncapping definitely sounds out of scope for wlr-screencopy. However, it could be something for niri msg indeed.

sodiboo:
I don't think it feels very out of scope. It's just a removal of 7 lines. With something like grim, this doesn't actually let you screenshot off-screen elements (it does not take advantage of regions, rather it handles them client-side), but Wayshot handles this case just fine. Ultimately, no sane client (except one made specifically for this) will end up requesting an out of bounds region. And of those that do, they likely don't expect the buffer to come in the wrong shape compared to the region they expected. Grim handles this by filling white in the rest of the area, while wayshot will distort the image into the resolution it wanted.

But, just removing the caps isn't the whole story. If i do that and take screenshots out of bounds, the background isn't completely filled in. Workspaces don't render on top of each other, unless you're switching. layer-shell components cover the windows while switching workspace. On winit, sometimes out of bounds windows are of the incorrect width (i.e. when sized as proportion and actual output size changes). There are just several issues that make it overall not very pleasant to use. So i don't think i'll implement that, as making the out-of-bounds screencopies look acceptable *is* out of scope.

Besides that, i guess i don't have any more plans for this PR. Basically done. I already run it on my own system.

YaLTeR:
> If i do that and take screenshots out of bounds, the background isn't completely filled in.

Yeah, naturally. That can't really work properly in many cases.

> On winit, sometimes out of bounds windows are of the incorrect width (i.e. when sized as proportion and actual output size changes).

You will also see stale window contents in general. That's because niri doesn't send frame callbacks to invisible surfaces to conserve CPU, so they don't update.

YaLTeR:
Actually, the protocol says:

> The region will be clipped to the output's extents.

YaLTeR:
I did a *very quick and dirty* benchmark and both ways of render_to_shm appear to take approximately the same time on my current machine.

YaLTeR:
I pushed my changes (quite extensive; all in one commit, sorry). They mainly fix working with output transforms (turns out, grim expects physical buffers, then transforms them by itself using transform info from wl_output). But also, importantly, I lowered the wlr-screencopy version to 1, because it seems that copy_with_damage() has different semantics from copy() in terms of when it should return the buffer: copy() should return immediately, whereas copy_with_damage() should wait until the next successful VBlank. The result is that grim still works, but some other tools I tried don't, because they generally request version 3 even if they don't actually need to use copy_with_damage(). I think this should be ok for the first implementation? I imagine grim covers the majority of uses here.

sodiboo:
hm. did not realize "review" button resulted in a feed entry which looks that important. this "started a review" i guess should've been a comment.

PR #245, opened 2024-03-05
Mouse warp

FluxTape:
Adds a mouse warp (aka mouse-follows-focus) config option. It can be activated by adding `warp-mouse-to-focus` to the input section of the config files.
There are some minor issues with this basic implementation documented with `FIXME` comments in `input.rs`, in particular:
1. `queue_redraw_all()` is sometimes called twice, which is unnecessary. I've decided to not fix this for now as more granular updating of the display is planned anyway.
2. the cursor currently doesn't warp to the focused window when spawning, closing or resizing windows, as a naive implementation doesn't work. To fix this the mouse warp code would need to be executed after some delay (when resizing/spawning/closing has finished), or by pre-calculating the final layout.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -67,6 +67,9 @@ input {
     // Uncomment this if you would like to configure the power button elsewhere
     // (i.e. logind.conf).
     // disable-power-key-handling
+
+    // Uncomment this to make the mouse warp to the center of newly focused windows.
+    // warp-mouse-to-focus
 }
 
 // You can configure outputs by their name, which you can find
--- a/src/input.rs
+++ b/src/input.rs
@@ -411,139 +411,166 @@ impl State {
             }
             Action::MoveColumnLeft => {
                 self.niri.layout.move_left();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveColumnRight => {
                 self.niri.layout.move_right();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveColumnToFirst => {
                 self.niri.layout.move_column_to_first();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveColumnToLast => {
                 self.niri.layout.move_column_to_last();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowDown => {
                 self.niri.layout.move_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowUp => {
                 self.niri.layout.move_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowDownOrToWorkspaceDown => {
                 self.niri.layout.move_down_or_to_workspace_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowUpOrToWorkspaceUp => {
                 self.niri.layout.move_up_or_to_workspace_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::ConsumeOrExpelWindowLeft => {
                 self.niri.layout.consume_or_expel_window_left();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::ConsumeOrExpelWindowRight => {
                 self.niri.layout.consume_or_expel_window_right();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusColumnLeft => {
                 self.niri.layout.focus_left();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusColumnRight => {
                 self.niri.layout.focus_right();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusColumnFirst => {
                 self.niri.layout.focus_column_first();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusColumnLast => {
                 self.niri.layout.focus_column_last();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWindowDown => {
                 self.niri.layout.focus_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWindowUp => {
                 self.niri.layout.focus_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWindowOrWorkspaceDown => {
                 self.niri.layout.focus_window_or_workspace_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWindowOrWorkspaceUp => {
                 self.niri.layout.focus_window_or_workspace_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowToWorkspaceDown => {
                 self.niri.layout.move_to_workspace_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowToWorkspaceUp => {
                 self.niri.layout.move_to_workspace_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveWindowToWorkspace(idx) => {
                 let idx = idx.saturating_sub(1) as usize;
                 self.niri.layout.move_to_workspace(idx);
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveColumnToWorkspaceDown => {
                 self.niri.layout.move_column_to_workspace_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveColumnToWorkspaceUp => {
                 self.niri.layout.move_column_to_workspace_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::MoveColumnToWorkspace(idx) => {
                 let idx = idx.saturating_sub(1) as usize;
                 self.niri.layout.move_column_to_workspace(idx);
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWorkspaceDown => {
                 self.niri.layout.switch_workspace_down();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWorkspaceUp => {
                 self.niri.layout.switch_workspace_up();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::FocusWorkspace(idx) => {
                 let idx = idx.saturating_sub(1) as usize;
                 self.niri.layout.switch_workspace(idx);
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
@@ -559,11 +586,14 @@ impl State {
             }
             Action::ConsumeWindowIntoColumn => {
                 self.niri.layout.consume_into_column();
+                // This does not cause immediate focus or window size change, so warping mouse to
+                // focus won't do anything here.
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::ExpelWindowFromColumn => {
                 self.niri.layout.expel_from_column();
+                self.maybe_warp_cursor_to_focus();
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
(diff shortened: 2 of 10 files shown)


resources/default-config.kdl
```diff
@@ -66,6 +66,9 @@ input {
     // Uncomment this if you would like to configure the power button elsewhere
     // (i.e. logind.conf).
     // disable-power-key-handling
+
+    // Uncomment this to make the mouse warp to the focused window
```
YaLTeR, line 70:
```suggestion
    // Uncomment this to make the mouse warp to the focused window.
```

src/input.rs
```diff
@@ -390,12 +391,17 @@ impl State {
...
+                    //FIXME: move_cursor_to_focused_tile doesn't work here
                 }
             }
             Action::FullscreenWindow => {
                 let focus = self.niri.layout.focus().cloned();
                 if let Some(window) = focus {
                     self.niri.layout.toggle_fullscreen(&window);
+                    if self.niri.config.borrow().input.warp_mouse_to_focus {
+                        self.move_cursor_to_focused_tile(CenteringMode::CoordsIndividually);
+                        //FIXME: usually calls queue_redraw_all (unnecessary)
```
YaLTeR, line 403:
```suggestion
```
I think we can get by without this comment everywhere

src/niri.rs
```diff
@@ -299,6 +299,11 @@ struct SurfaceFrameThrottlingState {
     last_sent_at: RefCell<Option<(Output, u32)>>,
 }
 
+pub enum CenteringMode {
+    CoordsIndividually,
+    IfNeeded,
```
YaLTeR, line 304: I see what you were going for with naming here, but the result is kinda weird. I suggest changing this to something like `CenterCoords::{Separately, Both}`.

src/niri.rs
```diff
@@ -386,6 +391,47 @@ impl State {
         self.niri.queue_redraw_all();
     }
 
+    // moves cursor within the specified rectangle, only adjusting coordinates if needed
+    fn move_cursor_within_rect(&mut self, rect: Rectangle<f64, Logical>, mode: CenteringMode) {
```
YaLTeR, line 395:
```suggestion
    fn move_cursor_to_rect(&mut self, rect: Rectangle<f64, Logical>, mode: CenteringMode) {
```

src/layout/workspace.rs
```diff
@@ -1077,6 +1086,20 @@ impl<W: LayoutElement> Workspace<W> {
         first.chain(rest)
     }
 
+    // location of the active tile in global visual coordinates
+    pub fn active_tile_visual_pos(&self) -> Point<i32, Logical> {
+        let output_loc = self
```
YaLTeR, line 1091: This output loc should be added in the function on Monitor, which always has an output.

src/layout/monitor.rs
```diff
@@ -575,6 +575,18 @@ impl<W: LayoutElement> Monitor<W> {
...
     }
 
+    // returns a rectangle describing where the focused tile is located in visual coordinates
+    pub fn active_tile_visual_rectangle(&mut self) -> Option<Rectangle<i32, Logical>> {
+        let workspace = self.active_workspace();
+        if workspace.columns.is_empty() {
+            return None;
+        }
+        let col = &mut workspace.columns[workspace.active_column_idx];
+        let tile_size = col.tiles[col.active_tile_idx].tile_size();
```
YaLTeR, line 585: This sounds like it should be inside the function on Workspace.

src/layout/monitor.rs
```diff
@@ -575,6 +575,18 @@ impl<W: LayoutElement> Monitor<W> {
...
+        let tile_pos = workspace.active_tile_visual_pos();
```
YaLTeR, line 586: This should probably take the workspace switch gesture into account?

src/layout/mod.rs
```diff
@@ -1219,6 +1219,14 @@ impl<W: LayoutElement> Layout<W> {
         mon.window_under(pos_within_output)
     }
 
+    // returns a rectangle where the currently focused tile is located (on the active output)
+    pub fn active_tile_visual_rectangle(&mut self) -> Option<Rectangle<i32, Logical>> {
+        let Some(monitor) = self.active_monitor() else {
+            return None;
+        };
+        monitor.active_tile_visual_rectangle()
```
YaLTeR, line 1227: What happens if this returns a rectangle out of bounds? For example, a window can be bigger than the workspace, and if it's 2x view size in width, the center will be out of bounds. Or, when you take the workspace switch into account, you can switch focus with keybinds while holding a touchpad workspace switch gesture. I guess some clipping should be done, or just returning None in that case?

src/input.rs
```diff
@@ -334,6 +334,7 @@ impl State {
             }
             Action::Spawn(command) => {
                 spawn(command);
+                //FIXME: move_cursor_to_focused_tile doesn't work here
```
YaLTeR, line 337: I guess instead of all these warps here, you could check if the active window changed somewhere around `refresh_and_flush_clients()`. That will also get spawning and closing and everything else, and remove all the code duplication.

It will also be a good place to put checks for when we don't want to move the mouse, e.g. when the screenshot UI is open.

src/niri.rs
FluxTape: fixed in commit [123c2f0](https://github.com/YaLTeR/niri/pull/245/commits/123c2f040e1891c7ce5d3d7c29d497f48da788cd)

src/niri.rs
FluxTape: fixed in commit [123c2f0](https://github.com/YaLTeR/niri/pull/245/commits/123c2f040e1891c7ce5d3d7c29d497f48da788cd)

src/layout/workspace.rs
FluxTape: fixed in commit [123c2f0](https://github.com/YaLTeR/niri/pull/245/commits/123c2f040e1891c7ce5d3d7c29d497f48da788cd)

src/layout/monitor.rs
FluxTape: fixed in commit [123c2f0](https://github.com/YaLTeR/niri/pull/245/commits/123c2f040e1891c7ce5d3d7c29d497f48da788cd)

src/layout/mod.rs
FluxTape: hopefully fixed by clipping in commit [123c2f0](https://github.com/YaLTeR/niri/pull/245/commits/123c2f040e1891c7ce5d3d7c29d497f48da788cd), however I am not able to test my updated code with multiple monitors/outputs at the moment

src/input.rs
FluxTape: implemented in [5310e88](https://github.com/YaLTeR/niri/pull/245/commits/5310e88ab540d38ed71afadbe33c7ed390f70935), again was not able to test multi monitor configuration yet.

src/layout/monitor.rs
FluxTape: I'm not quite sure what to change, to me it seems to work fine as is.

src/input.rs
```diff
@@ -645,163 +562,116 @@ impl State {
...
-                    self.move_cursor_to_focused_tile(CenterCoords::Seperately);
-                }
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
             Action::SwitchPresetColumnWidth => {
                 self.niri.layout.toggle_width();
-                //FIXME: move_cursor_to_focused_tile doesn't work here (reacts to old size)
+                if self.niri.config.borrow().input.warp_mouse_to_focus {
+                    self.move_cursor_to_focused_tile();
```
YaLTeR, line 576: Is it needed on these actions? The focus doesn't change here after all. (And either way, resizes will wait for the client to send a new buffer.)

FluxTape: I guess it depends on your workflow whether this is desirable or not. Most of the time nothing will happen anyway. However when resizing a window to a smaller size it can happen that your cursor ends up outside the focused window. Personally I find it convenient if in those cases the cursor warps back inside the focused window. From my testing resize actions don't cause a keyboard or pointer focus change (even in the cases where the cursor ends up outside the focused window), that's why I explicitly added the calls to `move_cursor_to_focused_tile()` here.

src/layout/mod.rs
FluxTape: Tested on dual monitor setup and works as expected

src/input.rs
FluxTape: Tested on dual monitor setup and also works as expected

src/input.rs
YaLTeR: > From my testing resize actions don't cause a keyboard or pointer focus change (even in the cases where the cursor ends up outside the focused window)

Hm, when the window does submit a new buffer (which is not immediately after a resize action) and the pointer ends up somewhere else, the pointer focus should change.

> that's why I explicitly added the calls to move_cursor_to_focused_tile() here.

Similarly to above, I'm not sure this will do anything because the window will only submit a new, resized buffer sometime later (if it does at all).

YaLTeR: Tracking things like "niri did a resize action => the window resized *because of that* (and not because it decided to resize on its own)" is not trivial but planned to some extent (will be needed for animations). Then maybe you would be able to put the "cursor ended up outside" check on top of that. For now might be better to do nothing here?

FluxTape: > Similarly to above, I'm not sure this will do anything because the window will only submit a new, resized buffer sometime later (if it does at all).

Weirdly enough it does fix it (at least on my machine), but I agree it's a bit of a hack and should probably be left out for now.

FluxTape: removed in commit #[94977a0](https://github.com/YaLTeR/niri/pull/245/commits/94977a016963deb9aa2f3dcce15a525357808411)

FluxTape:
Is there anything else that needs to be fixed before this can be merged?

YaLTeR:
Sorry, I've just been slowly reviewing other PRs. Do you mind if I make changes to this PR before merging?

FluxTape:
I don't mind at all as long as the basic workflow/behavior stays the same.

YaLTeR:
Taking a look at this now.

Question: why the coord-separate centering mode? If I'm not mistaken, in sway it always centers both coordinates, and it kind of seems to make more sense to me.

YaLTeR:
Also as far as I can tell, on sway, resizing the window in such a way that the pointer ends up outside does not cause it to be warped back in. Is my assumption correct that the separate-coordinate mode is mainly for this case? If so, is it ok if it only happens in this case, and changing focus always uses both coordinates?

YaLTeR:
When switching workspaces and scrolling the view, right now the pointer moves right away, as the animation starts. This actually causes it to end up on top of a different surface for a brief moment (whichever happened to be at the target location at the start of the animation). Would it be ok to instead warp the pointer only at the end of the animation? This would also be consistent with how the pointer focus changes normally.

YaLTeR:
When switching focus to a different output, should the cursor be centered in the output or in the focused window on that output? It seems that centering in the output makes more sense in at least a few cases, like when clicking the mouse on the empty area of the output, or when focus-follows-mouse is implemented and triggers upon moving the mouse into an output. In both of those cases centering in an output will do nothing since the pointer is already in the output, but this is expected (the mouse shouldn't jump in those cases). Currently both of those cases would result in the mouse jumping into the window.

FluxTape:
> Taking a look at this now.
> 
> Question: why the coord-separate centering mode? If I'm not mistaken, in sway it always centers both coordinates, and it kind of seems to make more sense to me.

This is something I added (and is quite important to me, although it is personal preference) because I found when actually using the feature that in a lot of cases it is easier to follow the pointer movement with the eyes if just one coordinate changes (in cases where that is possible). Also consider for example one common use case of switching back and forth between two columns:  if you don't need to move the pointer on the second column (for example because it's a terminal window or similar) and switch back to the first column your mouse will end up in a similar position to where you started. I can understand this behavior may seem a bit odd at first, but when actually using the feature, I just found it nicer to use this way. I don't mind whether the cursor uses coord-seperate centering on other actions however: for example I think always centering when switching workspaces would be ok too.

So in conclusion if you thnink the behavior should be different and always center on the focused window I would prefer if the coord-seperate centering mode (the current behavior) also stayed an option.

FluxTape:
> When switching workspaces and scrolling the view, right now the pointer moves right away, as the animation starts. This actually causes it to end up on top of a different surface for a brief moment (whichever happened to be at the target location at the start of the animation). Would it be ok to instead warp the pointer only at the end of the animation? This would also be consistent with how the pointer focus changes normally.

Sure either way is fine imo.

FluxTape:
> When switching focus to a different output, should the cursor be centered in the output or in the focused window on that output? It seems that centering in the output makes more sense in at least a few cases, like when clicking the mouse on the empty area of the output, or when focus-follows-mouse is implemented and triggers upon moving the mouse into an output. In both of those cases centering in an output will do nothing since the pointer is already in the output, but this is expected (the mouse shouldn't jump in those cases). Currently both of those cases would result in the mouse jumping into the window.

I agree the edge cases you mentioned are a problem. Especially the one with `focus-follows-mouse`. As you said it's probably best to just center on the output for now. If that ends up being annoying, a more fine-grained solution could be implemented in a followup PR.

YaLTeR:
After testing it more and adding checks for some edge cases, I honestly really want to go back from automatic warping on pointer/keyboard focus change to manually invoking warping on actions. There are just way too many edge cases involving windows resizing then opening layer-shell / screenshot UI / lock surface, it's not really possible to cleanly handle this.

Also, I found one really nasty edge case:
1. Move pointer outside focused window
2. Resize the focused window (this does not cause a warp because the pointer was already outside)
3. Have something appear underneath the pointer on its own (i.e. a constantly resizing window will do)

This will cause the pointer to warp to the focused window because:
1. The pointer focus had changed, triggering the warp check
2. The warp check succeeds because we resized the window earlier

However, this is undesired because it is not the original window resize that changed the pointer focus, and because the pointer was completely outside anyway.

This is once again not really possible to solve cleanly *at the moment*.

Therefore I suggest that I change things back to manual trigger on every action as appropriate. I will see if I can figure out a way to keep warping on resize, and I'll try to come up with a way to warp on spawn.

YaLTeR:
What's the expected behavior for moving the focused window (i.e. Mod+Ctrl+Left)? Only warp the pointer if it was inside and ended up outside, or also warp if the pointer was already outside?

YaLTeR:
I'm guessing always warp since that's consistent with changing focus.

Another case: when to warp if the focus changes due to the focused window closing? I'm guessing warping is desired in this case at least sometimes, so we'll have to accept the warp even if the client just randomly died and we were scrolling something unrelated in another unfocused window.

FluxTape:
> I'm guessing always warp since that's consistent with changing focus.

That would have been my answer as well, however I can also see the argument for the other behavior.

> Another case: when to warp if the focus changes due to the focused window closing? I'm guessing warping is desired in this case at least sometimes, so we'll have to accept the warp even if the client just randomly died and we were scrolling something unrelated in another unfocused window.

Yes I believe the case where the window was closed purposefully is a lot more common than the unintentional quit.

YaLTeR:
I remade it to manual and added warping on new window and on window close. I decided to postpone warping on resize until we can track that a resize originated from a user action; otherwise it's also getting quite edge-case-y I feel. However, I made it so warping works even if the focus action doesn't change focus (i.e. you press focus up while already on the topmost window), so that can help fix focus after resizing.

I think I got it working well now, things like focus monitor will jump straight to new window without a one frame stop at the output's center. When anything else has keyboard focus (layer-shell, screenshot UI), warping does nothing as intended.

Could you give this a try please?

YaLTeR:
I think warping on window close can be a bit disorienting, in cases like: clicking the X button to close a window, then your mouse goes somewhere else suddenly. Is that an OK trade-off here?

YaLTeR:
I guess it's ok and I just need a few mins getting used to it

YaLTeR:
Thanks again! I've been test driving this and it seems to work well. Merging as I've got more things in the queue now blocked by this.

PR #249, opened 2024-03-06
Workaround winit modifiers issue

sodiboo:
This fixes an issue described in the code comment. It is not a "permanent" fix, but it works just fine for me. The underlying issue is either in xkbcommon, smithay, or the niri compositor (i.e. not the winit backend, but the "main" one running in tty) depending on who exactly you want to blame (xkbcommon for entering a weird irreversible state, smithay for not handling this properly, or niri for not sending the correct events to its client).

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -13,6 +13,7 @@ use smithay::backend::input::{
     TabletToolProximityEvent, TabletToolTipEvent, TabletToolTipState, TouchEvent,
 };
 use smithay::backend::libinput::LibinputInputBackend;
+use smithay::input::keyboard::xkb::keysym_get_name;
 use smithay::input::keyboard::{keysyms, FilterResult, Keysym, ModifiersState};
 use smithay::input::pointer::{
     AxisFrame, ButtonEvent, CursorImageStatus, GestureHoldBeginEvent, GestureHoldEndEvent,
@@ -244,15 +245,54 @@ impl State {
     fn on_keyboard<I: InputBackend>(&mut self, event: I::KeyboardKeyEvent) {
         let comp_mod = self.backend.mod_key();
 
-        let serial = SERIAL_COUNTER.next_serial();
         let time = Event::time_msec(&event);
         let pressed = event.state() == KeyState::Pressed;
+        let keyboard = self.niri.seat.get_keyboard().unwrap();
 
-        let Some(Some(action)) = self.niri.seat.get_keyboard().unwrap().input(
+        // When running in the winit backend, we don't always get release events.
+        // Specifically, you can reproduce this issue reliably like so:
+        //
+        // 1. Open niri nested in niri
+        // 2. Focus nested niri
+        // 3. Press any modifier key, e.g. "Super" (and do not release)
+        //   -> nested niri gets a KeyState::Pressed event
+        // 4. Click with the mouse cursor on another window in the top-level compositor
+        // 5. Release the modifier key
+        //   -> nested niri gets *no* event
+        // 6. Focus nested niri again
+        //
+        // After you do this, xkbcommon will be in an annoying state,
+        // where it reports that modifier as being permanently pressed.
+        // (or at least, as far as smithay is concerned)
+        // Even release events do not "unpress" this modifier.
+        //
+        // As such, we detect this *before* it happens, like so.
+        if pressed && keyboard.pressed_keys().contains(&event.key_code().into()) {
+            // Then, we insert a fake KeyState::Released event.
+            keyboard
+                .input(
+                    self,
+                    event.key_code(),
+                    KeyState::Released,
+                    SERIAL_COUNTER.next_serial(),
+                    time,
+                    |_, _, keysym| {
+                        debug!(
+                            "Duplicate keypress of `{key}`. A fake release event was generated.",
+                            key = keysym_get_name(keysym.modified_sym())
+                        );
+                        // And lastly, we tell smithay to keep its mouth shut about this.
+                        FilterResult::Intercept(())
+                    },
+                )
+                .unwrap();
+        }
+
+        let Some(Some(action)) = keyboard.input(
             self,
             event.key_code(),
             event.state(),
-            serial,
+            SERIAL_COUNTER.next_serial(),
             time,
             |this, mods, keysym| {
                 let bindings = &this.niri.config.borrow().binds;


kchibisov:
The bug is in smithay and should be fixed there. In particular, no one will ever guarantee to send release events, especially given that it's wayland and you're not aware of things like that.

kchibisov:
@YaLTeR see above.

sodiboo:
Why was this never merged? I agree, that it's an issue in Smithay and not in niri, but it's also less trivial to fix in Smithay and it's not like the issue goes away by saying "It shouldn't be fixed here". Its been like four months, and this is a persistent annoyance for compositor development. Many times I've manually added this patch when working on something else, just to remove it before sending the PR. Is nobody else as annoyed by this existing behaviour as I am?

YaLTeR:
I guess because it's kinda cursed and while annoying, it's more or less work-aroundable (I try not to switch away from the nested window while holding Alt..). It also allocates an extra HashSet in that `pressed_keys()`. Though, I agree it's quite bad and unobvious for people just trying niri.

Could some simple workaround be implemented in Smithay maybe? I haven't looked into it myself. I remember someone did look into it some time ago in the Smithay Matrix room, actually, but I guess that never made it to pull requests?

sodiboo:
> it's more or less work-aroundable (I try not to switch away from the nested window while holding Alt..).

That's the thing though, it's kinda not possible to workaround just by being careful - because practically any keyboard-based window movement in the toplevel compositor will lead to this issue, since the sequence goes `Press(Super)`, then the arrow key input is supressed, and the release is never caught by the nested compositor.

The only way i can reliably prevent the issue from occurring (without an actual change to the code) is to rely entirely on mouse navigation to move focus in/out of the nested compositor. Which is... not exactly convenient when i'm iterating on code changes and using mostly just a terminal and a text editor. And more importantly, *my muscle memory doesn't allow me to do this 100% of the time*.

The reason i commented now in particular is because it *did* just bite me again but not even while working on niri, rather while working on a *client* that i was using nested niri to test (because it was working with session-lock and i can't debug the application if my real session is locked). So this bug doesn't even *only* affect niri contributors, but also niri users testing clients in a nested niri.

> It also allocates an extra HashSet in that `pressed_keys()`.

That's... an interesting observation i didn't think of or realize. This could probably be fixed in Smithay by changing the internal representation from `HashSet<u32>` to `HashSet<Keycode>` since like, [the actual implementation of that struct is `#[repr(transparent]`](https://docs.rs/xkeysym/0.2.1/src/xkeysym/lib.rs.html#48) to `u32`. Is it that problematic for us? This doesn't exactly run on every frame, only keypresses, and only on the downstroke.

kchibisov:
I mean, the only issue I can think of that winit doesn't do anything about is that it doesn't forward xkb state. However winit does forward a modifier state, which is just discarded by smithay entirely. So it has an option to at _least_ fixup the modifiers based on their common names... Probably will work in 99% of the cases, since users tend to not press both same modifiers at the same time when focusing a window.

Like I'm not sure what the issue with fixing it in smithay really is, when it has all the information to fix this.

YaLTeR:
> practically any keyboard-based window movement in the toplevel compositor will lead to this issue, since the sequence goes Press(Super), then the arrow key input is supressed, and the release is never caught by the nested compositor

Actually, Super binds shouldn't trigger this because the suppressed keys mechanism prevents the modifiers from reaching the client if they formed a host niri bind. I.e. pressing Super+Right which triggered some niri bind shouldn't send the Super press or release to the app.

sodiboo:
> pressing Super+Right which triggered some niri bind shouldn't send the Super press or release to the app.

But... it does? Super on its own doesn't form a full bind, and as such the downstroke isn't supressed. The upstrokes are to my understanding just exactly mirrored so the release would also not be supressed (but it is sent to the newly focused client, rather than the previous client).

also uhh HEY GUESS WHAT. while trying to reproduce this just now, i FOUND A CONSISTENT WORKAROUND.

for one, to observe the issue you can open a terminal in the nested compositor, press Super+Right to unfocus the nested compositor, then click back in it and try to type in the terminal. you will get stuff like `[100;9u`.

but HERE'S THE WORKAROUND:

1. Click outside of the compositor.
2. Focus the compositor using the Super+Arrow binds (such that it only receives downstroke events)
3. If the issue persists, go to step 1

This makes a lot of sense! xkbcommon keeps track of how many times the key has been pressed so that the modifier state means "this modifier is pressed at least once somewhere on any keyboard". This is how i thought it ought to work back when i made this PR, but somehow i didn't realize THAT IS HOW IT WORKS YOU DUMMY.

Either niri or Smithay has somehow indirectly fixed a bug without noticing, or i have completely misdiagnosed the issue in the past. I wrote that [the modifier is reported as permanently pressed](https://github.com/YaLTeR/niri/blob/259f98e6eb840a3b2497f5ff25afe8d17115f8b5/src/input.rs#L265) which is... Not True. because i could get the nested compositor out of the buggy state just now.

---

I think a real way to fix this is to probably [rely on the modifiers event](https://wayland.app/protocols/wayland#wl_keyboard:event:modifiers) sent by the parent compositor to compute the correct modifiers to use within the nested compositor and probably just forward that as-is to nested clients. This... isn't a Smithay bug after all? Maybe?

Either way, now that i have a usable workaround in practice that i can methodically execute, a programmatic hacky workaround is no longer as necessary as i thought it was. A more proper solution can be thought out an implemented in another PR; but it seems that this should be mainly fixed in niri.

YaLTeR:
> But... it does? Super on its own doesn't form a full bind, and as such the downstroke isn't supressed.

Uhh, actually good point, I forgot how exactly it works. But for some reason for me Super navigation doesn't generally break it? Maybe alacritty just ignores held Super or I'm otherwise lucky?

> HERE'S THE WORKAROUND

Oh so this actually works? I was sure I tried this in the past with no luck, huh.

> I think a real way to fix this is to probably rely on the modifiers event sent by the parent compositor to compute the correct modifiers to use within the nested compositor and probably just forward that as-is to nested clients. This... isn't a Smithay bug after all? Maybe?

As far as I understand, the issue is that the Smithay winit code only reacts to the key press/release events, but when you unfocus, release, and focus the window, you don't get a key release event, you only get the new modifiers state event. So Smithay should synthesize a release in this case (and press too if the modifier got pressed).

sodiboo:
> I was sure I tried this in the past with no luck, huh.

Absolutely! Same! I distinctly remember trying for just this thing, and concluding that, for some reason, the key release event must come from the same physical keystroke somehow, which I now realize is wrong. If you are sure you've tried the same thing, then I think we actually ought to go back in time and try niri from March to see if it genuinely was like this back then.

Gattag:
Hey everyone, so this bug still seems to exist and after reading all of the issues around it, my interpretation of what has been said (I could have easily misread) leads me to believe that no one wants to fix it and everyone wants to point the finger at someone else (niri, Smithay, & winit) have all said that another repo is responsible for the issue. Additionally the conversation seems to have died everywhere which sucks because the BUG doesn't exist any less.

kchibisov:
bug only relevant for development setup and will only be fixed by me at this point, probably in like 3-4 month, depending on my mood(time).

sodiboo:
> Hey everyone, so this bug still seems to exist [...]. Additionally the conversation seems to have died everywhere which sucks because the BUG doesn't exist any less.

for what it's worth, [this comment](https://github.com/Smithay/smithay/issues/1353#issuecomment-2343134990) describes a workaround for "unfucking" the state of the keyboard in the nested session: you can repeatedly focus and unfocus to "unpress" the keys. and anytime i do anything major with niri, i generally update [this branch](https://github.com/sodiboo/niri/tree/wayland-backend) and use that backend for development (which doesn't have this issue). in that sense, the bug kind of does exist less for me? that's why i haven't been contributing to the conversation anymore, since i don't really care anymore. doesn't affect me significantly. i've learned to live with it. :shrug:

Issue #88, opened 2023-12-04
`workspace_auto_back_and_forth` 


Kha:
Hi, let me preface this feature request with stating that I very much enjoy niri and it basically feels production-ready for me.

One nicety I've become used to in i3 and sway is [`workspace_auto_back_and_forth`](https://i3wm.org/docs/userguide.html#workspace_auto_back_and_forth) which lets you quickly switch to a workspace and back by hitting the same key combination twice. I think it would be just as handy to have in niri.

YaLTeR:
Hey, I assume this is only for Mod+digit binds and not for Mod+PgUp/Down in the default config? Would it make sense to have that option's behavior by default rather than as an option? I myself don't really use Mod+digit.

Kha:
Ah, I hadn't even seen Mod+Pg before. But I usually have around 5 workspaces open and then Mod+digit becomes more comfortable. Having it as the default behavior would be fine with me, yes!

YaLTeR:
> not allowing for gaps in the numbering

This is because niri uses dynamic workspaces like GNOME, where gaps don't really make sense, since empty workspaces in the middle disappear. It's not an inherent requirement to have dynamic workspaces, just that over my years of GNOME usage I've grown to like dynamic workspaces more than static ones. Mod+digit hotkeys just work on top of that.

> missing IPC for naming and displaying them in waybar

I looked at waybar and it seems that it uses special plugins for both sway and hyprland, i.e. there's no shared system for communicating workspaces between a bar and a compositor. Other than coming up with how an IPC could work in a way that makes the most sense, there isn't really anything else stopping this.

I was thinking of adding a built-in bar which would show workspaces which would allow animating the transitions in a better way compared to any external bar, but that also doesn't block adding an IPC for external bars.

YaLTeR:
I tried implementing this, and it seems that with the current code it'll need a ton of checks and edge case handling all over the layout code, because workspaces are tracked by index, and the index needs to change potentially any time a workspace is removed, added or moved inside or between monitors. I guess I need to come up with a better way to keep track of this.

Kha:
I see, it makes sense that this is more complicated and perhaps inadvisable with dynamic workspaces.

YaLTeR:
Hey, thank you!

> Does a new bar make sense, I mean, right now?

[...]

PR #253, opened 2024-03-11
Workspace back and forth

FluxTape:
Implements `workspace-back-and-forth` as described in this [#issue](https://github.com/YaLTeR/niri/issues/88#issue-2023429801) by generating a random ID for each workspace:
- Each workspace is assigned a random ID on creation (adds fastrand as dependency - should not matter as fastrand is already in the dependency tree)
- the ID of the previously active workspace is stored per output
- on workspace switch the previously active workspace ID is updated
- if the previous workspace ID is not in the list of workspaces, then a fallback behavior is executed.

When the `workspace-back-and-forth` action is triggered the list of workspaces on the active output is searched through to find the correct workspace index to switch to.


resources/default-config.kdl
```diff
@@ -472,6 +472,9 @@ binds {
     Mod+Ctrl+U         { move-column-to-workspace-down; }
     Mod+Ctrl+I         { move-column-to-workspace-up; }
 
+    // Switches focus between the current and the previous workspace.
+    Mod+Tab            { focus-workspace-back-and-forth; }
```
YaLTeR, line 476: I think the original idea was not a new action, but a setting that makes it so pressing `focus-workspace N` while you're on N will send you to the previous workspace? This way there's no fallback behavior needed either.

src/layout/monitor.rs
```diff
@@ -35,6 +35,8 @@ pub struct Monitor<W: LayoutElement> {
     pub workspaces: Vec<Workspace<W>>,
     /// Index of the currently active workspace.
     pub active_workspace_idx: usize,
+    /// ID of the previously active workspace.
+    pub previous_workspace_id: u64,
```
YaLTeR, line 39: It's probably worth it to wrap this in a newtype like `WorkspaceId` to avoid accidentally using it numerically.

YaLTeR (commented):
That is a very good idea to solve this with globally unique IDs. Though maybe instead of random, it's a better idea to use a simple u64 counter that you `.wrapping_add(1)`? It can be stored in the Layout, but if that's too annoying it could also be a static `AtomicU64`. Basically like https://smithay.github.io/smithay/smithay/utils/struct.SerialCounter.html.

YaLTeR:
Oh, no, I didn't mean to literally use SERIAL_COUNTER, that's reserved for event serials. I meant make a similar u64 for workspaces that you will `+ 1`.

resources/default-config.kdl
YaLTeR: What do you think about this? I could change it myself to work like this if you'd like.

FluxTape: I'm already working on an implementation. I'll also add `focus-workspace-no-auto` for the old behaviour.

YaLTeR: I think it should just be a setting, like `input { workspace-auto-back-and-forth; }` which will do that behavior if set. No need for a whole separate set of binds.

FreeFull: Even though sway is capable of doing back and forth by using the keybind for the current workspace, back when I was using sway I preferred to use the separate keybind. Using the current workspace keybind requires the user to actually know what the number of the current workspace is.

YaLTeR: Hm. I thought the use-case here is that you know your, say, IRC is on workspace 1 and you go check it so you hit Mod+1 twice quickly (and don't need to remember your current workspace then).

FluxTape:
This is simple and works, but I do wonder whether it would be worth it to wrap `active_workspace_idx` in a special type that then automatically updates `previous_workspace_id`. That would probably add quite a bit of extra code but might help avoid bugs in the future.

niri-config/src/lib.rs
```diff
@@ -745,6 +747,7 @@ pub enum Action {
     FocusWorkspaceDown,
     FocusWorkspaceUp,
     FocusWorkspace(#[knuffel(argument)] u8),
+    FocusWorkspaceBackAndForth,
```
YaLTeR, line 750: Maybe something like `FocusLastWorkspace`?

YaLTeR (commented):
Do you mind if I do some refactoring before merging?

niri-config/src/lib.rs
FluxTape: I feel like `FocusPreviousWorkspace` would be more clear since last workspace could be mistaken for the bottom most workspace.

FluxTape:
> Do you mind if I do some refactoring before merging?

Sure, go ahead

PR #256, opened 2024-03-13
Add `clickfinger` in touchpad config

uetcis:
This PR adds the [clickfinger behavior](https://wayland.freedesktop.org/libinput/doc/latest/clickpad-softbuttons.html#clickfinger-behavior) setting for touchpads.

I have tested live reload and it works well on my machine.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1492,14 +1492,13 @@ impl FromStr for ClickMethod {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "clickfinger" => Ok(Self::Clickfinger),
-            "button_areas" => Ok(Self::ButtonAreas),
+            "button-areas" => Ok(Self::ButtonAreas),
             _ => Err(miette!(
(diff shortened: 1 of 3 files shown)


YaLTeR:
Thanks! Hmm, so it seems that on my Apple Magic Trackpad 2 clickfinger is set by default, but on my laptop touchpad it's not. To preserve this behavior, I think it's better to make it into a `click-method` which can be set to `button-areas` or `clickfinger` (or left to default if unset). This is also consistent with what you can set in sway-input.

uetcis:
> Thanks! Hmm, so it seems that on my Apple Magic Trackpad 2 clickfinger is set by default, but on my laptop touchpad it's not. To preserve this behavior, I think it's better to make it into a `click-method` which can be set to `button-areas` or `clickfinger` (or left to default if unset). This is also consistent with what you can set in sway-input.

Sure, I'll change it tomorrow

uetcis:
@YaLTeR I've changed the `clickfinger` setting into `click-method`, could you take a look at it?

YaLTeR:
Thanks, works well! Did some minor tweaks.

PR #261, opened 2024-03-22
input: Add support for ISO level3 shift modifier

Trundle:
This modifier is typically called `AltGr` on keyboards or `Mod5` in xkb layouts. Requires a Smithay update.

I went with hyphens for the name in config because kebab case is already used for other settings, but will happily change it to something else. I also added a `Mod5` alias because `ISO-level3-shift` is quite a mouthful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3088,7 +3088,7 @@ checksum = "e6ecd384b10a64542d77071bd64bd7b231f4ed5940fba55e98c3de13824cf3d7"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/Smithay/smithay.git#e61db794643190d4628fbb045634827e4619ffdf"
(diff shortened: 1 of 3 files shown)


YaLTeR (commented):
Thanks, looks good.

> I went with hyphens for the name in config because kebab case is already used for other settings, but will happily change it to something else.

I think we should use underscores here, because other XKB keys also use underscores (like Page_Down).

Trundle:
:+1: Updated, the modifier is now called `ISO_Level3_Shift`

YaLTeR:
Thanks!

sodiboo:
When it's called `AltGr` on keyboards, why isn't that an alias in the config? That's the name i know this key by (as well as the way it shows up on my keyboards) and we have `Win` alias for Super, because that corresponds to the icon shown on most keyboards.

Trundle:
Mostly because nothing in the stack calls it `AltGr` and personally, all compositors and WMs I've used so far always called it `Mod5`.  I've opened #264, so now it's up to YaLTeR to decide.

PR #264, opened 2024-03-24
input: Add "AltGr" alias for ISO level3 shift modifier

Trundle:
It's what's written on keyboards, so perhaps a bit more accessible to users.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1574,6 +1574,7 @@ impl FromStr for Key {
                 modifiers |= Modifiers::SUPER;
             } else if part.eq_ignore_ascii_case("iso_level3_shift")
                 || part.eq_ignore_ascii_case("mod5")
+                || part.eq_ignore_ascii_case("altgr")
             {
                 modifiers |= Modifiers::ISO_LEVEL3_SHIFT;
             } else {
@@ -2145,19 +2146,14 @@ mod tests {
 
     #[test]
     fn parse_iso_level3_shift() {
-        assert_eq!(
-            "ISO_Level3_Shift+A".parse::<Key>().unwrap(),
-            Key {
-                trigger: Trigger::Keysym(Keysym::a),
-                modifiers: Modifiers::ISO_LEVEL3_SHIFT
-            },
-        );
-        assert_eq!(
-            "Mod5+A".parse::<Key>().unwrap(),
-            Key {
-                trigger: Trigger::Keysym(Keysym::a),
-                modifiers: Modifiers::ISO_LEVEL3_SHIFT
-            },
-        );
+        for input in ["ISO_Level3_Shift+A", "Mod5+A", "AltGr+A"] {
+            assert_eq!(
+                input.parse::<Key>().unwrap(),
+                Key {
+                    trigger: Trigger::Keysym(Keysym::a),
+                    modifiers: Modifiers::ISO_LEVEL3_SHIFT
+                },
+            );
+        }
     }
 }


YaLTeR:
cc @sodiboo who had the original concern.

So, I'm not sure this is such a good idea, because you need a specific xkb option (`lv3:ralt_switch` I think), or I guess a specific layout, to make AltGr actually emit the level 3 modifier. By default (on `us` with no special options) it emits a right alt, which does not trigger the level 3 modifier. (Compared to Win/Super, which sends that key by default.) So I think it would only be confusing to make AltGr an alias here.

Unless of course it is established practice in compositors to accept the AltGr alias for this.

Trundle:
The key is alt right on `us` layouts, but the key is also labeled `Alt` and not `AltGr` on US keyboards, isn't it? My non-scientific quick image search found keyboards with `Alt` and `AltGr` though :see_no_evil: 

As another data point, the `de` layout maps `AltGr` to `ISO Level3 Shift`.

Re other compositors: In my limited experience, it's `Mod5`-ish everywhere

kchibisov:
Mine `AltGr` was always a regular `Alt` on US layout, yes it's called `AltGr` on keyboard, but it's just _alt_. And when you compose it's a `Compose` key, not `AltGr`, so whether it's ISO level3 shift is a layout dependant, I guess.

Keep in mind that there's also a level 5 shift.

sodiboo:
> because you need a specific xkb option (`lv3:ralt_switch` I think), or I guess a specific layout, to make AltGr actually emit the level 3 modifier

I don't think the "layout" needs to be as specific as you think -- i've used the norwegian layout my whole life, that's what is physically printed on my keyboards whenever possible (really "nordic"  usually) and the only option i ever set for xkb is `layout "no"`; for me, on both my computers, this will lead to the AltGr key emitting the `ISO_Level3_Shift` keysym and my other alt key emitting the `Alt_L` keysym -- no special options here!

As far as i can tell by looking at a picture of a US layout, emitting `Alt_R` is correct there because those keys don't say AltGr, they say just Alt. so given that the configured keyboard layout matches the physical layout, i don't think there are any layouts where (with no special options) AltGr emits `Alt_R` or right alt emits `ISO_Level3_Shift`

Pajn:
Both UK and Swedish keyboard layouts also produce AltGr so I'd imagine it's very widespread in Europe at least

WilliButz:
> > because you need a specific xkb option (`lv3:ralt_switch` I think), or I guess a specific layout, to make AltGr actually emit the level 3 modifier
> 
> I don't think the "layout" needs to be as specific as you think -- i've used the norwegian layout my whole life, that's what is physically printed on my keyboards whenever possible (really "nordic" usually) and the only option i ever set for xkb is `layout "no"`; for me, on both my computers, this will lead to the AltGr key emitting the `ISO_Level3_Shift` keysym and my other alt key emitting the `Alt_L` keysym -- no special options here!
> 
> As far as i can tell by looking at a picture of a US layout, emitting `Alt_R` is correct there because those keys don't say AltGr, they say just Alt. so given that the configured keyboard layout matches the physical layout, i don't think there are any layouts where (with no special options) AltGr emits `Alt_R` or right alt emits `ISO_Level3_Shift`

@sodiboo  I think I'm experiencing an issue that is related to what you're saying, but I'm not quite sure how it would be related to this PR.

Until a few days ago I was using sway, some years before that i3. I'm used to the `altgr-intl` variant of the `us` layout, which I **never** used with any lv3 xkb option. With niri I'm currently unable to produce characters such as umlauts (for example `ä` should be produced by pressing `altgr + q`).

Inspecting events on niri with wev (without lv3 options, as under sway/i3), I saw the following:
```
[14:     wl_keyboard] key: serial: 19383; time: 90539102; key: 108; state: 1 (pressed)
                      sym: Alt_R        (65514), utf8: ''
[14:     wl_keyboard] modifiers: serial: 19383; group: 0
                      depressed: 00000008: Mod1
                      latched: 00000000
                      locked: 00000000
[14:     wl_keyboard] key: serial: 19384; time: 90539179; key: 108; state: 0 (released)
                      sym: ISO_Next_Group (65032), utf8: ''
[14:     wl_keyboard] modifiers: serial: 19384; group: 0
                      depressed: 00000000
                      latched: 00000000
                      locked: 00000000
```

Both from comparing the events from wev on sway and also from you're comment above, I'd expect to see `ISO_Leve3_Shift` instead of `Alt_R` and `ISO_Next_group`.

Now after finding this issue, I tried adding the `lv3:ralt_switch`. That seems closer to what I want, but also doesn't do what I expect:

```
[14:     wl_keyboard] key: serial: 35214; time: 91890908; key: 108; state: 1 (pressed)
                      sym: ISO_Level3_Shift (65027), utf8: ''
[14:     wl_keyboard] modifiers: serial: 35214; group: 0
                      depressed: 00000088: Mod1 Mod5
                      latched: 00000000
                      locked: 00000000
[14:     wl_keyboard] key: serial: 35215; time: 91891204; key: 24; state: 1 (pressed)
                      sym: adiaeresis   (228), utf8: 'ä'
[14:     wl_keyboard] key: serial: 35216; time: 91891291; key: 24; state: 0 (released)
                      sym: adiaeresis   (228), utf8: ''
[14:     wl_keyboard] key: serial: 35217; time: 91891403; key: 108; state: 0 (released)
                      sym: ISO_Level3_Shift (65027), utf8: ''
[14:     wl_keyboard] modifiers: serial: 35217; group: 0
                      depressed: 00000000
                      latched: 00000000
                      locked: 00000000
```

The events above (from pressing `altgr+q`) read like they might produce `ä`, but whatever appears in alacritty/firefox/other applications is not the expected utf-8 umlaut,  instead alacritty shows `<ffffffff>A`. The same happens for other combinations, such as `altgr+p` which should yield `ö` but also show as `<ffffffff>A` in alacritty.
There is an obvious difference between the events listed above on niri and those on sway, on niri it says
`depressed: 00000088: Mod1 Mod5`, while on sway `Mod1` is not part of the event.

I don't see anything suspicious on alacritty's side, but these are the events from `alacritty --print-events` for `altgr+q`:
```
[3.586631844s] [INFO ] [alacritty_winit_event] About to wait
[3.670478460s] [INFO ] [alacritty_winit_event] KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), event: KeyEvent { physical_key: Code(KeyQ), logical_key: Character("ä"), text: Some("ä"), location: Standard, state: Released, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: Some("ä"), key_without_modifiers: Character("q") } }, is_synthetic: false }
[3.670536911s] [INFO ] [alacritty_winit_event] About to wait
[3.719898981s] [INFO ] [alacritty_winit_event] KeyboardInput { device_id: DeviceId(Wayland(DeviceId)), event: KeyEvent { physical_key: Code(AltRight), logical_key: Named(AltGraph), text: None, location: Standard, state: Released, repeat: false, platform_specific: KeyEventExtra { text_with_all_modifiers: None, key_without_modifiers: Named(AltGraph) } }, is_synthetic: false }
[3.719956250s] [INFO ] [alacritty_winit_event] ModifiersChanged(Modifiers { state: ModifiersState(0x0), pressed_mods: ModifiersKeys(0x0) })
[3.719968553s] [INFO ] [alacritty_winit_event] About to wait
```

(My /etc/locale.conf is the same as on sway.)

WilliButz:
Follow-up to yesterday: now it works but I don't understand why it didn't before.
I was still missing the second layout that I'm using from time to time and since adding that everything works as expected, even without the lv3 option. This is now my keyboard section in the niri config:

```kdl
keyboard {
    xkb {
        layout "us,ru"
        variant "altgr-intl,phonetic"
        options "grp:caps_toggle"
    }
}
```

The wev events no longer include `Mod1 Mod5` as before, but instead they now match those on sway, i.e. only `Mod5` for the right alt key:

```
[14:     wl_keyboard] key: serial: 820035; time: 118098942; key: 108; state: 1 (pressed)
                      sym: ISO_Level3_Shift (65027), utf8: ''
[14:     wl_keyboard] modifiers: serial: 820035; group: 0
                      depressed: 00000080: Mod5
                      latched: 00000000
                      locked: 00000000
[14:     wl_keyboard] key: serial: 820036; time: 118099711; key: 24; state: 1 (pressed)
                      sym: adiaeresis   (228), utf8: 'ä'
[14:     wl_keyboard] key: serial: 820037; time: 118099782; key: 24; state: 0 (released)
                      sym: adiaeresis   (228), utf8: ''
[14:     wl_keyboard] key: serial: 820038; time: 118100965; key: 108; state: 0 (released)
                      sym: ISO_Level3_Shift (65027), utf8: ''
```

So I'm happy now and a bit confused :)

WilliButz:
Oh and something that seems more related to this PR, the keyboard in my laptop should be this one:
https://download.lenovo.com/Images/Parts/01HW644/01HW644_A.jpg

It's usually sold as "US English, International", `AltGr` is written on the right alt key and for example the `€` symbol on `5`.

PR #277, opened 2024-04-02
Hide pointer on touch interaction

YaLTeR:
Cannot test myself due to lack of a touch device.


regenman:
Tested briefly on a ThinkPad X1 Yoga, seems to be working well. 
Notably improves touch-and-hold actions, for example in web browsing, where previously the lingering pointer would interfere and trigger text selection.

YaLTeR:
Thanks!

PR #278, opened 2024-04-03
`niri_ipc::NiriSocket`; `niri msg version`; version checking on IPC

sodiboo:
~~This does *not* implement any sort of event stream, but rather implements basic support for the IPC client and server to handle multiple messages on each stream.~~

~~Previously, the server would ignore any additional events sent to it. This means every IPC request *must* start a new socket connection.~~

~~Now, with my changes, single IPC connection can handle arbitrarily many requests, separated by newlines.~~

~~The client can additionally handle them with arbitrary separation, as it relies on serde_json's `StreamDeserializer`. It cannot be used on the server-side, because it does not support async readers.~~

(no longer part of this PR)

A new API is made available in `niri_ipc`: `NiriSocket`. It handles the underlying `UnixStream` and allows a consumer to send multiple requests without closing it inbetween.

  - sidenote: wouldn't a more fitting API be that `Request` is a trait, with an associated type `Response`? That way, the request method could statically return the correct type (rather than an enum).
  
---

-  ### New command `niri msg version` to check the running compositor's version
  
    image

- ### All `niri msg` commands will now check the running compositor's version.

- If it matches, output is identical to before.
    
    image

- If it does not match, additional output is printed upon each `niri msg` invocation, informing the user of the discrepancy so they can understand any potential problem more easily.
  
    image

- This does not break existing scripts.
  
    image

- The "Did you forget?" hint is also available for compositors prior to this commit
  
    image

- The human-readable form of `niri msg version` does not duplicate the version information.
  
    image

- "Did you forget?" can be filtered out just as with json output

    image

- JSON output of `niri msg version` does duplicate the information, with rationale going that you may be piping its output into some other command that eventually discards the version numbers or puts it somewhere other than the terminal, but stderr should still be visible to the human

    image

- The exact wording of human-readable output in the screenshots above may vary from the real product.

---

Additionally, no attempt is made to do anything more than tell the user that the version string mismatches. This is because packagers can[1] and will[2] and do[3] overwrite the `version()` function with their own implementation, under the assumption that the result will only be shown to the user. This PR does not try to fundamentally change this assumption.

[1]: COPR at `yalter/niri/niri` \
[2]: Nix at `github:sodiboo/niri-flake` \
[3]: AUR at `niri-bin` depends on [1]

[1]: https://copr-dist-git.fedorainfracloud.org/cgit/yalter/niri/niri.git/tree/niri.spec?id=a22d516a8f477ae0e7ef0ae69dfc24878525eec6#n113
[2]: https://github.com/sodiboo/niri-flake/blob/a678ef2f4e4e95bff267823340c77df588bc5314/flake.nix#L139-L148
[3]: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=niri-bin#n29

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -7,14 +7,11 @@ use std::str::FromStr;
 use serde::{Deserialize, Serialize};
 
 mod socket;
-
-pub use socket::{NiriSocket, SOCKET_PATH_ENV};
+pub use socket::{Socket, SOCKET_PATH_ENV};
 
 /// Request from client to niri.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub enum Request {
-    /// Always responds with an error. (For testing error handling)
-    ReturnError,
     /// Request the version string for the running niri instance.
     Version,
     /// Request information about connected outputs.
@@ -23,6 +20,8 @@ pub enum Request {
     FocusedWindow,
     /// Perform an action.
     Action(Action),
+    /// Respond with an error (for testing error handling).
+    ReturnError,
 }
 
 /// Reply from niri to client.
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -6,12 +6,17 @@ use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
 
-/// Name of the environment variable containing the niri IPC socket path.
-pub const SOCKET_PATH_ENV: &str = "NIRI_SOCKET";
+mod socket;
+
+pub use socket::{NiriSocket, SOCKET_PATH_ENV};
 
 /// Request from client to niri.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub enum Request {
+    /// Always responds with an error. (For testing error handling)
+    ReturnError,
+    /// Request the version string for the running niri instance.
+    Version,
     /// Request information about connected outputs.
     Outputs,
     /// Request information about the focused window.
@@ -35,6 +40,8 @@ pub type Reply = Result<Response, String>;
 pub enum Response {
     /// A request that does not need a response was handled successfully.
     Handled,
+    /// The version string for the running niri instance.
+    Version(String),
     /// Information about connected outputs.
     ///
     /// Map from connector name to output info.
--- a/niri-ipc/src/socket.rs
+++ b/niri-ipc/src/socket.rs
@@ -1,0 +1,73 @@
+use std::io::{self, Write};
+use std::os::unix::net::UnixStream;
+use std::path::Path;
+
+use serde_json::de::IoRead;
+use serde_json::StreamDeserializer;
+
+use crate::{Reply, Request};
+
+/// Name of the environment variable containing the niri IPC socket path.
+pub const SOCKET_PATH_ENV: &str = "NIRI_SOCKET";
+
+/// A client for the niri IPC server.
+///
+/// This struct is used to communicate with the niri IPC server. It handles the socket connection
+/// and serialization/deserialization of messages.
+pub struct NiriSocket {
+    stream: UnixStream,
+    responses: StreamDeserializer<'static, IoRead<UnixStream>, Reply>,
+}
+
+impl TryFrom<UnixStream> for NiriSocket {
+    type Error = io::Error;
+    fn try_from(stream: UnixStream) -> io::Result<Self> {
+        let responses = serde_json::Deserializer::from_reader(stream.try_clone()?).into_iter();
+        Ok(Self { stream, responses })
+    }
+}
+
+impl NiriSocket {
+    /// Connects to the default niri IPC socket
+    ///
+    /// This is equivalent to calling [Self::connect] with the value of the [SOCKET_PATH_ENV]
+    /// environment variable.
+    pub fn new() -> io::Result<Self> {
+        let socket_path = std::env::var_os(SOCKET_PATH_ENV).ok_or_else(|| {
+            io::Error::new(
+                io::ErrorKind::NotFound,
+                format!("{SOCKET_PATH_ENV} is not set, are you running this within niri?"),
+            )
+        })?;
+        Self::connect(socket_path)
+    }
+
+    /// Connect to the socket at the given path
+    ///
+    /// See also: [UnixStream::connect]
+    pub fn connect(path: impl AsRef<Path>) -> io::Result<Self> {
+        Self::try_from(UnixStream::connect(path.as_ref())?)
+    }
+
+    /// Handle a request to the niri IPC server
+    ///
+    /// # Returns
+    /// Ok(Ok([Response](crate::Response))) corresponds to a successful response from the running
+    /// niri instance. Ok(Err([String])) corresponds to an error received from the running niri
+    /// instance. Err([std::io::Error]) corresponds to an error in the IPC communication.
+    pub fn send(mut self, request: Request) -> io::Result<Reply> {
+        let mut buf = serde_json::to_vec(&request).unwrap();
+        writeln!(buf).unwrap();
+        self.stream.write_all(&buf)?; // .context("error writing IPC request")?;
+        self.stream.flush()?;
+
+        if let Some(next) = self.responses.next() {
+            next.map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))
+        } else {
+            Err(io::Error::new(
+                io::ErrorKind::UnexpectedEof,
+                "no response from server",
+            ))
+        }
+    }
+}
--- a/niri-ipc/src/socket.rs
+++ b/niri-ipc/src/socket.rs
@@ -1,73 +1,63 @@
-use std::io::{self, Write};
+//! Helper for blocking communication over the niri socket.
+
+use std::env;
+use std::io::{self, Read, Write};
+use std::net::Shutdown;
 use std::os::unix::net::UnixStream;
 use std::path::Path;
-
-use serde_json::de::IoRead;
-use serde_json::StreamDeserializer;
 
 use crate::{Reply, Request};
 
 /// Name of the environment variable containing the niri IPC socket path.
 pub const SOCKET_PATH_ENV: &str = "NIRI_SOCKET";
 
-/// A client for the niri IPC server.
+/// Helper for blocking communication over the niri socket.
 ///
 /// This struct is used to communicate with the niri IPC server. It handles the socket connection
 /// and serialization/deserialization of messages.
-pub struct NiriSocket {
+pub struct Socket {
     stream: UnixStream,
-    responses: StreamDeserializer<'static, IoRead<UnixStream>, Reply>,
 }
 
-impl TryFrom<UnixStream> for NiriSocket {
-    type Error = io::Error;
-    fn try_from(stream: UnixStream) -> io::Result<Self> {
-        let responses = serde_json::Deserializer::from_reader(stream.try_clone()?).into_iter();
-        Ok(Self { stream, responses })
-    }
-}
-
-impl NiriSocket {
-    /// Connects to the default niri IPC socket
+impl Socket {
+    /// Connects to the default niri IPC socket.
     ///
-    /// This is equivalent to calling [Self::connect] with the value of the [SOCKET_PATH_ENV]
-    /// environment variable.
-    pub fn new() -> io::Result<Self> {
-        let socket_path = std::env::var_os(SOCKET_PATH_ENV).ok_or_else(|| {
+    /// This is equivalent to calling [`Self::connect_to`] with the path taken from the
+    /// [`SOCKET_PATH_ENV`] environment variable.
+    pub fn connect() -> io::Result<Self> {
+        let socket_path = env::var_os(SOCKET_PATH_ENV).ok_or_else(|| {
             io::Error::new(
                 io::ErrorKind::NotFound,
                 format!("{SOCKET_PATH_ENV} is not set, are you running this within niri?"),
             )
         })?;
-        Self::connect(socket_path)
+        Self::connect_to(socket_path)
     }
 
-    /// Connect to the socket at the given path
-    ///
-    /// See also: [UnixStream::connect]
-    pub fn connect(path: impl AsRef<Path>) -> io::Result<Self> {
-        Self::try_from(UnixStream::connect(path.as_ref())?)
+    /// Connects to the niri IPC socket at the given path.
+    pub fn connect_to(path: impl AsRef<Path>) -> io::Result<Self> {
+        let stream = UnixStream::connect(path.as_ref())?;
+        Ok(Self { stream })
     }
 
-    /// Handle a request to the niri IPC server
+    /// Sends a request to niri and returns the response.
     ///
-    /// # Returns
-    /// Ok(Ok([Response](crate::Response))) corresponds to a successful response from the running
-    /// niri instance. Ok(Err([String])) corresponds to an error received from the running niri
-    /// instance. Err([std::io::Error]) corresponds to an error in the IPC communication.
-    pub fn send(mut self, request: Request) -> io::Result<Reply> {
+    /// Return values:
+    ///
+    /// * `Ok(Ok(response))`: successful [`Response`](crate::Response) from niri
+    /// * `Ok(Err(message))`: error message from niri
+    /// * `Err(error)`: error communicating with niri
+    pub fn send(self, request: Request) -> io::Result<Reply> {
+        let Self { mut stream } = self;
+
         let mut buf = serde_json::to_vec(&request).unwrap();
-        writeln!(buf).unwrap();
-        self.stream.write_all(&buf)?; // .context("error writing IPC request")?;
-        self.stream.flush()?;
+        stream.write_all(&buf)?;
+        stream.shutdown(Shutdown::Write)?;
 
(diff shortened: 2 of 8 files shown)


YaLTeR:
> single IPC connection can handle arbitrarily many requests, separated by newlines

I wonder if we actually want this, at least currently? I'm not sure how much it brings over connecting twice, and it does add all this code to keep the socket open and be able to parse many messages at once, all with extra error conditions that are *rare* (like, it's easy to get an error when opening a socket and sending one message, but it's harder to get an error in-between messages).

Plus, this cannot be used with the actual event streaming, because then there's no way to tell whether a response from a server is the response to the request you just did, or just another event that got streamed in. LSPs pass explicit request ID to get around this problem, but this is even more complexity, all for seemingly little benefit in our case, when the client can just connect an extra time.

> New command niri msg version to check the running compositor's version

This is good.

> All niri msg commands will now check the running compositor's version.

I am a bit on the fence on this one, because it's adding a whole extra roundtrip to *every single* invocation of `niri msg` just to be able to show an error on a very rare case right after a niri update that made JSON incompatible. When is this even a problem currently? I believe adding new fields to response JSON will work correctly (I think serde already ignores unknown fields by default), so the only place this can happen is unknown action or unknown `niri msg` request, which means that `niri msg` can do the version request only after one of those returned an error.

sodiboo:
> I am a bit on the fence on this one, because it's adding a whole extra roundtrip to every single invocation of niri msg just to be able to show an error on a very rare case right after a niri update that made JSON incompatible. When is this even a problem currently?

In my use, this edge case is not infrequent. I usually do not restart niri often, even though the binary does update on my system. I find it nice to have a heads up, i suppose.

I didn't think too much of the impact of adding another request round-trip. One motivation for that specific implementation is to be "backwards compatible" with the previous versions of niri, without radically changing the protocol of those messages. This shouldn't be an issue, really, as we don't actually guarantee any ipc compatibility across commits (to my knowledge), but i personally dislike it quite a lot when a `niri_msg` ipc command fails for no apparent reason; and the error emitted does not suggest an intentional change.

Another way to do this, which would break the current protocol (but who cares? it'd work in the future), is to move the actual response/error into a field of a real "reply" object, where the compositor sends back `{"version": "...", "reply": <what was previously the whole message>}`. Then, the `version` subcommand is really a "noop" request that just requests the version. This eliminates the overhead of sending two requests; but i am not convinced this overhead is significant. When we do have "event streams" of sorts, we obviously would not be rechecking the version every time; only once when that client starts listening to the server.

It is also important to note that the `niri_ipc` crate that other programs could use does *not* enforce the version checking. It is only part of `niri msg`, so a consumer that wants to do anything complex with this data and who cares a lot about performance might instead prefer interfacing with the `niri_ipc` crate directly, which eliminates an extra roundtrip of json compared to parsing the output of `niri msg`.

Personally, i find value in having `niri msg` double-check version all the time. Though i just presented some reasons why i think it is reasonable to do unconditionally, i would also be fine with it being gated behind an environment variable so i can set it in niri config and forget about it.

> Plus, this cannot be used with the actual event streaming, because then there's no way to tell whether a response from a server is the response to the request you just did, or just another event that got streamed in

Sure it can! First of all, we might not want to send events by default, we might have some request to receive them. That makes it backwards compatible by default.

A refactor of the API might be needed to make it "nicer" (i.e. separating events from responses. see above suggestion of response trait), but even without radically changing the interface, the "event stream" should simply correspond to a `Response` for which there is no `Request` that creates it. `NiriSocket` could then handle these kinds of responses, and buffer them internally, skipping forward to the actual response that corresponds to the request we made. In other words, the client should ignore "event" objects sent over the socket.

While implementing some request ID could be necessary if we want to make all requests asynchronous; (i.e. handled in parallel and can be fulfilled in any order), currently we do not do that. they are sequential; one is read at a time, and fully handled, and responded to, before we even try reading another one. I believe this is how HTTP works.

If we consider a client that does not know every event and response it may receive, the "buffer events until we find a response" approach is less trivial. of course, we should just ignore unknown variants because an unknown response will never be received (because we can't possibly have requested it), but this is slightly incompatible with the current behaviour of "panic on incorrect response".

If we distinguish events and request responses (i.e. `Reply` will be an enum with variants `Event(Event)` and `Response(Result<Response, String>)`, the filtering described in the previous paragraph will depend a lot less on "assume it's an event"; since it'we do have that info already. This is better than adding new `Response` variants; because an event can not "fail" in the same way a request might. We didn't send a request, there's nothing to "fail". if there is no event, it should simply not be sent.

> all with extra error conditions that are rare (like, it's easy to get an error when opening a socket and sending one message, but it's harder to get an error in-between messages).

I suppose. But these error conditions do not occur if you use it as previously(?). They are also io errors; all io can fail, duh. doing "more" will obviously incur new failure states. io is inherently unreliable, but i don't see these changes creating any "unexpected" failures. any error between messages also applies to errors during messages. previously, a client can open a socket and never write a newline or close it, for example. then, the socket would stay open in compositor forever. `NiriSocket` should handle any possible io errors and provide a hard-to-misuse API, so i don't see this ultimately being a real problem? i don't get this point.

yes, a client can connect twice. but this too adds extra error conditions: the socket may, in an extreme case, actually be connected to a different instance of niri between consecutive calls.

YaLTeR:
> Adding Msg::FocusedWindow.

That should not have broken any other functionality though, right? So if you do a version check only on error, then that will work.

> Before that, i believe when some error handling was added

Yeah, that was a breaking change, that I intend to try not to do going forward.

> I guess? niri msg subcommands aren't necessarily the same as the underlying ipc protocol, and when discussing event streams above, i did not consider the command line interface that niri msg should expose, only the API it should interact with

I consider both `niri msg` interface and the IPC API at the same time when thinking about design, because it is a real and the primary consumer of the API, so it helps shape the API structure.

> But how is it better to force you to re-establish the connection every time? If you want to make a new socket for each connection, sending just one request then closing it, the compositor will handle that identically to before.

It's better because it preserves the error surface, whereas multiplexing introduces additional error surface that will also not be very well tested just because the conditions are rarer.

sodiboo:
I've pushed some more changes:

- A new `niri msg nonsense` / `Request::Nonsense` IPC request is added, that always returns an error. Akin to `niri panic`, it is useful for testing how the error handling will work
- The struct `Versions` defined before its singular usage is replaced with the `json!()` macro, which is much more fitting
- Error messages in `niri msg` are generally cleaned up to be more consistent. They will be shown to a user not digging through journalctl, after all. Likewise, the function signature of `niri::ipc::server::process()` has been changed to take a `Request` and return a `Reply`, more closely resembling what its return value should mean (rather than an `anyhow::Result`, which may contain additional error information that we strip from it anyways)
- The version request is now only double-checked if the intiial request fails. This, moreover, means that nothing is printed in `niri msg --json` because if the json command succeeds, the version is never checked to begin with. For non-json, it is also not printed nor checked.
- I also seem to have broken MSRV 1.72.0. will fix

niri-ipc/src/lib.rs
```diff
@@ -6,12 +6,17 @@ use std::str::FromStr;
...
-pub const SOCKET_PATH_ENV: &str = "NIRI_SOCKET";
+mod socket;
+
+pub use socket::{NiriSocket, SOCKET_PATH_ENV};
 
 /// Request from client to niri.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub enum Request {
+    /// Always responds with an error.
+    Nonsense,
```
YaLTeR, line 17:
```suggestion
    /// Always responds with an error (for testing).
    ReturnError,
```

src/ipc/client.rs
```diff
@@ -1,54 +1,106 @@
...
-            "{} is not set, are you running this within niri?",
-            niri_ipc::SOCKET_PATH_ENV
-        )
-    })?;
+    let mut client = NiriSocket::new()
+        .context("a communication error occured while trying to initialize the socket")?;
 
-    let mut stream =
-        UnixStream::connect(socket_path).context("error connecting to {socket_path}")?;
+    // Default SIGPIPE so that our prints don't panic on stdout closing.
```
YaLTeR, line 12: I intentionally put this after all IPC had been done and before any prints because I'm not sure how this will affect the IPC writes and reads. I suspect it might prevent read/write errors from going through.

src/ipc/server.rs
```diff
@@ -106,29 +108,40 @@ fn on_new_ipc_client(state: &mut State, stream: UnixStream) {
...
 
-    let reply = process(&ctx, &buf).map_err(|err| {
-        warn!("error processing IPC request: {err:?}");
-        err.to_string()
-    });
+    let mut lines = BufReader::new(read).lines();
 
-    let buf = serde_json::to_vec(&reply).context("error formatting reply")?;
-    write.write_all(&buf).await.context("error writing reply")?;
+    while let Some(line) = lines.next().await {
```
YaLTeR, line 118: Do you mind splitting out the "multiple requests per connection" into a separate PR? I'm not convinced that we need it *right now*, but I would like the rest of this PR in (version, return-error, version checks).

sodiboo: i made it no longer handle multiple requests on a connection; most of the inner workings to do so are still present, they're just unused.

src/ipc/client.rs
sodiboo: Hm. How should best to do it then? Handle all IPC logic first, encapsulate the results (including errors) in some variable, then set the signal handler, then do all error printing?

Why do we need it? Is it problematic if the `niri msg` process panics upon stdout hangup? If yes, wouldn't it be better to e.g. lock stdout and actually handle the err case in our code (rather than relying on an unsafe call to libc)?

YaLTeR: We need it because otherwise things like `niri msg -j focused-window | jq aaaa` (malformed jq command) will spew a Rust backtrace from niri msg in addition to the jq error.

I am not sure what happens to socket IO when it is set though.

Sidenote, there are discussions in Rust to make this more easily settable (e.g. https://github.com/rust-lang/rust/pull/120832), but they are considering CLI apps that don't do socket IO, from my understanding.

YaLTeR: I think it should be possible in this case to do all the requests before any of the printing, right?

src/ipc/server.rs
YaLTeR: Thanks. I think the `NiriSocket` struct is good, but I'd further roll back some logic complexity that will only be necessary for multiple requests per connection. I can do it myself if you'd prefer to save on the comments back and forth.

src/ipc/client.rs
sodiboo: It is possible, but somewhat annoying, particularly because of the edge case with version checking. In #294, where i've mainly focused on a more "correct" API (which is easier for e.g. scripting use cases that don't need to handle all possible requests and responses like niri msg does), i've also more clearly separated functionality for sending request, handling version checking, and "outputting" the results of these operations. and then, it's quite trivial to insert a call to ignore `SIGPIPE` at the appropriate point. https://github.com/YaLTeR/niri/blob/f1bfef9a36337696b57efa3e4e41ca04dece1f5b/src/ipc/client.rs#L56-L63

YaLTeR:
Adjusted this PR to my liking.

niri-ipc/src/socket.rs
```diff
@@ -1,73 +1,63 @@
...
 /// Name of the environment variable containing the niri IPC socket path.
 pub const SOCKET_PATH_ENV: &str = "NIRI_SOCKET";
 
-/// A client for the niri IPC server.
+/// Helper for blocking communication over the niri socket.
 ///
 /// This struct is used to communicate with the niri IPC server. It handles the socket connection
 /// and serialization/deserialization of messages.
-pub struct NiriSocket {
+pub struct Socket {
```
sodiboo, line 18: I personally don't like this generic name as much. When `use`d in a file, it's easier to confuse with more general I/O operations when skimming (compare to naming of `UnixStream`, `TcpStream`, `UdpSocket` in the std library). It's an unimportant point though; this was an exhaustive list of all the downsides that such a name brings.

YaLTeR: I get that, but I decided to do this because in other niri code I'm already using fully-qualified names for some types like `niri_config::Animation` and `niri_ipc::Output` (leaving non-qualified names for more frequently occurring types of the same name). So it's consistent this way.

src/ipc/client.rs
```diff
@@ -1,102 +1,98 @@
...
+                    eprintln!("Running niri compositor has a different version from the niri CLI:");
+                    eprintln!("Compositor version: {compositor_version}");
+                    eprintln!("CLI version:        {cli_version}");
+                    eprintln!("Did you forget to restart niri after an update?");
+                    eprintln!();
                 }
             }
-
-            return Ok(());
+            Some(_) => {
```
sodiboo, line 52: This block now handles the `Reply::Ok` case for non-`Response::Version`. Is that desirable? This wouldn't be caused by an outdated compositor (it would cause a `Reply::Err("error parsing request")`), and as such i made it not print any kind of error response. In reality, it should be unreachable since the compositor server Does Not Behave Like This

src/ipc/client.rs
```diff
@@ -1,102 +1,98 @@
...
+    let response = reply.map_err(|err_msg| {
+        // Check for CLI-server version mismatch to add helpful context.
+        match compositor_version {
+            Some(Ok(Response::Version(compositor_version))) => {
+                let cli_version = version();
+                if cli_version != compositor_version {
+                    eprintln!("Running niri compositor has a different version from the niri CLI:");
+                    eprintln!("Compositor version: {compositor_version}");
+                    eprintln!("CLI version:        {cli_version}");
+                    eprintln!("Did you forget to restart niri after an update?");
```
sodiboo, line 48: I prefer your phrasing to my original version. "Running niri compositor" is better than "the compositor".

src/ipc/client.rs
YaLTeR: Yeah, it is intentional for the reason that you mentioned. If it ever happens then something weird is going on; doesn't hurt to show the warning.

src/ipc/server.rs
```diff
@@ -108,42 +107,36 @@ fn on_new_ipc_client(state: &mut State, stream: UnixStream) {
...
-    };
-
-    let reply: Reply = serde_json::from_str(&line)
-        .map_err(|err| format!("error parsing request: {err}"))
-        .and_then(|req| process(&ctx, req));
+    let reply = request.and_then(|request| process(&ctx, request));
 
     if let Err(err) = &reply {
-        warn!("error processing IPC request: {err:?}");
+        if !requested_error {
```
sodiboo, line 126: Do we want to handle this error flow differently? It's not really problematic to print it; nobody's gonna spam `niri msg request-error` and then complain that the log output contains the same message; conversely, it's reasonable that someone may want to see what it looks like on the compositor side handling an error in which case this check somewhat obscures it.

YaLTeR: Well, no hard opinions here, but in my head this request is aimed much more at the client side, so as far as the server is concerned it shouldn't print a warning.

src/ipc/client.rs
```diff
@@ -1,54 +1,99 @@
...
-        .context("error writing IPC request")?;
-    stream
-        .shutdown(Shutdown::Write)
-        .context("error closing IPC stream for writing")?;
 
-    buf.clear();
-    stream
-        .read_to_end(&mut buf)
-        .context("error reading IPC response")?;
+    let socket = Socket::connect().context("error connecting to the niri socket")?;
```
sodiboo, line 17: re: naming of `niri_ipc::Socket` (what? is there a better mechanism to add suggestions as replies to existing review comments?)

in that case, given that we use qualified `niri_ipc::Window` (among others), it should be more consistent across all niri code.

```suggestion
    let socket = niri_ipc::Socket::connect().context("error connecting to the niri socket")?;
```

sodiboo (commented):
Interestingly i cannot mark my approval for others' changes on my own pull request? (focused on your commit; not the whole PR overview)

image

but yeah, looks good to me. concerns are pretty much exclusively with naming of some items or exact behaviour of error printing flow. they don't matter too much and i don't see a problem with merging as-is

src/ipc/client.rs
YaLTeR: In this file specifically, since it's under `src/ipc/` and doesn't import any of the conflicting types, I am importing and using all `niri_ipc` types directly. The only exception is `Transform` which seems like an oversight; I'll also import it. (In `src/ipc/server.rs` it's clearer to fully qualify, because it's filling info from a Smithay `Window`).

YaLTeR:
Thanks!

src/ipc/client.rs
sodiboo: Although in general, for other types, i agree with that justification; Qualify them to indicate it's the `niri_ipc` version of that type when not in the context of IPC. But, see point about preferring `NiriSocket` over `Socket` (and i stand by that as it does avoid the need to qualify the name at all): In this context, while it's obvious that we're working with niri_ipc, it's not obvious what `Socket` means without viewing the imports or having seen the definition. Qualifying the name here is to disambiguate (for the reader) the concept of a "niri IPC socket" from "a generic type of socket with arbitrary data flowing in each direction" (e.g. they are different kinds of objects in the semantic sense); not to disambiguate the specific variant of "a transform but with serde attributes" from "a transform but the one Smithay likes" (i.e. they are the same thing semantically but distinct only in type systems and specific trait impls that both could just as well do)

Issue #221, opened 2024-02-22
Input method popup candidate window is placed in wrong position


HardGraphite:
*I reported this issue to the IME's repo. They said that it is a compositor bug. See [here](https://github.com/fcitx/fcitx5/issues/971).*

I'm trying to use an input method with the `text-input-v*` protocol. I found that the popup candidate window covers the input line. The upper left corner of the popup window is placed exactly where the text will be entered. The popup window is expected to be under the input line (see the last screenshot).

Here are two examples of this issue:

Alacritty (winit):

![1](https://github.com/YaLTeR/niri/assets/64914256/67afb655-25c0-4b6d-95ad-1a3fd92a2fb6)

Firefox (GTK):

![2](https://github.com/YaLTeR/niri/assets/64914256/f0543088-94fe-40be-9131-46b14513d8eb)

But Neovide (winit) some how **works well**:

![3](https://github.com/YaLTeR/niri/assets/64914256/3bf16b35-0250-4bd1-af27-9ed29bbee597)

There is another problem I found when using Firefox. I don't know whether it is related to the previous one. In Firefox, when I start typing for a word, the popup window always shows up at the place of the previous input word.

### System Information

* niri version: niri 0.1.2 (62892d6)
* GPU: Intel(R) HD Graphics 620
* CPU: Intel(R) Core(TM) i5-7200U

YaLTeR:
Thanks for the report! @kchibisov told me this may need Smithay changes to fix.

kchibisov:
Ah, no, it's not really a duplicate...

PR #295, opened 2024-04-19
Unconstrain InputMethod's PopupSurface

kchibisov:
Make IME popup to be visible inside the parent and not obscure the
text input rectangle region.

Fixes https://github.com/YaLTeR/niri/issues/221

--

depends on https://github.com/Smithay/smithay/pull/1397


kchibisov:
Haven't tested the grab changes from smithay, but should be good to go.

YaLTeR:
Thanks

PR #307, opened 2024-04-22
Add is_active_in_column

TheZoq2:
This adds a `is-active-in-column` window rule. My primary use case is to "minimize" windows in columns using

```
window-rule {
    match is-active-in-column=false
    max-height 50
    min-height 50
}
```

### DIFF
--- a/wiki/Configuration:-Window-Rules.md
+++ b/wiki/Configuration:-Window-Rules.md
@@ -158,12 +158,12 @@ window-rule {
 ```
 
 #### `is-active-in-column`
+
 Can be `true` or `false`.
-Matches any window that is the "active" window in a column.
-Contrary to
-`is-active` there is one `is-active-in-column` window per column. It is the
-window that was last focused in each column, i.e. the one that will gain focus
-if focus is moved to the column.
+Matches the window that is the "active" window in its column.
+
+Contrary to `is-active`, there is always one `is-active-in-column` window in each column.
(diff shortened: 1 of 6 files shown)


wiki/Configuration:-Window-Rules.md
```diff
@@ -157,6 +157,19 @@ window-rule {
 }
 ```
 
+#### `is-active-in-column`
+Can be `true` or `false`.
+Matches any window that is the "active" window in a column.
+Contrary to
+`is-active` there is one `is-active-in-column` window per column. It is the
+window that was last focused in each column, i.e. the one that will gain focus
+if focus is moved to the column.
```
YaLTeR, line 166:
```suggestion

Can be `true` or `false`.
Matches the window that is the "active" window in its column.

Contrary to `is-active`, there is always one `is-active-in-column` window in each column.
It is the window that was last focused in the column, i.e. the one that will gain focus if this column is focused.
```

YaLTeR (commented):
Looks good, just have a few stylistic nits.

YaLTeR:
And make the tests and rustfmt pass too, please :)

YaLTeR:
Do you mind if I finish this myself before merging?

TheZoq2:
Go ahead, I have been busy with other stuff so I forgot to take care of this. I'm just happy to have it upstreamed!

YaLTeR:
Merged in 47a8e75fd585e351dd2846a58152a7766403ca11. Thanks!

PR #315, opened 2024-04-28
Add support for using a different amount of fingers for swiping.

MagneFire:
This removes the hardcoded prerequisite of using three fingers to use the swipe gesture. Instead one can configure it by setting `gesture-swipe-fingers` in the `input` section.

A working example is:
```
input {
    gesture-swipe-fingers 4
}
```

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -77,6 +77,8 @@ pub struct Input {
     pub focus_follows_mouse: bool,
     #[knuffel(child)]
     pub workspace_auto_back_and_forth: bool,
+    #[knuffel(child, unwrap(argument), default = 3)]
+    pub gesture_swipe_fingers: u32,
 }
 
 #[derive(knuffel::Decode, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


YaLTeR:
Sorry for taking a while to respond. The reason is that I was thinking of different ways to configure this that don't impact future extensibility. For example, I'm thinking of possibly adding a different swipe gesture that will use 4 fingers, in addition to the current gestures using 3.

What's your use case for this, why do you want to change it?

MagneFire:
> Sorry for taking a while to respond. The reason is that I was thinking of different ways to configure this that don't impact future extensibility. For example, I'm thinking of possibly adding a different swipe gesture that will use 4 fingers, in addition to the current gestures using 3.

Hehe, no worries. I'm thankful for you giving the time to respond :smile: 

I thought about this too, as in creating a `Gesture` struct to store all gesture related configuration (not too dissimilar to Hyprland https://wiki.hyprland.org/Configuring/Variables/#gestures).

But as this struct would only contain one option decided to keep it simple and perhaps discuss this in this PR :wink: 

> What's your use case for this, why do you want to change it?

I'm currently using `fusuma` to have the three finger horizontal gesture setup for previous and next actions (`alt+Right` and `alt+Left`). The vertical gesture is used for volume control. The four fingers gesture is setup for workspace related gestures.

This works well for me with GNOME and Hyprland, which is why I wanted the same thing for niri.

PR #347, opened 2024-05-16
Basic Workspace IPC

rustysec:
I just took a hack at getting monitor + workspace data via IPC to facilitate some scripting in AGS/EWW/Waybar. It is pretty sparse, but it has the basic information I wanted. If there is a better/different way to implement this functionality, I am more than willing to go back to the drawing board.

`niri msg -j workspaces | jq`:

```json
[
  {
    "id": 0,
    "display": 0,
    "active": true
  },
  {
    "id": 2,
    "display": 0,
    "active": false
  },
  {
    "id": 1,
    "display": 1,
    "active": true
  },
  {
    "id": 3,
    "display": 1,
    "active": false
  }
]
```

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -31,6 +31,8 @@ pub enum Request {
         /// Configuration to apply.
         action: OutputAction,
     },
+    /// Request information about workspaces.
+    Workspaces,
     /// Respond with an error (for testing error handling).
     ReturnError,
 }
@@ -60,6 +62,8 @@ pub enum Response {
     FocusedWindow(Option<Window>),
     /// Output configuration change result.
     OutputConfigChanged(OutputConfigChanged),
+    /// Information about workspaces.
+    Workspaces(Vec<Workspace>),
 }
 
 /// Actions that niri can perform.
@@ -482,6 +486,23 @@ pub enum OutputConfigChanged {
     Applied,
     /// The target output was not found, the change will be applied when it is connected.
     OutputWasMissing,
+}
+
+/// A workspace.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct Workspace {
+    /// Index of the workspace on its monitor.
+    ///
+    /// This is the same index you can use for requests like `niri msg action focus-workspace`.
+    pub idx: u8,
+    /// Optional name of the workspace.
+    pub name: Option<String>,
+    /// Name of the output that the workspace is on.
+    ///
+    /// Can be `None` if no outputs are currently connected.
+    pub output: Option<String>,
+    /// Whether the workspace is currently active on its output.
+    pub is_active: bool,
 }
 
 impl FromStr for WorkspaceReferenceArg {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -78,6 +78,8 @@ pub enum Msg {
         #[command(subcommand)]
         action: OutputAction,
     },
+    /// List workspaces.
+    Workspaces,
     /// Request an error from the running niri instance.
     RequestError,
 }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -32,12 +32,6 @@ pub struct Cli {
 
 #[derive(Subcommand)]
 pub enum Sub {
-    /// Validate the config file.
-    Validate {
-        /// Path to config file (default: `$XDG_CONFIG_HOME/niri/config.kdl`).
-        #[arg(short, long)]
-        config: Option<PathBuf>,
-    },
     /// Communicate with the running niri instance.
     Msg {
         #[command(subcommand)]
@@ -46,16 +40,22 @@ pub enum Sub {
         #[arg(short, long)]
         json: bool,
     },
+    /// Validate the config file.
+    Validate {
+        /// Path to config file (default: `$XDG_CONFIG_HOME/niri/config.kdl`).
+        #[arg(short, long)]
+        config: Option<PathBuf>,
+    },
     /// Cause a panic to check if the backtraces are good.
     Panic,
 }
 
 #[derive(Subcommand)]
 pub enum Msg {
-    /// Print the version of the running niri instance.
-    Version,
     /// List connected outputs.
     Outputs,
+    /// List workspaces.
+    Workspaces,
     /// Print information about the focused window.
     FocusedWindow,
     /// Perform an action.
@@ -78,8 +78,8 @@ pub enum Msg {
         #[command(subcommand)]
         action: OutputAction,
     },
-    /// List workspaces.
-    Workspaces,
+    /// Print the version of the running niri instance.
+    Version,
     /// Request an error from the running niri instance.
     RequestError,
 }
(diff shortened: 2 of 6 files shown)


rustysec:
Well this does not work as well as I thought it did. Working on some improvements.

rustysec:
Fixed the workspace id mapping, seems to work as intended.

niri-ipc/src/lib.rs
```diff
@@ -60,6 +62,8 @@ pub enum Response {
     FocusedWindow(Option<Window>),
     /// Output configuration change result.
     OutputConfigChanged(OutputConfigChanged),
+    /// Information about workspaces
```
YaLTeR, line 65:
```suggestion
    /// Information about workspaces.
```

niri-ipc/src/lib.rs
```diff
@@ -628,3 +632,14 @@ impl FromStr for ScaleToSet {
         Ok(Self::Specific(scale))
     }
 }
+
+/// Display mapped workspace information
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct IpcWorkspace {
```
YaLTeR, line 638:
```suggestion
pub struct Workspace {
```

niri-ipc/src/lib.rs
```diff
@@ -628,3 +632,14 @@ impl FromStr for ScaleToSet {
...
+    /// Id of the workspace
+    pub id: u32,
+    /// Monitor id the workspace is on
+    pub monitor: usize,
+    /// If the workspace is currently active
+    pub active: bool,
```
YaLTeR, line 644: So, a couple of thoughts.

First, let's rename `id` to `idx`. So that we can expose `id` later when we have a more complete picture of the IPC.

Second, we don't actually have any stable output mapping either (`niri msg outputs` returns a `HashMap`). Let's use the output name. Plus, let's make it an `Option` so it works for the "no connected monitors" case too.

Third, let's add a newly-added workspace `name`.

```suggestion
    /// Index of the workspace on its monitor.
    ///
    /// This is the same index you can use for requests like `niri msg action focus-workspace`.
    pub idx: u8,
    /// Optional name of the workspace.
    pub name: Option<String>,
    /// Name of the output that the workspace is on.
    ///
    /// Can be `None` if no outputs are currently connected.
    pub output: Option<String>,
    /// Whether the workspace is currently active on its output.
    pub is_active: bool,
```

src/cli.rs
```diff
@@ -78,6 +78,8 @@ pub enum Msg {
         #[command(subcommand)]
         action: OutputAction,
     },
+    /// List information about workspaces.
```
YaLTeR, line 81:
```suggestion
    /// List currently existing workspaces.
```

niri-ipc/src/lib.rs
```diff
@@ -60,6 +62,8 @@ pub enum Response {
     FocusedWindow(Option<Window>),
     /// Output configuration change result.
     OutputConfigChanged(OutputConfigChanged),
+    /// Information about workspaces
+    Workspaces(Vec<IpcWorkspace>),
```
YaLTeR, line 66: I thought to maybe return an enum with a HashMap from output name to a vector of workspaces and a bare vector of workspaces as the second variant, but on a second thought this seems to work better (less chance to mess it up in the script with the rare "no outputs" case).

src/ipc/client.rs
```diff
@@ -260,6 +261,25 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+                println!("{response}");
+                return Ok(());
+            }
+
+            for workspace in response {
+                let display = workspace.monitor;
+                let active = if workspace.active { " *" } else { "" };
+                let workspace = workspace.id;
+                println!("{display}: {workspace}{active}");
+            }
```
YaLTeR, line 281: This should sort and display workspaces per monitor. If you don't mind, I will just change this to my liking myself after the other fixes.

src/niri.rs
```diff
@@ -3934,6 +3935,10 @@ impl Niri {
             self.queue_redraw_all();
         }
     }
+
+    pub fn ipc_workspaces(&self) -> Arc<Mutex<Vec<IpcWorkspace>>> {
+        Arc::new(Mutex::new(self.layout.ipc_workspaces()))
```
YaLTeR, line 3940: The other arc-mutexes in the IPC Context struct point at long-lived data that is constantly updated on any change. This function here on the other hand returns a newly constructed piece of data that is never updated, so it can simply return it as a `Vec<IpcWorkspace>`.

(Or actually, I don't think this function is needed because you can just call the layout method directly for now.)

src/layout/workspace.rs
```diff
@@ -113,6 +113,10 @@ impl WorkspaceId {
     fn next() -> WorkspaceId {
         WorkspaceId(WORKSPACE_ID_COUNTER.next())
     }
+
+    pub fn inner(&self) -> u32 {
+        self.0
+    }
```
YaLTeR, line 119:
```suggestion
```
No longer used.

src/layout/mod.rs
```diff
@@ -188,6 +188,33 @@ pub struct Layout<W: LayoutElement> {
...
+                        monitor: idx,
+                        id: (workspace + 1) as _,
+                        active: monitor.active_workspace_idx == workspace,
+                    })
+                }
+            }
+
+            workspaces
+        } else {
+            Default::default()
```
YaLTeR, line 213: Return the list of workspaces here too, with outputs set to `None`.

src/ipc/server.rs
```diff
@@ -198,6 +200,7 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
 
             Response::OutputConfigChanged(response)
         }
+        Request::Workspaces => Response::Workspaces(ctx.ipc_workspaces.lock().unwrap().clone()),
```
YaLTeR, line 203:
```suggestion
        Request::Workspaces => {
            let (tx, rx) = async_channel::bounded(1);
            ctx.event_loop.insert_idle(move |state| {
                let workspaces = state.niri.layout.ipc_workspaces();
                let _ = tx.send_blocking(workspaces);
            });
            let workspaces = rx.recv().await;
            Response::Workspaces(workspaces)
        }
```

YaLTeR (changes requested):
Thanks! Some changes.

src/niri.rs
rustysec: I agree... I originally didn't know what i was getting myself into when I started hacking away, so I made it "match" the signature of the other props. I should have re-worked that when I better understood what I was doing.

src/ipc/client.rs
rustysec: Totally fine with me, have at it.

niri-ipc/src/lib.rs
rustysec: I also started with a `HashMap` but thought this would be cleaner in the long run.

YaLTeR (approved):
Alright, did some minor changes, should be good now.

PR #353, opened 2024-05-18
Update custom-shader example for easy copying and pasting

lpnh:
A slight edit to the [close_custom_shader.frag](https://github.com/YaLTeR/niri/blob/v0.1.6/wiki/examples/close_custom_shader.frag)  example file, making it more straightforward to copy and paste.


YaLTeR:
Thanks! How did I miss this

PR #379, opened 2024-05-22
Added actions to allow focusing up or down that wraps to the next column

minego:
This adds new actions:

```
focus-window-down-or-column-left
focus-window-down-or-column-right
focus-window-up-or-column-left
focus-window-up-or-column-right
```

These behave just like `focus-window-down` and `focus-window-up` but if there is no window in the requested direction then this wraps to the next column. Both directions have been added so that each user can control which direction the wrap moves in.

This change helps me a lot with transitioning from DWL, without changing behavior for those who don't like this.


YaLTeR (approved):
Thanks, looks good, just fix rustfmt.

minego:
Thanks!

I'll update it first thing in the morning. I don't know rust that well, so I missed that...

I'm really enjoying niri so far. I will probably have more PRs in the future.

minego:
I think I got that right this time :-) 

Sorry it took so long. I had hoped to push that first thing this morning, but day job was a lot today.

YaLTeR:
> Sorry it took so long. I had hoped to push that first thing this morning, but day job was a lot today.

No worries, there's no rush

YaLTeR:
Thanks!

PR #382, opened 2024-05-22
active-display

### LINKED ISSUE #362: Expose active display in `niri msg outputs`
### Discussed in https://github.com/YaLTeR/niri/discussions/361

<div type='discussions-op-text'>

<sup>Originally posted by **KiaraGrouwstra** May 20, 2024</sup>
i would like to make a keybind for scaling displays. to this end, i would like to be able to check which display is in focus, such as to be able to use that info to pass to `niri msg output $OUTPUT scale 2`.
to facilitate this use-case, it would be nice if `niri msg outputs` exposed this info.</div>

rustysec:
exposes `is_active: bool` on the `niri_ipc::Output` struct. if there is a better way to do this, please let me know.

resolves #362


src/ipc/server.rs
```diff
@@ -89,6 +89,8 @@ fn on_new_ipc_client(state: &mut State, stream: UnixStream) {
         }
     };
 
+    state.backend.refresh_ipc_outputs(&mut state.niri);
```
YaLTeR, line 92: Computing outputs on every request is something I wanted to avoid with this Arc-Mutex. If we compute outputs like this then we might as well drop the whole Arc-Mutex and do the same as `niri msg workspaces`.

That said, I'm not entirely sure yet how I want to do this. Currently `niri msg outputs` contains *mostly* purely backend-specific stuff (with the exception of logical output info), and this is adding a layout property to it. One way to solve this would be to keep the Arc-Mutex as is, then query only the layout properties at the time of handling the IPC request (i.e. with a channel like `niri msg workspaces`). Another way is to have a separate IPC call for the layout info. Haven't decided what would be the best yet.

rustysec: makes total sense, i can hop in there and try to reduce the frequency of refreshes.

rustysec: i was just thinking through these options, and sway includes a `focused` field in the `get_outputs` ipc response, which was my frame of reference. however, a `focused-output` call would semantically match the `focused-window` ipc message, so that might be the simplest solution at the moment.

YaLTeR: That would be the simplest I guess.

src/ipc/server.rs
```diff
@@ -89,6 +89,8 @@ fn on_new_ipc_client(state: &mut State, stream: UnixStream) {
         }
     };
 
+    // state.backend.refresh_ipc_outputs(&mut state.niri);
+
```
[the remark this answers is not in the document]
rustysec: oof :man_facepalming:

rustysec:
sorry, i went on auto pilot there, didn't rebase that properly :disappointed:

src/ipc/client.rs
```diff
@@ -313,3 +241,98 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+        let qualifier = match (is_current, is_preferred) {
+            (true, true) => " (current, preferred)",
+            (true, false) => " (current)",
+            (false, true) => " (preferred)",
+            (false, false) => "",
+        };
+
+        println!("    {width}x{height}@{refresh:.3}{qualifier}");
+    }
+    println!();
```
YaLTeR, line 336: Nitpick, please move this line outside because in `focused-output` it isn't needed.

src/ipc/server.rs
```diff
@@ -208,6 +208,21 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
...
+                let active_output = state
+                    .niri
+                    .layout
+                    .active_output()
+                    .map(|output| output.name());
+                let _ = tx.send_blocking(active_output);
+            });
+            let result = rx.recv().await;
+            let name = result.map_err(|_| String::from("error getting active output info"))?;
+            let output = name.and_then(|name| ctx.ipc_outputs.lock().unwrap().get(&name).cloned());
```
YaLTeR, line 223: Do this inside the `insert_idle` closure to guarantee that the output exists. Otherwise it can get disconnected between when the idle runs, and when this line runs. Instead of `ctx.ipc_outputs` you can extract it from `state.niri.global_space.outputs()`.

rustysec: i have been looking at this for a little bit and the `global_space.outputs()` returns smithay `Output`s instead of `niri_ipc::Output`s. what am i missing?

YaLTeR: Oh, right. Sorry, got confused. Then get it from the arc-mutex, but within the closure. The arc-mutex should be available somewhere in the backend or in the niri state.

rustysec: that is what i ended up doing, thanks

src/ipc/server.rs
```diff
@@ -208,6 +208,31 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
...
+        Request::FocusedOutput => {
+            let (tx, rx) = async_channel::bounded(1);
+            ctx.event_loop.insert_idle(move |state| {
+                let active_output = state
+                    .niri
+                    .layout
+                    .active_output()
+                    .map(|output| output.name());
+
+                let output = active_output.clone().and_then(|active_output| {
```
YaLTeR, line 220: I don't think this clone is necessary?

rustysec: yep, just left over from me debugging.

YaLTeR:
Thanks!

PR #391, opened 2024-05-25
Add `focus-column-right-or-first`, `focus-column-left-or-last`

sullyj3:
Adds `focus-column-next`, which focuses the next column to the right, looping back to the start if we're at the last column.

In my config, I bind it to `Mod+Tab`, it's useful when your hand is on the mouse.

No pressure, just whipped it up to get feedback. What do you think?

Alternatives:
- Instead of adding a new action, modify the behaviour of `focus-column-right`.
 Decided against this because I think most people are used to the current non-looping behaviour and would prefer it to stay the same. It's more predictable.
- different name: `focus-column-right-loop` or similar
- could also add the symmetric `focus-column-prev`. I think this makes sense, I'll do this if you're interested in this feature.


YaLTeR:
Hm, well, if we were to add this we could call it `focus-column-right-or-first` for consistency with other actions. And add `focus-column-left-or-last`.

Another idea is to add a `loop=true` property, but no other actions do this currently, so I'm leaning against it.

Vermoot:
I'd love to see a `loop=true` property, I would definitely use it for `focus-workspace-down`, `focus-column-next`, and `focus-window-down-or-column-right` (and their symmetric counterparts)

sullyj3:
I don't have a strong preference. Up to you @YaLTeR. If you pick `loop=true` I'll close this, since it seems like a more substantial change that I don't want to do myself. I can implement `focus-column-right-or-first`/`focus-column-left-or-last` otherwise.

Given that niri is pre 1.0, it doesn't seem that big a deal to me to implement `focus-column-right-or-first`/`focus-column-left-or-last` now, and deprecate in favour of `loop=true` later if you decide you want to go that route.

sullyj3:
Ok, this is ready for review

src/layout/workspace.rs
```diff
@@ -1536,6 +1536,28 @@ impl<W: LayoutElement> Workspace<W> {
...
+    pub fn focus_column_left_or_last(&mut self) {
+        if self.columns.is_empty() {
+            return;
+        }
+
+        let column_idx = if self.active_column_idx == 0 {
+            self.columns.len() - 1
+        } else {
+            self.active_column_idx - 1
+        };
```
sullyj3, line 1557: Could also do
```rust
let column_idx = match self.active_column_idx.overflowing_sub(1) {
    (i, false) => i,
    (_, true) => self.columns.len() - 1,
}
```
I prefer the `if` though

YaLTeR: Yeah, if is good.

YaLTeR:
Thanks!

sullyj3:
thank you!

PR #397, opened 2024-05-26
Add scroll-method property to pointer devices

yuja:
My use case is to enable middle-button scroll on my keyboard with pointing stick. The device is recognized as USB mouse.

~Variant names are copied from Sway. I'm not sure if we should rename "none" to "no-scroll" (according to input::ScrollMethod), or rename our config enum to None.~
> input <identifier> scroll_method none|two_finger|edge|on_button_down
>     Changes the scroll method for the specified input device.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -170,6 +170,8 @@ pub struct Touchpad {
     #[knuffel(child, unwrap(argument, str))]
     pub accel_profile: Option<AccelProfile>,
     #[knuffel(child, unwrap(argument, str))]
+    pub scroll_method: Option<ScrollMethod>,
+    #[knuffel(child, unwrap(argument, str))]
     pub tap_button_map: Option<TapButtonMap>,
     #[knuffel(child)]
     pub left_handed: bool,
@@ -183,6 +185,8 @@ pub struct Mouse {
     pub accel_speed: f64,
     #[knuffel(child, unwrap(argument, str))]
     pub accel_profile: Option<AccelProfile>,
+    #[knuffel(child, unwrap(argument, str))]
+    pub scroll_method: Option<ScrollMethod>,
     #[knuffel(child)]
     pub left_handed: bool,
 }
@@ -195,6 +199,8 @@ pub struct Trackpoint {
     pub accel_speed: f64,
     #[knuffel(child, unwrap(argument, str))]
     pub accel_profile: Option<AccelProfile>,
+    #[knuffel(child, unwrap(argument, str))]
+    pub scroll_method: Option<ScrollMethod>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


niri-config/src/lib.rs
```diff
@@ -183,6 +185,8 @@ pub struct Mouse {
     pub accel_speed: f64,
     #[knuffel(child, unwrap(argument, str))]
     pub accel_profile: Option<AccelProfile>,
+    #[knuffel(child, unwrap(argument, str))]
+    pub scroll_method: Option<ScrollMethod>,
```
YaLTeR, line 189: So this does make sense on a mouse device? Rather than just touchpad and trackpoint.

wiki/Configuration:-Input.md
```diff
@@ -124,6 +127,9 @@ A few settings are common between `touchpad`, `mouse` and `trackpoint`:
 - `natural-scroll`: if set, inverts the scrolling direction.
 - `accel-speed`: pointer acceleration speed, valid values are from `-1.0` to `1.0` where the default is `0.0`.
 - `accel-profile`: can be `adaptive` (the default) or `flat` (disables pointer acceleration).
+- `scroll-method`: when to generate scroll events instead of pointer motion
+  events, can be `none`, `two-finger`, `edge`, or `on-button-down`. The default
+  and supported methods vary depending on the device type.
```
YaLTeR, line 132: I wrap the md files in this repo by sentence.

```suggestion
- `scroll-method`: when to generate scroll events instead of pointer motion events, can be `none`, `two-finger`, `edge`, or `on-button-down`.
  The default and supported methods vary depending on the device type.
```

YaLTeR (changes requested):
> I'm not sure if we should rename "none" to "no-scroll" (according to input::ScrollMethod)

Let's change to no-scroll for consistency with libinput.

niri-config/src/lib.rs
yuja: > So this does make sense on a mouse device?

Yes. The supported methods vary depending on devices, though. For example, my USB keyboard has pointing stick, but the device is recognized as mouse (on both Niri and Sway.)
```
Device:           PFU Limited HHKB-Studio
Kernel:           /dev/input/event5
Group:            6
Seat:             seat0, default
Capabilities:     pointer 
Tap-to-click:     n/a
Tap-and-drag:     n/a
Tap drag lock:    n/a
Left-handed:      disabled
Nat.scrolling:    disabled
Middle emulation: disabled
Calibration:      n/a
Scroll methods:   button
Click methods:    none
Disable-w-typing: n/a
Disable-w-trackpointing: n/a
Accel profiles:   flat *adaptive custom
Rotation:         0.0
```

Perhaps, the device type can be fixed by udev hwdb, but I guess physical mouse also supports `on-button-down` method.

YaLTeR:
Thanks!

PR #400, opened 2024-05-26
Support `$NIRI_CONFIG` environment variable.

sodiboo:
On NixOS, it is a common pattern for config files to be in the read-only `/nix/store/`. With my current niri-flake, that is where it ends up, and it is symlinked to `~/.config/niri/config.kdl`.

There are various reasons to not symlink that, mainly if you need multiple configs for a single user. When supplying config files from the nix store, it is standard practice to pass them with an environment variable using the standard `makeWrapper` helper function ([example in `openrefine`](https://github.com/NixOS/nixpkgs/blob/5a279778af727927f638dcf136a1a04cd550b3e2/pkgs/applications/science/misc/openrefine/default.nix#L117)). This is because the CLI options on most programs are not flexible enough.

it is easy to use a custom config by passing the `--config` flag, to either `niri` or `niri validate`, but it becomes a problem to prefix it to all commands. This is because (1) it will be rejected when invoking subcommands, so you can no longer call `niri validate` or `niri msg`, and (2) it cannot be passed multiple times, so prefixing it like this is essentially disabling the `--config` flag for all users of this wrapper. Essentially, it will cause all niri commands to be invalid except `niri` and `niri --session`.

Instead of bodging something strange onto the cli arguments, i've instead added an environment variable. Then, it can be easily substituted with the standard wrapper approach.

This allows me to easily create a `pkgs.niri-unstable.with-config` function that can add this in a wrapper. I've been thinking of doing this for a long time, but didn't implement anything because i was unsure of any real use cases for it. Since then, i've started using niri on my login screen, something which currently relies [heavily](https://github.com/sodiboo/nix-config/blob/8ca2b21c61a4f23052c67e52276bb673a574e17c/login.mod.nix#L23-L25) on [implementation details](https://github.com/sodiboo/nix-config/blob/8ca2b21c61a4f23052c67e52276bb673a574e17c/login.mod.nix#L66) of `niri-flake`. Additionally, i believe such a function (for which this environment variable is necessary) would also resolve https://github.com/sodiboo/niri-flake/issues/278.


minego:
I love this change and am very excited for it. This will greatly simplify my nixOS setup. Thank you.

wiki/Configuration:-Overview.md
```diff
@@ -13,7 +13,7 @@ You can find documentation for various sections of the config on these wiki page
 
 ### Loading
 
-Niri will load configuration from `$XDG_CONFIG_HOME/.config/niri/config.kdl` or `~/.config/niri/config.kdl`.
+Niri will load configuration from `$XDG_CONFIG_HOME/niri/config.kdl` or `~/.config/niri/config.kdl`.
```
YaLTeR, line 16: Good catch!

YaLTeR (commented):
I see what the problem is. You mentioned that some programs have a config environment variable for this reason. However, I'd like to know, how do other compositors handle this? Since compositors have unique questions in this regard, namely:

Should `NIRI_CONFIG` be propagated to children?
 * If yes, other programs developed on NixOS may start relying on the presence of this. We might need to start force setting the variable then.
 * If not, running nested development niri (i.e. `./target/debug/niri`) won't pick it up.

sodiboo:
> I'd like to know, how do other compositors handle this?

With Sway, the `sway` command is the only one that accepts a config argument. The problems with prefixing an argument don't apply because (1) all other, non-"start the compositor" commands (e.g. `swaymsg`), are not affected and (2) there are no subcommands that "conflict" with the config argument. Additionally, passing `--config` mulitple times isn't an issue, it will just accept the last one. It has a couple "subcommands" but they also don't conflict with `--config`, so e.g. `sway --config a --config b --validate` will validate the config at `b` ignore `a`, and `sway --config a --config b --version` will print the version. (remember: `--config a` is injected by a wrapper, so the user really typed `sway --config b --validate`, which ought to work, and it does for Sway but not in the current state of niri)

Hyprland works similarly: the `Hyprland` executable has only three possible arguments, `--help`, `--config` and `--i-am-really-stupid` (allows to run as root). It's kind of strange, in that it accepts the *first* config and also doesn't ignore configs if passed `--help`. Nonetheless, this is easily "wrapped" by adding arguments to the *end* of the invocation. `hyprctl` is separate executable, and a wrapper wouldn't touch.

Wayfire supports "last passed `--config` is active" like Sway and Hyprland, but also reads from `$WAYFIRE_CONFIG_FILE` if no argument is passed to it. This environment variable seems unnecessary for use with `makeWrapper`, because a prefix argument works just as well with wayfire, but they might have other use cases in mind. **Wayfire propagates the environment variable to processes spawned in its `autostart` config section.**

Reworking the CLI arguments such that a prefix of `--config` is always valid for all invocations, and having it at the start also affects `niri validate` would be fine. All of the above supports this, actually. But they don't have IPC on the same executable.

> Should `NIRI_CONFIG` be propagated to children?

Intuitively, I'd say no. `makeWrapper` should typically be completely opaque and really quite indistinguishable from a search/replace for the default path in a source code. Generally, propagating a config to child processes isn't a consideration for most programs, so this isn't something i had thought of before.

Ideally, if `NIRI_CONFIG` is set globally, it should be propagated, but not if invoked from `makeWrapper`. This is basically impossible to detect, so it's probably easier to just accept that we should just accept it as a global.

> If yes, other programs developed on NixOS may start relying on the presence of this. We might need to start force setting the variable then.

If other programs need to read from the niri config, i reckon it'd make more sense to implement as a `niri msg` subcommand. Either to dump the path, or something more elaborate that can query settings in a better structured way (supporting defaults, etc). I don't think we should support programs trying to rely on this environment variable to read config values, and the documentation could be updated to reflect such a stance. We can enforce such a stance by deliberately not setting `NIRI_CONFIG` when it is passed as a command line argument, making it an unreliable way to find the "active config".

> If not, running nested development niri (i.e. ./target/debug/niri) won't pick it up.

If we don't propagate it, but it is desired to use for a nested dev niri, it is fairly easy (on NixOS at least) to pass the config path as a variable. If we say that the config lives a `/nix/store/c0ffee-config.kdl` (hashes are longer than `c0ffee`), then it can contain a line like `environment { NIRI_CONFIG "/nix/store/c0ffee-config.kdl"; }`. This is because nix store hashes (for regular derivations) are based on inputs, so it is known in advance at "build-time" (strange terminology for a config file; but this essentially means "the point at which this config is written to the nix store") and can be circularly referenced like that.

Note that i did not say this is trivial; it requires some restructuring in `niri-flake` to achieve this easily. Such functionality could also just be written into niri itself with some `propagate-config-env` flag in the config file. But it's not strictly necessary for NixOS (and of course, on systems without a `/nix/store/`-like path, it's also easy to include the path in the config as `/etc/niri/config.kdl` or wherever you're putting it)

I would however argue that the nested compositor not picking it up is desirable in some cases too. I've personally been working on a branch once or twice that is based on an older version of niri, where my current configuration is invalid. It would honestly be fine for such a use case to have `NIRI_CONFIG` not propagate, and reserve `$XDG_CONFIG_HOME/niri/config.kdl` for a "testing config" that only applies to the nested instance. This is the reason why it's ignored when set to an invalid value, such as an empty string, so that it *can* be explicitly unset.

---

Ultimately, i don't think it's a significant problem either way. It doesn't affect my use case. The two main arguments in favor of either way are:

We don't want other programs to rely on this environment variable => unset it

and

We do want nested development niri to be able to rely on it => propagate it.

(note that general "nested niri" doesn't apply really, at least not on NixOS, because in that case it would still be executed in the wrapper and get the config no matter what)

---

As for adding a `niri msg` subcommand to get the config path, it could make sense to use this in the niri compositor as well. As in, either make the default config path first try to get it from `niri msg`, or add a flag like `--inherit-config` that will try to run with the same config as the outer compositor. This feels strange? Not sure we should be doing it, but worth considering to some extent. Maybe pointless, since a similar result could be achieved with `NIRI_CONFIG=$(niri msg config-path)` before any invocation.

YaLTeR:
Thanks for the detailed explanation! Hmm, there's probably a relatively straightforward way to always accept `--config` and take the last value in case multiple are set. Maybe that's indeed better, if only because it doesn't add an extra way of passing the config?

> niri msg subcommand to get the config path

Could be added if anyone needs it I guess.

> add a flag like --inherit-config that will try to run with the same config as the outer compositor. This feels strange?

I agree it's kinda strange; don't think it's necessary.

sodiboo:
> there's probably a relatively straightforward way to always accept `--config` and take the last value in case multiple are set. Maybe that's indeed better, if only because it doesn't add an extra way of passing the config?

I don't wish to implement this personally, as it's kind of annoying to restructure the entire cli for what is seemingly one use case. That being said, if it were done and `--config` could reliably be passed to all `niri` invocations unconditionally, that would be entirely sufficient for my NixOS use-cases.

I also think that for niri in particular, it's not the most intuitive solution, because then you can pass `--config` to `niri msg` which doesn't really make sense. The thing in common for most other compositors is that the command that accepts `--config` is *only* for launching the compositor. All possible permutations of the command (except `--help` and `--version`) would actually use the config in a meaningful way.

src/main.rs
```diff
@@ -315,6 +316,12 @@ fn import_environment() {
     }
 }
 
+fn env_config_path() -> Option<PathBuf> {
+    env::var_os("NIRI_CONFIG")
+        .map(PathBuf::from)
+        .and_then(|p| p.is_file().then_some(p))
```
YaLTeR, line 322: We probably want to log a warning here.

wiki/Configuration:-Overview.md
```diff
@@ -25,6 +25,8 @@ You can run `niri validate` to parse the config and see any errors.
 
 To use a different config file path, pass it in the `--config` or `-c` argument to `niri`.
 
+You can also set `$NIRI_CONFIG` to the path of the config file. `--config` always takes precedence. If `--config` doesn't point to a real file, the config will not be loaded. If `$NIRI_CONFIG` doesn't point to a real file, it is ignored and the default path is used instead.
```
YaLTeR, line 28:
```suggestion
You can also set `$NIRI_CONFIG` to the path of the config file.
`--config` always takes precedence.
If `--config` doesn't point to a real file, the config will not be loaded.
If `$NIRI_CONFIG` doesn't point to a real file, it is ignored and the default path is used instead.
```

YaLTeR, line 28: Why is `$NIRI_CONFIG` ignored if missing, instead of acting the same way as `--config`?

src/main.rs
```diff
@@ -113,7 +114,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Load the config.
     let mut config_created = false;
-    let path = cli.config.or_else(|| {
+    let path = cli.config.or_else(env_config_path).or_else(|| {
```
YaLTeR, line 117: `NIRI_CONFIG` should be removed from the environment somewhere below the last use.

wiki/Configuration:-Overview.md
sodiboo: My main thought here is that a user can wish to remove this on an individual invocation, which is slightly more annoying if you need to unset the environment variable as opposed to something like `NIRI_CONFIG="" niri` which sets it to an empty string; i.e. an invalid path. It could act the same as `--config`, it's just not what i felt would be the easiest for me to use, personally. Something like `env -u NIRI_CONFIG niri` should be an actual equivalent way to run it by unsetting `NIRI_CONFIG`, the only downside to this being that i had to actually look at documentation to find what magic incantation unsets an environment variable (whereas the env prefix is standard and common in all shells). That's a very artificial reason to do it this way, but it is my thought process.

YaLTeR: Hm, that is a valid point.

YaLTeR: But also, it can explicitly check for is_empty(), rather than path existing?

YaLTeR: Did that

Issue #371, opened 2024-05-21
Specific input device configuration


YaLTeR:
Ability to configure specific input devices.

Syntax: something along
```
input {
    mouse "identifier" {
        // Settings for this device.
    }
}
```

Resolution for a given device:
1. Search for settings block matching this device's identifier.
2. If not found, use the general block for this device's type.

Identifier: check what other compositors use, guess it's something libinput should be exposing.

For keyboards, it should support per-keyboard XKB config.

FreeFull:
This should include the ability to disable the device, too

rustysec:
I was looking through the current implementation and at the `smithay` code, and I don't see how to accomplish this at the moment. 100% possible I'm just missing something obvious, still trying to get to know the code base.

YaLTeR:
Are you referring to something in particular? For libinput settings, it should be possible to simply get the device name and find the right config section in [`apply_libinput_settings()`](https://github.com/YaLTeR/niri/blob/efb39e466b5248eb894745e899de33661493511d/src/input/mod.rs#L2193).

For dynamic XKB, you need to look in  `on_keyboard()`, extract the libinput device [like this](https://github.com/YaLTeR/niri/blob/efb39e466b5248eb894745e899de33661493511d/src/input/mod.rs#L236) (pretty cursed, but possible), then switch the XKB settings similar to how config hot-reload works I guess?

nebulosa2007:
> This should include the ability to disable the device, too

I'm expand the request: it should be permanent/toggle (via config file) or eventually (via `niri msg action ...`). 

Use cases are: hotkeys to on/off touchpad, temporary off touchpad while mouse is connected.

YaLTeR:
This can work similarly to `niri msg output`: apply a config change transiently until the input section is modified in the real config.

rustysec:
@YaLTeR Aha, that was the "obvious" hint I was missing. My cursory look through the config and input code just missed this. Thanks!

YaLTeR:
So that we don't lose it: there's a close-to-complete PR for per-keyboard configuration at #417.

expectocode:
I just wanted to chime in and say - this project is amazing! I've been stuck on X forever on my setup (although I've loved PaperWM in the past), and I was sure that per-keyboard configuration would be the kind of niche thing that would prevent me from using such a cool project as Niri. It's awesome to see that you're already working on it :D Nice job, YaLTeR.

YaLTeR:
[...]

PR #407, opened 2024-05-28
Add "off" and "disabled-on-external-mouse" properties to input devices

yuja:
This is called "events <mode>" in Sway, ~but I choose "send-events-mode" for consistency with libinput.~

I think this option also applies to keyboards, but I'm not going to add the one because we don't have libinput machinery for the keyboard config, and it's unlikely that user wants to disable _all_ keyboards. OTOH, pointer devices can be disabled per type. Perhaps, this should be revisited after implementing #371.

### DIFF
--- a/wiki/Configuration:-Input.md
+++ b/wiki/Configuration:-Input.md
@@ -25,6 +25,7 @@ input {
     }
 
     touchpad {
+        // off
         tap
         // dwt
         // dwtp
@@ -35,9 +36,11 @@ input {
         // tap-button-map "left-middle-right"
         // click-method "clickfinger"
         // left-handed
+        // disabled-on-external-mouse
     }
 
     mouse {
+        // off
         // natural-scroll
         // accel-speed 0.2
         // accel-profile "flat"
@@ -46,6 +49,7 @@ input {
     }
 
     trackpoint {
+        // off
         // natural-scroll
         // accel-speed 0.2
         // accel-profile "flat"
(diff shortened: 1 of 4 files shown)


YaLTeR:
Hmm, is this how one is supposed to implement disabling the input device in general (with the `disabled` mode)? I'm asking because general disabling should probably use `off` for consistency. So how do we do this then

yuja:
> Hmm, is this how one is supposed to implement disabling the input device in general (with the `disabled` mode)? I'm asking because general disabling should probably use `off` for consistency. So how do we do this then

Okay, I didn't notice `off` is widely used as an abstraction of enabled/disabled items. In that case, maybe we can add `off` (to all input devices) and boolean `disabled-on-external-mouse` (to touchpads). Fortunately, the default of `send_events_mode` is hard-coded to `ENABLED`, so we don't need an explicit `on` flag.

YaLTeR:
Yeah, that sounds like a better idea.

wiki/Configuration:-Input.md
```diff
@@ -132,6 +141,7 @@ A few settings are common between `touchpad`, `mouse` and `trackpoint`:
 
 Settings specific to `touchpad`s:
 
+- `disabled-on-external-mouse`: do not send events while external pointer device is plugged in.
```
YaLTeR, line 144: Could you move this to the bottom of this list?

YaLTeR (commented):
Thanks, looks good, just a minor change

wiki/Configuration:-Input.md
yuja: Sure. Updated struct field, too.

Issue #333, opened 2024-05-08
Support Custom modeline


FlyingWombat:
My display needs custom modeline timings set to work properly (my gpu can't reclock its memory). I'm just checking out Niri, but even if I don't end up using Niri long-term, this will be helpful for others.

An example config could be something like this:
```
output "DP-1" {
    mode "1332.75 3840 3888 3920 4000 2160 2163 2168 2314 +HSync -VSync"
}
```

YaLTeR:
Hm, I'm not sure off the bat how to do that, might need Smithay support.

FlyingWombat:
Sway/wlroots supports it with `output DP-1 modeline <modeline>`. That might be useful for reference.

Should I create a related issue in Smithay?

YaLTeR:
I'll ask in the Smithay room. Maybe it can already do it.

FlyingWombat:
FYI, I found out that linux can [load custom edid firmware in KMS](https://wiki.archlinux.org/title/Kernel_mode_setting#Forcing_modes_and_EDID) -- so it isn't strictly necessary to configure custom modelines from the compositor. That said, it will be much more user friendly to be able to set it from the compositor.

------------

Here is some more info for anyone else who comes across this thread.

Having the kernel load custom edid firmware was easy. Just put the .bin in `/usr/lib/firmware/edid/` and add `drm.edid_firmware=edid/your_edid.bin` to kernel cmdline.

The trouble comes in getting a proper edid binary. You can try [edid-generator](https://github.com/akatrevorjay/edid-generator) or building one from scratch, but most will be better off editing a copy of the existing firmware. You can grab a connection's edid with e.g. `cat /sys/class/drm/card2-DP-5/edid > edid.bin`.

In my case, I had to edit it by hand with a hex editor. None of the edid editors I found worked properly. [wxEDID](https://sourceforge.net/projects/wxedid/) doesn't support DisplayID extension blocks -- though it may work for edid bins only containing EDID with CTA extension. [CRU](https://www.monitortests.com/forum/Thread-Custom-Resolution-Utility-CRU) on wine supported all the exensions, but rewrote the entire edid assuming my display was a generic analog CRT. There is also edid-generator, but I didn't try that because I wanted to keep the other display properties from the original edid.

[...]

PR #413, opened 2024-05-31
feat: wlr output management

### LINKED ISSUE #370: Support wlr-output-management-unstable-v1
It should apply the output configuration transiently, exactly like `niri msg output`. I'm not sure how exactly testing a configuration should work, but I think cosmic-comp has that, so can reference that.

gmorer:
This implement the [wlr output management](https://wayland.app/protocols/wlr-output-management-unstable-v1) which allow other program to control outputs.

My goal was to use niri with [kanshi](https://sr.ht/~emersion/kanshi/) which succeeded.

I don't know any other up to date software (it implement the v4 of the protocol) to further test the implementation.

should fix #370

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -555,7 +555,7 @@ impl OutputManagementHandler for State {
         &mut self.niri.output_management_state
     }
 
-    fn apply_output_config(&mut self, config: Vec<niri_config::Output>) {
+    fn apply_output_config(&mut self, config: niri_config::Outputs) {
         self.niri.config.borrow_mut().outputs = config;
         self.reload_output_config();
     }
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -68,9 +68,13 @@ use crate::protocols::foreign_toplevel::{
     self, ForeignToplevelHandler, ForeignToplevelManagerState,
 };
 use crate::protocols::gamma_control::{GammaControlHandler, GammaControlManagerState};
+use crate::protocols::output_management::{OutputManagementHandler, OutputManagementManagerState};
 use crate::protocols::screencopy::{Screencopy, ScreencopyHandler};
 use crate::utils::{output_size, send_scale_transform};
-use crate::{delegate_foreign_toplevel, delegate_gamma_control, delegate_screencopy};
+use crate::{
+    delegate_foreign_toplevel, delegate_gamma_control, delegate_output_management,
+    delegate_screencopy,
+};
 
 impl SeatHandler for State {
     type KeyboardFocus = WlSurface;
@@ -545,3 +549,15 @@ delegate_xdg_activation!(State);
 
 impl FractionalScaleHandler for State {}
 delegate_fractional_scale!(State);
+
+impl OutputManagementHandler for State {
+    fn output_management_state(&mut self) -> &mut OutputManagementManagerState {
+        &mut self.niri.output_management_state
+    }
+
+    fn apply_output_config(&mut self, config: Vec<niri_config::Output>) {
+        self.niri.config.borrow_mut().outputs = config;
+        self.reload_output_config();
+    }
+}
+delegate_output_management!(State);
--- a/src/protocols/output_management.rs
+++ b/src/protocols/output_management.rs
@@ -38,7 +38,7 @@ pub struct OutputManagementManagerState {
     serial: u32,
     clients: HashMap<ClientId, ClientData>,
     current_state: HashMap<OutputId, niri_ipc::Output>,
-    current_config: Vec<niri_config::Output>,
+    current_config: niri_config::Outputs,
 }
 
 pub struct OutputManagementManagerGlobalData {
@@ -47,7 +47,7 @@ pub struct OutputManagementManagerGlobalData {
 
 pub trait OutputManagementHandler {
     fn output_management_state(&mut self) -> &mut OutputManagementManagerState;
-    fn apply_output_config(&mut self, config: Vec<niri_config::Output>);
+    fn apply_output_config(&mut self, config: niri_config::Outputs);
 }
 
 #[derive(Debug)]
@@ -84,11 +84,11 @@ impl OutputManagementManagerState {
             clients: HashMap::new(),
             serial: 0,
             current_state: HashMap::new(),
-            current_config: Vec::new(),
+            current_config: Default::default(),
         }
     }
 
-    pub fn on_config_changed(&mut self, new_config: Vec<niri_config::Output>) {
+    pub fn on_config_changed(&mut self, new_config: niri_config::Outputs) {
         self.current_config = new_config;
     }
 
@@ -405,8 +405,7 @@ where
                     Entry::Vacant(entry) => {
                         let mut config = g_state
                             .current_config
-                            .iter()
-                            .find(|o| o.name.eq_ignore_ascii_case(&current_config.name))
+                            .find(&current_config.name)
                             .cloned()
                             .unwrap_or_else(|| niri_config::Output {
                                 name: current_config.name.clone(),
@@ -455,8 +454,7 @@ where
                     Entry::Vacant(entry) => {
                         let mut config = g_state
                             .current_config
-                            .iter()
-                            .find(|o| o.name.eq_ignore_ascii_case(&current_config.name))
+                            .find(&current_config.name)
                             .cloned()
                             .unwrap_or_else(|| niri_config::Output {
                                 name: current_config.name.clone(),
--- a/src/protocols/output_management.rs
+++ b/src/protocols/output_management.rs
@@ -1,0 +1,891 @@
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::iter::zip;
+use std::mem;
+
+use niri_config::FloatOrInt;
+use niri_ipc::Transform;
+use smithay::reexports::wayland_protocols_wlr::output_management::v1::server::{
+    zwlr_output_configuration_head_v1, zwlr_output_configuration_v1, zwlr_output_head_v1,
+    zwlr_output_manager_v1, zwlr_output_mode_v1,
+};
+use smithay::reexports::wayland_server::backend::ClientId;
+use smithay::reexports::wayland_server::protocol::wl_output::Transform as WlTransform;
+use smithay::reexports::wayland_server::{
+    Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource, WEnum,
+};
+use zwlr_output_configuration_head_v1::ZwlrOutputConfigurationHeadV1;
+use zwlr_output_configuration_v1::ZwlrOutputConfigurationV1;
+use zwlr_output_head_v1::{AdaptiveSyncState, ZwlrOutputHeadV1};
+use zwlr_output_manager_v1::ZwlrOutputManagerV1;
+use zwlr_output_mode_v1::ZwlrOutputModeV1;
+
+use crate::backend::OutputId;
+use crate::niri::State;
+use crate::utils::ipc_transform_to_smithay;
+
+const VERSION: u32 = 4;
+
+#[derive(Debug)]
+struct ClientData {
+    heads: HashMap<OutputId, (ZwlrOutputHeadV1, Vec<ZwlrOutputModeV1>)>,
+    confs: HashMap<ZwlrOutputConfigurationV1, OutputConfigurationState>,
+    manager: ZwlrOutputManagerV1,
+}
+
+pub struct OutputManagementManagerState {
+    display: DisplayHandle,
+    serial: u32,
+    clients: HashMap<ClientId, ClientData>,
+    current_state: HashMap<OutputId, niri_ipc::Output>,
+    current_config: Vec<niri_config::Output>,
+}
+
+pub struct OutputManagementManagerGlobalData {
+    filter: Box<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
+}
+
+pub trait OutputManagementHandler {
+    fn output_management_state(&mut self) -> &mut OutputManagementManagerState;
+    fn apply_output_config(&mut self, config: Vec<niri_config::Output>);
+}
+
+#[derive(Debug)]
+enum OutputConfigurationState {
+    Ongoing(HashMap<OutputId, niri_config::Output>),
+    Finished,
+}
+
+pub enum OutputConfigurationHeadState {
+    Cancelled,
+    Ok(OutputId, ZwlrOutputConfigurationV1),
+}
+
+impl OutputManagementManagerState {
+    pub fn new<D, F>(display: &DisplayHandle, filter: F) -> Self
+    where
+        D: GlobalDispatch<ZwlrOutputManagerV1, OutputManagementManagerGlobalData>,
+        D: Dispatch<ZwlrOutputManagerV1, ()>,
+        D: Dispatch<ZwlrOutputHeadV1, OutputId>,
+        D: Dispatch<ZwlrOutputConfigurationV1, u32>,
+        D: Dispatch<ZwlrOutputConfigurationHeadV1, OutputConfigurationHeadState>,
+        D: Dispatch<ZwlrOutputModeV1, ()>,
+        D: OutputManagementHandler,
+        D: 'static,
+        F: for<'c> Fn(&'c Client) -> bool + Send + Sync + 'static,
+    {
+        let global_data = OutputManagementManagerGlobalData {
+            filter: Box::new(filter),
+        };
+        display.create_global::<D, ZwlrOutputManagerV1, _>(VERSION, global_data);
+
+        Self {
+            display: display.clone(),
+            clients: HashMap::new(),
+            serial: 0,
+            current_state: HashMap::new(),
+            current_config: Vec::new(),
+        }
+    }
+
+    pub fn on_config_changed(&mut self, new_config: Vec<niri_config::Output>) {
+        self.current_config = new_config;
+    }
+
+    pub fn notify_changes(&mut self, new_state: HashMap<OutputId, niri_ipc::Output>) {
+        let mut changed = false; /* most likely to end up true */
+        for (output, conf) in new_state.iter() {
+            if let Some(old) = self.current_state.get(output) {
+                if old.vrr_enabled != conf.vrr_enabled {
+                    changed = true;
+                    for client in self.clients.values() {
+                        if let Some((head, _)) = client.heads.get(output) {
+                            if head.version() >= zwlr_output_head_v1::EVT_ADAPTIVE_SYNC_SINCE {
+                                head.adaptive_sync(match conf.vrr_enabled {
+                                    true => AdaptiveSyncState::Enabled,
+                                    false => AdaptiveSyncState::Disabled,
+                                });
+                            }
+                        }
+                    }
+                }
+
+                // TTY outputs can't change modes I think, however, winit and virtual outputs can.
+                let modes_changed = old.modes != conf.modes;
+                if modes_changed {
+                    changed = true;
+                    if old.modes.len() != conf.modes.len() {
+                        error!("output's old mode count doesn't match new modes");
+                    } else {
+                        for client in self.clients.values() {
+                            if let Some((_, modes)) = client.heads.get(output) {
+                                for (wl_mode, mode) in zip(modes, &conf.modes) {
+                                    wl_mode.size(i32::from(mode.width), i32::from(mode.height));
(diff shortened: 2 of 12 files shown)


src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+    fn bind(
+        state: &mut D,
+        display: &DisplayHandle,
+        client: &Client,
+        manager: New<ZwlrOutputManagerV1>,
+        _manager_state: &OutputManagementManagerGlobalData,
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        let manager = data_init.init(manager, ());
+        let outputs = state.get_outputs();
```
YaLTeR, line 102: The way I think this should work is:
* `OutputManagementManagerState` stores the "current state" as far as it knows. I suppose it already does, in `current_out`.
* When sending the state to the clients, it uses its stored "current state". Not the "newest state", to avoid any potential desync problems.
* The "current state" is updated to "newest state" only when niri calls `notify_changes`. This is also when state changes are pushed to the clients.

It already seems to work mostly like this, just need to replace some uses of "newest state" with the "current state".

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            zwlr_output_manager_v1::Request::CreateConfiguration { id, serial } => {
+                let new_out = state.get_outputs().clone();
+                let g_state = state.output_management_state();
+                let conf = data_init.init(id, ());
+                if let Some(client_data) = g_state.clients.get_mut(&client.id()) {
+                    if serial != g_state.serial {
+                        conf.cancelled();
```
YaLTeR, line 145: Should this mark `conf` as inert somehow and ignore future requests to it?

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+            zwlr_output_configuration_head_v1::Request::SetCustomMode {
+                width,
+                height,
+                refresh,
+            } => {
+                let (width, height, refresh): (u16, u16, u32) =
+                    match (width.try_into(), height.try_into(), refresh.try_into()) {
+                        (Ok(width), Ok(height), Ok(refresh)) => (width, height, refresh),
+                        _ => {
+                            error!("SetCustomMode: invalid input data");
```
YaLTeR, line 410: In niri, `error` [is reserved for bugs](https://github.com/YaLTeR/niri/wiki/Developing-niri#logging-levels) in the niri code, so this should be a `warn` or a `debug`.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+                            return;
+                        }
+                    };
+                if let Some(index) = new_out.modes.iter().position(|m| {
```
YaLTeR, line 414: I guess what this request is intended to do is more like https://github.com/YaLTeR/niri/issues/333 rather than picking one of the existing modes? I suppose this logic is fine until that's implemented.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+                            position: None,
+                            mode: None,
+                            variable_refresh_rate: false,
+                        },
+                    });
+                }
+                state.apply_new_conf(new_conf);
+                conf.succeeded();
+            }
+            zwlr_output_configuration_v1::Request::Test => conf.succeeded(),
```
YaLTeR, line 322: Did you look at cosmic-comp how hard it would be to do at least some of this?

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+                let output = match conf_data.new_out.get_mut(head_name) {
+                    Some(out) => out,
+                    None => {
+                        error!("EnableHead: output configuration missing requested new head",);
+                        let _fail = data_init.init(id, conf.clone());
+                        return;
+                    }
+                };
+                if output.current_mode.is_none() {
+                    output.current_mode = output.modes.iter().position(|a| a.is_preferred);
```
YaLTeR, line 226: Shouldn't this do a similar logic to how the TTY backend picks a mode?

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+                    None => {
+                        error!("SetScale: head is disabled");
+                        return;
+                    }
+                };
+                logical.scale = scale;
+            }
+            zwlr_output_configuration_head_v1::Request::SetAdaptiveSync { state } => match state {
+                WEnum::Value(AdaptiveSyncState::Enabled) => new_out.vrr_enabled = true,
+                WEnum::Value(AdaptiveSyncState::Disabled) => new_out.vrr_enabled = true,
```
YaLTeR, line 470:
```suggestion
                WEnum::Value(AdaptiveSyncState::Disabled) => new_out.vrr_enabled = false,
```

src/protocols/output_management.rs
```diff
@@ -0,0 +1,785 @@
...
+    new_head.model(output.model.clone());
+    new_head.adaptive_sync(match output.vrr_enabled {
+        true => AdaptiveSyncState::Enabled,
+        false => AdaptiveSyncState::Disabled,
+    });
+
+    new_head.adaptive_sync(match output.vrr_enabled {
+        true => AdaptiveSyncState::Enabled,
+        false => AdaptiveSyncState::Disabled,
+    });
```
YaLTeR, line 648: Copy-paste?
```suggestion
```

YaLTeR (commented):
Thanks! Haven't looked too deep yet, but here are some initial comments.

src/protocols/output_management.rs
gmorer: That what is started doing, sadly apparently when receiving an `EnableHead` it crash if no `ZwlrOutputConfigurationHeadV1` is created via `data_init.init()` so even if the conf is outdated it has to continue handling the requests. :/

YaLTeR: Yeah, the object must always be created, but you can add a bool on it and then check if it's set and ignore further requests accordingly

src/protocols/output_management.rs
gmorer: So from what i see:
- Verify the existence and validity of every head (mode...)
- They dont allow all the user to have no enabled outputs
- check if all the outputs have surfaces

src/protocols/output_management.rs
gmorer: like the `pick_mode()` in `backend/tty.rs` ?

YaLTeR: Yeah. But I haven't looked too deep into this yet to say that it's the right thing to do here.

gmorer:
ATM, `kanshi` manage to change the configuration but it crashes with the following error message:
`not a valid new object id (4278190081), message mode(n)` after that

src/protocols/output_management.rs
gmorer: I use the serial for that, of the conf serial is not the same as the global one I ignore most of the request, check the `outdated` variables created before most of the request.

gmorer:
I just pushed some modifications that moves data from the global state to the each of the Zwlr object, i tested it.
working but still the same issue around kanshi mode, not sure if it comes from kanshi or the server side

src/protocols/output_management.rs
```diff
@@ -0,0 +1,774 @@
...
+                    }
+                };
+                if output.current_mode.is_none() {
+                    output.current_mode = output.modes.iter().position(|a| a.is_preferred);
+                }
+                let current_mode = match output.current_mode {
+                    Some(current) => current,
+                    None => {
+                        error!("EnableHead: missing output prefered mode");
+                        return;
```
YaLTeR, line 224: A Wayland new ID must always be initialized with `data_init.init()`. This code path is missing it, probably where the errors are coming from.

gmorer:
I found the issue i had with mode, it now work fine for me i'l remove the draft status.
In `notify_changes()`:
```diff
-  head.mode(&modes[new_index]);
+  head.current_mode(&modes[new_index]);
```
which make sense

src/handlers/mod.rs
```diff
@@ -544,3 +549,30 @@ delegate_xdg_activation!(State);
...
+            .find(|output| output.name() == *name)
+            .map(logical_output)
+    }
+
+    fn apply_new_conf(&mut self, conf: Vec<niri_config::Output>) {
+        self.niri.config.borrow_mut().outputs = conf;
+        self.backend.on_output_config_changed(&mut self.niri);
+    }
+}
+
```
YaLTeR, line 577:
```suggestion
```

src/protocols/output_management.rs
YaLTeR: I see you reacted with +1 and re-requested a review, but I'm not sure this change has been completed yet, i.e. I think the `get_outputs()` call should be removed from the trait, and instead `current_out` should be used.

YaLTeR (commented):
Could you also please resolve all conversations which you had addressed? Will make it easier to review.

src/protocols/output_management.rs
gmorer: So i gave the `outputs` at the initialization.

src/protocols/output_management.rs
gmorer: It check for at least one enabled output now

gmorer:
Hello,
so everything is ready expect the 2 last opened discussions, I'm not sure for the expected behavior wanted for answering to the configuration test, atm it just check that at least one output is enabled.

For the mode in https://github.com/YaLTeR/niri/pull/413#discussion_r1623164198 I cannot use the tty has it is now, since it might be running via winit.
Or i could use the way it is now, and go check the devices/surfaces and find the one with the good name. Not sure whats the best solution.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+                            false => AdaptiveSyncState::Disabled,
+                        });
+                    }
+                }
+            }
+            match (old.current_mode, new.current_mode) {
+                (Some(old_index), Some(new_index)) => {
+                    if old.modes.len() <= old_index
+                        || old.modes.len() <= new_index
+                        || new.modes.len() <= new_index
```
YaLTeR, line 682: Could you just check `old.modes == new.modes` and error!() out otherwise? I am not sure modes can change over the lifetime of an output.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+    // new_head.serial_number(output.serial);
+    client_data
+        .heads
+        .insert(output.name.clone(), (new_head, new_modes));
+}
+
+pub fn notify_changes(state: &mut OutputManagementManagerState, new_out: IpcOutputMap) {
+    let mut changed = false; /* most likely to endup true */
+    for (name, new) in new_out.iter() {
+        if let Some(old) = state.current_out.get(name) {
```
YaLTeR, line 666: Instead of indexing by name, this should really index by something more bound to the actual output like an `Output` or `WeakOutput` from Smithay. Because if, say, in a single cycle, you unplug one monitor from HDMI-A-1 and plug in a completely different monitor, it should really be advertised as a removed head + completely new head, rather than a diff. This should also make the modes code easier to deal with.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+                }
+                (None, None) => {}
+            }
+            match (old.logical, new.logical) {
+                (Some(old_logical), Some(new_logical)) => {
+                    if old_logical != new_logical {
+                        changed = true;
+                        for (_, client) in &state.clients {
+                            if let Some((head, _)) = client.heads.get(name) {
+                                if old_logical.x != new_logical.x || old_logical.y != old_logical.x
```
YaLTeR, line 733:
```suggestion
                                if old_logical.x != new_logical.x || old_logical.y != new_logical.y
```
copy paste error

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+                (None, Some(new_logical)) => {
+                    changed = true;
+                    for (_, client) in &state.clients {
+                        if let Some((head, _)) = client.heads.get(name) {
+                            head.enabled(0);
```
YaLTeR, line 753: First of all this should probably be 1, but also is this really needed here? It's already sent in the modes check above.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+                            head.transform(ipc_transform_to_smithay(new_logical.transform).into());
+                            head.scale(new_logical.scale);
+                        }
+                    }
+                }
+                (Some(_), None) => {
+                    changed = true;
+                    for (_, client) in &state.clients {
+                        if let Some((head, _)) = client.heads.get(name) {
+                            head.enabled(0);
```
YaLTeR, line 764: Same, is this needed? Already sent from the mode.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+        _manager: &ZwlrOutputManagerV1,
+        request: zwlr_output_manager_v1::Request,
+        _data: &(),
+        _display: &DisplayHandle,
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            zwlr_output_manager_v1::Request::CreateConfiguration { id, serial } => {
+                let new_out = state.output_management_state().current_out.clone();
+                let g_state = state.output_management_state();
```
YaLTeR, line 147:
```suggestion
                let g_state = state.output_management_state();
                let new_out = g_state.current_out.clone();
```

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+                    }
+                    client_data.confs.insert(conf);
+                } else {
+                    error!("CreateConfiguration: missing client data");
+                }
+            }
+            zwlr_output_manager_v1::Request::Stop => {
+                let clients = &mut state.output_management_state().clients;
+                clients.get(&client.id()).map(|c| c.manager.finished());
+                clients.remove(&client.id());
```
YaLTeR, line 164: if let Some(client) = clients.remove?

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+    D: 'static,
+    D: Dispatch<ZwlrOutputHeadV1, String>,
+    D: Dispatch<ZwlrOutputModeV1, ()>,
+    D: 'static,
+{
+    let new_head = client
+        .create_resource::<ZwlrOutputHeadV1, _, D>(
+            display,
+            client_data.manager.version(),
+            output.name.clone(),
```
YaLTeR, line 611: This should also probably be WeakOutput (see the other comment about using name as key).

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+                let Some(head_name) = head.data::<String>() else {
+                    error!("EnableHead: Missing attached output head name");
+                    let _fail = data_init.init(id, OutputConfigurationHeadState::Cancelled);
+                    return;
+                };
+                if outdated {
+                    info!("EnableHead: request from an outdated configuration");
+                    let _fail = data_init.init(id, OutputConfigurationHeadState::Cancelled);
+                    return;
+                }
```
YaLTeR, line 205: Probably should be at the top of the block, and debug!()

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+        state: &mut D,
+        client: &Client,
+        _output_head: &ZwlrOutputHeadV1,
+        request: zwlr_output_head_v1::Request,
+        data: &String,
+        _display: &DisplayHandle,
+        _data_init: &mut DataInit<'_, D>,
+    ) {
+        match request {
+            zwlr_output_head_v1::Request::Release => {
```
YaLTeR, line 501: All handling of release requests should instead be moved to destructors (another fn in Dispatch). This way, even if the client dies, it doesn't linger in the data structures. The release requests themselves can just do nothing.

YaLTeR (commented):
Again sorry for the delays, phew, here's a more in-depth round of review.

src/protocols/output_management.rs
```diff
@@ -0,0 +1,791 @@
...
+        conf: &ZwlrOutputConfigurationV1,
+        request: zwlr_output_configuration_v1::Request,
+        conf_data: &OutputConfigurationData,
+        _display: &DisplayHandle,
+        data_init: &mut DataInit<'_, D>,
+    ) {
+        let g_state = state.output_management_state();
+        let outdated = conf_data.lock().unwrap().serial != g_state.serial;
+        match request {
+            zwlr_output_configuration_v1::Request::EnableHead { id, head } => {
```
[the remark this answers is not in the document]
gmorer: `It is a protocol error to send two enable_head or disable_head requests with the same head.` should I check it ?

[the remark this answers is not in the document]
YaLTeR: Yeah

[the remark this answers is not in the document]
gmorer: Because `It is a protocol error to omit a head in a configuration.` The initial configuration is just a copy of the old one, we some head enabled. 
So i guess i guess with the actual implementation, it will succeed and the client will be able to control the same output from 2 different objects.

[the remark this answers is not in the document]
gmorer: Keeping the heads enable allows to keep the `LogicalOutput` since there us no default logic for this (https://github.com/YaLTeR/niri/pull/413#discussion_r1658994165) so it's not really possible to now if and head as already been enable. Or i can put a boolean.

gmorer:
So i used `WeakOutput` as keys for the hashmaps and as wlr object's data, it does feel better than String.

src/handlers/mod.rs
```diff
@@ -569,5 +570,18 @@ impl OutputManagementHandler for State {
...
         self.backend.on_output_config_changed(&mut self.niri);
     }
+
+    fn get_current_outputs(&self) -> OutputsConf {
+        let ipc_outputs = self.backend.ipc_outputs();
+        let ipc_outputs = ipc_outputs.lock().unwrap();
+        let mut res = HashMap::new();
+        for (name, output) in &self.niri.output_by_name {
+            let Some(conf) = ipc_outputs.get(name) else {
+                continue;
```
YaLTeR, line 580: I think this can have an error!() message too?

src/protocols/output_management.rs
```diff
@@ -635,58 +639,62 @@ fn send_new_head<D>(
...
-                        error!("notify_changes: out of bound mode");
-                    } else if old.modes[old_index] != new.modes[new_index] {
+                    if old.modes != conf.modes {
+                        error!("NotifyChange: output's old modes dosnt match new modes");
+                    } else if old.modes[old_index] != conf.modes[new_index] {
                         changed = true;
-                        if old.modes[new_index] == new.modes[new_index] {
-                            for (_, client) in &state.clients {
-                                if let Some((head, modes)) = client.heads.get(name) {
+                        if old.modes[new_index] == conf.modes[new_index] {
```
YaLTeR, line 695: This is now implied by the `old.modes == conf.modes` condition

src/protocols/output_management.rs
```diff
@@ -697,22 +705,22 @@ pub fn notify_changes(state: &mut OutputManagementManagerState, new_out: IpcOutp
...
+                    for (_, client) in &g_state.clients {
+                        if let Some((head, _)) = client.heads.get(output) {
                             head.enabled(1);
-                            if old.modes.len() <= new_index || new.modes.len() <= new_index {
+                            if old.modes.len() <= new_index || conf.modes.len() <= new_index {
                                 error!("notify_changes: out of bound mode");
-                            } else if old.modes[new_index] == new.modes[new_index] {
-                                for (_, client) in &state.clients {
-                                    if let Some((head, modes)) = client.heads.get(name) {
+                            } else if old.modes[new_index] == conf.modes[new_index] {
```
YaLTeR, line 721: Also should be a `old.modes == conf.modes` check with an error!()

src/handlers/mod.rs
gmorer: yeah i had this change locally, didn't committed it, mb

src/protocols/output_management.rs
gmorer: the whole mode check was a bit of a mess, mb.

YaLTeR:
Okay, I *think* I've got everything now.
* Added OutputId mapping so disabled outputs can be manipulated.
* Added support for modes changing size and refresh because that happens with the winit backend (or in the future with virtual monitors).
* Added errors for double EnableHead, EnableHead after Apply, etc.
* Made it use niri_config::Output as the backbone, so it is able to keep unset fields unset, while preserving everything as it was.

Now it just needs a bunch more testing to make sure everything works as expected.

YaLTeR:
Thanks!

PR #417, opened 2024-06-01
per keyboard device config

rustysec:
this PR allows keyboards to be configured by name. the following config sets my laptop's built in keyboard layout to colemak while allowing my external keyboard (already flashed with colemak) to work as expected.

```
input {
    keyboard "AT Translated Set 2 keyboard" {
        xkb {
            layout "us(colemak),us"
            options "caps:escape_shifted_capslock,lv3:menu_switch,lv3:ralt_alt"
        }
    }

    keyboard {
        xkb {
            // layout "us(colemak),us"
            // options "caps:escape_shifted_capslock,lv3:menu_switch,lv3:ralt_alt"
        }
    }
}
```

the last used configuration is cached between calls to `on_keyboard()` in an attempt to squeeze some performance gains. based on the feedback on this implementation, the same can be quickly done for all other supported input types.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -269,27 +269,7 @@ impl State {
     }
 
     fn on_keyboard<I: InputBackend>(&mut self, event: I::KeyboardKeyEvent) {
-        {
-            let device = event.device();
-
-            let keyboard_config = self
-                .niri
-                .config
-                .borrow()
-                .input
-                .keyboard_named(device.name());
-
-            let keyboard = self.niri.seat.get_keyboard().unwrap();
-
-            if let Err(err) = keyboard.set_xkb_config(self, keyboard_config.xkb.to_xkb_config()) {
-                warn!("error updating xkb config: {err:?}");
-            }
-
-            keyboard.change_repeat_info(
-                keyboard_config.repeat_rate.into(),
-                keyboard_config.repeat_delay.into(),
-            );
-        }
+        self.apply_keyboard_config::<I>(&event);
 
         let comp_mod = self.backend.mod_key();
 
@@ -339,6 +319,32 @@ impl State {
         }
 
         self.handle_bind(bind);
+    }
+
+    fn apply_keyboard_config<I: InputBackend>(&mut self, event: &I::KeyboardKeyEvent) {
+        let device = event.device();
+
+        let keyboard_config = self
+            .niri
+            .config
+            .borrow()
+            .input
+            .keyboard_named(device.name());
+
+        if keyboard_config != self.niri.current_keyboard {
+            let keyboard = self.niri.seat.get_keyboard().unwrap();
+
+            if let Err(err) = keyboard.set_xkb_config(self, keyboard_config.xkb.to_xkb_config()) {
+                warn!("error updating xkb config: {err:?}");
+            }
+
+            keyboard.change_repeat_info(
+                keyboard_config.repeat_rate.into(),
+                keyboard_config.repeat_delay.into(),
+            );
+
+            self.niri.current_keyboard = keyboard_config;
+        }
     }
 
     pub fn handle_bind(&mut self, bind: Bind) {
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -271,19 +271,24 @@ impl State {
     fn on_keyboard<I: InputBackend>(&mut self, event: I::KeyboardKeyEvent) {
         {
             let device = event.device();
-            let xkb = self
+
+            let keyboard_config = self
                 .niri
                 .config
                 .borrow()
                 .input
-                .keyboard_named(device.name())
-                .xkb;
+                .keyboard_named(device.name());
 
             let keyboard = self.niri.seat.get_keyboard().unwrap();
 
-            if let Err(err) = keyboard.set_xkb_config(self, xkb.to_xkb_config()) {
+            if let Err(err) = keyboard.set_xkb_config(self, keyboard_config.xkb.to_xkb_config()) {
                 warn!("error updating xkb config: {err:?}");
             }
+
+            keyboard.change_repeat_info(
+                keyboard_config.repeat_rate.into(),
+                keyboard_config.repeat_delay.into(),
+            );
         }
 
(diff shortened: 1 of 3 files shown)


rustysec:
whoops! fixing the errors.

src/input/mod.rs
```diff
@@ -305,6 +307,32 @@ impl State {
...
 
+    fn apply_keyboard_config<I: InputBackend>(&mut self, event: &I::KeyboardKeyEvent) {
+        let device = event.device();
+
+        let keyboard_config = self
+            .niri
+            .config
+            .borrow()
+            .input
+            .keyboard_named(device.name());
```
YaLTeR, line 318: This `device.name()` allocates a new `String` every time, not a big fan of that... Maybe we should have a cache `HashMap<Device, niri_config::Keyboard>`?

src/input/mod.rs
```diff
@@ -305,6 +307,32 @@ impl State {
...
+
+        if keyboard_config != self.niri.current_keyboard {
+            let keyboard = self.niri.seat.get_keyboard().unwrap();
+
+            if let Err(err) = keyboard.set_xkb_config(self, keyboard_config.xkb.to_xkb_config()) {
```
YaLTeR, line 323: Would it be worth to check the xkb config and the repeat info for equality individually too?

src/niri.rs
```diff
@@ -269,6 +269,8 @@ pub struct Niri {
     // Casts are dropped before PipeWire to prevent a double-free (yay).
     pub casts: Vec<Cast>,
     pub pipewire: Option<PipeWire>,
+
+    pub current_keyboard: niri_config::Keyboard,
```
YaLTeR, line 273: Put this above near devices/tablets/touch.

src/niri.rs
```diff
@@ -806,7 +808,15 @@ impl State {
                 }
             }
 
-            if self.niri.config.borrow().input.keyboard.track_layout == TrackLayout::Window {
+            if self
+                .niri
+                .config
+                .borrow()
+                .input
+                .default_keyboard()
```
YaLTeR, line 816: I guess we should track this per-keyboard? Then should also be careful which layout to switch to in `on_keyboard()` when switching keyboards.

src/niri.rs
```diff
@@ -906,19 +916,23 @@ impl State {
...
         }
 
+        let default_keyboard = config.input.default_keyboard();
+
+        let old_keyboard = old_config.input.default_keyboard();
+
         // We need &mut self to reload the xkb config, so just store it here.
-        if config.input.keyboard.xkb != old_config.input.keyboard.xkb {
-            reload_xkb = Some(config.input.keyboard.xkb.clone());
+        if default_keyboard.xkb != old_keyboard.xkb {
```
YaLTeR, line 924: This should check against the current keyboard I guess?

src/input/mod.rs
rustysec: i don't think it's currently possible since `Device` supers `PartialEq` which can't be a trait object. please correct me if i'm wrong.

rustysec: ahh can use a `HashMap<input::Device, niri_config::Keyboard>` and just downcast.

src/input/mod.rs
rustysec: it now checks the key repeat settings and xkb independently and only applies what needs changing.

src/niri.rs
rustysec: i have no idea what i was doing here, sorting that out

src/niri.rs
rustysec: hrm not sure i have thought this through completely :thinking:

rustysec: > Then should also be careful which layout to switch to in `on_keyboard()` when switching keyboards.

meaning we should only switch to `Global` layouts in `on_keyboard()`?

YaLTeR: If you type with a track_layout = window keyboard, then type with a second keyboard, then switch windows, then type with the first keyboard, it should pick the correct layout for the first keyboard taking the window switch into account.

rustysec: new changes address the `on_keyboard()` handling, and it seems to work in my tests. i might be missing something, but i'm not currently tracking any state per-keyboard.

src/niri.rs
```diff
@@ -912,19 +914,23 @@ impl State {
...
+            || default_keyboard.repeat_delay != current_keyboard.repeat_delay
         {
             let keyboard = self.niri.seat.get_keyboard().unwrap();
             keyboard.change_repeat_info(
-                config.input.keyboard.repeat_rate.into(),
-                config.input.keyboard.repeat_delay.into(),
+                default_keyboard.repeat_rate.into(),
+                default_keyboard.repeat_delay.into(),
             );
         }
```
YaLTeR, line 935: This should also drop the keyboard cache and set `self.niri.current_keyboard` to `default_keyboard`. Although, ideally, only do all of this if anything in the keyboard configs changed (`config.input.keyboards != old_config.input.keyboards`).

src/niri.rs
```diff
@@ -1502,7 +1508,9 @@ impl Niri {
             root_surface: HashMap::new(),
             monitors_active: true,
 
+            current_keyboard: Default::default(),
```
YaLTeR, line 1511: This should be the fallback keyboard

src/niri.rs
```diff
@@ -812,7 +814,7 @@ impl State {
                 }
             }
 
-            if self.niri.config.borrow().input.keyboard.track_layout == TrackLayout::Window {
+            if self.niri.current_keyboard.track_layout == TrackLayout::Window {
```
YaLTeR, line 817: The way I see it is that instead of one `KeyboardLayout` per surface, we should store a `KeyboardLayout` for every keyboard per surface. Then, here we will update all of them at once for those keyboards which have `track_layout == Window`, and then in `apply_keyboard_config()` we will pick the right layout for the current surface.

And also this should somehow be cleared upon a change to the keyboard config.

YaLTeR:
FYI I was planning to finish this at some point (after I'm done with the interactive move PR), if you don't mind.

PR #428, opened 2024-06-07
PointerMotionAbsolute: use union rect of all outputs

galister:
this makes niri behave more like other desktops that i've tested with absolute input (xorg, kde, gnome, wlroots).

tested screenshots as well, they don't seem to be affected and work on either screen via the absolute pointer.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -23,7 +23,7 @@ use smithay::input::pointer::{
     GrabStartData as PointerGrabStartData, MotionEvent, RelativeMotionEvent,
 };
 use smithay::input::touch::{DownEvent, MotionEvent as TouchMotionEvent, UpEvent};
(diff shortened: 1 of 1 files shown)


YaLTeR:
Is it intended that this can cause the pointer to end up outside all outputs? I.e. when you have a smaller output beside a larger one, or just not adjacent. Or should it clamp to the closest output kinda like what happens in the regular pointer motion function?

galister:
as far as i understand, the point of absolute input is that you can take screen pos + mouse pos relative to that screen and the cursor will show up right where you think it'll show up. there's no way to do that without potential dead areas. also, this seems to be the norm across a majority of desktops.

YaLTeR:
So on other desktops the cursor also just goes out of bounds and is not clamped?

YaLTeR:
Alright, I think that's what happens with this PR too.

YaLTeR:
Thanks

PR #440, opened 2024-06-14
Handle KDE decorations in Mapped::has_ssd

kchibisov:
This fixes an issue with default CSD border being drawn for SSD rendering firefox, because only xdg decorations were checked.


YaLTeR:
Thanks. At a glance seems to work for Firefox with "Title Bar" enabled, and for GTK 3 apps. Haven't noticed anything breaking yet.

PR #453, opened 2024-06-20
Event stream IPC

YaLTeR:
Some basic scaffolding for an event stream IPC.

1. Changed the client to read only a single line worth of response.

    This mirrors how the server works, and will be necessary to tell apart further messages. Events will arrive at one per line.
2. Added `Request::EventStream` that replies with `Response::Handled` and then proceeds to write one `Event` per line as they occur.
3. Events will be designed in such a way that the initial burst will synchronize the current state to the client, then subsequent events will modify that state in a consistent way. E.g. the initial burst of events will create all current outputs and workspaces.
4. Compositor-side, event stream clients have a buffer of events for sending. If this buffer runs too large (the client hanged and isn't reading), this client will be dropped. This is kinda similar to what Wayland itself does.

The events are all subject to change until this PR is merged. Currently only the WindowFocused is actually implemented, and I'm planning to change its type too.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1233,6 +1233,7 @@ impl From<niri_ipc::Action> for Action {
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum WorkspaceReference {
+    Id(u64),
     Index(u8),
     Name(String),
 }
@@ -1240,6 +1241,7 @@ pub enum WorkspaceReference {
 impl From<WorkspaceReferenceArg> for WorkspaceReference {
     fn from(reference: WorkspaceReferenceArg) -> WorkspaceReference {
         match reference {
+            WorkspaceReferenceArg::Id(id) => Self::Id(id),
             WorkspaceReferenceArg::Index(i) => Self::Index(i),
             WorkspaceReferenceArg::Name(n) => Self::Name(n),
         }
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -308,10 +308,12 @@ pub enum SizeChange {
     AdjustProportion(f64),
 }
 
-/// Workspace reference (index or name) to operate on.
+/// Workspace reference (id, index or name) to operate on.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
 pub enum WorkspaceReferenceArg {
+    /// Id of the workspace.
+    Id(u64),
     /// Index of the workspace.
     Index(u8),
     /// Name of the workspace.
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -21,6 +21,8 @@ pub enum Request {
     Outputs,
     /// Request information about workspaces.
     Workspaces,
+    /// Request information about open windows.
+    Windows,
     /// Request information about the configured keyboard layouts.
     KeyboardLayouts,
     /// Request information about the focused output.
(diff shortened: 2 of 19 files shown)


SidharthArya:
Hi, when is this planned to be merged? 
Can pick up other tasks (except WindowFocused) if needed ?
Would love to see WorkspaceFocused, MonitorFocused, WorkspaceAdded and MonitorAdded.

YaLTeR:
No plan yet, this is just scaffolding for now. It's in a too early of a state to get help from others on specific tasks.

jakesarjeant:
> No plan yet, this is just scaffolding for now. It's in a too early of a state to get help from others on specific tasks.

Has that changed yet? I'd be happy to pick up some work too if needed

YaLTeR:
I was fairly busy with uni; I think I am about to have sorted most of it out though so I will look into this more soon.

YaLTeR:
Update, I did some more sketching out. Now, the following events are implemented: `WorkspacesChanged`, `WorkspaceSwitched`, `KeyboardLayoutsChanged`, `KeyboardLayoutSwitched`.

I was able to make a Waybar workspace indicator with the following Python script:
```python
#!/usr/bin/python3

import json
import os
import sys
from socket import AF_UNIX, SHUT_WR, socket

niri_socket = socket(AF_UNIX)
niri_socket.connect(os.environ["NIRI_SOCKET"])
file = niri_socket.makefile("rw")

file.write('"EventStream"')
file.flush()
niri_socket.shutdown(SHUT_WR)

output = None
if len(sys.argv) > 1:
    output = sys.argv[1]

workspaces = {}


def print_workspaces():
    vals = workspaces.values()
    if output is not None:
        vals = filter(lambda ws: ws["output"] == output, vals)

    vals = sorted(vals, key=lambda ws: ws["idx"])
    text = "\n".join(("󰪥" if ws["is_active"] else "󰄰" for ws in vals))
    print(json.dumps({"text": text}), flush=True)


for line in file:
    event = json.loads(line)

    if changed := event.get("WorkspacesChanged"):
        workspaces = {ws["id"]: ws for ws in changed["workspaces"]}
        print_workspaces()
    elif switched := event.get("WorkspaceSwitched"):
        for ws in workspaces.values():
            if ws["output"] == switched["output"]:
                ws["is_active"] = ws["id"] == switched["id"]
        print_workspaces()
```

(Note the `\n`: my bar is vertical, so it uses newlines.)

And this is the custom module setup:
```json
    "custom/workspaces": {
        "exec": "~/source/py/niri-ipc/workspaces.py \"$WAYBAR_OUTPUT_NAME\"",
        "return-type": "json"
    },
```

image

YaLTeR:
I started implementing built-in support in Waybar using this event stream draft PR. You can try my fork here: https://github.com/YaLTeR/Waybar/tree/niri

https://github.com/user-attachments/assets/a7a0591f-0ee3-4944-825e-3ded54b84989

So far I added `niri/workspaces` and `niri/language` with configuration roughly matching some subset of the same from `sway/` and `hyprland/`.

YaLTeR:
Pushed another breaking change, along with an implementation of Window events:
- WorkspaceActiveWindowChanged
- WindowsChanged
- WindowOpenedOrChanged
- WindowClosed
- WindowFocusChanged

Waybar fork updated to use this, and also I implemented a `niri/window` module similar to `hyprland/window`.

This completes my sketching so far; I'm fairly sure that this is enough IPC to cover the remaining window icon support in the workspaces Waybar module (I'm too lazy to implement that), and anything past that should be fairly straightforward to add without breaking changes.

That said, I'd like others to take a look at this before merging, give it a try with Waybar or by implementing their own scripts, in case I missed something important.

PR #456, opened 2024-06-20
feat: add `focus-column-or-monitor-left`, `focus-column-or-monitor-right`

lpnh:
Partially addresses: https://github.com/YaLTeR/niri/discussions/296

This is a preliminary sketch for implementing `focus-column-or-monitor-left` and `focus-column-or-monitor-right`, which expand the default left and and right focus behaviors. If a window is available in the target direction, it receives focus; otherwise, the focus shifts to the corresponding monitor.

Current state: it kinda works, but there are cases that still need to be fixed. I will try to provide a better description soon.

To be completely honest, I'm new to the Niri codebase, so there's a lot I'm still unsure about. I was hoping to provide a good starting point for this PR, but it seems like I still have a lot to learn.

Any feedback would be greatly appreciated.


lpnh:
Well, it seems both `focus-window-or-monitor-left` and `focus-window-or-monitor-right` are now working as expected.

A couple of things about the implementation:

`maybe_focus_window_or_monitor_*`
  - I'm not entirely sure if using a boolean here is the most idiomatic approach, but it seems like a reasonable solution to avoid adding further complexity.
  - The "maybe" prefix try to suggest that this function depends on the availability of the output and that it's not solely responsible for handling the focus.

`FocusWindowOrMonitor*`
  - Given the need to manage the output and handle three different possible cases, the structure ended up quite clear.
  - The main challenge seems to be handling both focus and cursor at the same time.

Now, about the test... I'm not quite sure how to write it properly. I attempted to provide some, but I am unsure if they make any sense.

This was very fun, so I wouldn't mind rewriting anything you think could be improved. Also, feel free to take it from here.

src/layout/mod.rs
```diff
@@ -1320,6 +1320,41 @@ impl<W: LayoutElement> Layout<W> {
         monitor.focus_window_or_workspace_up();
     }
 
+    pub fn maybe_focus_window_or_monitor_left(&mut self, output: &Output) -> bool {
+        let Some(monitor) = self.active_monitor() else {
+            return false;
```
YaLTeR, line 1325: I guess what we care about is whether we focused an output or not. So this should `return true;`. Alternatively, I suggest flipping the return value, so this will return `false`, focus_left will return `false` and only focus_output will return `true`.

src/layout/mod.rs
```diff
@@ -2705,6 +2740,8 @@ mod tests {
         FocusWindowUpOrColumnRight,
         FocusWindowOrWorkspaceDown,
         FocusWindowOrWorkspaceUp,
+        FocusWindowOrMonitorLeft(#[proptest(strategy = "0..=1u8")] u8), // no idea still
+        FocusWindowOrMonitorRight(#[proptest(strategy = "0..=1u8")] u8), // no idea still
```
YaLTeR, line 2744:
```suggestion
        FocusWindowOrMonitorLeft(#[proptest(strategy = "1..=2u8")] u8),
        FocusWindowOrMonitorRight(#[proptest(strategy = "1..=2u8")] u8),
```

YaLTeR (changes requested):
Sorry for taking so long to review, just some minor changes needed.

> I'm not entirely sure if using a boolean here is the most idiomatic approach

It's alright.

> Now, about the test... I'm not quite sure how to write it properly. I attempted to provide some, but I am unsure if they make any sense.

Seems correct. Just random index should be from 1 because the output random indices are 1 to 5.

YaLTeR (approved):
Looks good.

YaLTeR (commented):
Wait, I just realized. Could you rename Window to Column throughout? Left-Right focus goes by columns, not by windows, and other actions are also focus-column-left, etc.

lpnh:
> Wait, I just realized. Could you rename Window to Column throughout? Left-Right focus goes by columns, not by windows, and other actions are also focus-column-left, etc.

Sure! I had the same thought initially. We might have been influenced by the initial request. Thanks for catching this.

lpnh:
May I move them next to the other Column variables?

YaLTeR:
Yeah

PR #462, opened 2024-06-22
Update winit to fix failing build on arm linux

itsjunetime:
Version 0.30.2 of winit fails to build on arm linux, and they just released 0.30.3 to fix that (https://github.com/rust-windowing/winit/releases/tag/v0.30.3), so this just updates that (and also downgrades proc-macro-crate, I guess? I just ran `cargo update winit` so I guess that's a needed side-effect) to make it build again on arm linux.


YaLTeR:
Thanks. 

> and also downgrades proc-macro-crate, I guess?

something's causing this version to go back and forth every time, not sure what exactly, doesn't seem to affect anything.

PR #475, opened 2024-06-28
Fix dead links and add FAQ entry

aspizu:


wiki/Important-Software.md
```diff
@@ -8,13 +8,13 @@ Many apps need one. For example, [mako](https://github.com/emersion/mako) works
 
 These provide a cross-desktop API for apps to use for various things like file pickers or UI settings. Flatpak apps in particular require working portals.
 
-Portals **require** [running niri as a session](https://github.com/YaLTeR/niri#session), which means through the `niri-session` script or from a display manager. You will want the following portals installed:
+Portals **require** [running niri as a session](./Getting-Started.md#installation), which means through the `niri-session` script or from a display manager. You will want the following portals installed:
```
YaLTeR, line 11: We don't have a good anchor there to link to that sentence in the new place. Since it's at the top level, let's just link the wiki page itself.
```suggestion
Portals **require** [running niri as a session](./Getting-Started.md), which means through the `niri-session` script or from a display manager. You will want the following portals installed:
```

wiki/FAQ.md
```diff
@@ -1,3 +1,10 @@
+### How to disable Client-Side Decorations?
+
+See below.
+
+(Note that this will disable resizing windows using mouse.)
+
```
YaLTeR, line 6:
```suggestion
### How to disable client-side decorations/make windows rectangular?

Uncomment the `prefer-no-csd` setting at the [top level](./Configuration:-Miscellaneous.md) of the config.
Then niri will ask windows to omit client-side decorations, and also inform them that they are being tiled (which makes some windows rectangular, even if they cannot omit the decorations).

Note that currently this will prevent edge window resize handles from showing up.
You can still resize windows by holding <kbd>Mod</kbd> and the right mouse button.
```

YaLTeR (changes requested):

YaLTeR:
Thanks

Issue #345, opened 2024-05-15
Implement key repeat for compositor bindings


YaLTeR:
### Discussed in https://github.com/YaLTeR/niri/discussions/344

YaLTeR:
Apparently sway does key repeat for all bindings; GNOME does key repeat at least for custom spawn bindings. So it's probably worth to do key repeat for everything by default and have a flag to disable it?

SalmanFarooqShiekh:
I think no key-repeat makes sense more of the times than key-repeat does. What are the bindings which one would preferably want to have repeat enabled on? This is without thinking much but I can only think of volume and brightness relative adjustments. Most other binds are for stuff like launch an app, run a command, etc. which I think don't need repeats.

Really, I don't have any preference for the default of binding key-repeat enabled/disabled. Either way will be good.

And, thank you so much for the awesome compositor. I have been dailying it since this weekend for backend and frontend web-development and it is working really nicely especially with multi-monitors. Touchpad gestures and the animations are really nice and it is obvious considerable work went into properly tuning them.

SalmanFarooqShiekh:
I guess some people would like to have Mod+H/J/K/L etc. (in the default config) to repeat too when held.

YaLTeR:
I think it's pretty standard to have repeat for changing focus and stuff (which is a lot of the commands), then for spawning GNOME seems to also do it, so why not.

YaLTeR:
This is now implemented.

PR #487, opened 2024-06-30
feat: key repeat for binds

SalmanFarooqShiekh:
Addresses https://github.com/YaLTeR/niri/issues/345

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -294,6 +294,19 @@ impl State {
         let time = Event::time_msec(&event);
         let pressed = event.state() == KeyState::Pressed;
 
+        // Stop bind key repeat on any release. This won't work 100% correctly in cases like:
+        // 1. Press Mod
+        // 2. Press Left (repeat starts)
+        // 3. Press PgDown (new repeat starts)
+        // 4. Release Left (PgDown repeat stops)
+        // But it's good enough for now.
+        // FIXME: handle this properly.
+        if !pressed {
+            if let Some(token) = self.niri.bind_repeat_timer.take() {
+                self.niri.event_loop.remove(token);
+            }
+        }
+
         let Some(Some(bind)) = self.niri.seat.get_keyboard().unwrap().input(
             self,
             event.key_code(),
@@ -330,12 +343,44 @@ impl State {
             return;
         };
 
-        // Filter actions when the key is released or the session is locked.
         if !pressed {
             return;
         }
 
-        self.handle_bind(bind);
+        self.handle_bind(bind.clone());
+
+        // Start the key repeat timer if necessary.
+        if !bind.repeat {
+            return;
+        }
+
+        // Stop the previous key repeat if any.
+        if let Some(token) = self.niri.bind_repeat_timer.take() {
+            self.niri.event_loop.remove(token);
+        }
+
+        let config = self.niri.config.borrow();
+        let config = &config.input.keyboard;
+
+        let repeat_rate = config.repeat_rate;
+        if repeat_rate == 0 {
+            return;
+        }
+        let repeat_duration = Duration::from_secs_f64(1. / f64::from(repeat_rate));
+
+        let repeat_timer =
+            Timer::from_duration(Duration::from_millis(u64::from(config.repeat_delay)));
+
+        let token = self
+            .niri
+            .event_loop
+            .insert_source(repeat_timer, move |_, _, state| {
+                state.handle_bind(bind.clone());
+                TimeoutAction::ToDuration(repeat_duration)
+            })
+            .unwrap();
+
+        self.niri.bind_repeat_timer = Some(token);
     }
 
     pub fn handle_bind(&mut self, bind: Bind) {
@@ -2132,6 +2177,7 @@ fn should_intercept_key(
                         modifiers: Modifiers::empty(),
                     },
                     action,
+                    repeat: true,
                     cooldown: None,
                     allow_when_locked: false,
                 });
@@ -2181,6 +2227,7 @@ fn find_bind(
                 modifiers: Modifiers::empty(),
             },
             action,
+            repeat: true,
             cooldown: None,
             allow_when_locked: false,
         });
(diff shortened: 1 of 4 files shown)


SalmanFarooqShiekh:
```
env RUN_SLOW_TESTS=1 PROPTEST_CASES=200000 PROPTEST_MAX_GLOBAL_REJECTS=200000 RUST_BACKTRACE=1 cargo test --release --all
```

This shows all tests passed when run locally.

YaLTeR:
Are you sure you're not skipping any tests when running locally by accident?

SalmanFarooqShiekh:
Sorry, I made some mistake, I deleted that comment. The same test on CI fails locally too. Working on it.

SalmanFarooqShiekh:
Maybe we should have a hard coded upper limit on repeat rate for this or maybe separate config?

src/input/mod.rs
```diff
@@ -2578,7 +2641,13 @@ mod tests {
...
+        assert!(matches!(
+            filter,
+            FilterResult::Intercept(Some(Bind {
+                action: Action::CloseWindow,
+                ..
+            }))
+        ));
 
         let filter = none_key_event(&mut suppressed_keys, mods, false);
         assert!(matches!(filter, FilterResult::Forward));
```
YaLTeR, line 2653: I can't comment on the lines below, but you can see in the test right below this line that if you press Mod then Q then release Mod, you won't get a release event with the bind, so the key repeat will go on forever.

src/input/mod.rs
```diff
@@ -330,12 +332,63 @@ impl State {
             return;
         };
 
-        // Filter actions when the key is released or the session is locked.
-        if !pressed {
-            return;
-        }
+        let config = self.niri.config.clone();
```
YaLTeR, line 335: You don't need this, it's accessible from `state.niri.config`.

src/input/mod.rs
```diff
@@ -330,12 +332,63 @@ impl State {
...
+                Entry::Occupied(_) => {
+                    // already repeating, nothing to do
+                }
+            }
+        } else {
+            let value = self.niri.bind_repeat_timers.remove(&bind.key);
+
+            if value.is_some() {
+                self.niri.event_loop.remove(value.unwrap())
+            }
```
YaLTeR, line 390:
```suggestion
            if let Some(value) = self.niri.bind_repeat_timers.remove(&bind.key) {
                self.niri.event_loop.remove(value)
            }
```

src/input/mod.rs
```diff
@@ -330,12 +332,63 @@ impl State {
...
+                    let repeat_delay_timer = Timer::from_duration(Duration::from_millis(
+                        config.borrow().input.keyboard.repeat_delay as u64,
+                    ));
+                    let token = self
+                        .niri
+                        .event_loop
+                        .insert_source(repeat_delay_timer, move |_, _, state| {
+                            state.handle_bind(bind.clone());
+                            repeat_handler(state, config.clone(), bind.clone());
+                            TimeoutAction::Drop
```
YaLTeR, line 375: Instead of re-creating the timer every time, you can return `TimeoutAction::Duration(repeat delay)` and it will re-fire for you. So the whole `repeat_handler()` is no longer needed.

YaLTeR (changes requested):

SalmanFarooqShiekh:
Thank you for the feedback, I will get on to it soon.

src/input/mod.rs
SalmanFarooqShiekhEsketcher: resolved

src/input/mod.rs
```diff
@@ -2596,7 +2636,7 @@ mod tests {
 
         mods = Default::default();
         let filter = close_key_event(&mut suppressed_keys, mods, false);
-        assert!(matches!(filter, FilterResult::Intercept(None)));
+        assert!(matches!(filter, FilterResult::Forward));
```
YaLTeR, line 2639: This must not change. This is what prevents the client from getting the release of a compositor key, for which it did not see the press.

src/input/mod.rs
```diff
@@ -327,15 +327,51 @@ impl State {
...
+                    // already repeating, nothing to do
+                }
+            }
+        } else {
+            self.remove_all_bind_repeat_timers();
         }
+    }
 
-        self.handle_bind(bind);
+    fn remove_all_bind_repeat_timers(&mut self) {
```
YaLTeR, line 367: At this point instead of a HashMap, might as well just have a single Option<(Key, Timer)> field.

src/input/mod.rs
```diff
@@ -327,15 +327,51 @@ impl State {
...
+                    ));
+
+                    let token = self
+                        .niri
+                        .event_loop
+                        .insert_source(repeat_delay_timer, move |_, _, state| {
+                            state.handle_bind(bind.clone());
+
+                            TimeoutAction::ToDuration(Duration::from_millis(max(
+                                1000 / state.niri.config.borrow().input.keyboard.repeat_rate as u64,
```
YaLTeR, line 350: Compute this TimeoutAction once, before inserting the timer, then just use it. Also instead of a max() on the 1000 / rate, instead check `if rate == 0 { TimeoutAction::Drop } else { TimeoutAction::ToDuration(...) }`.

YaLTeR (changes requested):

src/input/mod.rs
YaLTeR: So, I don't think there should be any hardcoded limit here. E.g. in GNOME spawn repeats just like anything else, and it's useful if you want to stress-test the compositor with a bunch of terminals. Instead, what I was thinking, is to add a property to binds, something like `repeat=false` (same place as the existing `cooldown-ms=150`) that will exclude it from key repeat entirely.

YaLTeR (approved):
Good job adding the repeat property. I adjusted the code a bit (turns out, duplicate repeat is possible, but it's not a big deal to handle it) and added a wiki mention. Let's merge this and see if any issues come up.

As for performance, in my testing only repeat spawning seemed laggy, but it didn't seem to be any specific in the compositor (spawning in general takes a while, and there's some competition for CPU threads when it happens). Other actions like focus change or consume-to-column are smooth.

YaLTeR:
Thanks

ocharles:
Nice work everyone - been really missing this for changing laptop brightness! :partying_face:

PR #488, opened 2024-06-30
fix monitor reactivation for broken monitor firmware

it-a-me:
My monitor has a bug that causes it to be briefly re-detected as a new monitor while turning off. This allows power-off-monitors to work as intended. Powering on monitors causes a redraw so I don't believe this *should* cause any issues, but I'm open to advice about other ways I could attempt to fix this.


YaLTeR:
Could you check if GNOME Shell keeps monitors powered off for you or also powers them back on right away?

it-a-me:
It looks line gnome powers them right back on. I'm not all that familiar with it, but when I run ```busctl --user set-property org.gnome.Mutter.DisplayConfig /org/
gnome/Mutter/DisplayConfig org.gnome.Mutter.DisplayConfig PowerSaveMode i 1``` gnome exhibits the same behavior of pre-patch niri where the display starts to turn off, then flickers back on.

src/backend/tty.rs
```diff
@@ -970,11 +970,14 @@ impl Tty {
...
-        // Power on all monitors if necessary and queue a redraw on the new one.
-        niri.event_loop.insert_idle(move |state| {
-            state.niri.activate_monitors(&mut state.backend);
-            state.niri.queue_redraw(&output);
-        });
+        // redraw is always performed when when monitors are reactivated
+        if niri.monitors_active {
+            // Power on all monitors if necessary and queue a redraw on the new one.
+            niri.event_loop.insert_idle(move |state| {
+                state.niri.activate_monitors(&mut state.backend);
```
YaLTeR, line 977:
```suggestion
        // Some buggy monitors replug upon powering off,
        // so powering on here would prevent such monitors from powering off.
        if niri.monitors_active {
            // Power on all monitors if necessary and queue a redraw on the new one.
            niri.event_loop.insert_idle(move |state| {
```

src/backend/tty.rs
```diff
@@ -970,11 +970,14 @@ impl Tty {
...
+                state.niri.queue_redraw(&output);
+            });
+        }
```
YaLTeR, line 980: Something like this I guess
```suggestion
        } else {
                set_crtc_active(&device.drm, *crtc, false);
                if let Err(err) = surface.compositor.reset_state() {
                    warn!("error resetting surface state: {err:?}");
                }
        }
```

YaLTeR (commented):
Well, I suppose we can keep monitors disabled, but that means we need to disable the new crtc too

YaLTeR:
Upon closer inspection, I don't think the reset call is needed, so I removed it. Seems to work for me; do you mind testing once more? Then we can merge.

it-a-me:
The new version works great.  Thanks for the help

YaLTeR:
Thanks!

edwrap:
Thanks for this, solves the issue for my HP Omen monitor as well!

PR #489, opened 2024-06-30
Add install location instructions for manual installation

NickHastings:
Hi just giving niri a try. Looks very nice. I noticed that the install instructions suggest using paths that are usually reserved for system packages not software installed manually by the sysadmin. This PR updates the wiki with what I think are more appropriate install paths.

### DIFF
--- a/wiki/Getting-Started.md
+++ b/wiki/Getting-Started.md
@@ -164,21 +164,38 @@ If you're not on NixOS, you may need [NixGL](https://github.com/nix-community/ni
 nix run --impure github:guibou/nixGL -- ./results/bin/niri
 ```
 
-### Installation
+### Packaging
 
-The recommended way to install and run niri is as a standalone desktop session.
-To do that, put files into the correct directories according to this table.
+The recommended way to package niri  is to have niri run as a standalone desktop
+session. To do that, put files into the correct directories according to this table.
 
-| File | Package | Manual |
-| ---- | ----------- | ----------- |
-| `target/release/niri` | `/usr/bin/` | `/usr/local/bin/` |
-| `resources/niri-session` | `/usr/bin/` | `/usr/local/bin/` |
-| `resources/niri.desktop` | `/usr/share/wayland-sessions/` |`/usr/local/share/wayland-sessions/` |
-| `resources/niri-portals.conf` | `/usr/share/xdg-desktop-portal/` |`/usr/local/share/xdg-desktop-portal/` |
+| File | Destination |
+| ---- | ----------- |
+| `target/release/niri` | `/usr/bin/` | 
+| `resources/niri-session` | `/usr/bin/` |
+| `resources/niri.desktop` | `/usr/share/wayland-sessions/` |
+| `resources/niri-portals.conf` | `/usr/share/xdg-desktop-portal/` |
+| `resources/niri.service` | `/usr/lib/systemd/user/` |
+| `resources/niri-shutdown.target` | `/usr/lib/systemd/user/` |
+
+Doing this will make niri appear in GDM and other display managers.
+
+### Manual Installation
+
+If installing directly without a package, the recommended file destinations area
+slightly different. In this case, put the files in the directories indicated in 
+the table below.
+
+| File | Destination |
+| ---- | ----------- |
+| `target/release/niri` | `/usr/local/bin/` |
+| `resources/niri-session` | `/usr/local/bin/` |
+| `resources/niri.desktop`  | `/usr/local/share/wayland-sessions/` |
+| `resources/niri-portals.conf` | `/usr/local/share/xdg-desktop-portal/` |
 | `resources/niri.service` | `/usr/lib/systemd/user/` |`/etc/systemd/user/` |
 | `resources/niri-shutdown.target` | `/usr/lib/systemd/user/` |`/etc/systemd/user/` |
 
-Doing this will make niri appear in GDM and other display managers.
+
 
 [Alacritty]: https://github.com/alacritty/alacritty
 [fuzzel]: https://codeberg.org/dnkl/fuzzel
--- a/wiki/Getting-Started.md
+++ b/wiki/Getting-Started.md
@@ -182,9 +182,9 @@ Doing this will make niri appear in GDM and other display managers.
 
 ### Manual Installation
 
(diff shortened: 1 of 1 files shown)


NickHastings:
I just noticed that `niri.service` specifies the full path as `ExecStart=/usr/bin/niri`. This should probably be `ExectStart=niri`.

YaLTeR:
Hey, the intention for these instructions is to be used when making a distro package. I don't really want to recommend installing without a package because it can lead to all sorts of problems.

NickHastings:
I see. What sort of problems are you anticipating? Would you be open to the idea of having two sets of instructions? One for packaging and one for non-packed installation?

YaLTeR:
> What sort of problems are you anticipating?

Installing software without a package is generally not recommended. It's very hard to clean up afterwards since the files aren't tracked by anything.

> Would you be open to the idea of having two sets of instructions?

Not sure that's necessary; if you want to install without a package you likely already know what paths to change. Maybe we could add a sentence there that these paths are for packages, and otherwise need to be adjusted.

NickHastings:
> > What sort of problems are you anticipating?
> 
> Installing software without a package is generally not recommended. It's very hard to clean up afterwards since the files aren't tracked by anything.

Ahh, I thought you were referring to something specific to niri. For this reason many projects ship build systems with install and uninstall targets. Alternatively providing clear instructions on the correct place to install things also effectively provides the required information on how to uninstall them. 

> > Would you be open to the idea of having two sets of instructions?
> 
> Not sure that's necessary; if you want to install without a package you likely already know what paths to change. Maybe we could add a sentence there that these paths are for packages, and otherwise need to be adjusted.

I'd expect that there would be more people wanting to try out a nice new wayland compositor manually than there are distro package managers. I also expect such people would be less likely to know where to install the components than the package managers would be.

NickHastings:
I updated the table in my branch to show both. Perhaps something like that could work?

YaLTeR:
> For this reason many projects ship build systems with install and uninstall targets.

I briefly considered it, but I figured that for now we don't have *that* complicated of a setup. It's easier to manually install files in a package than deal with the potential interaction with some build system and cargo.

> I'd expect that there would be more people wanting to try out a nice new wayland compositor manually than there are distro package managers. I also expect such people would be less likely to know where to install the components than the package managers would be.

Well, we've got packages for a few popular distros already, so maybe it's not a big deal.

> I updated the table in my branch to show both. Perhaps something like that could work?

Eh, it becomes a bit wider, which will lead to awkward word wrapping on narrower screens. Maybe if you really think it's necessary, it can be a second separate table below?

NickHastings:
> > I updated the table in my branch to show both. Perhaps something like that could work?
> 
> Eh, it becomes a bit wider, which will lead to awkward word wrapping on narrower screens. Maybe if you really think it's necessary, it can be a second separate table below?

Personally I don't mind if the information is not in there, I already know where to install stuff like this, I just thought it would be helpful for those who don't.

I tried it as two separate tables, but then switched to a single table since I thought being more concise would be better.

NickHastings:
I just pushed a version with two tables.

wiki/Getting-Started.md
```diff
@@ -164,14 +164,14 @@ If you're not on NixOS, you may need [NixGL](https://github.com/nix-community/ni
 nix run --impure github:guibou/nixGL -- ./results/bin/niri
 ```
 
-### Installation
+### Packaging
 
-The recommended way to install and run niri is as a standalone desktop session.
-To do that, put files into the correct directories according to this table.
+The recommended way to package niri  is to have niri run as a standalone desktop
+session. To do that, put files into the correct directories according to this table.
```
YaLTeR, line 170:
```suggestion
The recommended way to package niri is so that it runs as a standalone desktop session.
To do that, put files into the correct directories according to this table.
```

wiki/Getting-Started.md
```diff
@@ -164,14 +164,14 @@ If you're not on NixOS, you may need [NixGL](https://github.com/nix-community/ni
...
 
 | File | Destination |
 | ---- | ----------- |
-| `target/release/niri` | `/usr/bin/` |
+| `target/release/niri` | `/usr/bin/` |
```
YaLTeR, line 174:
```suggestion
| `target/release/niri` | `/usr/bin/` |
```

wiki/Getting-Started.md
```diff
@@ -180,5 +180,22 @@ To do that, put files into the correct directories according to this table.
 
 Doing this will make niri appear in GDM and other display managers.
 
+### Manual Installation
+
+If installing directly without a package, the recommended file destinations area
+slightly different. In this case, put the files in the directories indicated in 
+the table below.
```
YaLTeR, line 187:
```suggestion
If installing directly without a package, the recommended file destinations are slightly different.
In this case, put the files in the directories indicated in the table below.
These may vary depending on your distribution.
```

wiki/Getting-Started.md
```diff
@@ -180,5 +180,22 @@ To do that, put files into the correct directories according to this table.
...
+
+| File | Destination |
+| ---- | ----------- |
+| `target/release/niri` | `/usr/local/bin/` |
+| `resources/niri-session` | `/usr/local/bin/` |
+| `resources/niri.desktop`  | `/usr/local/share/wayland-sessions/` |
+| `resources/niri-portals.conf` | `/usr/local/share/xdg-desktop-portal/` |
+| `resources/niri.service` | `/usr/lib/systemd/user/` |`/etc/systemd/user/` |
+| `resources/niri-shutdown.target` | `/usr/lib/systemd/user/` |`/etc/systemd/user/` |
```
YaLTeR, line 196:
```suggestion
| `resources/niri.service` |`/etc/systemd/user/` |
| `resources/niri-shutdown.target` | `/etc/systemd/user/` |
```

wiki/Getting-Started.md
```diff
@@ -180,5 +180,22 @@ To do that, put files into the correct directories according to this table.
...
+
+
```
YaLTeR, line 198:
```suggestion
```

NickHastings: Oops, that was sloppy of me.

YaLTeR:
Thanks

PR #497, opened 2024-07-01
Added Commnads to focus windows or Monitors above/below the active window

TheAngusMcFire:
I use a multi monitor setup with monitors on top of each other, so I added commands to move up/down or switch to the next monitor.
I used the existing FocusColumnOrMonitorLeft/Right commands as Inspiration.


YaLTeR:
Wouldn't it make more sense for it to be `focus-window-or-workspace-or-monitor-down` that also goes through workspaces?

TheAngusMcFire:
The idea is to naturally navigate  through the available screen space.
e.g. I am on my main monitor and want to switch to my upper monitor which just runs netflix I just go up without thinking which workspace.
If I would use something like  focus-window-or-workspace-or-monitor-down on the 4 the workspace on my main monitor, I would  need to move through 4 workspaces before I would get on the upper monitor.

Niri already implements:     

```
FocusColumnOrMonitorLeft,
FocusColumnOrMonitorRight,
```
so I figured I just add the same behavior for Monitors above and below the main monitor.

YaLTeR:
> The idea is to naturally navigate through the available screen space.
e.g. I am on my main monitor and want to switch to my upper monitor which just runs netflix I just go up without thinking which workspace.

Right, I see, that makes sense.

niri-ipc/src/lib.rs
```diff
@@ -120,6 +120,10 @@ pub enum Action {
     FocusColumnRightOrFirst,
     /// Focus the next column to the left, looping if at start.
     FocusColumnLeftOrLast,
+    /// Focus the column or the monitor above.
```
YaLTeR, line 123:
```suggestion
    /// Focus the window or the monitor above.
```

niri-ipc/src/lib.rs
```diff
@@ -120,6 +120,10 @@ pub enum Action {
...
+    FocusWindowOrMonitorUp,
+    /// Focus the column or the monitor below.
```
YaLTeR, line 125:
```suggestion
    /// Focus the window or the monitor below.
```

src/layout/mod.rs
```diff
@@ -1264,6 +1264,43 @@ impl<W: LayoutElement> Layout<W> {
...
+                    workspace.focus_down();
+                    return false;
+                }
+            }
+        }
+
+        self.focus_output(output);
+        true
+    }
+
```
YaLTeR, line 1303: You should add the two new actions to tests at the bottom of this file. At least into the Op struct which participates in randomized tests.

TheAngusMcFire: Done, sorry I missed this I just saw It in the documentation :see_no_evil:

niri-ipc/src/lib.rs
TheAngusMcFire: Done

niri-ipc/src/lib.rs
TheAngusMcFire: Done

YaLTeR:
Thanks!

PR #504, opened 2024-07-02
Add: more border colorspaces

CaliOn2:
implementing css style color interpolation


wiki/Configuration:-Layout.md
```diff
@@ -169,7 +169,7 @@ layout {
         inactive-color "#505050"
 
         // active-gradient from="#ffbb66" to="#ffc880" angle=45 relative-to="workspace-view"
-        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view"
+        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view" gradient-type="linear"
```
YaLTeR, line 172: I think it would make sense for this to instead be spelled like `in="srgb"`, `in="srgb-linear"`, `in="oklch shorter hue"`, mirroring [the CSS syntax](https://www.w3.org/TR/css-color-4/#color-interpolation-method).

niri-config/src/lib.rs
```diff
@@ -419,6 +421,15 @@ pub enum GradientRelativeTo {
     WorkspaceView,
 }
 
+#[derive(knuffel::DecodeScalar, Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub enum GradientType {
```
YaLTeR, line 428: `GradientColorSpace` maybe?

src/render_helpers/border.rs
```diff
@@ -43,6 +44,7 @@ impl BorderRenderElement {
     pub fn new(
         size: Size<f64, Logical>,
         gradient_area: Rectangle<f64, Logical>,
+        gradient_format: f32,
```
YaLTeR, line 47: Pass the enum here, and convert right before passing into the uniform. Render elements know what shaders they use and their input format; whoever uses the render elements doesn't.

src/render_helpers/shaders/color_interpol.frag
```diff
@@ -0,0 +1,42 @@
+
```
YaLTeR, line 1:
```suggestion
```

src/render_helpers/shaders/color_interpol.frag
```diff
@@ -0,0 +1,42 @@
...
+vec4 color_linear(vec4 color) {
```
YaLTeR, line 2:
```suggestion
vec4 srgb_to_linear(vec4 color) {
```

src/render_helpers/shaders/color_interpol.frag
```diff
@@ -0,0 +1,42 @@
...
+  return vec4(
+    pow(color.r, 2.0),
+    pow(color.g, 2.0),
+    pow(color.b, 2.0),
+    color.a
+  );
+}
+
+vec4 color_root(vec4 color) {
```
YaLTeR, line 11:
```suggestion
vec4 linear_to_srgb(vec4 color) {
```

src/render_helpers/shaders/color_interpol.frag
```diff
@@ -0,0 +1,42 @@
...
+  return vec4(
+    sqrt(color.r),
+    sqrt(color.g),
+    sqrt(color.b),
+    color.a
+  );
+}
+
+vec4 color_mix(vec4 color1, vec4 color2, float color_ratio) {
+
```
YaLTeR, line 21: Could you format this file properly please? See how other shaders are formatted. Remove weird empty lines, add missing spaces, fix indentation, make the comments follow the same style.

src/render_helpers/shaders/color_interpol.frag
```diff
@@ -0,0 +1,42 @@
...
+vec4 color_linear(vec4 color) {
+  return vec4(
+    pow(color.r, 2.0),
+    pow(color.g, 2.0),
+    pow(color.b, 2.0),
+    color.a
+  );
```
YaLTeR, line 8: I wonder if it's more efficient to do something like `vec4(color.rgb * color.rgb, color.a)`

src/render_helpers/shaders/resize.frag
```diff
@@ -5,6 +5,6 @@ vec4 resize_color(vec3 coords_curr_geo, vec3 size_curr_geo) {
     vec3 coords_tex_next = niri_geo_to_tex_next * coords_curr_geo;
     vec4 color_next = texture2D(niri_tex_next, coords_tex_next.st);
 
-    vec4 color = mix(color_prev, color_next, niri_clamped_progress);
+    vec4 color = color_mix(color_prev, color_next, niri_clamped_progress);
```
YaLTeR, line 8:
```suggestion
    vec4 color = mix(color_prev, color_next, niri_clamped_progress);
```
This is incorrect to change, this is the window contents crossfade. There's no gradient rendering in the resize shader.

You can also remove the include from the resize shader compilation.

Actually, you can even just inline the functions into `border.frag` for now since that's the only place that draws the gradients. Thereby avoiding the whole problem.

wiki/Configuration:-Layout.md
```diff
@@ -209,7 +209,7 @@ There's also a *deprecated* syntax for setting colors with four numbers represen
 
 Similarly to colors, you can set `active-gradient` and `inactive-gradient`, which will take precedence.
 
-Gradients are rendered the same as CSS [`linear-gradient(angle, from, to)`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient).
+Gradients can be rendered the same as CSS [`linear-gradient(angle, from, to)`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient).
```
YaLTeR, line 212:
```suggestion
Gradients are rendered the same as CSS [`linear-gradient(angle, from, to)`](https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient).
```
They will still be rendered the same as CSS linear-gradient, just if you set the `in`.

wiki/Configuration:-Layout.md
```diff
@@ -221,6 +221,18 @@ layout {
 }
 ```
 
+Gradients can be rendered with different kinds of color interpolation, this doesen't mean that the arguments the gradient takes
```
YaLTeR, line 224: Markdown files put newline after every sentence and not in other places. For lists use proper markdown lists.

YaLTeR (changes requested):
I'd also like to see a test with a static gradient of every interpolation type added into niri-visual-tests (copy the gradient test that's there multiple times and remove the animation). This way I will be able to compare with browser rendering and make sure it matches.

niri-config/src/lib.rs
```diff
@@ -422,6 +422,8 @@ pub struct Gradient {
     pub angle: i16,
     #[knuffel(property, default)]
     pub relative_to: GradientRelativeTo,
+    #[knuffel(property, str)]
+    pub _in: GradientInterpolation,
```
YaLTeR, line 426: Prefixing with `_` means "allow unused"
```suggestion
    pub in_: GradientInterpolation,
```

niri-config/src/lib.rs
```diff
@@ -431,6 +433,43 @@ pub enum GradientRelativeTo {
     WorkspaceView,
 }
 
+#[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
```
YaLTeR, line 436: You don't need `knuffel::Decode` on this since you're using `FromStr`.

niri-config/src/lib.rs
```diff
@@ -431,6 +433,43 @@ pub enum GradientRelativeTo {
...
+pub struct GradientInterpolation {
+    #[knuffel(property, default)]
+    pub color_space: GradientColorSpace,
+    #[knuffel(property, default)]
+    pub hue_interpol: HueInterpolation
+}
+
+#[derive(knuffel::DecodeScalar, Debug, Default, Clone, Copy, PartialEq, Eq)]
```
YaLTeR, line 444: You don't need `knuffel::DecodeScalar` on this since you're doing it manually.

niri-config/src/lib.rs
```diff
@@ -431,6 +433,43 @@ pub enum GradientRelativeTo {
...
+    Rec2020,
+    Lab,
+    Oklab,
+    Xyz,
+    XyzD50,
+    XyzD65,
+    Hsl,
+    Hwb,
+    Lch,
+    Oklch
```
YaLTeR, line 461: We don't need all of these here, just the ones we actually support. And I think it makes sense to support only srgb, srgb-linear, oklab and oklch, not sure any other would be useful in our setting?

niri-config/src/lib.rs
```diff
@@ -1401,6 +1440,55 @@ impl CornerRadius {
...
+                "hwb" => GradientColorSpace::Hwb,
+                "lch" => GradientColorSpace::Lch,
+                "oklch" => GradientColorSpace::Oklch,
+                &_ => return Err(miette!("Invalid color-space: {in_str}"))
+            }
+        } else {
+            GradientColorSpace::Srgb
+        };
+
+        let interpolation = if in_part2 != None {
```
YaLTeR, line 1475: Should this also check that a polar space was parsed on the step above? Because this only makes sense for polar spaces

src/layout/focus_ring.rs
```diff
@@ -99,6 +103,9 @@ impl FocusRing {
             GradientRelativeTo::WorkspaceView => view_rect,
         };
 
+        let gradient_format = gradient._in;
```
YaLTeR, line 106: You can pass this directly now, no need to extract a variable.

niri-config/src/lib.rs
YaLTeR, line 436: Add `Default` to derives here

src/layout/focus_ring.rs
```diff
@@ -91,6 +91,10 @@ impl FocusRing {
             to: color,
             angle: 0,
             relative_to: GradientRelativeTo::Window,
+            _in: GradientInterpolation {
```
YaLTeR, line 94: Default::default()

src/layout/tile.rs
```diff
@@ -757,6 +757,10 @@ impl<W: LayoutElement> Tile<W> {
                         return BorderRenderElement::new(
                             geo.size,
                             Rectangle::from_loc_and_size((0., 0.), geo.size),
+                            GradientInterpolation{
```
YaLTeR, line 760: GradientInterpolation::default()

src/render_helpers/border.rs
```diff
@@ -77,6 +80,10 @@ impl BorderRenderElement {
             params: Parameters {
                 size: Default::default(),
                 gradient_area: Default::default(),
+                gradient_format: GradientInterpolation {
```
YaLTeR, line 83: Default::default()

src/render_helpers/shaders/border.frag
```diff
@@ -21,6 +23,50 @@ uniform vec2 geo_size;
...
+  return vec4(
+    sqrt(color.r) * color.a,
+    sqrt(color.g) * color.a,
+    sqrt(color.b) * color.a,
+    color.a
+  );
+}
+
+vec4 color_mix(vec4 color1, vec4 color2, float color_ratio) {
+
```
YaLTeR, line 43:
```suggestion
```

src/render_helpers/shaders/mod.rs
```diff
@@ -66,7 +68,10 @@ impl Shaders {
             })
             .ok();
 
-        let resize = compile_resize_program(renderer, include_str!("resize.frag"))
+        let resize = compile_resize_program(
+                renderer,
+                include_str!("resize.frag")
+            )
```
YaLTeR, line 74: This can be reverted

src/window/mapped.rs
```diff
@@ -289,6 +289,10 @@ impl Mapped {
                     return BorderRenderElement::new(
                         geo.size,
                         Rectangle::from_loc_and_size((0., 0.), geo.size),
+                        GradientInterpolation{
+                            color_space: GradientColorSpace::Srgb,
+                            hue_interpol: HueInterpolation::Shorter
+                        },
```
YaLTeR, line 295: GradientInterpolation::default()

niri-config/src/lib.rs
```diff
@@ -422,6 +422,8 @@ pub struct Gradient {
     pub angle: i16,
     #[knuffel(property, default)]
     pub relative_to: GradientRelativeTo,
+    #[knuffel(property, str)]
```
YaLTeR, line 425:
```suggestion
    #[knuffel(property(name="in"), str)]
```
Something like this I think

niri-config/src/lib.rs
```diff
@@ -1401,6 +1440,55 @@ impl CornerRadius {
...
+
+        let interpolation = if in_part2 != None {
+            let in_str = in_part2.unwrap();
+            match in_str {
+                "shorter" => HueInterpolation::Shorter,
+                "longer" => HueInterpolation::Longer,
+                "increasing" => HueInterpolation::Increasing,
+                "decreasing" => HueInterpolation::Decreasing,
+                &_ => return Err(miette!("Invalid hue-interpolation: {in_str}"))
+            }
```
YaLTeR, line 1483: Let's also verify that it ends with ` hue` and there's nothing else following it. For any potential future compat, and to reduce confusion.

YaLTeR (changes requested):
This is shaping up nicely

niri-config/src/lib.rs
CaliOn2: i should probably implement that for user simplicity however on the rendering end it makes little difference as the hue interpolation type is ignored for non polar spaces

src/render_helpers/shaders/border.frag
```diff
@@ -21,6 +23,178 @@ uniform vec2 geo_size;
...
+  float divisor =
+    max(max(color.r, color.g), color.b) > 1.0 ?
+    max(max(color.r, color.g), color.b) :
+    1.0 ;
+  color.rgb /= divisor;
+  return color;
+}
+
+vec4 linear_to_oklab(vec4 color){
+  float l = color.r * 0.4122214708 + color.g * 0.5363325363 + color.b * 0.0514459929;
```
YaLTeR, line 75: Idk if it makes any difference in the compiled shader bytecode, but these operations can probably be written as something like:
```
mat3x3 rgb_to_lms = mat3x3(/* the numbers */);
mat3x3 lms_to_lab = mat3x3(/* the numbers */);

vec3 lms = pow(color.rgb * rgb_to_lms, 3.0);
return vec4(lms * lms_to_lab, color.a);
```

niri-visual-tests/src/cases/gradient_srgb.rs
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            delta = delta.div_f64(slowdown);
+        }
+
+        self.time = delta.as_secs_f32() ;
+
+        if self.time >= 26.9 {
+            self.time = 0.0
+        }
+    }
```
YaLTeR, line 57: In static tests (no animation) these two functions can be removed.

resources/default-config.kdl
```diff
@@ -153,6 +153,7 @@ layout {
         // The angle is the same as in linear-gradient, and is optional,
         // defaulting to 180 (top-to-bottom gradient).
         // You can use any CSS linear-gradient tool on the web to set these up.
+        // for gradient types visit the wiki
```
YaLTeR, line 156:
```suggestion
        // Changing the color space is also supported, check the wiki for more info.
```

YaLTeR (commented):
Thanks for the visual tests. Giving it a compare against Firefox using the following code on https://developer.mozilla.org/en-US/play:

```html
<div></div>
```
```css
div {
  width: 459px;
  height: 146px;
  background: linear-gradient(90deg in srgb-linear, #00ff00, #ff0000);
  position: absolute;
  top: 100px;
  left: 100px;
}
```

1. srgb matches exactly (as it used to).
2. srgb-linear is slightly darker than it should be:
    | niri | firefox |
    | --- | --- |
    | image | image |
3. oklab, oklch-shorter and oklch-increasing are also slightly darker (guess due to the srgb-linear problem)
4. oklch-longer and oklch-decrearsing has way darker blue:
    | niri | firefox |
    | --- | --- |
    | image | image |

niri-config/src/lib.rs
```diff
@@ -422,6 +422,8 @@ pub struct Gradient {
     pub angle: i16,
     #[knuffel(property, default)]
     pub relative_to: GradientRelativeTo,
+    #[knuffel(property(name="in"), str)]
```
YaLTeR, line 425:
```suggestion
    #[knuffel(property(name="in"), str, default)]
```

CaliOn2:
should be done now

wiki/Configuration:-Layout.md
```diff
@@ -32,7 +32,7 @@ layout {
         active-color "#ffc87f"
         inactive-color "#505050"
         // active-gradient from="#ffbb66" to="#ffc880" angle=45 relative-to="workspace-view"
-        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view"
+        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view" gradient-type="linear"
```
YaLTeR, line 35:
```suggestion
        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view" in="srgb-linear"
```

wiki/Configuration:-Layout.md
YaLTeR, line 172:
```suggestion
        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view" in="srgb-linear"
```

YaLTeR (commented):
Thanks, it seems that the visual tests match Firefox now.

src/render_helpers/shaders/border.frag
```diff
@@ -21,6 +23,174 @@ uniform vec2 geo_size;
...
+  //  srgb
+  
+  if (colorspace == 0.0) {
+    return mix(color1, color2, color_ratio);
+  }
+  
+  vec4 color_out;
+
+  color1.rgb /= color1.a != 0.0 ? color1.a : 1.0;
+  color2.rgb /= color2.a != 0.0 ? color2.a : 1.0;
```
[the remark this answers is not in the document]
YaLTeR: Actually, is this even defined anywhere in the standard? I'm not seeing any green in `linear-gradient(270deg in srgb-linear, #f00f, #0f00)` which implies that this one worked with premul colors. I think this only uses unpremul for oklch and not for the other three?

niri-config/src/lib.rs
```diff
@@ -509,7 +535,7 @@ impl Color {
 impl From<Color> for [f32; 4] {
     fn from(c: Color) -> Self {
         let [r, g, b, a] = [c.r, c.g, c.b, c.a].map(|x| x as f32 / 255.);
-        [r * a, g * a, b * a, a]
+        [r, g, b, a]
```
YaLTeR, line 538: Note to self: change this to two methods instead before merging.

YaLTeR (commented):
Hmm, so it seems that the rectangular color space edge case is not just alpha = 0, but all low alpha values in general. For example, if I set target alpha to 11 then niri will have way more green than firefox:

| niri | firefox |
| --- | --- |
|  image | image |

```css
div {
  width: 459px;
  height: 146px;
  background: linear-gradient(270deg in srgb, #ff0000ff, #00ff0011);
  position: absolute;
  top: 100px;
  left: 100px;
}

body {
  background: rgb(51, 51, 51, 1.0)
}
```
```
active-gradient from="#ff0000ff" to="#00ff0011" angle=270 in="srgb-linear"
```

This is a problem for srgb, srgb-linear, oklab. But oklch looks just fine.

YaLTeR:
This seems to have the CSS rules for what is premultiplied when: https://www.w3.org/TR/css-color-4/#interpolation-alpha

CaliOn2:
im not going to work on this anymore as I really dislike the way firefox displays this and am not willing to desicrate the way the gradients look to fit some css specs as that defeats the entire point of me writing this i hope you find some way to implement this on your own good luck!

YaLTeR:
Thanks a lot for the work! I'll finish this myself. 

FWIW if there's any flag in CSS to avoid the premultiplication, we can add that too. Or if there isn't we can add it anyway, something like premul=false that preserves the current behavior.

I just want the *default* to match CSS so that there's some reference on how it should look to compare against, and so that people can use familiar tools to configure it.

YaLTeR:
GitHub + LFS was being uncooperative, so I had to make a new PR: https://github.com/YaLTeR/niri/pull/548

I did the following:

- added premultiplication to match CSS, added more visual tests for this
- made Color more explicit about premultiplication, added helper methods
- fixed some issues with parsing, added tests
- did some renames and formatting fixes
- rewrote the wiki part

Too lazy to extract things into commits so I'll just squash it all into one.

PR #513, opened 2024-07-03
tty: do not crash on failing to scan connectors

cmeissl:
depends on https://github.com/Smithay/smithay/pull/1477


YaLTeR:
Closing since a while ago we updated Smithay so this change is already in.

PR #522, opened 2024-07-06
Pin `smithay` version to prevent auto-updates from causing compilation errors

itsjunetime:
When updating with `cargo install-update`, cargo tries to automatically update all dependencies to their latest 'compatible' versions (ignoring the lockfile). However, that means that the `smithay` dependency, since it's a git repo with no specified commit, will just get auto-updated to the latest revision, even if that's not compatible (since semver doesn't really apply to git revisions). So this just pins the version of `smithay` in Cargo.toml so that updating via `cargo install-update` doesn't automatically update `smithay` to an incompatible version from the git repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3355,7 +3355,7 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/Smithay/smithay.git#234586dbea6bc084cb72a32f164997e47ea36b2b"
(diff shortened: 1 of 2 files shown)


YaLTeR:
I mean... That's what the lock file is for? Niri isn't the only project that would break if you ignore the lock file. 

Besides, I'm not sure cargo install (and therefore install-update) is a very expected use case in the first place. You don't get any of the session files that way.

dragonnn:
cargo install does support `--locked` flag with does take into account the lock file. As far I know it is now consider using it without the --locked flag a bad idea and it not being the default is only due bad backward compatibility

kchibisov:
>it not being the default is only due bad backward compatibility

It was default in the past, they flipped it.

PR #532, opened 2024-07-08
Implement Binding Modes

therealolie:
https://github.com/YaLTeR/niri/discussions/529

example config:
```
binds {
	Mod+A { binding-mode "test"; }
	Mod+T { spawn "alacritty"; }
	Mod+Shift+E { quit; }
	Mod+Shift+Slash { show-hotkey-overlay; }
}
binds "test" {
	B { binding-mode "default"; }
	Mod+E { quit; }
	Mod+Shift+Slash { show-hotkey-overlay; }
}
```
`niri msg binding-mode` -> `default`

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -48,8 +48,8 @@ pub struct Config {
     pub environment: Environment,
     #[knuffel(children(name = "window-rule"))]
     pub window_rules: Vec<WindowRule>,
-    #[knuffel(child, default)]
-    pub binds: Binds,
+    #[knuffel(children(name = "binds"))]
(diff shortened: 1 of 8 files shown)


YaLTeR:
The implementation seems fairly straightforward (though there are some things I'd fix). But, I'm still not entirely convinced we want this?

YaLTeR:
Let's discuss that further in the discussion, leaving the PR for code comments

vncnz:
Not OP, but can be useful for shortcuts when an overlay is open (for example with EWW). It's the only user case I can think about.
With overlay closed normal keybinds are active with modifier keys, with overlay open you can have more keybinds on normal letters too.

PR #542, opened 2024-07-12
Add metadata for generate deb package with cargo deb

tglman:
Hi,

This add the metadata to generate a .deb with `cargo deb` in the same way that is done for the rpm package.


YaLTeR:
Thanks

PR #557, opened 2024-07-25
Background Color Option in Config 

anant-357:
Hi, please take a look. I have also implemented recoloring on config reload, although I am not sure if it is the optimised way.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -18,6 +18,8 @@ use smithay::input::keyboard::xkb::{keysym_from_name, KEYSYM_CASE_INSENSITIVE};
 use smithay::input::keyboard::{Keysym, XkbConfig};
 use smithay::reexports::input;
 
+pub const DEFAULT_BACKGROUND_COLOR: Color = Color::from_array_unpremul([0.2, 0.2, 0.2, 1.]);
+
 #[derive(knuffel::Decode, Debug, PartialEq)]
 pub struct Config {
     #[knuffel(child, default)]
@@ -323,6 +325,8 @@ pub struct Output {
     pub mode: Option<ConfiguredMode>,
     #[knuffel(child)]
     pub variable_refresh_rate: bool,
+    #[knuffel(child, default = DEFAULT_BACKGROUND_COLOR)]
+    pub background_color: Color,
 }
 
 impl Default for Output {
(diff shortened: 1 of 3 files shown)


niri-config/src/lib.rs
```diff
@@ -2879,6 +2882,7 @@ mod tests {
                         refresh: Some(144.),
                     }),
                     variable_refresh_rate: true,
+                    background_color: Color::from_rgba8_unpremul(51, 51, 51, 255),
```
YaLTeR, line 2885: Please add some non-default color to the test config.

src/niri.rs
```diff
@@ -1120,7 +1121,21 @@ impl State {
...
+
+            let background_color = config
+                .map(|c| c.background_color)
+                .unwrap_or(Color::from_array_unpremul(CLEAR_COLOR));
+
+            if self.niri.output_state.get(output).is_some_and(|os| {
+                os.background_buffer.color() != background_color.to_array_unpremul()
+            }) {
+                recolored_outputs.push((output.clone(), background_color));
+            }
```
YaLTeR, line 1133: Can't this just set the color directly here instead of doing the whole `recolored_outputs`? Also avoid the `output_recolored()` function.

niri-config/src/lib.rs
```diff
@@ -335,6 +337,7 @@ impl Default for Output {
             position: None,
             mode: None,
             variable_refresh_rate: false,
+            background_color: Color::from_rgba8_unpremul(51, 51, 51, 255),
```
YaLTeR, line 340: I feel like since we use this default color both here and in niri.rs, we should move the `CLEAR_COLOR` constant here and call it something like `DEFAULT_BACKGROUND_COLOR`.

src/niri.rs
anant-357: I'll remove the `output_recolored()` function but I have used recolored_outputs because I cannot call `self.niri.queue_redraw(&output)` (mutable borrow) inside output loop which also borrows `self.niri` immutably. Is there a particular way of doing this? Thanks

YaLTeR: I guess you could leave `recolored_outputs` as a `Vec<Output>` for this to call `queue_redraw` after the loop.

anant-357: Pushed! Please let me know if everything is done correctly.

YaLTeR:
Thanks. I'm doing some more fixes on top, then will merge.

PR #562, opened 2024-07-30
Screencopy version 3

my4ng:


src/protocols/screencopy.rs
```diff
@@ -187,24 +200,23 @@ where
...
-        //         buffer_size.w as u32,
-        //         buffer_size.h as u32,
-        //     );
-        //
-        //     // Notify client that all supported buffers were enumerated.
-        //     frame.buffer_done();
-        // }
+        // Send desired DMA buffer parameters.
+        frame.linux_dmabuf(
+            Fourcc::Argb8888 as u32,
```
YaLTeR, line 205: I think we want at least Xrgb too here. Might be worth checking what formats clients typically query.

src/protocols/screencopy.rs
```diff
@@ -275,44 +287,70 @@ where
...
             buffer,
             frame: frame.clone(),
             info: info.clone(),
             submitted: false,
         });
+
+        if with_damage {
+            if let Some(damages) = damages {
+                for Rectangle { loc, size } in damages {
+                    frame.damage(loc.x as u32, loc.y as u32, size.w as u32, size.h as u32);
```
YaLTeR, line 335: This should be sent before ready:

> This event is sent right before the ready event when copy_with_damage is requested.

And ready is currently sent in state.frame() above.

src/protocols/screencopy.rs
```diff
@@ -275,44 +287,70 @@ where
...
             );
             return;
-        }
+        };
 
         copied.store(true, Ordering::SeqCst);
 
-        state.frame(Screencopy {
-            with_damage,
+        let damages = state.frame(Screencopy {
```
YaLTeR, line 325: So, the semantics for copy_with_damage() is a bit different from copy() (which is one of the reasons why I haven't implemented v3 initially). Whereas copy() can always copy the current frame immediately, as it does here, copy_with_damage() cannot:

> Same as copy, except it waits until there is damage to copy.

As far as I understand this PR, if a client issues copy_with_damage() before there's been any damage, niri will copy the frame and submit it right away, and not submit any damage. This is wrong, as far as I can tell from the quoted description from the protocol.

Instead, what should happen, is that every time the corresponding output is rendered, it should also get checked against the screencopy damage tracker, and if there's any damage, then it should get submitted to screencopy (damage event goes first).

YaLTeR (changes requested):

src/protocols/screencopy.rs
my4ng: `wf-recorder` accepts a number of [formats](https://github.com/ammen99/wf-recorder/blob/9978c3570436eb5358e4578421d83a7f3ffd2683/src/frame-writer.cpp#L149). I will add XRGB8888 for now.

my4ng: The documentation of the protocol seems to suggest that only one format can be supported by the compositor:
> Provides information about linux-dmabuf buffer parameters that need to be used for this frame. This event is sent _once_ after the frame is created if linux-dmabuf buffers are supported.

YaLTeR: Hm, okay then. I'm just thinking that maybe Xrgb is a better choice, but idk. Might be worth checking what sway sends in that case

src/protocols/screencopy.rs
```diff
@@ -16,12 +20,13 @@ use smithay::reexports::wayland_server::{
...
-use smithay::wayland::shm;
+use smithay::wayland::{dmabuf, shm};
 
-// We do not support copy_with_damage() semantics yet.
-const VERSION: u32 = 1;
+const VERSION: u32 = 3;
 
-pub struct ScreencopyManagerState;
+pub struct ScreencopyManagerState {
+    damage_trackers: HashMap<Output, ScreencopyDamageTracker>,
```
YaLTeR, line 28: There needs to be a damage tracker per screencopy manager instance, rather than per output. (It can stay per-output, but every output can have multiple independent damage trackers.)

This is because multiple screencopy clients can be connected at once, and they will all have separate damage accumulation.

The protocol says about damage:

> area that has changed since the last copy request that was derived from the current screencopy manager instance

src/niri.rs
```diff
@@ -3730,45 +3745,78 @@ impl Niri {
...
+            return Ok(());
+        }
 
-                Ok(())
-            })
-            .context("primary renderer is missing")?
+        let region_loc = screencopy.region_loc();
+        let elements = elements.into_iter().rev().map(|element| {
+            RelocateRenderElement::from_element(element, region_loc.upscale(-1), Relocate::Relative)
+        });
```
YaLTeR, line 3782: This should happen before the damage tracker because this will affect the damage. (If you screencopy a small rectangle of the output, all damage should be computed only in this rectangle and relative to it.)

src/niri.rs
```diff
@@ -3730,45 +3745,78 @@ impl Niri {
...
+
+        let scale = output.current_scale().fractional_scale().into();
+        let transform = if is_winit_backend {
+            Transform::Normal
```
YaLTeR, line 3786: Having to do this seems a bit weird to me. Like, didn't `grim` work fine without this on the winit backend? If so then with this change `grim` will no longer work fine on the winit backend.

Also, this sounds like it will corrupt the damage reported to wlr-screencopy. I'm not sure if there's any client that uses damage from screencopy though to test it... Could you search if there's a client like that? Because otherwise we'll have to test damage by comparing numerical values.

src/protocols/screencopy.rs
```diff
@@ -42,7 +47,22 @@ impl ScreencopyManagerState {
...
+
+    pub fn damage_tracker(&mut self, output: &Output) -> Option<&mut ScreencopyDamageTracker> {
+        self.damage_trackers.get_mut(output)
+    }
+
+    pub fn damage_tracker_or_insert(&mut self, output: &Output) -> &mut ScreencopyDamageTracker {
+        self.damage_trackers
+            .entry(output.clone())
+            .or_insert_with(|| ScreencopyDamageTracker {
+                damage_tracker: OutputDamageTracker::from_output(output),
```
YaLTeR, line 63: I guess if a region copy is requested then this should be a small region-sized damage tracker, rather than a full output sized damage tracker? Or alternatively you can clip the damage when rendering I guess with `Rectangle::intersect`.

src/protocols/screencopy.rs
```diff
@@ -181,23 +201,21 @@ where
...
-        //     );
-        //
-        //     // Notify client that all supported buffers were enumerated.
-        //     frame.buffer_done();
-        // }
+        // Send desired DMA buffer parameters.
+        frame.linux_dmabuf(
+            Fourcc::Xrgb8888 as u32,
+            buffer_size.w as u32,
+            buffer_size.h as u32,
```
YaLTeR, line 214: When an output resizes, or the output transform changes, or the output disconnects, it should destroy the screencopy objects for this output.

src/protocols/screencopy.rs
YaLTeR: Also, I wonder what should happen if a client creates two screencopy frames at once and calls copy_with_damage() on both of them without waiting for each other? I guess in this case we can fail the second one? Or, if I'm understanding the protocol right, we can queue up the second one to be processed as if it was called right after we send the first one to the client.

src/protocols/screencopy.rs
my4ng: Right now we are checking for output existence but not the other changes. Do you mean destroying the screencopy manager as well?

src/niri.rs
my4ng: The reason why I had to take it into account is `wl-recorder` would output a flipped-180 version instead. There might be a transform part that I forgot to add, I will have to dig more into it.

src/niri.rs
my4ng: Indeed since the x and y are both uint's. I will clip and transform relative to the region_loc then.

src/protocols/screencopy.rs
YaLTeR: Might be worth checking what sway does. Because I'm not sure what `damage` you're supposed to send after an output resize. And destroying the screencopy manager solves that problem. Though, if sway doesn't destroy the manager, then clients won't expect it.

src/niri.rs
YaLTeR: This is likely a wf-recorder problem. I check against grim as that should be a more established implementation.

src/protocols/screencopy.rs
my4ng: > Also, I wonder what should happen if a client creates two screencopy frames at once and calls copy_with_damage() on both of them without waiting for each other? 

If that's the same frame then that's already used, but if different I think it is currently serializing them?

src/render_helpers/mod.rs
```diff
@@ -233,49 +234,49 @@ pub fn render_to_vec(
...
-    dmabuf: smithay::backend::allocator::dmabuf::Dmabuf,
+    dmabuf: Dmabuf,
     size: Size<i32, Physical>,
     scale: Scale<f64>,
     transform: Transform,
     elements: impl Iterator<Item = impl RenderElement<GlesRenderer>>,
 ) -> anyhow::Result<SyncPoint> {
     let _span = tracy_client::span!();
+    ensure!(
+        dmabuf.format().code == Fourcc::Xrgb8888
```
YaLTeR, line 247: This should not be here, render_to_dmabuf() can render to any format. This should be checked in screencopy upon receiving the buffer.

src/niri.rs
```diff
@@ -3728,47 +3734,160 @@ impl Niri {
...
+            &screencopy,
+        )
+        .map(|_| ());
+
+        match render_result {
+            Ok(()) => screencopy.submit(false),
+            Err(_) => {
+                // recreate damage tracker to report full damage next check.
+                *damage_tracker = OutputDamageTracker::new((0, 0), 1.0, Transform::Normal);
+                screencopy.invalid_buffer();
```
YaLTeR, line 3827: post_error() kills the screencopy client, so it should only be used for protocol violations. This should instead send the failed event.

src/protocols/screencopy.rs
```diff
@@ -275,44 +351,50 @@ where
 
         let (buffer, with_damage) = match request {
             zwlr_screencopy_frame_v1::Request::Copy { buffer } => (buffer, false),
-            // zwlr_screencopy_frame_v1::Request::CopyWithDamage { buffer } => (buffer, true),
+            zwlr_screencopy_frame_v1::Request::CopyWithDamage { buffer } => (buffer, true),
             _ => unreachable!(),
         };
 
-        if !shm::with_buffer_contents(&buffer, |_buf, shm_len, buffer_data| {
```
YaLTeR, line 282: The size and format check should remain here because this is the only place where we can clearly check for a protocol violation (we asked for a buffer of specific format and size, and the client sent something different). If subsequent rendering finds that the buffer size (that we asked for) is now wrong, that should produce a failed event.

src/handlers/mod.rs
```diff
@@ -419,14 +420,36 @@ impl ForeignToplevelHandler for State {
...
         }
     }
+
+    fn bind(&mut self, manager: &ZwlrScreencopyManagerV1) {
+        self.niri.screencopy_state.bind_manager(manager);
+    }
+
+    fn destroy(&mut self, manager: &ZwlrScreencopyManagerV1) {
+        self.niri.screencopy_state.destroy_manager(manager);
+    }
```
YaLTeR, line 452: Instead of making these methods public, usually you add one method `fn screencopy_state(&mut self) -> &mut ScreencopyManager`. Check other protocol impls for an example.

YaLTeR (commented):
Haven't looked fully yet

src/handlers/mod.rs
my4ng: I was hoping to restrict the interface so it won't be able to call other `ScreencopyManagerState` public methods, but I guess this is fine too.

src/render_helpers/mod.rs
my4ng: I did it because that's what `render_to_shm` was and is doing.

YaLTeR: Because that one has to render and download in a specific pixel format for it to match up the data layout (and it's also only used for screencopy)

src/niri.rs
my4ng: The docs say `invalid_buffer` is for when:
> buffer attributes are invalid

Isn't that when the format or size of the buffer is not the same as the ones sent by the manager? Those are checked in the `render_to_` functions, but I guess I need to separate protocol violation from other kinds of errors.

src/protocols/screencopy.rs
my4ng: The rendering is checking the buffer size and format against specification though. I want to avoid double checking if possible.

src/niri.rs
YaLTeR: I thought that this branch could trigger if the output resized between copy_with_damage() and rendering? Might be mistaken

src/protocols/screencopy.rs
YaLTeR: I don't think you can avoid the double size checking since the output can resize at any time (which shouldn't kill the screencopy client with a protocol error). It's not a big deal to check it twice

src/niri.rs
```diff
@@ -3821,6 +3944,18 @@ impl Niri {
...
 
+    pub fn remove_screencopy_output(&mut self, output: &Output) {
+        for queue in self.screencopy_state.queues_mut() {
+            if queue
+                .split()
+                .1
+                .is_some_and(|screencopy| screencopy.output() == output)
+            {
+                queue.pop();
+            }
```
YaLTeR, line 3955: I guess this needs to check *all* frames in the queue. If any queued frame matches the output, pop all frames starting from that one to the end of the queue.

src/niri.rs
```diff
@@ -3728,47 +3736,162 @@ impl Niri {
...
+            trace!("screencopy manager destroyed already");
+            return Ok(());
+        };
+        let damage_tracker = queue.split().0;
+
+        let render_result = Self::render_for_screencopy_internal(
+            renderer,
+            output,
+            &elements,
+            screencopy.with_damage(),
```
YaLTeR, line 3821: I guess this will be always false?

src/niri.rs
```diff
@@ -3728,47 +3736,162 @@ impl Niri {
...
+                        }
+                    }
+                };
+            }
+        }
+
+        self.screencopy_state = screencopy_state;
+    }
+
     pub fn render_for_screencopy(
```
YaLTeR, line 3789: Since this handles only non-damage screencopies, maybe rename this to something like `render_for_screencopy_no_damage`

src/render_helpers/mod.rs
```diff
@@ -233,18 +233,18 @@ pub fn render_to_vec(
...
 pub fn render_to_dmabuf(
     renderer: &mut GlesRenderer,
-    dmabuf: smithay::backend::allocator::dmabuf::Dmabuf,
+    dmabuf: Dmabuf,
     size: Size<i32, Physical>,
     scale: Scale<f64>,
     transform: Transform,
     elements: impl Iterator<Item = impl RenderElement<GlesRenderer>>,
-) -> anyhow::Result<SyncPoint> {
+) -> anyhow::Result<()> {
```
YaLTeR, line 243: Hm, why'd you remove this? Sure it's not used currently, but it may become needed, especially with explicit sync.

src/render_helpers/mod.rs
```diff
@@ -255,27 +255,14 @@ pub fn render_to_shm(
...
-                && buffer_data.stride == size.w * 4
-                && buffer_data.height == size.h
-                && shm_len as i32 == buffer_data.stride * buffer_data.height,
-            "invalid buffer format or size"
-        );
+        let size = Size::<_, Physical>::from((buffer_data.width, buffer_data.height));
+        let mapping =
+            render_and_download(renderer, size, scale, transform, Fourcc::Xrgb8888, elements)?;
 
-        ensure!(bytes.len() == shm_len, "mapped buffer has wrong length");
```
YaLTeR, line 278: These checks must remain here to ensure the unsafe code below doesn't access invalid memory

src/niri.rs
```diff
@@ -3728,47 +3736,162 @@ impl Niri {
...
+            unreachable!("output always has static mode");
+        };
 
-                screencopy.submit(false);
+        let size = screencopy.buffer_size();
+        let scale: Scale<f64> = output.current_scale().fractional_scale().into();
+        let transform = output.current_transform();
+
+        if size != last_size || scale != last_scale || transform != last_transform {
+            *damage_tracker = OutputDamageTracker::new(size, scale, transform);
```
YaLTeR, line 3859: As far as I understand, this will always use the screencopy buffer size for rendering. However, when an output resizes after copy_with_damage(), it should fail the screencopy frame, instead of using an incorrect size for rendering.

This makes me realize that after capture_output_region() there's no way to tell if an output resized or not from the sizes alone. So, perhaps, after all, we should just fail all screencopies for the output from `Niri::output_resized()`?

src/niri.rs
```diff
@@ -3728,47 +3736,162 @@ impl Niri {
...
+                render_to_dmabuf(renderer, dmabuf.clone(), size, scale, transform, elements)
+                    .context("error rendering to screencopy dmabuf")
+                    .map(|_| ())
+            }
+            ScreencopyBuffer::Shm(wl_buffer) => {
+                render_to_shm(renderer, wl_buffer, scale, transform, elements)
+                    .context("error rendering to screencopy shm buffer")
+            }
+        }?;
+        renderer.unbind().context("error unbinding screencopy")?;
```
YaLTeR, line 3893: Manual unbind shouldn't be necessary?

src/protocols/screencopy.rs
```diff
@@ -275,44 +351,63 @@ where
...
+            ScreencopyBuffer::Dmabuf(dmabuf::get_dmabuf(&buffer).unwrap().clone())
+        } else if shm::with_buffer_contents(&buffer, |_, shm_len, buffer_data| {
+            buffer_data.format == Format::Xrgb8888
+                && buffer_data.width == size.w
+                && buffer_data.height == size.h
+                && buffer_data.stride == size.w * 4
+                && shm_len == buffer_data.stride as usize * buffer_data.height as usize
         })
-        .unwrap_or(false)
+        .unwrap_or_default()
```
YaLTeR, line 372: I'd rather keep this explicitly `false`

YaLTeR (commented):
We're getting there. Let's leave the pointer always rendered for now; there's a similar FIXME in PW screencasts, I'll deal with it in both at some point.

src/render_helpers/mod.rs
my4ng: This is currently only used by screencopy which does the checks already. Though indeed shm can be mutated anytime by the client. Again I am trying to avoid double checking the format and size, and I am still not convinced it is necessary to do so.

YaLTeR: This is a *safety* matter and we're writing Rust here :p Besides, these checks are trivial, there's no problem in checking them twice.

src/niri.rs
my4ng: I don't think it is specified what to do when an output changes size. Right now it is just reporting full damage. We can save the output size at the time of request in the info and then compare it and only fail if it has changed.

src/niri.rs
my4ng: I will double check this but it did cause the ghosting issue we talked about before.

src/niri.rs
YaLTeR: Right now, as far as I read what this code will do, is it will render the output with the wrong size (the size from screencopy, whereas the real output size is already different).

src/protocols/screencopy.rs
```diff
@@ -42,7 +85,28 @@ impl ScreencopyManagerState {
...
+            queues: HashMap::new(),
+        }
+    }
+
+    pub fn bind(&mut self, manager: &ZwlrScreencopyManagerV1) {
+        self.queues.insert(manager.clone(), ScreencopyQueue::new());
+    }
+
+    pub fn destroy(&mut self, manager: &ZwlrScreencopyManagerV1) {
+        self.queues.remove(manager);
```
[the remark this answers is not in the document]
my4ng: Well I might be wrong but I thought removing the queue does not destroy the frame objects? This will fail all the frame requests of course but that is expected.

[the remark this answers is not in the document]
YaLTeR: You are correct, but as far as I read the protocol, this is not what should happen. I.e. if the client creates a manager, then calls capture_output(), then destroys a manager, the frame can still get rendered, since "All objects created by the manager will still remain valid, until their appropriate destroy request has been called."

src/niri.rs
my4ng: Yes it is still rendering at the initial size. I meant the doc is not clear on what to do in this case. What do you propose?

YaLTeR: In `Niri::output_resized()` if the size changed, fail the screencopies (maybe also if transform/scale changed but idk)

src/protocols/screencopy.rs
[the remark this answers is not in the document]
my4ng: Mhmmm I am not sure. If the client wants to `copy_with_damage` which will put it in the queue, but the manger instance has already been destroyed, hence it would not be possible to compute the damage.

src/niri.rs
my4ng: Indeed I think it should fail if any of the three changes since whats rendered in screencopy is now different from what's displayed.

src/protocols/screencopy.rs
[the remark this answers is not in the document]
YaLTeR: Why not? The manager is not alive, but the Rust object still exists. You just can't send any events to it.

src/render_helpers/mod.rs
my4ng: Yes I agree. I will put them back in.

src/protocols/screencopy.rs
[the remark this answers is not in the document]
my4ng: I think you are right. I will implement it as described in your first comment. I blame the underspecified documentation :p

src/niri.rs
my4ng: Without manual unbind there is still the ghosting like artifacts. I suspect there might be a bug in the existing code where it is not being rebound properly.

src/protocols/screencopy.rs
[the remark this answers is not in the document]
my4ng: This is a bit tricky since we have to add it to every site that might remove a screencopy from a queue, and can be  error prone to keep track of them all. :(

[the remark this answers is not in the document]
YaLTeR: You could make a wrapper method around queue.pop() that also checks if the manager is still alive?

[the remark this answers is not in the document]
my4ng: Yeah well the problem is that ScreencopyQueue of course can't remove itself from the hashmap it is in. The best way I can think of is let `ScreencopyManagerState::get_queue_mut` take a closure, run it and then check for the emptiness at the end.

YaLTeR (approved):

PR #579, opened 2024-08-08
Fix `nix build`

feathecutie:
Currently, running `nix build` in this repo to build niri using the `flake.nix` file fails because the tests during the check phase can't find certain runtime dependencies (like `libxkbcommon`).

This PR fixes this by simply passing the required runtime dependencies to the phases themselves using the `LD_LIBRARY_PATH` environment variable, fixing the tests without changing the resulting output.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
       system: let
         pkgs = nixpkgs.legacyPackages.${system};
         toolchain = fenix.packages.${system}.complete.toolchain;
-        craneLib = crane.lib.${system}.overrideToolchain toolchain;
+        craneLib = (crane.mkLib pkgs).overrideToolchain toolchain;
 
         craneArgs = {
           pname = "niri";
@@ -80,6 +80,7 @@
           ];
 
           LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
+          LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath craneArgs.runtimeDependencies; # Needed for tests to find libxkbcommon
         };
 
         cargoArtifacts = craneLib.buildDepsOnly craneArgs;
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1720226507,
-        "narHash": "sha256-yHVvNsgrpyNTXZBEokL8uyB2J6gB1wEx0KOJzoeZi1A=",
+        "lastModified": 1722960479,
+        "narHash": "sha256-NhCkJJQhD5GUib8zN9JrmYGMwt4lCRp6ZVNzIiYCl0Y=",
         "owner": "ipetkov",
         "repo": "crane",
-        "rev": "0aed560c5c0a61c9385bddff471a13036203e11c",
+        "rev": "4c6c77920b8d44cd6660c1621dea6b3fc4b4c4f4",
         "type": "github"
       },
       "original": {
@@ -28,11 +28,11 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1719815435,
-        "narHash": "sha256-K2xFp142onP35jcx7li10xUxNVEVRWjAdY8DSuR7Naw=",
+        "lastModified": 1722493751,
+        "narHash": "sha256-l7/yMehbrL5d4AI8E2hKtNlT50BlUAau4EKTgPg9KcY=",
         "owner": "nix-community",
         "repo": "fenix",
-        "rev": "ebfe2c639111d7e82972a12711206afaeeda2450",
+        "rev": "60ab4a085ef6ee40f2ef7921ca4061084dd8cf26",
         "type": "github"
       },
       "original": {
@@ -77,11 +77,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1720368505,
-        "narHash": "sha256-5r0pInVo5d6Enti0YwUSQK4TebITypB42bWy5su3MrQ=",
+        "lastModified": 1723572004,
+        "narHash": "sha256-U5gKtbKuPahB02iGeGHFPlKr/HqrvSsHlEDEXoVyaPc=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "ab82a9612aa45284d4adf69ee81871a389669a9e",
+        "rev": "19674872444bb3e0768249e724d99c8649c3bd78",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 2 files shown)


YaLTeR:
@sodiboo hi

feathecutie:
Sadly this is probably unrelated to the issues @sodiboo is facing, since they use a custom build process instead of the one supplied by this repo's `flake.nix`.
This PR simply fixes the build for people using this flake, not https://github.com/sodiboo/niri-flake

YaLTeR:
Well, I still want someone who knows nix to take a look

sodiboo:
> Sadly this is probably unrelated to the issues @sodiboo is facing, since they use a custom build process instead of the one supplied by this repo's `flake.nix`.
> This PR simply fixes the build for people using this flake, not https://github.com/sodiboo/niri-flake

It's worth noting that I was investigating this issue the other day. I got it down to a change made in Rust nightly `2024-05-18`, and was going to bisect rustc but didn't quite get around to doing that yet (I was going down the path of "why did this suddenly change??"), and as such this probably is relevant to me even though I mainly deal with that other repo. I'm also a niri contributor, and I *do* happen to make use of this repo's flake when working on niri (which is hard, when it doesn't build lol).

I'm exhausted right now, though, and I will probably take a look at this PR tomorrow.

flake.nix
```diff
@@ -83,7 +83,10 @@
         };
 
         cargoArtifacts = craneLib.buildDepsOnly craneArgs;
-        niri = craneLib.buildPackage (craneArgs // {inherit cargoArtifacts;});
+        niri = craneLib.buildPackage (craneArgs // {
+          inherit cargoArtifacts;
+          LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath craneArgs.runtimeDependencies; # Needed for tests to find libxkbcommon
```
sodiboo, line 88: Why is this being added here instead of to `craneArgs`, like with `LIBCLANG_PATH`?

feathecutie: I didn't add it in `craneArgs` because `craneArgs` is also used for creating `cargoArtifacts`, and it's technically not necessary there.
I could also just move it to `craneArgs` though, it shouldn't make any difference, I simply tried to keep everything as clean as possible

feathecutie: Also, if I put `LD_LIBRARY_PATH` into `craneArgs`, I'd have to turn `craneArgs` into a recursive attrset in order to access the `runtimeDependencies`

sodiboo: >  I'd have to turn `craneArgs` into a recursive attrset in order to access the `runtimeDependencies`
 
 `craneArgs` is in scope though. you can just move the line up without changing a single char on it

feathecutie: Huh, I didn't know that was possible. I rebased it now to use `craneArgs`, and also fixed the deprecation warning:

```
warning: `crane.lib.${system}` is deprecated. please use `(crane.mkLib nixpkgs.legacyPackages.${system})` instead
```

sodiboo (approved):
looks good to me! does what it says on the tin (fixes `nix build`) and now the diff is also really pretty :3. i'd ask to fix `cargo build` on NixOS in the same go but uhh, that's a whole 'nother can of worms. i tried. it's awful. i'll submit another PR sometime. this one could do with maybe a `nix flake update` and then let's merge it!

feathecutie:
Should be ready to merge now, but what do you mean by fixing `cargo build` on NixOS?
A simple `cargo build` command in the devShell seems to work fine for me, the main problem seems to be that the devShell doesn't correctly set up runtime dependencies for `cargo run`.
I was able to get `cargo run` to work too using the following diff:
```diff
diff --git a/flake.nix b/flake.nix
index f8c4b63..9b7562e 100644
--- a/flake.nix
+++ b/flake.nix
@@ -77,6 +77,15 @@
             xorg.libXcursor
             xorg.libXi
             libxkbcommon
+            pipewire
+            pango
+            glib
+            cairo
+            seatd
+            systemd
+            libinput
+            pixman
+            mesa # For libgbm
           ];
 
           LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
```
(I simply added these packages to `runtimeDependencies`).
It isn't really obvious why this would make any differences, as most of these packages are already part of `buildInputs`, and `buildInputs` also gets converted to `LD_LIBRARY_PATH`, same as `runtimeDependencies`.
However, when inspecting the resulting `LD_LIBRARY_PATH` environment variable in the devShell, you can see that (for whatever reason), the resulting path uses different derivation outputs for the derivations from `nativeBuildInputs` and `buildInputs` as compared to `runtimeDependencies` (as can be seen by the `-dev` suffix on some of the package paths).
I don't know if there's a more elegant solution to this problem, and I'd like to understand the underlying reason for this, but at least this seems to get `cargo run` working (for me).

feathecutie:
More experimentation:

Running `builtins.map toString packages.x86_64-linux.default.buildInputs` in a `nix repl` results in the following:

```nix
nix-repl> builtins.map toString packages.x86_64-linux.default.buildInputs
[
  "/nix/store/dhafvd6slhk8hz8g0hvqnqh5i9fmqma4-wayland-1.23.0-dev"
  "/nix/store/prf4i4plz0k43jz0ghyzlf7gwmi8rjvg-systemd-256.2-dev"
  "/nix/store/x4ai78x8i92qz2bxpvp9wzgl0k31jvw7-seatd-0.8.0-dev"
  "/nix/store/8pbh705kblpkhlg25lax7hcfh2zcv2ab-libxkbcommon-1.7.0-dev"
  "/nix/store/gj93mj1hks0xrsj20n7f3dfsdcd471s9-libinput-1.26.1-dev"
  "/nix/store/xgpdv93w3kkvp45ip1xym55z72crz7v0-mesa-24.1.4-dev"
  "/nix/store/zs4v7bsmxvblgv02nak0wcxpjncw80jf-fontconfig-2.15.0-dev"
  "/nix/store/kgmfgzb90h658xg0i7mxh9wgyx0nrqac-gcc-13.3.0-lib"
  "/nix/store/cw78ndjp827zan6hpdk41c45ynfwqrvk-pipewire-1.2.1-dev"
  "/nix/store/khwdaapk8l3kfy764j71cir1lafr5bpm-pango-1.52.2-dev"
]
```

This shows that, for whatever reason, the `buildInputs` attribute on the `niri` package evaluates to the `dev` output of the contained derivations (which seems to not contain the files that `cargo run` needs, I guess).
This is probably completely normal behaviour, so it might make sense to simply add the above packages to `runtimeDependencies` for now in order to fix `cargo run`.

YaLTeR:
Thanks; let's do the cargo run fix in a separate PR

sodiboo:
> I was investigating this issue the other day. I got it down to a change made in Rust nightly `2024-05-18`, and was going to bisect rustc but didn't quite get around to doing that yet

yoo turns out I don't need to do that anymore. it's almost certainly because of `rust-lld`. see blog post date:  https://blog.rust-lang.org/2024/05/17/enabling-rust-lld-on-linux.html

thanks to @getchoo for pointing me in the right direction for that (unbeknownst to him)

this breaks builds with Fenix, such as our package at the time:

 - https://github.com/nix-community/fenix/issues/159

Issue #563, opened 2024-07-30
VRR window rule


YaLTeR:
A new property makes VRR on an output enable on-demand (i.e. disabled unless enabled by a window rule):
```
output "DP-1" {
    variable-refresh-rate on-demand=true
}
```

The window rule:
```
window-rule {
    match app-id="^mpv$"
    variable-refresh-rate true
}
```

The behavior: when a surface matching the VRR window rule is visible anywhere on the output, enable on-demand VRR on the output. When no surfaces matching the VRR window rule are visible on an output, disable on-demand VRR on the output.

Alternatively, use a surface's primary output rather than "surface visible on output". For most cases these will be equivalent, so whichever is easier to implement.

Names are up for discussion but I'm fairly content with these ones.

YaLTeR:
Implemented in #586.

PR #586, opened 2024-08-11
feature: add on-demand vrr

my4ng:
Implements #563.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -324,9 +324,23 @@ pub struct Output {
     #[knuffel(child, unwrap(argument, str))]
     pub mode: Option<ConfiguredMode>,
     #[knuffel(child)]
-    pub variable_refresh_rate: bool,
+    pub variable_refresh_rate: Option<Vrr>,
     #[knuffel(child, default = DEFAULT_BACKGROUND_COLOR)]
     pub background_color: Color,
+}
+
+impl Output {
+    pub fn is_vrr_always_on(&self) -> bool {
+        self.variable_refresh_rate == Some(Vrr { on_demand: false })
+    }
+
+    pub fn is_vrr_on_demand(&self) -> bool {
+        self.variable_refresh_rate == Some(Vrr { on_demand: true })
+    }
+
+    pub fn is_vrr_always_off(&self) -> bool {
+        self.variable_refresh_rate.is_none()
+    }
 }
 
 impl Default for Output {
@@ -338,7 +352,7 @@ impl Default for Output {
             transform: Transform::Normal,
             position: None,
             mode: None,
-            variable_refresh_rate: false,
+            variable_refresh_rate: None,
             background_color: DEFAULT_BACKGROUND_COLOR,
         }
     }
@@ -350,6 +364,12 @@ pub struct Position {
     pub x: i32,
     #[knuffel(property)]
     pub y: i32,
+}
+
+#[derive(knuffel::Decode, Debug, Clone, PartialEq, Default)]
+pub struct Vrr {
+    #[knuffel(property, default = false)]
+    pub on_demand: bool,
 }
 
 // MIN and MAX generics are only used during parsing to check the value.
@@ -896,6 +916,8 @@ pub struct WindowRule {
     pub clip_to_geometry: Option<bool>,
     #[knuffel(child, unwrap(argument))]
     pub block_out_from: Option<BlockOutFrom>,
+    #[knuffel(child, unwrap(argument))]
+    pub variable_refresh_rate: Option<bool>,
 }
 
 // Remember to update the PartialEq impl when adding fields!
@@ -2705,7 +2727,7 @@ mod tests {
                 transform "flipped-90"
                 position x=10 y=20
                 mode "1920x1080@144"
-                variable-refresh-rate
+                variable-refresh-rate on-demand=true
                 background-color "rgba(25, 25, 102, 1.0)"
             }
 
@@ -2889,7 +2911,7 @@ mod tests {
                         height: 1080,
                         refresh: Some(144.),
                     }),
-                    variable_refresh_rate: true,
+                    variable_refresh_rate: Some(Vrr { on_demand: true }),
                     background_color: Color::from_rgba8_unpremul(25, 25, 102, 255),
                 }]),
                 layout: Layout {
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -434,7 +434,7 @@ pub struct VrrToSet {
         ),
     )]
     pub vrr: bool,
-    /// Only enable if there is a matched VRR window rule visible on the output.
+    /// Only enable when the output shows a window matching the variable-refresh-rate window rule.
     #[cfg_attr(feature = "clap", arg(long))]
     pub on_demand: bool,
 }
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -352,18 +352,11 @@ pub enum OutputAction {
         #[cfg_attr(feature = "clap", command(subcommand))]
         position: PositionToSet,
     },
-    /// Toggle variable refresh rate.
+    /// Set the variable refresh rate mode.
     Vrr {
-        /// Whether to enable variable refresh rate.
-        #[cfg_attr(
-            feature = "clap",
-            arg(
-                value_name = "ON|OFF",
-                action = clap::ArgAction::Set,
-                value_parser = clap::builder::BoolishValueParser::new(),
-            ),
-        )]
-        enable: bool,
+        /// Variable refresh rate mode to set.
+        #[cfg_attr(feature = "clap", command(flatten))]
+        vrr: VrrToSet,
     },
 }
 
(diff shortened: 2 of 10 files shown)


niri-config/src/lib.rs
```diff
@@ -323,8 +323,8 @@ pub struct Output {
     pub position: Option<Position>,
     #[knuffel(child, unwrap(argument, str))]
     pub mode: Option<ConfiguredMode>,
-    #[knuffel(child)]
-    pub variable_refresh_rate: bool,
+    #[knuffel(child, default)]
+    pub variable_refresh_rate: Vrr,
```
YaLTeR, line 327: I think it would be clearer, code- and type-wise, if this worked like FocusFollowsMouse. That is, `pub variable_refresh_rate: Option<Vrr>` where None means disabled, Some means enabled, then on_demand is a bool with default value false.

niri-ipc/src/lib.rs
```diff
@@ -425,6 +418,19 @@ pub struct ConfiguredPosition {
...
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub enum VrrToSet {
+    /// Always OFF.
+    AlwaysOff,
+    /// Always ON.
+    AlwaysOn,
+    /// Only ON when a surface matching a VRR window rule is visible on the output.
+    OnDemand,
+}
```
YaLTeR, line 432: I wonder if it's better to match the config here, i.e. instead of `niri msg output vrr always-on/always-off/on-demand` have it be:
- `niri msg output vrr off`
- `niri msg output vrr on`
- `niri msg output vrr on --on-demand`

src/backend/mod.rs
```diff
@@ -153,6 +153,13 @@ impl Backend {
         }
     }
 
+    pub fn set_output_vrr(&mut self, niri: &mut Niri, output: &Output, enable_vrr: bool) {
+        match self {
+            Backend::Tty(tty) => tty.set_output_vrr(niri, output, enable_vrr),
+            Backend::Winit(_) => todo!(),
```
YaLTeR, line 159:
```suggestion
            Backend::Winit(_) => (),
```

niri-config/src/lib.rs
```diff
@@ -352,6 +352,32 @@ pub struct Position {
...
 }
 
+#[derive(knuffel::Decode, Debug, Clone, PartialEq, Default)]
+pub struct Vrr {
+    #[knuffel(property, default = Some(false))]
+    on_demand: Option<bool>,
+}
+
+impl Vrr {
+    pub fn always_on(&self) -> bool {
```
YaLTeR, line 362: The convention is to name checker methods starting from `is_`

src/backend/tty.rs
```diff
@@ -1707,7 +1738,7 @@ impl Tty {
                 };
 
                 let change_mode = surface.compositor.pending_mode() != mode;
-                let change_vrr = surface.vrr_enabled != config.variable_refresh_rate;
+                let change_vrr = surface.vrr_enabled != config.variable_refresh_rate.always_on();
```
YaLTeR, line 1741: This seems like it should check whether some surface is enabling VRR currently on this output.

src/niri.rs
```diff
@@ -1200,15 +1201,71 @@ impl State {
...
+                        .invert()
+                        .transform_size(output_size),
+                );
+                let current = workspace.windows().any(|mapped| {
+                    mapped.rules().variable_refresh_rate == Some(true)
+                        && mapped
+                            .window
+                            .geometry()
+                            .to_physical_precise_round(output.current_scale().fractional_scale())
+                            .overlaps(output_geo)
```
YaLTeR, line 1259: Window geometry is relative to the window buffer, not a global coordinate system, so this check doesn't work properly.

Either way, I feel like this should check if any surface of the window has this output as its primary output. That will deal with subsurfaces outside the window, and when we get floating, it will also deal with cases like a few pixels of a window overlapping one output (in which case we don't want to enable VRR there). It already works for things like frame callbacks, I think it would be appropriate if the VRR window rule used the same logic.

niri-config/src/lib.rs
my4ng: That was my initial implementation but the downside is no help function, which is why I wrapped it around a new type.

src/backend/mod.rs
my4ng: I blame autocomplete template :p

src/backend/tty.rs
my4ng: Mhmm I am not sure the best way to do that. This just makes it false until the next refresh.

niri-ipc/src/lib.rs
my4ng: I am not sure if this is possible under clap.

src/niri.rs
my4ng: I will have to have a deeper look into the frame callback logic. One issue is that the window may not be in the active workspace and secondly it could be not visible.

niri-config/src/lib.rs
```diff
@@ -324,11 +324,31 @@ pub struct Output {
...
+    pub variable_refresh_rate: Option<Vrr>,
     #[knuffel(child, default = DEFAULT_BACKGROUND_COLOR)]
     pub background_color: Color,
 }
 
+pub trait VrrOutput {
+    fn is_always_on(&self) -> bool;
+    fn is_on_demand(&self) -> bool;
+    fn is_always_off(&self) -> bool;
+}
```
YaLTeR, line 336: I think it's a better idea to have these methods on `impl Output` directly, and have them contain the word "vrr", i.e. `is_vrr_always_on()`

src/backend/tty.rs
```diff
@@ -1661,6 +1659,39 @@ impl Tty {
...
+                        error!("surface pending connectors is empty");
+                        return;
+                    };
+                    let Some(connector) = device.drm_scanner.connectors().get(&connector) else {
+                        error!("missing enabled connector in drm_scanner");
+                        return;
+                    };
+                    let output_state = niri.output_state.get_mut(output).unwrap();
+
+                    try_to_change_vrr(
```
YaLTeR, line 1680: If the surface already has VRR enabled/disabled as we want, then this should just return instead of trying.

src/niri.rs
```diff
@@ -3113,6 +3121,49 @@ impl Niri {
...
+                    let mut visible = false;
+                    mapped.window.with_surfaces(|_, states| {
+                        if !visible {
+                            let surface_primary_scanout_output = states
+                                .data_map
+                                .get_or_insert_threadsafe(Mutex::<PrimaryScanoutOutput>::default);
+                            if surface_primary_scanout_output
+                                .lock()
+                                .unwrap()
+                                .current_output()
```
YaLTeR, line 3149: There's a helper for this: `surface_primary_scanout_output(surface, states)`

src/niri.rs
```diff
@@ -3113,6 +3121,49 @@ impl Niri {
...
+            .config
+            .borrow()
+            .outputs
+            .find(&output.name())
+            .map(|output| output.is_on_demand())
+        else {
+            warn!("error getting output config for {}", output.name());
+            return;
+        };
+        if on_demand {
```
YaLTeR, line 3136: Would be better to do `if !on_demand { return; }`

src/backend/tty.rs
```diff
@@ -1661,6 +1659,39 @@ impl Tty {
...
+                    let Some(connector) =
+                        surface.compositor.pending_connectors().into_iter().next()
+                    else {
+                        error!("surface pending connectors is empty");
+                        return;
+                    };
+                    let Some(connector) = device.drm_scanner.connectors().get(&connector) else {
+                        error!("missing enabled connector in drm_scanner");
+                        return;
+                    };
```
YaLTeR, line 1677: Would it make sense to just store the connector on the `Surface`? I forget already how the code is structured.

src/backend/tty.rs
YaLTeR: Maybe something like, put a bool into OutputState that indicates whether, to the best of our knowledge, VRR should be enabled? Update it in `refresh_on_demand_vrr()`.

my4ng: The problem is that `refresh_on_demand_vrr` which is called in redraw does not occur before this function, so this won't do any good. I still do not see what the issue is with how it is now, where it just defaults to false. If there is an enabling surface visible, then the display will to VRR upon next redraw anyway.

YaLTeR: I'm just thinking of a case like changing the output scale or position while there's a surface forcing VRR. Currently this will also toggle the VRR KMS property off and on. Not a big deal I guess, but also when this happens, the output state is not erased, so if you put the last frame's VRR flag there, it should work?

my4ng: Uh yes I forgot about that case. I was thinking when switching from always on to on demand. I will add a flag there in the output state.

my4ng: I don't think a simple VRR flag is enough. The window rule config could have changed at the same time, and so what enabled the VRR previously may not apply anymore. There are also edge cases with the other transitions. Overall, I don't think it is worth the trouble to implement it.

YaLTeR:
Did some minor changes + wiki. After a quick test, this appears to work!

YaLTeR (approved):

YaLTeR:
Thanks!

PR #628, opened 2024-08-24
fix `cargo run` on nixos

sodiboo:
this boils down to adding some extra dependencies to the shell environment. they're also inherited from craneArgs because the ones from the package are actually transformed into the WRONG outputs of the packages. also refactors to use craneLib.devShell because it's somewhat cleaner.

---

the formatting changes around L45-L47 are caused by `nix fmt`. this is what `alejandra` wants to format it like, [which is already the formatter set to be used for the nix flake](https://github.com/YaLTeR/niri/blob/b2c7d3ad401574a3db01ed20acd67ebc8d331cbc/flake.nix#L89)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1722960479,
-        "narHash": "sha256-NhCkJJQhD5GUib8zN9JrmYGMwt4lCRp6ZVNzIiYCl0Y=",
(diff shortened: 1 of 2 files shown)


sodiboo:
This PR is a follow-up to https://github.com/YaLTeR/niri/pull/579#pullrequestreview-2236763014

YaLTeR:
Thanks

Issue #198, opened 2024-02-15
Implement keyboard-shortcuts-inhibit-unstable-v1


YaLTeR:
https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1

Looks like this is the protocol used for shortcut inhibiting.

Mutter impl: https://gitlab.gnome.org/GNOME/mutter/-/blob/8677e36c4c3c457a974c71b36fd2bfcbc9e53f6c/src/wayland/meta-wayland-inhibit-shortcuts.c

PR #630, opened 2024-08-26
implement `keyboard-shortcuts-inhibit` and `wlr-virtual-pointer`

sodiboo:
My motivation for this is the use of KVM switching software, in particular i noticed that [`lan-mouse`](https://github.com/feschber/lan-mouse) requires them (hence my branch name).

Two things need to be done:

- Please someone check that what i am doing with pointer axis frames is sane. This was painful. It works. Smithay docs refer to a nonexistent function so i had to check 2-year-old git blames to find out what replaced it. We just ignore the axis source if it's done before an arbitrary other thing?? Help.
- Need to implement a keyboard-shortcuts-inhibit release bind. The current implementation already doesn't inhibit the `ChangeVt` binds ever.

You can try the functionality of both of these protocols using `lan-mouse --test-emulation` (will cause the pointer to move in a lemniscate) or `lan-mouse --test-capture` (WARNING: YOUR FOCUS WILL BE UNRECOVERABLE FROM WITHIN NIRI. use `Ctrl+Alt+F2` and then `pkill -f test-capture` to recover. the `Ctrl+Alt+{F1..F12}` binds are the `ChangeVt` binds i've mentioned already)

Another thing i've noticed is that virtual-keyboard inputs aren't ever handled by the compositor (i.e. `Super+Left` is passed to whatever client, instead of focusing the column to the left)? This is irrelevant to this PR, but something i noticed while using lan-mouse. This is probably a Smithay issue? Since it's uhh mostly implemented in Smithay.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1191,6 +1191,7 @@ pub struct Bind {
     pub repeat: bool,
     pub cooldown: Option<Duration>,
     pub allow_when_locked: bool,
+    pub allow_inhibiting: bool,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
@@ -3015,6 +3016,7 @@ where
         let mut cooldown = None;
         let mut allow_when_locked = false;
         let mut allow_when_locked_node = None;
+        let mut allow_inhibiting = true;
         for (name, val) in &node.properties {
             match &***name {
                 "repeat" => {
@@ -3028,6 +3030,9 @@ where
                 "allow-when-locked" => {
                     allow_when_locked = knuffel::traits::DecodeScalar::decode(val, ctx)?;
                     allow_when_locked_node = Some(name);
+                }
+                "allow-inhibiting" => {
+                    allow_inhibiting = knuffel::traits::DecodeScalar::decode(val, ctx)?;
                 }
                 name_str => {
                     ctx.emit_error(DecodeError::unexpected(
@@ -3050,6 +3055,7 @@ where
             repeat: true,
             cooldown: None,
             allow_when_locked: false,
+            allow_inhibiting: true,
         };
 
         if let Some(child) = children.next() {
@@ -3078,6 +3084,7 @@ where
                         repeat,
                         cooldown,
                         allow_when_locked,
+                        allow_inhibiting,
                     })
                 }
                 Err(e) => {
@@ -3470,7 +3477,7 @@ mod tests {
                 Mod+Comma { consume-window-into-column; }
                 Mod+1 { focus-workspace 1; }
                 Mod+Shift+1 { focus-workspace "workspace-1"; }
-                Mod+Shift+E { quit skip-confirmation=true; }
+                Mod+Shift+E allow-inhibiting=false { quit skip-confirmation=true; }
                 Mod+WheelScrollDown cooldown-ms=150 { focus-workspace-down; }
             }
 
@@ -3788,6 +3795,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: true,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3798,6 +3806,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3808,6 +3817,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3818,6 +3828,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3828,6 +3839,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3838,6 +3850,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3850,6 +3863,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                     Bind {
                         key: Key {
@@ -3860,6 +3874,7 @@ mod tests {
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
+                        allow_inhibiting: false,
                     },
                     Bind {
                         key: Key {
@@ -3870,6 +3885,7 @@ mod tests {
                         repeat: true,
                         cooldown: Some(Duration::from_millis(150)),
                         allow_when_locked: false,
+                        allow_inhibiting: true,
                     },
                 ]),
                 switch_events: SwitchBinds {
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1279,6 +1279,7 @@ pub enum Action {
         id: u64,
         write_to_disk: bool,
     },
+    ToggleKeyboardShortcutsInhibit,
     CloseWindow,
     #[knuffel(skip)]
     CloseWindowById(u64),
@@ -3078,6 +3079,12 @@ where
                         }
                     }
 
+                    // The toggle-inhibit action must always be uninhibitable.
+                    // Otherwise, it would be impossible to trigger it.
+                    if matches!(action, Action::ToggleKeyboardShortcutsInhibit) {
+                        allow_inhibiting = false;
+                    }
+
                     Ok(Self {
                         key,
                         action,
@@ -3470,6 +3477,8 @@ mod tests {
             }
 
             binds {
+                Mod+Escape { toggle-keyboard-shortcuts-inhibit; }
+                Mod+Shift+Escape allow-inhibiting=true { toggle-keyboard-shortcuts-inhibit; }
                 Mod+T allow-when-locked=true { spawn "alacritty"; }
                 Mod+Q { close-window; }
                 Mod+Shift+H { focus-monitor-left; }
@@ -3786,6 +3795,28 @@ mod tests {
                     },
                 ],
                 binds: Binds(vec![
+                    Bind {
+                        key: Key {
+                            trigger: Trigger::Keysym(Keysym::Escape),
+                            modifiers: Modifiers::COMPOSITOR,
+                        },
+                        action: Action::ToggleKeyboardShortcutsInhibit,
+                        repeat: true,
+                        cooldown: None,
+                        allow_when_locked: false,
+                        allow_inhibiting: false,
+                    },
+                    Bind {
+                        key: Key {
+                            trigger: Trigger::Keysym(Keysym::Escape),
+                            modifiers: Modifiers::COMPOSITOR | Modifiers::SHIFT,
+                        },
+                        action: Action::ToggleKeyboardShortcutsInhibit,
+                        repeat: true,
+                        cooldown: None,
+                        allow_when_locked: false,
+                        allow_inhibiting: false,
+                    },
                     Bind {
                         key: Key {
                             trigger: Trigger::Keysym(Keysym::t),
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -536,6 +536,16 @@ binds {
     Ctrl+Print { screenshot-screen; }
     Alt+Print { screenshot-window; }
 
+    // Applications such as remote-desktop clients and software KVM switches may
+    // request that niri stops processing the keyboard shortcuts defined here
+    // so they may, for example, forward the key presses as-is to a remote machine.
+    // It's a good idea to bind an escape hatch to toggle the inhibitor,
+    // so a buggy application can't hold your session hostage.
+    //
+    // The allow-inhibiting=false property can be applied to other binds as well,
+    // which ensures niri always processes them, even when an inhibitor is active.
+    Mod+Escape allow-inhibiting=false { toggle-keyboard-shortcuts-inhibit; }
+
     // The quit action will show a confirmation dialog to avoid accidental exits.
     Mod+Shift+E { quit; }
     Ctrl+Alt+Delete { quit; }
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -11,7 +11,7 @@ use std::time::Duration;
 
 use smithay::backend::allocator::dmabuf::Dmabuf;
 use smithay::backend::drm::DrmNode;
-use smithay::backend::input::TabletToolDescriptor;
+use smithay::backend::input::{InputEvent, TabletToolDescriptor};
 use smithay::desktop::{PopupKind, PopupManager};
 use smithay::input::pointer::{
     CursorIcon, CursorImageStatus, CursorImageSurfaceData, PointerHandle,
@@ -79,7 +79,11 @@ use crate::protocols::gamma_control::{GammaControlHandler, GammaControlManagerSt
 use crate::protocols::mutter_x11_interop::MutterX11InteropHandler;
 use crate::protocols::output_management::{OutputManagementHandler, OutputManagementManagerState};
 use crate::protocols::screencopy::{Screencopy, ScreencopyHandler, ScreencopyManagerState};
-use crate::protocols::virtual_pointer::{VirtualPointerHandler, VirtualPointerManagerState};
+use crate::protocols::virtual_pointer::{
+    VirtualPointerAxisEvent, VirtualPointerButtonEvent, VirtualPointerHandler,
+    VirtualPointerInputBackend, VirtualPointerManagerState, VirtualPointerMotionAbsoluteEvent,
+    VirtualPointerMotionEvent,
+};
 use crate::utils::{output_size, send_scale_transform, with_toplevel_role};
 use crate::{
     delegate_foreign_toplevel, delegate_gamma_control, delegate_mutter_x11_interop,
@@ -591,6 +595,24 @@ delegate_screencopy!(State);
 impl VirtualPointerHandler for State {
     fn virtual_pointer_manager_state(&mut self) -> &mut VirtualPointerManagerState {
         &mut self.niri.virtual_pointer_state
+    }
+
+    fn on_virtual_pointer_motion(&mut self, event: VirtualPointerMotionEvent) {
+        self.process_input_event(InputEvent::<VirtualPointerInputBackend>::PointerMotion { event });
+    }
+
+    fn on_virtual_pointer_motion_absolute(&mut self, event: VirtualPointerMotionAbsoluteEvent) {
+        self.process_input_event(
+            InputEvent::<VirtualPointerInputBackend>::PointerMotionAbsolute { event },
+        );
+    }
+
+    fn on_virtual_pointer_button(&mut self, event: VirtualPointerButtonEvent) {
+        self.process_input_event(InputEvent::<VirtualPointerInputBackend>::PointerButton { event });
+    }
+
+    fn on_virtual_pointer_axis(&mut self, event: VirtualPointerAxisEvent) {
+        self.process_input_event(InputEvent::<VirtualPointerInputBackend>::PointerAxis { event });
     }
 }
 delegate_virtual_pointer!(State);
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -254,11 +254,17 @@ impl KeyboardShortcutsInhibitHandler for State {
     }
 
     fn new_inhibitor(&mut self, inhibitor: KeyboardShortcutsInhibitor) {
-        info!("new inhibitor: {inhibitor:?}");
+        // FIXME: show a confirmation dialog with a "remember for this application" kind of toggle.
+        inhibitor.activate();
+        self.niri
+            .keyboard_shortcuts_inhibiting_surfaces
+            .insert(inhibitor.wl_surface().clone(), inhibitor);
     }
 
     fn inhibitor_destroyed(&mut self, inhibitor: KeyboardShortcutsInhibitor) {
-        info!("inhibitor destroyed: {inhibitor:?}");
+        self.niri
+            .keyboard_shortcuts_inhibiting_surfaces
+            .remove(&inhibitor.wl_surface().clone());
     }
 }
 
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -35,6 +35,9 @@ use smithay::wayland::fractional_scale::FractionalScaleHandler;
 use smithay::wayland::idle_inhibit::IdleInhibitHandler;
 use smithay::wayland::idle_notify::{IdleNotifierHandler, IdleNotifierState};
 use smithay::wayland::input_method::{InputMethodHandler, PopupSurface};
+use smithay::wayland::keyboard_shortcuts_inhibit::{
+    KeyboardShortcutsInhibitHandler, KeyboardShortcutsInhibitState, KeyboardShortcutsInhibitor,
+};
 use smithay::wayland::output::OutputHandler;
 use smithay::wayland::pointer_constraints::{with_pointer_constraint, PointerConstraintsHandler};
 use smithay::wayland::security_context::{
@@ -59,11 +62,12 @@ use smithay::wayland::xdg_activation::{
 use smithay::{
     delegate_cursor_shape, delegate_data_control, delegate_data_device, delegate_dmabuf,
     delegate_drm_lease, delegate_fractional_scale, delegate_idle_inhibit, delegate_idle_notify,
-    delegate_input_method_manager, delegate_output, delegate_pointer_constraints,
-    delegate_pointer_gestures, delegate_presentation, delegate_primary_selection,
-    delegate_relative_pointer, delegate_seat, delegate_security_context, delegate_session_lock,
-    delegate_single_pixel_buffer, delegate_tablet_manager, delegate_text_input_manager,
-    delegate_viewporter, delegate_virtual_keyboard_manager, delegate_xdg_activation,
+    delegate_input_method_manager, delegate_keyboard_shortcuts_inhibit, delegate_output,
+    delegate_pointer_constraints, delegate_pointer_gestures, delegate_presentation,
+    delegate_primary_selection, delegate_relative_pointer, delegate_seat,
+    delegate_security_context, delegate_session_lock, delegate_single_pixel_buffer,
+    delegate_tablet_manager, delegate_text_input_manager, delegate_viewporter,
+    delegate_virtual_keyboard_manager, delegate_xdg_activation,
 };
 
 pub use crate::handlers::xdg_shell::KdeDecorationsModeState;
@@ -75,10 +79,11 @@ use crate::protocols::gamma_control::{GammaControlHandler, GammaControlManagerSt
 use crate::protocols::mutter_x11_interop::MutterX11InteropHandler;
 use crate::protocols::output_management::{OutputManagementHandler, OutputManagementManagerState};
 use crate::protocols::screencopy::{Screencopy, ScreencopyHandler, ScreencopyManagerState};
+use crate::protocols::virtual_pointer::{VirtualPointerHandler, VirtualPointerManagerState};
 use crate::utils::{output_size, send_scale_transform, with_toplevel_role};
 use crate::{
     delegate_foreign_toplevel, delegate_gamma_control, delegate_mutter_x11_interop,
-    delegate_output_management, delegate_screencopy,
+    delegate_output_management, delegate_screencopy, delegate_virtual_pointer,
 };
 
 pub const XDG_ACTIVATION_TOKEN_TIMEOUT: Duration = Duration::from_secs(10);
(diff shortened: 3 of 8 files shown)


src/input/mod.rs
```diff
@@ -2221,6 +2247,18 @@ fn should_intercept_key(
         }
     }
 
+    if is_inhibiting_shortcuts
```
YaLTeR, line 2250: I have a feeling that this should interact in a more complex manner with the `match` below because starting to inhibit while some modifier is held (and in `suppressed_keys`) then releasing the modifier will currently leave it in `suppressed_keys`, which does not seem right.

This behavior should also be tested in the unit tests below.

src/input/mod.rs
```diff
@@ -2176,6 +2201,7 @@ fn should_intercept_key(
     mods: ModifiersState,
     screenshot_ui: &ScreenshotUi,
     disable_power_key_handling: bool,
+    is_inhibiting_shortcuts: bool,
```
YaLTeR, line 2204: I wonder, do other compositors have any hard-coded key to escape the input inhibition?

YaLTeR (commented):
> Smithay docs refer to a nonexistent function so i had to check 2-year-old git blames to find out what replaced it. We just ignore the axis source if it's done before an arbitrary other thing??

I haven't looked in detail but if it involves axis v120 events then yeah it's like this pretty much because of legacy, v120 events replaced the old ones: https://wayland.freedesktop.org/libinput/doc/latest/wheel-api.html

src/input/mod.rs
sodiboo: Not necessarily that i know of, but [the protocol description](https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1#zwp_keyboard_shortcuts_inhibitor_v1) mentions that it is fully compliant for the compositor to keep such a keyboard shortcut active even when other shortcuts are generally inhibited. I'm imagining this shouldn't be hardcoded, but rather configurable. Like any other bind. I feel i'd want it on my system to be something like `Mod+Ctrl+Shift+Escape` (and maybe just `Super+Escape` for the application-specific release. e.g. lan-mouse has its own built-in release that doesn't need compositor support, which is very useful already, and for such well-behaved clients i don't even need a compositor release)

src/input/mod.rs
sodiboo: I've since updated the handling in a way that i think works with `supressed_keys` properly ([and i've commented the behaviour](https://github.com/YaLTeR/niri/pull/630/files#diff-306fabc3ade1a58278e40c0c99da8c0d45c5f2b7e7e43c3ead13d64099194aa2R2401-R2408)), but to be honest, thinking about the order of events kinda makes my head hurt and i'm dreading writing the requested tests. I'm also not entirely confident that my intuition of how it should behave exactly aligns with what you think it should do? Do you wanna take a shot at writing tests for this? :pleading_face::point_right::point_left:

sodiboo:
I've implemented the `allow-inhibiting` property for binds, and the `toggle-keyboard-shortcuts-inhibit` action, which weren't present in the initial draft of this PR. This is basically the only thing that was missing from making this PR "complete". `keyboard-shortcuts-inhibit` is now fully implemented (should probably have some more tests).

I've also made sure you can't bind `toggle-keyboard-shortcuts-inhibit` but allow that bind to be inhibited.  [the `allow-inhibiting` property is ignored for this action](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/niri-config/src/lib.rs#L2656-L2660)

`wlr-virtual-pointer` worked flawlessly right off the bat for my use case, and you wouldn't be able to tell from using, but the underlying code is definitely janky!!! I'd like someone to verify that [this is how the `Axis*` events are suppposed to work](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L448-L526). Basically if we get more than one timestamp in a single frame, all but the first are ignored. And... `AxisDiscrete` overrides the values set by `Axis`? It feels very dirty, but judging by the `AxisFrame` struct in Smithay, it feels like the protocol should not really be allowing any of these requests to be sent twice in one frame. It feels very jank, and i know little about how axis events actually work, i just plugged together the parts of Smithay that "looked" about right to me.

The uhh... `AxisSource` request is optional (i think?) but `PointerAxisEvent::source()` does not allow a `None` value, so [this is definitely wrong](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L227-L232). The `AxisFrame` struct cannot be constructed without a timestamp, so [for the `AxisSource` event, which doesn't include a timestamp](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L487), it's just [dropped because no `AxisFrame` can be started](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L71); a different `Axis*` event must be sent first for the `AxisSource` to amend it.

resources/default-config.kdl
```diff
@@ -459,6 +459,16 @@ binds {
...
 
+    // Applications such as remote-desktop clients and software KVM switches may
+    // request that niri stops processing the keyboard shortcuts defined here
+    // so they may, for example, forward the key presses as-is to a remote machine.
+    // It's a good idea to bind an escape hatch to toggle the inhibitor,
+    // so a buggy application can't hold your session hostage.
+    //
+    // The allow-inhibiting=false property can be applied to other binds as well,
+    // which ensures niri always processes them, even when an inhibitor is active.
+    Mod+Shift+Escape allow-inhibiting=false { toggle-keyboard-shortcuts-inhibit; }
```
YaLTeR, line 470: I think we should bind this to `Mod+Escape` by default because Super+Escape is the default in GNOME: https://gitlab.gnome.org/GNOME/mutter/-/blob/d7d92c68bd5c76525752feab65c44b3157deb7f1/data/org.gnome.mutter.wayland.gschema.xml.in#L58-61

src/handlers/mod.rs
```diff
@@ -183,7 +192,27 @@ impl InputMethodHandler for State {
     }
 }
 
+impl KeyboardShortcutsInhibitHandler for State {
+    fn keyboard_shortcuts_inhibit_state(&mut self) -> &mut KeyboardShortcutsInhibitState {
+        &mut self.niri.keyboard_shortcuts_inhibit_state
+    }
+
+    fn new_inhibitor(&mut self, inhibitor: KeyboardShortcutsInhibitor) {
+        inhibitor.activate();
```
YaLTeR, line 201:
```suggestion
        // FIXME: show a confirmation dialog with a "remember for this application" kind of toggle.
        inhibitor.activate();
```

src/input/mod.rs
```diff
@@ -287,6 +288,28 @@ impl State {
...
     }
 
+    fn get_focused_keyboard_shortcuts_inhibitor(&self) -> Option<&KeyboardShortcutsInhibitor> {
+        if let KeyboardFocus::LayerShell { surface }
+        | KeyboardFocus::LockScreen {
+            surface: Some(surface),
+        }
+        | KeyboardFocus::Layout {
+            surface: Some(surface),
+        } = &self.niri.keyboard_focus
```
YaLTeR, line 298: If only there was a convenient `KeyboardFocus::surface()` method that did exactly this :p

niri-config/src/lib.rs
```diff
@@ -2983,14 +2997,16 @@ mod tests {
...
+                Mod+Shift+Escape allow-inhibiting=true { toggle-keyboard-shortcuts-inhibit; }
                 Mod+T allow-when-locked=true { spawn "alacritty"; }
                 Mod+Q { close-window; }
                 Mod+Shift+H { focus-monitor-left; }
                 Mod+Ctrl+Shift+L { move-window-to-monitor-right; }
                 Mod+Comma { consume-window-into-column; }
                 Mod+1 { focus-workspace 1; }
                 Mod+Shift+1 { focus-workspace "workspace-1"; }
-                Mod+Shift+E { quit skip-confirmation=true; }
+                Mod+Shift+E allow-inhibiting=false { quit skip-confirmation=true; }
```
YaLTeR, line 3009: I wonder if `allow-inhibiting=false` is a bit hard to wrap one's head around? Maybe something along `prevent-inhibiting=true` would be better, human config file reading and writing wise?

src/input/mod.rs
YaLTeR: It's configurable in mutter it seems

src/input/mod.rs
```diff
@@ -2344,17 +2380,29 @@ fn should_intercept_key(
                     repeat: true,
                     cooldown: None,
                     allow_when_locked: false,
+                    // The screenshot UI owns the focus anyays, so this doesn't really matter?
+                    // But logically, nothing can inhibit its actions. Only opening it can be
+                    // inhibited.
```
YaLTeR, line 2385:
```suggestion
                    // The screenshot UI owns the focus anyway, so this doesn't really matter.
                    // But logically, nothing can inhibit its actions. Only opening it can be
                    // inhibited.
```
Also yeah, either way it's better to safeguard against errors by passing false

src/input/mod.rs
```diff
@@ -2344,17 +2380,29 @@ fn should_intercept_key(
...
+            } else {
+                suppressed_keys.insert(key_code);
+                FilterResult::Intercept(Some(bind))
+            }
         }
         (_, false) => {
+            // By this point, we know that the key was supressed on press. Even if we're inhibiting
+            // shortcuts, we should still suppress the release. But we don't need to check for
+            // shortcuts inhibition here, because if it wasn't inhibited on press, it wouldn't be
+            // suppressed, so it would already have been forwarded at the start of this function.
```
YaLTeR, line 2405:
```suggestion
            // shortcuts inhibition here, because if it was inhibited on press (forwarded to the client), it wouldn't be
            // suppressed, so it would already have been forwarded at the start of this function.
```
Is this right?

src/protocols/virtual_pointer.rs
```diff
@@ -0,0 +1,554 @@
...
+
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct VirtualPointer {
+    pointer: ZwlrVirtualPointerV1,
+}
+
+#[derive(Debug)]
+pub struct VirtualPointerUserData {
+    seat: Option<WlSeat>,
+    output: Option<WlOutput>,
```
YaLTeR, line 44: What is this supposed to do?

src/protocols/virtual_pointer.rs
```diff
@@ -0,0 +1,554 @@
...
+                    state,
+                };
+                handler.on_virtual_pointer_button(event);
+            }
+            zwlr_virtual_pointer_v1::Request::Axis { time, axis, value } => {
+                let axis = match axis {
+                    WEnum::Value(wl_pointer::Axis::VerticalScroll) => Axis::Vertical,
+                    WEnum::Value(wl_pointer::Axis::HorizontalScroll) => Axis::Horizontal,
+                    _ => {
+                        warn!("Axis: invalid axis");
```
YaLTeR, line 453:
```suggestion
                        warn!("virtual pointer Axis: invalid axis");
```
and others similarly I guess

src/handlers/mod.rs
```diff
@@ -448,6 +477,32 @@ impl ScreencopyHandler for State {
...
+
+    fn on_virtual_pointer_button(&mut self, event: VirtualPointerButtonEvent) {
+        self.process_input_event(InputEvent::<VirtualPointerInputBackend>::PointerButton { event });
+    }
+
+    fn on_virtual_pointer_axis(&mut self, event: VirtualPointerAxisEvent) {
+        self.process_input_event(InputEvent::<VirtualPointerInputBackend>::PointerAxis { event });
+    }
+}
+
```
YaLTeR, line 503:
```suggestion
```

src/input/mod.rs
YaLTeR: Seems fine to me. For the tests, just add a few cases with the shortcuts inhibited = true. For pressing and releasing allow-inhibit, !allow-inhibit, which also check suppress keys. Basically check that a shortcut that would return Intercept() returns instead Forward if inhibited = true. And maybe also something which first does a press with inhibited = true, then release with inhibited = false in case that logic would change later.

YaLTeR (commented):
> I'd like someone to verify that [this is how the Axis* events are suppposed to work](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L448-L526). Basically if we get more than one timestamp in a single frame, all but the first are ignored. And... AxisDiscrete overrides the values set by Axis? It feels very dirty, but judging by the AxisFrame struct in Smithay, it feels like the protocol should not really be allowing any of these requests to be sent twice in one frame. It feels very jank, and i know little about how axis events actually work, i just plugged together the parts of Smithay that "looked" about right to me.
>
> The uhh... AxisSource request is optional (i think?) but PointerAxisEvent::source() does not allow a None value, so [this is definitely wrong](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L227-L232). The AxisFrame struct cannot be constructed without a timestamp, so [for the AxisSource event, which doesn't include a timestamp](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L487), it's just [dropped because no AxisFrame can be started](https://github.com/YaLTeR/niri/blob/62982c44ea84dd21374fe45f940e6cb14cc5cc9b/src/protocols/virtual_pointer.rs#L71); a different Axis* event must be sent first for the AxisSource to amend it.

Yeah idk tbh, I would just test with several wlr-virtual-pointer clients to see that they work. And if there's something weird, check in WAYLAND_DEBUG=1 what events they send.

(btw, what wlr-virtual-pointer clients are there apart from lan-mouse? I'll need a few to test this I guess)

sodiboo:
> (btw, what wlr-virtual-pointer clients are there apart from lan-mouse? I'll need a few to test this I guess)

just about any KVM or RDP software. Something like [deskflow](https://github.com/deskflow/deskflow) or [input leap](https://github.com/input-leap/input-leap) should do the same thing as lan-mouse. I also noticed that [WayVNC](https://github.com/any1/wayvnc) works just fine with niri as of this PR, as it has several capture protocols including `wlr-screencopy`, so you can connect to it from `wlvncc`.

Most of these should probably also be using keyboard shortcuts inhibit? Unsure. I know `lan-mouse` does, at least.

src/input/mod.rs
sodiboo: yeah. that seems correct. i think i wrote that thinking "if it wasn't intercepted" (== YES inhibited). or thinking that the keypress itself is inhhibited; but what that really means here is that the compositor's handling of the keypress is inhibited.

src/protocols/virtual_pointer.rs
sodiboo: Uh,, it stores the parameters the virtual pointer was created with? I don't think i used it for anything (and looking at references, i just noticed that i kept in `#[allow(dead_code)]` for the getters); but this is like, "generic protocol code" that could (should?) be part of Smithay or copied to other compositors, so i figured it would be weird to not keep it in.

Something we *should* be doing, though, is properly handling absolute pointer events when `output` is `Some(_)`, which i don't think i did. This would just happen to work by chance on single-monitor layouts, but i think on multi-monitor it would look janky without special care taken (as is done for tablets' and touchscreens' `map-to-output`)

sodiboo: > i just noticed that i kept in #[allow(dead_code)] for the getters

fixed this now by making them pub instead, which also silences the warnings. since i never ended up using them, ig i forgot that i should've made them pub. lol.

YaLTeR: I guess it's for absolute motion, yeah. How hard would that be to implement?

src/protocols/virtual_pointer.rs
YaLTeR: ah, okay, cool, then not necessary

src/protocols/virtual_pointer.rs
sodiboo: From a glance at the input handling, it looks like it would be nontrivial to make this track the correct output. We could fairly easily just bypass the existing input handling; but I think it'd be best to have some trait we implement on each input backend to allow us to query further device information about each event, like which output this absolute position event logically refers to, if any. This should apply equally to all devices, and would probably imply the ability to configure individual devices in the config, and virtual pointers are just one kind of device that is configured via wlr-virtual-pointer instead.

This is... maybe too much just for this PR? should I open another one? Alternatively, maybe Smithay should be able to handle providing an output for an absolute motion event? At the very least, that Wayland backend I've been working on does require the ability to specify which output a motion event belongs to, if we're ever gonna hope to see it use multiple outputs mapped as multiple toplevels. That could make more sense to implement in Smithay than in niri.

Likewise, it makes a lot of sense for individual keyboard devices to carry their xkb states and keymaps, for e.g. virtual keyboard, which currently Smithay just sneaks past the compositor and directly into its surface tree, but ideally they should go through the normal input backend system, and propagate their keymaps properly. The same goes for my Wayland backend, which can have multiple keyboards in different outer seats with different layouts. A lot of that could/should make its way into Smithay, while at the same time it wouldn't really be too difficult to implement such extension traits in niri.

src/input/mod.rs
sodiboo: yeah. if only.

src/protocols/virtual_pointer.rs
YaLTeR: There's already a map-to-output system in niri for tablets and touchscreens, and there's a downcast to libinput event to figure out what device the event comes from. Maybe you could similarly downcast to a virtual pointer event, and then get the output from it?

sodiboo: > There's already a map-to-output system in niri for tablets and touchscreens

Yeah, I know. But these are applied unilaterally to all touch and tablet position events, based on the event kind; it's not very useful to apply that to virtual pointers since there isn't a single output they all should map to. What I'm suggesting is we replace this system with one that can associate the mapped output with the device or the event data.

YaLTeR: Yeah but if you downcast the absolute motion pointer event to its virtual absolute motion pointer event type, then you can grab the right output from that event. Should be simple enough until proper systems are in place in Smithay or wherever.

ConsularParadi:
Hi, I was trying to setup wayvnc and having some issues with wlr-virtual-pointer. Has this issue been resolved in  niri-0.1.9-1. If yes, how can I set it up?

YaLTeR:
wlr-virtual-pointer is part of this PR. It is not yet merged

sodiboo:
> I was trying to setup wayvnc [...]

Note that `wayvnc` still won't work smoothly once this is merged, at the very least because Smithay doesn't handdle `ext-virtual-keyboard` properly; all keyboard inputs are sent to the focused applications, and the compositor can't currently intercept them for purposes such as window management (e.g. you can only change focus by clicking with the mouse; over a VNC connection it will behave as if you have zero binds configured)

Currently, i'm fairly busy with a bunch of exams and other assignments at school, which is why i haven't been very actively working on this PR (and others) for some time. I'll get to it sometime, and also improve the virtual keyboard situation such that WayVNC will work properly, as i do actually also wish to use WayVNC eventually. I have no ETA on when all of this will be finished; it's a fairly broad scope of work requiring changes in Smithay's input backend and not just niri.

If you don't find the lack of keyboard shortcuts to be problematic, and still wish to use WayVNC and want to do it right now, you can just run this PR as-is, before it's merged. Its not necessarily up-to-date with the most recent version of niri, so you'd be "downgrading" slightly (not by much). I am running [a branch based on this PR](https://github.com/sodiboo/niri/commits/sodi-bonus-features/) live on my machines and it works fine.

sodiboo:
by the way:

- closes #198

i didn't realize this was an issue that existed but it should definitely be linked to this PR.

current status: i have a lot of stuff in my backlog, including a lot of schoolwork. i'll get back to this (and my other PRs) eventually but right now it's not my top priority

the keyboard shortcuts inhibition part is actually stable and worked well last i recall, so like, if you want it right now just go merge it manually. i guess i could open another PR to separate these, because i think that one requires less work to "finish" and actually be good to merge.

resources/default-config.kdl
sodiboo: The reasoning for *not* doing this is so:

The primary release mechanism should be implemented in the client. It will know best what keys are in use and easily accessible. In many cases, we'd expect the client to release on `Super+Escape` (because inhibiting shortcuts means you're free to use `Super` for whatever binds, and `Super+Escape` is unlikely to mean anything else), but if it does mean something else then the client knows best what is an appropriate release shortcut.

The reason why i chose `Mod+Shift+Escape` is because that's the shortcut to open Task Manager on Windows; so it already has the connotation of "An app is misbehaving!! Help me, operating system!" in some users' minds (including my own).

The difference is, however, ultimately arbitrary. So if you disagree with the above, i'll change it to `Mod+Escape`.

niri-config/src/lib.rs
sodiboo: i don't feel like that change is an improvement, because `prevent-inhibiting=true` is a double negative. but also, i agree with your concern, and i think it's slightly confusing because `allow-inhibiting=false` is also a double negative, because the value it will be set to is false.

is there a way to name it that isn't a double negative? `always-available=true`, but more descriptive? borrow an idiom from CSS and call it `important=true`? maybe with the exclamation point to drive home the correlation? `!important=true`. though it reads as "not".

src/handlers/mod.rs
sodiboo: I'm not quite sure how such a dialog would work in niri (how do we remember it? just based off `app-id`? what about when it's done on a layer shell surface? also, how do we display the dialog? with pangocairo or though gnome portal somehow?), or if it even makes sense to have (this isn't exactly a "permission" a client asks for as on iOS/Android; if you have a client that abuses it, that client should be fixed really), but i've added the comment nonetheless because these questions don't need to be resolved in this PR if the implementation of such a feature is done separately

sodiboo:
Okay, so respecting the `output` parameter was actually surprisingly less work than expected. It's fairly trivial to define an extension trait for the input backend, such that we can easily expose the `VirtualPointer::output()` to the event handlers.

I've implemented this in the commit titled "add InputBackend extensions; use Device::output() for absolute pos events". This is the latest commit.

Virtual Pointer absolute motion when provided with an output should now work, and actually give output-relative coordinates.

I've also hooked up the touch and tablet positions to this function, falling back to the previous behaviour when it returns `None`, which is actually **always** for these event types at this time. But, this makes it easier to implement the `ei` protocol and also the potential to expand the config to allow per-tablet configuration of which output it maps to. Just change the impls in `backend_ext.rs` to make sure they return the correct `Output`, and everything else *should* just work as is. (and, maybe need to change its signature to take additional params, but besides that the callsites already work).

src/protocols/virtual_pointer.rs
sodiboo: https://github.com/YaLTeR/niri/pull/630#issuecomment-2495646090

niri-config/src/lib.rs
YaLTeR: Hmmm, difficult question indeed. I like the word "inhibit" being there because it's the technical term for this (unless it isn't always? do some apps re-phrase it?).

If we drop inhibit, maybe something like "always-accessible" is better than "always-available"? That's kind of still not very descriptive, while it does control a very specific thing, so it would be good if it were descriptive...

salman-farooq-sh:
Thanks @sodiboo and @YaLTeR. What work is left in this now?

salman-farooq-sh:
I am very excited for this.

YaLTeR:
Remind me, what is left to do here? (Apart from the whole "virtual keyboard doesn't do compositor shortcuts" which is not really related to these code changes.) Just review and merge?

sodiboo:
uihhh yeah, just review i guess. everything works i think.

resources/default-config.kdl
YaLTeR:
```suggestion
    Mod+Escape allow-inhibiting=false { toggle-keyboard-shortcuts-inhibit; }
```
so yeah

src/input/backend_ext.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn output(&self, _state: &State) -> Option<Output> {
+        // FIXME: Allow specifying the output per-device?
+        None
+    }
+}
+
+impl NiriInputDevice for WinitVirtualDevice {
+    fn output(&self, state: &State) -> Option<Output> {
+        match state.backend {
+            Backend::Winit(ref winit) => Some(winit.single_output().clone()),
```
YaLTeR, line 37: Does this actually do anything? Maybe just replace the whole thing with `None` and remove the `single_output()` function?

src/input/mod.rs
```diff
@@ -2588,6 +2648,12 @@ fn find_bind(
             repeat: true,
             cooldown: None,
             allow_when_locked: false,
+            // In a worst-case scenario, the user has no way to unlock the compositor and a
```
YaLTeR, line 2651: Speaking of, I think ToggleInhibit should be added to `allowed_when_locked()`

src/input/mod.rs
```diff
@@ -2538,17 +2585,30 @@ fn should_intercept_key(
...
                 });
             }
         }
     }
 
     match (final_bind, pressed) {
         (Some(bind), true) => {
-            suppressed_keys.insert(key_code);
-            FilterResult::Intercept(Some(bind))
+            if is_inhibiting_shortcuts && bind.allow_inhibiting {
```
YaLTeR, line 2599: Bumping my comment to add a few test cases for this that got lost due to code changes.

src/protocols/virtual_pointer.rs
```diff
@@ -0,0 +1,552 @@
...
+                pointer.mutate_axis_frame(Some(time), |frame| {
+                    frame.value(axis, value).v120(axis, discrete)
+                });
+            }
+            zwlr_virtual_pointer_v1::Request::Destroy => {
+                handler.destroy_virtual_pointer(pointer);
+                handler
+                    .virtual_pointer_manager_state()
+                    .virtual_pointers
+                    .remove(resource);
```
YaLTeR, line 530: This should be moved to destructor I think

YaLTeR (changes requested):
Some minor things left, otherwise looks good.

src/input/backend_ext.rs
sodiboo: Basically, it asserts that the mouse position on the winit backend *will* break if/when it ever adds another output. Like, it'd try to scale it relative to the global extents of all monitors instead of just the one. That's obviously like, super wrong, for everything on the winit backend.

---

I guess it might also correctly transform the tablet position? If we return None, we hit that code path:

https://github.com/YaLTeR/niri/blob/f337a272a5c94b657429656608921f930a56cabc/src/input/mod.rs#L290

(the `Transform::Normal` here looks suspicious to me)

---

in general, it just makes sure the more relevant known-output code paths are hit. It feels more correct, if nothing else.

src/protocols/virtual_pointer.rs
sodiboo: o. good catch. that is true, because the `Destroy` request won't be called if a client abruptly disconnects

src/input/mod.rs
sodiboo: done

YaLTeR:
I added a few tests for inhibited shortcuts. Manually tested inhibiting with SDL3 test clients, seems to work fine.

However, I noticed that pointer input in winit is vertically inverted. What's up with that, does it work on your machine?

sodiboo:
> However, I noticed that pointer input in winit is vertically inverted. What's up with that, does it work on your machine?

(resolved on Matrix, up to my push just now; but summarizing here for future reference)

No, this didn't work on my machine. I don't know how i missed that.

It's because returning the `Output` from the winit backend made the pointer event code actually respect the `Transform::Flipped180` that is unilaterally applied to the winit output.
This wasn't an issue before, because previously it was handled as a global coordinate space (where the output transform is not considered).
The *correct* way to solve the *root cause* of this is to apply a buffer transform at the end of rendering (e.g. [in my Wayland backend, i tell the parent compositor to flip the buffer](https://github.com/sodiboo/niri/blob/9fa6bd00ed5857815b4fb2e8855252628fd98fef/src/backend/wayland.rs#L212-L216)), but since that's in Smithay (and actually even winit, not just Smithay), we cannot easily do that right now,
so the *current* resolution is to just revert to the old behaviour. it's indistinguishable anyway, so it's whatever. buh. it works i guess. i just don't like that it's essentially two bugs cancelling each other out

YaLTeR:
Let's assume this works fine now, and i'll also pretend i didn't see how you added that new State parameter. Still, please re-test with lan-mouse (ideally with transformed outputs too)

YaLTeR:
Thanks!

sodiboo:
> and i'll also pretend i didn't see how you added that new State parameter

hey there was already a comment about that in the previous commit, that we'd need it to implement per-device config (e.g. multiple touch screens, or multiple tablets mapped to different outputs). this is just preparing for *that* inevitability, at least

sodiboo:
I've tested it now with lan-mouse, and it seems to still be fully functional! both `wlr-virtual-pointer` and `keyboard-shortcuts-inhibit` work fine!

i did notice, however, that disabling the keyboard shortcuts inhibitor is *not* very helpful against lan-mouse because, you see, lan-mouse still has keyboard focus, and if i open a fuzzel, i can't type into it; those keypresses go to my other machine. so best i can do is like, exit niri or switch VTs (though the latter doesn't require the inhibitor anyways). I guess that is basically a `lan-mouse` bug, because it should release its focus whenever the inhibitor is disabled methinks (it ought to run the same function as if a peer tells it that "hey you, you have the focus now, you're the main machine to control", but it doesn't)

of course, virtual keyboard still can't do compositor binds, but that's like, super not our fault, and not related to this PR. but this is the main thing limiting the usability of `lan-mouse` and similar such software now

YaLTeR:
Alright, thanks for testing

Froidoh:
@sodiboo Sorry for being dumb, but: lan-mouse should just work now, right?

I am trying niri built from git and lan mouse built from git and I can't get it to work at all :(

vilhalmer:
I wasn't able to get lan-mouse-git to work either, I think because of the encryption work that's in progress. The current release of lan-mouse works though.

Froidoh:
@vilhalmer thanks a bunch! I got it to work with the newest version, but just in one direction... I am sure it's a layer 8 problem at this point (pebkac)... :D

PR #638, opened 2024-08-29
implement `always_center_single_column` layout option

elkowar:
This option basically configures niri such that, regardless of the center focused column option, windows are centered if there's only one column in the workspace.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -397,6 +397,8 @@ pub struct Layout {
     pub default_column_width: Option<DefaultColumnWidth>,
     #[knuffel(child, unwrap(argument), default)]
     pub center_focused_column: CenterFocusedColumn,
+    #[knuffel(child)]
+    pub always_center_single_column: bool,
     #[knuffel(child, unwrap(argument), default = Self::default().gaps)]
     pub gaps: FloatOrInt<0, 65535>,
     #[knuffel(child, default)]
@@ -411,6 +413,7 @@ impl Default for Layout {
             preset_column_widths: Default::default(),
             default_column_width: Default::default(),
             center_focused_column: Default::default(),
+            always_center_single_column: false,
             gaps: FloatOrInt(16.),
             struts: Default::default(),
         }
@@ -3057,6 +3060,7 @@ mod tests {
                         bottom: FloatOrInt(0.),
                     },
                     center_focused_column: CenterFocusedColumn::OnOverflow,
+                    always_center_single_column: false,
                 },
                 spawn_at_startup: vec![SpawnAtStartup {
                     command: vec!["alacritty".to_owned(), "-e".to_owned(), "fish".to_owned()],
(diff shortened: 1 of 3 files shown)


niri-config/src/lib.rs
```diff
@@ -389,6 +389,8 @@ pub struct Layout {
     pub default_column_width: Option<DefaultColumnWidth>,
     #[knuffel(child, unwrap(argument), default)]
     pub center_focused_column: CenterFocusedColumn,
+    #[knuffel(child, unwrap(argument), default = false)]
```
YaLTeR, line 392:
```suggestion
    #[knuffel(child)]
```
Bool flags can work like this, then if they are present it counts as true, and if absent, it counts as false.

src/layout/workspace.rs
```diff
@@ -718,11 +718,17 @@ impl<W: LayoutElement> Workspace<W> {
         idx: usize,
         prev_idx: Option<usize>,
     ) -> f64 {
-        match self.options.center_focused_column {
-            CenterFocusedColumn::Always => {
+        match (
+            self.options.always_center_single_column,
+            self.options.center_focused_column,
+        ) {
+            (true, _) if self.columns.len() <= 1 => {
```
YaLTeR, line 725: I'd rather write it as a separate `if` at the front. `if self.options.always_center_single_column && self.columns.len() <= 1 { return ...; }`

YaLTeR:
How does this interact with the rest of centered column special casing? E.g. horizontal resizing goes twice as fast in centered mode.

elkowar:
Fixed both style issues, and made it such that the new flag is also respected in resizing and the other places where `CenterFocusedColumn::Always` is being checked 👍🏻

vreon:
Lurking Niri user here, hello all! :wave:

One behavior I like from [Karousel](https://github.com/peterfajdiga/karousel) is that it centers the view whenever all columns can fit on the display. I'm wondering if that's something we eventually want to replicate in Niri and whether or not that affects this PR.

Video from Karousel's README:

https://github.com/peterfajdiga/karousel/assets/22796326/2ab62d18-09c7-45f9-8fda-e5e36b8d7a02

YaLTeR:
Well, it's out of scope for this PR at least, because it's quite a bit different. 

I thought about it a while ago but decided against it, at least by default, since on my setup I find it convenient to have windows flushed to the side even when they total to shorter than the screen width.

YaLTeR:
There's still one issue left: `Column::tile_offsets_iter()` also has a centering check which now needs to somehow know whether that column is the only column on the workspace. This check is needed for interactive resizing without transactions (or with really slow windows), to know whether to center the slow window in the column on size mismatch, or flush it left/right.

I'll need to think of the best way to pass this flag in.

YaLTeR:
Actually, I just left a FIXME comment because with resize transactions that case is quite hard to trigger. Thanks!

PR #642, opened 2024-08-31
send frame callbacks for active screencast windows

### LINKED ISSUE #539: Off-screen windows are not refreshed for screencasts
### Description

When screencasting windows that aren't visible (e.g. scrolled far away, or on a different workspace entirely), they are only refreshed at 1Hz. This is because of https://github.com/YaLTeR/niri/commit/ed8a6afe806fbc3438b1fd467a71919900d04852. Prior to that commit, they wouldn't be refreshed at all. Windows that aren't visible aren't refreshed every frame, because that's wasteful. This calculation doesn't take into consideration the new per-window screencasting [introduced just before 0.1.7 was tagged](https://github.com/YaLTeR/niri/commit/dc92d80b9f8de761df2aa0bfc90b61d53b9c0831). So, those captured windows, when on a separate workspace or otherwise not within the current viewport, look really stuttery on a screencast. But they should be considered as visible. This likely went unnoticed due to testing windowed screencasting on mostly static windows that change when interacted with (and always have focus as a result). To reproduce it, play a youtube video, and open Discord on a separate workspace. Screenshare that browser tab in Discord and observe the stuttery mess it becomes.

### Discovery

Originally discovered by @yuu-fur ([`@yuu.dev:catgirl.cloud`](https://matrix.to/#/@yuu.dev:catgirl.cloud)), who uses a laptop with a NVIDIA-brand dedicated GPU and a hybrid setup. We both thought it might be an NVIDIA issue, but it turned out not to be the case. Changing which GPU anything runs on doesn't help. I later reproduced the issue on my own system, which runs an AMD GPU, completely ruling out the possibility of it being an NVIDIA issue. Since this was happening on Vesktop, i thought it might be https://github.com/Vencord/Vesktop/issues/629 which is caused by an underlying [Chromium bug](https://issues.chromium.org/issues/333945842), but his issue persists in Firefox-based browsers too (Librewolf; tested with web Discord) so that's not it either. We eventually realized it was highly correlated to window focus, and i remembered [that commit from way back about refreshing off-screen windows](https://github.com/YaLTeR/niri/commit/ed8a6afe806fbc3438b1fd467a71919900d04852), and i tested various ways for a window to leave the viewport, all of which resulted in the same stutter. This led me to look into the underlying code.

### Workaround

I wrote this patch to workaround the issue. It does not fundamentally solve the problem, because it just increases the refresh rate of off-screen windows. You can change out the `/ 60` with whatever your refresh rate is. This patch will cause more resource usage, including degraded battery life. It basically disables a power-saving feature (or rather, heavily nerfs it). I created it by looking at https://github.com/YaLTeR/niri/commit/ed8a6afe806fbc3438b1fd467a71919900d04852 and just changing the timer values. [A `Duration::ZERO` is valid for how this is used](https://smithay.github.io/smithay/smithay/desktop/utils/fn.send_frames_surface_tree.html). I didn't try this on my own system, but @yuu-fur did and reports that it worked well and seems to solve the issue (yay! high-refresh-rate screencasting!)

I don't recommend anyone really run this patch, especially not on a battery-constrained device (think of your battery life! your poor battery!), but it basically confirms the root cause is what i described at the top. You can apply it in my `niri-flake` by saving this patch to a file `niri.patch` and then setting `programs.niri.package = pkgs.niri-unstable.override { patches = [ ./niri.patch ]; };`

```patch
index 664aea6..9ae8505 100644
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -147,7 +147,7 @@ const CLEAR_COLOR_LOCKED: [f32; 4] = [0.3, 0.1, 0.1, 1.];
 // We'll try to send frame callbacks at least once a second. We'll make a timer that fires once a
 // second, so with the worst timing the maximum interval between two frame callbacks for a surface
 // should be ~1.995 seconds.
-const FRAME_CALLBACK_THROTTLE: Option<Duration> = Some(Duration::from_millis(995));
+const FRAME_CALLBACK_THROTTLE: Option<Duration> = Some(Duration::ZERO);
 
 pub struct Niri {
     pub config: Rc<RefCell<Config>>,
@@ -1607,10 +1607,10 @@ impl Niri {
 
         event_loop
             .insert_source(
-                Timer::from_duration(Duration::from_secs(1)),
+                Timer::from_duration(Duration::from_secs(1) / 60),
                 |_, _, state| {
                     state.niri.send_frame_callbacks_on_fallback_timer();
-                    TimeoutAction::ToDuration(Duration::from_secs(1))
+                    TimeoutAction::ToDuration(Duration::from_secs(1) / 60)
                 },
             )
             .unwrap();


```

### System Information

* niri version: niri unstable 2024-07-08 (commit [1af9f9bd95283544b068dbbe33e923ab9e836243](https://github.com/YaLTeR/niri/commit/1af9f9bd95283544b068dbbe33e923ab9e836243)) -- we both use recent niri unstable package

* Distro: NixOS 24.05 Uakari & NixOS 24.11 Vicuña

* GPU: NVIDIA GeForce GTX 1650 Mobile & AMD Radeon RX 7800 XT

* CPU: Intel Core i7-9750H & AMD Ryzen 5 7600X

---

###### how many times can you mention your favorite commit in just one issue?

sodiboo:
Fixes #539. cc @yuu-fur who originally brought this issue to my attention, he might find this interesting.

I don't really know what i'm doing here. Frame callbacks are confusing. But this looks entirely reasonable to my eyes, and i've tested it on my own system and it seems to work.

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -3170,16 +3170,13 @@ impl Niri {
         // However, this should probably be restricted to sending frame callbacks to more surfaces,
         // to err on the safe side.
         self.send_frame_callbacks(output);
+        #[cfg(feature = "xdp-gnome-screencast")]
+        self.send_frame_callbacks_for_cast(output);
         backend.with_primary_renderer(|renderer| {
             #[cfg(feature = "xdp-gnome-screencast")]
             {
                 // Render and send to PipeWire screencast streams.
                 self.render_for_screen_cast(renderer, output, target_presentation_time);
-
-                // FIXME: when a window is hidden, it should probably still receive frame callbacks
-                // and get rendered for screen cast. This is currently
-                // unimplemented, but happens to work by chance, since output
-                // redrawing is more eager than it should be.
                 self.render_windows_for_screen_cast(renderer, output, target_presentation_time);
             }
 
@@ -3573,6 +3570,39 @@ impl Niri {
                 FRAME_CALLBACK_THROTTLE,
                 |_, _| None,
             );
+        }
+    }
+
+    #[cfg(feature = "xdp-gnome-screencast")]
+    fn send_frame_callbacks_for_cast(&self, output: &Output) {
+        use smithay::wayland::seat::WaylandFocus;
+
+        let _span = tracy_client::span!("Niri::send_frame_callbacks_for_cast");
+        let frame_callback_time = get_monotonic_time();
+
+        for mapped in self.layout.windows_for_output(output) {
+            for cast in &self.casts {
+                if !cast.is_active.get() {
+                    continue;
+                }
+                let CastTarget::Window { id } = cast.target else {
+                    continue;
+                };
+                if mapped.id().get() != id {
+                    continue;
+                };
+                let Some(surface) = mapped.window.wl_surface() else {
+                    continue;
+                };
+
+                send_frames_surface_tree(
+                    surface.as_ref(),
+                    output,
+                    frame_callback_time,
+                    FRAME_CALLBACK_THROTTLE,
+                    |_, _| Some(output.clone()),
+                );
+            }
         }
     }
 


src/niri.rs
```diff
@@ -3576,6 +3573,39 @@ impl Niri {
...
+                let Some(surface) = mapped.window.wl_surface() else {
+                    continue;
+                };
+
+                send_frames_surface_tree(
+                    surface.as_ref(),
+                    output,
+                    frame_callback_time,
+                    FRAME_CALLBACK_THROTTLE,
+                    |_, _| Some(output.clone()),
```
YaLTeR, line 3603: I think instead of being a separate function, this cast check should be integrated into `send_frame_callbacks()` where it should be part of the `if current_primary_output.as_ref() != Some(output)` check. The stuff in `send_frame_callbacks()` does throttling to at most one frame callback per output refresh; without it windows will busy-loop frame callbacks (especially so off-screen windows, which is the primary case here).

YaLTeR (changes requested):

nukdokplex:
Are there any updates on this?

YaLTeR:
no

sodiboo:
there was review feedback and then I didn't respond and then I kinda forgot. that's the update lol. waiting on me to do literally anything i guess

YaLTeR:
More like this isn't really a proper fix so it's back to waiting for me to sometime in the future do a proper fix or something

raum-dellamorte:
A `render-offscreen` window rule might be a good solution. I suffer from being a VTuber, and I do apologise for that, but if I could make a window rule for VSeeFace to always render as if it were visible, that would reduce my suffering. It currently works as a floating window, which is not ideal.

In my head, a window rule would be easier than keeping track of windows that OBS or any other program is watching, but I've not looked at the code. How many windows could one need to maintain a normal refresh rate offscreen? (Never ask questions like that, ppl take it as a challenge)

YaLTeR:
> A `render-offscreen` window rule might be a good solution.

It would require the exact same solution as just fixing this for window screencasts would require

PR #655, opened 2024-09-05
SwitchPresetWindowHeight and WindowHeight presets

TheAngusMcFire:
Implements: #653


YaLTeR:
Thanks. This looks mostly good! I'll take it from here.

PR #670, opened 2024-09-15
flake: update inputs and remove crane.inputs.nixpkgs override

sodiboo:
general health check on nix flake.

- updated the inputs
- noticed a new warning about incorrectly overriding inputs.
  ```
  input 'crane' has an override for a non-existent input 'nixpkgs'
  ```
- i removed the override of `crane.inputs.nixpkgs`, because that input was removed in https://github.com/ipetkov/crane/pull/692
- see also https://github.com/YaLTeR/niri/pull/579 ([specific diff](https://github.com/YaLTeR/niri/pull/579/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0L36-R36)) where we stopped depending on the override being present

---

- there's still an evaluation warning:
  ```
  `rust.toRustTarget platform` is deprecated. Use `platform.rust.rustcTarget` instead.
  ```
- that warning originates in https://github.com/NixOS/nixpkgs/pull/339072 ([specific diff](https://github.com/NixOS/nixpkgs/pull/339072/files#diff-be600723a84d07032a964b03a4e1076e178f64507a8cbaedb8fa26ecb86cb7bbR90))
- it is triggered by fenix and fixed in https://github.com/nix-community/fenix/pull/163
- we use fenix monthly, which is from 3 days before that fix. no further action possible on our end.
- this warning will go away on the 1st of october. no further action necessary on our end.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,17 +1,12 @@
 {
   "nodes": {
     "crane": {
-      "inputs": {
-        "nixpkgs": [
-          "nixpkgs"
-        ]
-      },
       "locked": {
-        "lastModified": 1724533099,
-        "narHash": "sha256-ZIDtvVQHoCkNoBlLUB3wmqbqCb0Es3DfdUGeDI/58aY=",
+        "lastModified": 1725409566,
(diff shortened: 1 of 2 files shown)


YaLTeR:
Thanks

PR #686, opened 2024-09-27
Fix focus-window-up-or-column-right focusing the left column

MarwinKreuzig:
I'm currently trying out niri and noticed that focus-window-up-or-column-right actually focuses the left column. Seems like it's just a mix up with left and right.


YaLTeR:
Thanks. Looks like a copy-paste error indeed.

PR #697, opened 2024-09-29
Remove one unnecessary .clone() call and reorder another

spazzylemons:
I was reading through the code and just stumbled upon a call to .clone() that isn't needed. I also moved another call in another function, so it doesn't always need to run. This would hardly affect performance but I don't see why not to address it


YaLTeR:
Thanks!

PR #709, opened 2024-10-02
Implement Cursor Hiding While Typing & Cursor Inactive Timeout (#284)

yzy-1:
This pull request addresses the discussion on #284 regarding cursor management in niri. Specifically, I have implemented the second solution: hiding the cursor while typing.

The implementation of cursor timeout (solution 1) would require active polling to check the remaining time, which is not straightforward given our current design where all cursor-related updates are triggered by events. Implementing this approach would involve finding a suitable place in the codebase to handle the periodic checks, which I found challenging due to the existing event-driven architecture.

In contrast, hiding the cursor while typing (solution 2) can be seamlessly integrated as it aligns with the existing event-driven model. This implementation is more straightforward and reduces potential complexity.

Please review the changes and let me know if any further adjustments are needed.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -615,6 +615,10 @@ pub struct Cursor {
     pub xcursor_theme: String,
     #[knuffel(child, unwrap(argument), default = 24)]
     pub xcursor_size: u8,
+    #[knuffel(child)]
+    pub hide_on_key_press: bool,
+    #[knuffel(child, unwrap(argument))]
+    pub hide_after_inactive_ms: Option<u32>,
 }
 
 impl Default for Cursor {
@@ -622,6 +626,8 @@ impl Default for Cursor {
         Self {
             xcursor_theme: String::from("default"),
             xcursor_size: 24,
+            hide_on_key_press: false,
+            hide_after_inactive_ms: None,
         }
     }
 }
@@ -2953,6 +2959,8 @@ mod tests {
             cursor {
                 xcursor-theme "breeze_cursors"
                 xcursor-size 16
+                hide-on-key-press
+                hide-after-inactive-ms 3000
             }
 
             screenshot-path "~/Screenshots/screenshot.png"
@@ -3154,6 +3162,8 @@ mod tests {
                 cursor: Cursor {
                     xcursor_theme: String::from("breeze_cursors"),
                     xcursor_size: 16,
+                    hide_on_key_press: true,
+                    hide_after_inactive_ms: Some(3000),
                 },
                 screenshot_path: Some(String::from("~/Screenshots/screenshot.png")),
                 hotkey_overlay: HotkeyOverlay {
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2,7 +2,7 @@ use std::any::Any;
 use std::cmp::min;
 use std::collections::hash_map::Entry;
 use std::collections::HashSet;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 use calloop::timer::{TimeoutAction, Timer};
 use input::event::gesture::GestureEventCoordinates as _;
@@ -87,8 +87,8 @@ impl State {
             }
         }
 
-        if should_update_cursor_timeout(&event) {
-            self.niri.last_cursor_movement = Instant::now();
+        if should_reset_pointer_inactivity_timer(&event) {
+            self.niri.reset_pointer_inactivity_timer();
         }
 
         let hide_hotkey_overlay =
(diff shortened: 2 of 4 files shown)


src/input/mod.rs
```diff
@@ -360,16 +364,15 @@ impl State {
         }
 
         let config = self.niri.config.borrow();
-        let config = &config.input.keyboard;
```
YaLTeR, line 363: Seems like unrelated changes?

YaLTeR (commented):
Seems rather straightforward and simple. As long as it works for the people who want this functionality.

> Implementing this approach would involve finding a suitable place in the codebase to handle the periodic checks, which I found challenging due to the existing event-driven architecture.

Sounds similar enough to the idle notifier code. That one's refreshed on every input event, but the cursor one could be refreshed on every pointer event instead.

yzy-1:
> Sounds similar enough to the idle notifier code. That one's refreshed on every input event, but the cursor one could be refreshed on every pointer event instead.

Thanks for the feedback. You're right that it does have similarities with the idle notifier code. However, I believe cursor hiding should occur after a period of inactivity, which means there would be no corresponding pointer event (or even `input::State::process_input_event` will not be called) during that time.

I tried to implement this here, but the effect is that the cursor wouldn't hide until an input event occurred (like key press or something), essentially making the timeout ineffective.

yzy-1:
I tried it again, checking the cursor timeout in `refresh_pointer_outputs` and now it succeeds and works perfectly fine.

niri-config/src/lib.rs
```diff
@@ -2953,6 +2976,10 @@ mod tests {
             cursor {
                 xcursor-theme "breeze_cursors"
                 xcursor-size 16
+                hide-on-key-press
+                inactive-timeout {
+                    duration-ms 3000
+                }
```
YaLTeR, line 2982: I think a better way is to replace this block with something like `hide-after-inactive-ms 3000`

src/input/mod.rs
```diff
@@ -307,6 +311,8 @@ impl State {
             }
         }
 
+        self.hide_cursor_if_needed();
```
YaLTeR, line 314: This should be inside `if pressed {}`

src/input/mod.rs
```diff
@@ -383,6 +392,13 @@ impl State {
         self.niri.bind_repeat_timer = Some(token);
     }
 
+    fn hide_cursor_if_needed(&mut self) {
+        let config = self.niri.config.borrow();
+        if config.cursor.hide_on_key_press {
+            self.niri.pointer_hidden = true;
```
YaLTeR, line 398:
```suggestion
            self.niri.pointer_hidden = true;
            self.niri.queue_redraw_all();
```

src/niri.rs
```diff
@@ -2663,7 +2665,25 @@ impl Niri {
...
+        let config = &config.cursor.inactive_timeout;
+
+        if config.off {
+            return;
+        }
+
+        let now = Instant::now();
+        let timeout = Duration::from_millis(config.duration_ms as u64);
+
+        if now - self.last_cursor_movement >= timeout {
```
YaLTeR, line 2679:
```suggestion
        if self.last_cursor_movement.elapsed() >= timeout {
```

YaLTeR (changes requested):

niri-config/src/lib.rs
yzy-1: Should the inactive-timeout block of the previous layer be removed?

What I mean is that if a user needs to disable this feature, they should:

```plain
cursor {
  duration-ms -1 // Oh, I hate Magic Numbers!
}
```

or

```plain
cursor {
  inactive-timeout {
    off
    // hide-after-inactive-ms 3000
  }
}
```

YaLTeR: To disable the timeout, just don't put it in the config. This is consistent with other things like `focus-follows-mouse` and so on. In the struct, it should be an Option, then None would mean it's disabled and Some(u32) means it's enabled with that timeout.

yzy-1: Oh I know what you mean. We can use `Options<T>`. My bad.

yzy-1:
OK. I resolved them all.

YaLTeR:
I did some fixes, and I also had to reimplement the inactivity hiding as a timer after all, since otherwise it would fire late due to `refresh()` simply not being called when nothing is happening. Please test that this still works for you; otherwise it's good to merge.

YaLTeR (approved):

yzy-1:
It worked perfectly! Thank you!

YaLTeR:
Thanks

PR #715, opened 2024-10-05
Fix several clippy lints; most notably by using cstr literals

sodiboo:
I noticed the [`clippy::manual_c_str_literals`](https://rust-lang.github.io/rust-clippy/master/index.html#/manual_c_str_literals) lint started firing now for some reason, even though [it was added in Rust 1.78.0](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-178) (from May 2024), so it should've been firing on that code earlier. This PR primarily fixes those literals into their equivalent `c""` literals. This makes the code subtly safer to editing by moving the checks performed by that `expect()` to the compiler itself; although it doesn't actually change anything tangible for this code. It now looks kinda dumb to assign string literals to variables; but they are used several times throughout, so i haven't decided to inline them.

I have. however, removed a bunch of redundant pointer casts. [`CStr::as_ptr()` returns a `*const ::core::ffi::c_char`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.as_ptr), which is [exactly the same type as `*const ffi::types::GLchar`](https://smithay.github.io/smithay/smithay/backend/renderer/gles/ffi/types/type.GLchar.html) (defined as an alias of `::std::os::raw::c_char`, which is an alias of `::core:.ffi::c_char`, which is the full path of the type `CStr` gives us) on all architectures, so the cast that was there was already redundant.

I also fixed another lint that's been bugging me for a while where the layout code is matching on an option with `None => ()`

### DIFF
--- a/src/layout/workspace.rs
+++ b/src/layout/workspace.rs
@@ -3102,14 +3102,11 @@ impl<W: LayoutElement> Column<W> {
     }
 
     pub fn advance_animations(&mut self, current_time: Duration) {
-        match &mut self.move_animation {
-            Some(anim) => {
-                anim.set_current_time(current_time);
-                if anim.is_done() {
-                    self.move_animation = None;
-                }
+        if let Some(anim) = &mut self.move_animation {
+            anim.set_current_time(current_time);
+            if anim.is_done() {
+                self.move_animation = None;
             }
-            None => (),
         }
 
         for tile in &mut self.tiles {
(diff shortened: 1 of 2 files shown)


YaLTeR:
Thanks. I was very confused trying to remember where I used C literals and GL pointer casts in niri, then opened the diff and realized it's in the render element code copied from Smithay.

PR #718, opened 2024-10-06
Inverse touchpad scroll

ProMix0:
Currently when scrolling down with two fingers on touchpad I can move cursor down or scroll page down. But scrolling workspaces with three-finger swipe doing opposite - it scrolls up

This patch aims to achieve more consistency in interaction with the WM


YaLTeR:
Huh? They always go into the same direction for me.

ProMix0:
Yes, I messed up somewhere

PR #730, opened 2024-10-13
Implement `scroll_factor` for `Touchpad`

elipp:
Quick, dirty and poorly researched, but implementing it like this seems to behave quite the same as in `sway`.

I gathered that `libinput` doesn't directly support this, so seems like there's no relevant function to call in the  `apply_libinput_settings` function found in `src/input/mod.rs`.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -188,8 +188,8 @@ pub struct Touchpad {
     pub disabled_on_external_mouse: bool,
     #[knuffel(child)]
     pub middle_emulation: bool,
-    #[knuffel(child, unwrap(argument), default = 1.0)]
-    pub scroll_factor: f64,
+    #[knuffel(child, unwrap(argument), default = FloatOrInt(1.0))]
+    pub scroll_factor: FloatOrInt<0, 100>,
 }
 
 #[derive(knuffel::Decode, Debug, Default, PartialEq)]
@@ -210,8 +210,8 @@ pub struct Mouse {
     pub left_handed: bool,
     #[knuffel(child)]
     pub middle_emulation: bool,
-    #[knuffel(child, unwrap(argument), default = 1.0)]
-    pub scroll_factor: f64,
+    #[knuffel(child, unwrap(argument), default = FloatOrInt(1.0))]
+    pub scroll_factor: FloatOrInt<0, 100>,
 }
 
 #[derive(knuffel::Decode, Debug, Default, PartialEq)]
@@ -3175,7 +3175,7 @@ mod tests {
                         left_handed: false,
                         disabled_on_external_mouse: true,
                         middle_emulation: false,
-                        scroll_factor: 0.9,
+                        scroll_factor: FloatOrInt(0.9),
                     },
                     mouse: Mouse {
                         off: false,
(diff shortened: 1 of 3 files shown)


niri-config/src/lib.rs
```diff
@@ -182,6 +182,8 @@ pub struct Touchpad {
     pub disabled_on_external_mouse: bool,
     #[knuffel(child)]
     pub middle_emulation: bool,
+    #[knuffel(child, unwrap(argument), default)]
```
YaLTeR, line 185:
```suggestion
    #[knuffel(child, unwrap(argument), default = 1.0)]
```

src/input/mod.rs
```diff
@@ -1817,14 +1817,18 @@ impl State {
...
-            // Winit backend, discrete scrolling.
-            horizontal_amount_v120.unwrap_or(0.0) / 120. * 15.
-        });
-        let vertical_amount = vertical_amount.unwrap_or_else(|| {
-            // Winit backend, discrete scrolling.
-            vertical_amount_v120.unwrap_or(0.0) / 120. * 15.
-        });
+        let scroll_factor = self.niri.config.borrow().input.touchpad.scroll_factor;
+
+        let horizontal_amount = scroll_factor
```
YaLTeR, line 1822: Two problems doing the multiplication here:
1. It affects both mouse wheel and touchpad when the setting is in the `touchpad {}` section. I'd expect `mouse {}` to have the same, separate scroll factor setting, and them to apply only to the correct input device (the `source` check like above).
2. It doesn't affect the scroll bindings above. I don't think it should for the mouse wheel but maybe it should for touchpad? Not sure.

elipp:
Any further suggestions?

src/input/mod.rs
```diff
@@ -1815,22 +1815,30 @@ impl State {
...
+            _ => 1.0,
+        };
+
+        let horizontal_amount = scroll_factor
+            * horizontal_amount
+                // Winit backend, discrete scrolling.
+                .or(horizontal_amount_v120.map(|a| a / 120. * 15.))
+                .unwrap_or(0.0);
+
+        let vertical_amount = scroll_factor
```
YaLTeR, line 1837: Could you move `scroll_factor *` in the end so it's `vertical_amount.unwrap_or_else(...) * scroll_factor;`? I think that should preserve the previous formatting.

niri-config/src/lib.rs
```diff
@@ -3100,6 +3106,7 @@ mod tests {
                         left_handed: false,
                         disabled_on_external_mouse: true,
                         middle_emulation: false,
+                        scroll_factor: 1.0,
```
YaLTeR, line 3109: I guess these shouldn't be 1.0.

elipp:
Synced with upstream, resolved based on YaLTeR's commentary

src/input/mod.rs
```diff
@@ -1873,22 +1873,27 @@ impl State {
                         }
                     }
                 }
-
```
YaLTeR, line 1876: Accidentally deleted line

niri-config/src/lib.rs
YaLTeR: Well this is still not resolved as you can see from the failing test.

YaLTeR:
Thanks

PR #735, opened 2024-10-14
flake: use nightly rust-analyzer and add rust-src component

sodiboo:
this also improves the application of overlays to be more uniform, in particular [this like applying the overlay](https://github.com/YaLTeR/niri/blob/be7fbd418fcee2d2ff9f80aa15f1eb8ff0d83afb/flake.nix#L84) was just weird and basically wrong (that's not how overlays work)

there's also some subtlety with using `selectLatestNightlyWith`. from rust-overlay's README:
> Note: Don't use `rust-bin.nightly.latest`. Your build would fail when some components missing on some days. Always use `selectLatestNightlyWith` instead.

and uhh... i *think* [`packages.${system}.rust-nightly` maps to `rust-bin.nightly.latest`](https://github.com/oxalica/rust-overlay/blob/bdbe1611c2029de90bca372ce0b1e3b4fa65f55a/flake.nix#L71-L72)? so we were using the one that the README says not to use? the code is slightly hard to read. Either way, this subtlety wasn't causing any issues previously, but is fixed now to prevent it from causing any issues in the future as well.

the main benefit is that we're now using `rust-analyzer` also from `rust-overlay`, as well as including the `rust-src` component which is required for `rust-analyzer` to actually be useful. this is **a required change** because rust-analyzer is basically useless in the devshell since #687 because uhh Nobody Thought To Test It. the other stuff in this PR is mostly just refactoring.

cc @getchoo you touched this last and idk if you want to review it or anything, but feel free to leave comments. or don't, i don't know how much you want to be involved in the future

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,6 +22,130 @@
       rust-overlay,
     }:
     let
+      niri-package =
+        {
+          lib,
+          cairo,
+          clang,
+          dbus,
+          libGL,
+          libclang,
+          libdisplay-info,
+          libinput,
+          libseat,
+          libxkbcommon,
+          mesa,
+          pango,
+          pipewire,
+          pkg-config,
+          rustPlatform,
+          systemd,
+          wayland,
+          withDbus ? true,
+          withSystemd ? true,
+          withScreencastSupport ? true,
+          withDinit ? false,
+        }:
+
+        rustPlatform.buildRustPackage {
+          pname = "niri";
+          version = self.shortRev or self.dirtyShortRev or "unknown";
+
+          src = nix-filter.lib.filter {
+            root = self;
+            include = [
+              "niri-config"
+              "niri-ipc"
+              "niri-visual-tests"
+              "resources"
+              "src"
+              ./Cargo.lock
+              ./Cargo.toml
+            ];
+          };
+
+          postPatch = ''
+            patchShebangs resources/niri-session
+            substituteInPlace resources/niri.service \
+              --replace-fail '/usr/bin' "$out/bin"
+          '';
+
+          cargoLock = {
+            # NOTE: This is only used for Git dependencies
+            allowBuiltinFetchGit = true;
+            lockFile = ./Cargo.lock;
+          };
+
+          strictDeps = true;
+
+          nativeBuildInputs = [
+            clang
+            pkg-config
+          ];
+
+          buildInputs =
+            [
+              cairo
+              dbus
+              libGL
+              libdisplay-info
+              libinput
+              libseat
+              libxkbcommon
+              mesa # libgbm
+              pango
+              wayland
+            ]
+            ++ lib.optional (withDbus || withScreencastSupport || withSystemd) dbus
+            ++ lib.optional withScreencastSupport pipewire
+            # Also includes libudev
+            ++ lib.optional withSystemd systemd;
+
+          buildFeatures =
+            lib.optional withDbus "dbus"
+            ++ lib.optional withDinit "dinit"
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -27,7 +27,16 @@
...
+      nixpkgsFor = forAllSystems (
+        system:
+        import nixpkgs {
+          inherit system;
+          overlays = [
+            self.overlays.default
+            rust-overlay.overlays.rust-overlay
+          ];
+        }
+      );
```
getchoo, line 39: Don't instantiate our own Nixpkgs here. There isn't any real benefit from doing so besides making things more "uniform" (FYI the current method of using the overlay as a function is [quite well known](https://jade.fyi/blog/flakes-arent-real/) and not "incorrect"), but there are [massive costs for consumers of this Flake](https://zimbatm.com/notes/1000-instances-of-nixpkgs). Specifically, this adds 100mb+ and a few seconds to any given evaluation

***This is a footgun and was avoided purposefully***

getchoo (changes requested):
everything else LGTM

flake.nix
sodiboo: ohh...

yeah, so i've read [that article](https://zimbatm.com/notes/1000-instances-of-nixpkgs) before but apparently not closely enough. my takeaway was about storage space. wow.

this explains why my system takes ages to eval. i think i've got some spring cleaning to do. this is surely not the only reason, but i do make extensive use of overlays which i can't imagine helps.

sodiboo: > FYI the current method of using the overlay as a function is [,.,] not "incorrect"

i mean yeah. overlay is just a function and since we weren't using `prev:` it's whatever, but i still thought it looked very ugly and bad to be passing `null` where it expects an instance of nixpkgs. i've refactored the code to move the package definition higher up, and not rely on the overlay at all internally, so we can just call `pkgs.callPackage` instead. this avoids the need to even consider how best to use the overlay here; because the overlay isn't for internal use in this flake.

flake.nix
```diff
@@ -22,6 +22,130 @@
       rust-overlay,
     }:
     let
+      niri-package =
+        {
+          lib,
```
sodiboo, line 27: it's easy to miss in the diff but when moving this package declaration up here, i've also separately added the `lib` parameter so that comes from `pkgs.callPackage` instead of our outermost `inherit (nixpkgs) lib;`.

getchoo (approved):
> i don't know how much you want to be involved in the future

And feel free to ping me here any time :)

flake.nix
```diff
@@ -39,17 +163,31 @@
         system:
         let
           pkgs = nixpkgsFor.${system};
-          inherit (self.packages.${system}) niri;
+          niri = pkgs.callPackage niri-package { };
```
getchoo, line 166:
```suggestion
          inherit (self.packages.${system}) niri;
```
We don't need to re-evaluate the package

getchoo (changes requested):
Almost missed this

flake.nix
sodiboo: o yea. lol. happened 'cause i removed `inherit (self.packages.${system}) niri;` when applying the overlay straightup, and then returned to a different point when undoing that change. nice catch

getchoo (approved):

YaLTeR:
Thanks

sodiboo:
> feel free to ping me here any time :)

I think I just might. It'll be nice to have more than one pair of eyes actively overlooking the code.

PR #736, opened 2024-10-15
Add scroll-button property to Touchpad, Mouse, and Trackpoint

chillinbythetree:
* Added scroll-button option to touchpad, mouse, and trackpoint
* Added to default config an example touchpad config that uses the `on-button-down` scroll method with an assigned scroll button
* Added tests
* Updated wiki to describe `scroll-button` config option

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -40,6 +40,16 @@ input {
         // scroll-method "no-scroll"
     }
 
+    trackpoint {
+        // off
+        // natural-scroll
+        // accel-speed 0.2
+        // accel-profile "flat"
+        // scroll-method "on-button-down"
+        // scroll-button 273
+        // middle-emulation
+    }
+
     // Uncomment this to make the mouse warp to the center of newly focused windows.
     // warp-mouse-to-focus
 
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2651,6 +2651,10 @@ pub fn apply_libinput_settings(config: &niri_config::Input, device: &mut input::
 
         if let Some(method) = c.scroll_method {
             let _ = device.config_scroll_set_method(method.into());
+
+            if method == niri_config::ScrollMethod::OnButtonDown {
+                let _ = device.config_scroll_set_button(c.scroll_button);
+            }
         } else if let Some(default) = device.config_scroll_default_method() {
             let _ = device.config_scroll_set_method(default);
         }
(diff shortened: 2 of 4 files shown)


resources/default-config.kdl
```diff
@@ -40,6 +40,16 @@ input {
         // scroll-method "no-scroll"
     }
 
+    trackpoint {
+        // off
+        // natural-scroll
+        // accel-speed 0.2
+        // accel-profile "flat"
+        // scroll-method "on-button-down"
+        // scroll-button 273
```
YaLTeR, line 49: What's the 273 button?

src/input/mod.rs
```diff
@@ -2651,6 +2651,10 @@ pub fn apply_libinput_settings(config: &niri_config::Input, device: &mut input::
 
         if let Some(method) = c.scroll_method {
             let _ = device.config_scroll_set_method(method.into());
+
+            if method == niri_config::ScrollMethod::OnButtonDown {
+                let _ = device.config_scroll_set_button(c.scroll_button);
+            }
```
YaLTeR, line 2657: I guess these should be outside of the `if` in case the default scroll method for the device is already `OnButtonDown`?

wiki/Configuration:-Input.md
```diff
@@ -140,6 +141,7 @@ A few settings are common between `touchpad`, `mouse` and `trackpoint`:
 - `accel-speed`: pointer acceleration speed, valid values are from `-1.0` to `1.0` where the default is `0.0`.
 - `accel-profile`: can be `adaptive` (the default) or `flat` (disables pointer acceleration).
 - `scroll-method`: when to generate scroll events instead of pointer motion events, can be `no-scroll`, `two-finger`, `edge`, or `on-button-down`.
+- `scroll-button`: the button key (numerical form) used for the `on-button-down` scroll method
```
YaLTeR, line 144:
```suggestion
- `scroll-button`: the button code (numerical form) used for the `on-button-down` scroll method. You can find it in `libinput debug-events`.
```

resources/default-config.kdl
YaLTeR: I found that it's BTN_RIGHT, is that consistent for every mouse?

chillinbythetree: It seemsto be.

Source: https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h

PR #737, opened 2024-10-15
input: apply output transform for tablet input

cmeissl:
when mapping a tablet input to an output apply
the output transform just like we already do for
touch input.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -24,7 +24,7 @@ use smithay::input::pointer::{
     GrabStartData as PointerGrabStartData, MotionEvent, RelativeMotionEvent,
 };
 use smithay::input::touch::{DownEvent, MotionEvent as TouchMotionEvent, UpEvent};
-use smithay::utils::{Logical, Point, Rectangle, SERIAL_COUNTER};
+use smithay::utils::{Logical, Point, Rectangle, Transform, SERIAL_COUNTER};
 use smithay::wayland::pointer_constraints::{with_pointer_constraint, PointerConstraint};
 use smithay::wayland::tablet_manager::{TabletDescriptor, TabletSeatTrait};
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
I just tried on main and tablet input seems to work fine with transformed outputs as is? I.e. if I set transform 90 and rotate the monitor upright, the tablet input will work as if I had a vertical monitor with transform normal.

cmeissl:
> I just tried on main and tablet input seems to work fine with transformed outputs as is? I.e. if I set transform 90 and rotate the monitor upright, the tablet input will work as if I had a vertical monitor with transform normal.

Interesting, for me main is not working properly. Thinking about that a bit I guess you have an external tablet while I have
it integrated into the display. So I am actually physically rotating the tablet with the monitor itself. 
Not sure how this could be detected, maybe libinput provides some way to detect that the tablet is mounted on the monitor?

YaLTeR:
Hmm, right, yeah, I have an external tablet. Let's check how GNOME works in this case... For my external tablet it seems to apply the output transform, so the same as your PR. I guess then that's the expected outcome?

cmeissl:
I just pushed an update because I forgot to adapt the `keep_aspect` logic, breaking 90 and 270 on my laptop.

cmeissl:
> Hmm, right, yeah, I have an external tablet. Let's check how GNOME works in this case... For my external tablet it seems to apply the output transform, so the same as your PR. I guess then that's the expected outcome?

Good question, I haven't found much information on that topic. There is https://wayland.freedesktop.org/libinput/doc/latest/configuration.html#rotation but `TabletTool::rotation()` seems to always return `0.0 ` for me.

YaLTeR:
Gave this a test, something is broken with the transform in `flipped-90` mode. It works fine for some region of the screen and gets inverted in another.

cmeissl:
> Gave this a test, something is broken with the transform in `flipped-90` mode. It works fine for some region of the screen and gets inverted in another.

Strange, just tried and it seems to work consistently over the whole screen for me. Did you try with `7803fbe7bcd6ea93fd87a3ac9feb151b2a54be15`? Because I messed up the `keep_aspect` ratio option before.

YaLTeR:
Yeah, I tested 7803fbe. 180 and 90 works fine, flipped-90 is glitched.

cmeissl:
Wait, hm, with flipped-90/270 input is strange... The cursor also behaves strange...
Touch seems to be correct, tablet input itself also seems work as expected, but the cursor is rendered wrongly for tablet and pointer? Or maybe I am mixing things up, I hate this output rotation stuff...

cmeissl:
flipped-90/270 seem to already behave strange on current main?

YaLTeR:
Huh, indeed. Weird, I'm fairly sure it worked at some point. Well, this PR should be good then.

YaLTeR:
Thanks

PR #741, opened 2024-10-15
Add scroll-button property for Touchpad, Mouse, and Trackpoint

chillinbythetree:
* Added scroll-button option to touchpad, mouse, and trackpoint
* Added to default config an example touchpad config that uses the on-button-down scroll method with an assigned scroll button
* Added tests
* Updated wiki to describe scroll-button config option

Applied suggestions. Scroll-button property will also apply if default scroll-method is 'on-button-down'.


YaLTeR:
There's an extra mod.rs-fmt file now

chillinbythetree:
> There's an extra mod.rs-fmt file now

Removed it. Accidentally left it there when testing rustfmt.

niri-config/src/lib.rs
```diff
@@ -176,6 +176,8 @@ pub struct Touchpad {
     pub accel_profile: Option<AccelProfile>,
     #[knuffel(child, unwrap(argument, str))]
     pub scroll_method: Option<ScrollMethod>,
+    #[knuffel(child, unwrap(argument), default)]
+    pub scroll_button: u32,
```
YaLTeR, line 180:
```suggestion
    #[knuffel(child, unwrap(argument))]
    pub scroll_button: Option<u32>,
```
and others too.

src/input/mod.rs
```diff
@@ -2657,8 +2657,17 @@ pub fn apply_libinput_settings(config: &niri_config::Input, device: &mut input::
 
         if let Some(method) = c.scroll_method {
             let _ = device.config_scroll_set_method(method.into());
+
+            if method == niri_config::ScrollMethod::OnButtonDown && Some(c.scroll_button).is_some()
+            {
+                let _ = device.config_scroll_set_button(c.scroll_button);
+            }
```
YaLTeR, line 2664:
```suggestion
            if method == niri_config::ScrollMethod::OnButtonDown {
                if let Some(button) = c.scroll_button {
                    let _ = device.config_scroll_set_button(button);
                }
            }
```
and others too.

niri-config/src/lib.rs
```diff
@@ -216,6 +220,8 @@ pub struct Trackpoint {
...
     #[knuffel(child)]
     pub middle_emulation: bool,
 }
```
YaLTeR, line 227: I guess now that we have `Trackball`, it should also get the setting.

YaLTeR (changes requested):

PR #743, opened 2024-10-16
Implement input configuration for trackballs

tazjin:
This is almost the same as mouse/trackpad configuration. I've verified each setting with libinput's CLI to see if it makes sense for trackballs, and applied the changes locally to verify that everything works correctly.


YaLTeR:
Thanks

PR #744, opened 2024-10-16
Add scroll-button property for Touchpad, Mouse, Trackpoint, Trackball

chillinbythetree:
* Added scroll-button option to touchpad, mouse, trackpoint, and trackball
* Added to default config an example touchpad config that uses the on-button-down scroll method with an assigned scroll button
* Added tests
* Updated wiki to describe scroll-button config option

Applied suggestions from prior pull requests.


YaLTeR:
Thanks

PR #745, opened 2024-10-17
Allow creation of new workspace above top most workspace

FluxTape:
Add the layout option allow-workspace-above-first to allow adding new workspaces above the top most workspace. When this option is enabled the first and last workspace on a monitor are always empty (compared to just the last one) allowing the movement of windows/columns above the first workspace.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -444,6 +444,8 @@ pub struct Layout {
     pub center_focused_column: CenterFocusedColumn,
     #[knuffel(child)]
     pub always_center_single_column: bool,
+    #[knuffel(child)]
+    pub empty_workspace_above_first: bool,
     #[knuffel(child, unwrap(argument), default = Self::default().gaps)]
     pub gaps: FloatOrInt<0, 65535>,
     #[knuffel(child, default)]
@@ -460,6 +462,7 @@ impl Default for Layout {
             default_column_width: Default::default(),
             center_focused_column: Default::default(),
             always_center_single_column: false,
+            empty_workspace_above_first: false,
             gaps: FloatOrInt(16.),
             struts: Default::default(),
             preset_window_heights: Default::default(),
@@ -3310,6 +3313,7 @@ mod tests {
                     },
                     center_focused_column: CenterFocusedColumn::OnOverflow,
                     always_center_single_column: false,
+                    empty_workspace_above_first: false,
                 },
                 spawn_at_startup: vec![SpawnAtStartup {
                     command: vec!["alacritty".to_owned(), "-e".to_owned(), "fish".to_owned()],
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -517,7 +517,20 @@ impl<W: LayoutElement> Layout<W> {
                             workspaces.push(ws);
                         }
 
-                        if i <= primary.active_workspace_idx {
+                        if i <= primary.active_workspace_idx
+                            // Generally when moving the currently active workspace, we want to
+                            // fall back to the workspace above, so as not to end up on the last
+                            // empty workspace. However, with empty workspace above first, when
+                            // moving the workspace at index 1 (first non-empty), we want to stay
+                            // at index 1, so as once again not to end up on an empty workspace.
+                            //
+                            // This comes into play at compositor startup when having named
+                            // workspaces set up across multiple monitors. Without this check, the
+                            // first monitor to connect can end up with the first empty workspace
+                            // focused instead of the first named workspace.
+                            && !(self.options.empty_workspace_above_first
+                                && primary.active_workspace_idx == 1)
+                        {
                             primary.active_workspace_idx =
                                 primary.active_workspace_idx.saturating_sub(1);
                         }
@@ -541,6 +554,7 @@ impl<W: LayoutElement> Layout<W> {
                 if let Some(idx) = &mut active_workspace_idx {
                     *idx = workspaces.len() - *idx - 1;
                 }
+                let mut active_workspace_idx = active_workspace_idx.unwrap_or(0);
 
                 // Make sure there's always an empty workspace.
                 workspaces.push(Workspace::new(
@@ -554,6 +568,7 @@ impl<W: LayoutElement> Layout<W> {
                         0,
                         Workspace::new(output.clone(), self.clock.clone(), self.options.clone()),
                     );
+                    active_workspace_idx += 1;
                 }
 
                 for ws in &mut workspaces {
@@ -562,7 +577,7 @@ impl<W: LayoutElement> Layout<W> {
 
                 let mut monitor =
                     Monitor::new(output, workspaces, self.clock.clone(), self.options.clone());
-                monitor.active_workspace_idx = active_workspace_idx.unwrap_or(0);
+                monitor.active_workspace_idx = active_workspace_idx;
                 monitors.push(monitor);
 
                 MonitorSet::Normal {
@@ -579,15 +594,16 @@ impl<W: LayoutElement> Layout<W> {
                     self.options.clone(),
                 ));
 
+                let mut active_workspace_idx = 0;
                 if self.options.empty_workspace_above_first && workspaces.len() > 1 {
                     workspaces.insert(
                         0,
                         Workspace::new(output.clone(), self.clock.clone(), self.options.clone()),
                     );
+                    active_workspace_idx += 1;
                 }
 
                 let ws_id_to_activate = self.last_active_workspace_id.remove(&output.name());
-                let mut active_workspace_idx = 0;
 
                 for (i, workspace) in workspaces.iter_mut().enumerate() {
                     workspace.set_output(Some(output.clone()));
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -3294,6 +3294,8 @@ impl<W: LayoutElement> Layout<W> {
                     }
                 }
 
+                // needed because empty_workspace_above_first could have modified the idx
+                let ws_idx = mon.active_workspace_idx();
                 let ws = &mut mon.workspaces[ws_idx];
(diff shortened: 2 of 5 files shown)


niri-config/src/lib.rs
```diff
@@ -429,6 +429,8 @@ pub struct Layout {
     pub center_focused_column: CenterFocusedColumn,
     #[knuffel(child)]
     pub always_center_single_column: bool,
+    #[knuffel(child)]
+    pub allow_workspace_above_first: bool,
```
YaLTeR, line 433: "allow" feels weird here. Maybe `empty-workspace-above-first`? Or `add-empty-workspace-above-first`?

src/layout/monitor.rs
```diff
@@ -105,6 +105,18 @@ impl<W: LayoutElement> Monitor<W> {
         }
     }
 
+    /// adds a workspace above the top most one. Also modifies self.active_workspace_idx
```
YaLTeR, line 108:
```suggestion
```

YaLTeR (changes requested):
Please add checks into `verify_invariants()` that when this flag is active, there is an empty workspace above first, similarly to checks about there being an empty workspace below last. See https://github.com/YaLTeR/niri/blob/8d08782eba427a46607a478d088744d023a292c3/src/layout/mod.rs#L1819

Also run the randomized tests as described here: https://github.com/YaLTeR/niri/wiki/Developing-niri#running-tests

I expect you'll need to fix several more places that assume there's only an empty workspace in the end.

niri-config/src/lib.rs
FluxTape: I agree, I like `empty-workspace-above-first`

FluxTape:
I've made the necessary changes to `verify_invariants()` and fixed any issues that popped up. Anything else that needs changing?

proptest-regressions/layout/mod.txt
```diff
@@ -0,0 +1,8 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
YaLTeR, line 1: Don't push this file; instead, convert it into a regular unit test. As far as I understand, these seeds change when changing the Op enum (happens frequently).

src/layout/mod.rs
```diff
@@ -1819,16 +1822,34 @@ impl<W: LayoutElement> Layout<W> {
                 monitor.workspaces.last().unwrap().columns.is_empty(),
                 "monitor must have an empty workspace in the end"
             );
+            if monitor.options.empty_workspace_above_first {
+                assert!(monitor.workspaces.last().unwrap().columns.is_empty(),
```
YaLTeR, line 1826:
```suggestion
                assert!(monitor.workspaces.first().unwrap().columns.is_empty(),
```

src/layout/mod.rs
```diff
@@ -1819,16 +1822,34 @@ impl<W: LayoutElement> Layout<W> {
...
+                "first workspace must be empty workspace when empty_workspace_above_first is set")
+            }
 
             assert!(
                 monitor.workspaces.last().unwrap().name.is_none(),
                 "monitor must have an unnamed workspace in the end"
```
YaLTeR, line 1832: Needs a corresponding assert for `first()`

src/layout/mod.rs
```diff
@@ -1819,16 +1822,34 @@ impl<W: LayoutElement> Layout<W> {
...
             if monitor.workspace_switch.is_none() {
-                for (idx, ws) in monitor.workspaces.iter().enumerate().rev().skip(1) {
+                for (idx, ws) in monitor
+                    .workspaces
+                    .iter()
+                    .enumerate()
+                    .skip(pre_skip)
+                    .rev()
+                    .skip(1)
+                // skip last
```
YaLTeR, line 1851:
```suggestion
                    // skip last
                    .skip(1)
```

src/layout/mod.rs
```diff
@@ -3950,6 +3971,27 @@ mod tests {
...
                     }
 
+                    first.apply(&mut layout);
+                    layout.verify_invariants();
+                    second.apply(&mut layout);
+                    layout.verify_invariants();
+                    third.apply(&mut layout);
+                    layout.verify_invariants();
+                }
+                // same test with empty-workspace-above-first set to true
```
YaLTeR, line 3981: Not sure I want to add this since it doubles the test time (which is already the longest I believe). Let's let proptest catch these issues, together with manual test cases.

src/layout/mod.rs
```diff
@@ -4315,12 +4357,18 @@ mod tests {
         };
 
         let mon = monitors.into_iter().next().unwrap();
+        let expected_workspace_offset = if mon.options.empty_workspace_above_first {
```
YaLTeR, line 4360: Why this check? The test is done with default options.

FluxTape: Ups, when testing I temporarily changed the Default, this is a remnant of that. I'll refactor it to it's own test.

src/layout/mod.rs
FluxTape: ok removed it again

proptest-regressions/layout/mod.txt
FluxTape: removed in latest commit

src/layout/mod.rs
FluxTape: should be fixed in latest commit

FluxTape: looks like there are still more issues, I'll investigate them later

FluxTape: Ok all regressions should be fixed now, can you take another look?

YaLTeR (changes requested):
I'm so sorry, I forgot to submit the review several days ago..

FluxTape:
Implemented all your suggestions, let me know if there are any more issues

YaLTeR:
If you don't mind, I will postpone this to after the upcoming release. The logic has some remaining bugs, which are not difficult to fix, but they cannot get caught by the current testing system. I want to extend the testing system a bit to make sure they are caught and that there are no other similar bugs left.

YaLTeR:
So that I don't forget, the current bugs are:
1. Interactive move uses add_tile() but assumes the workspace idx doesn't change, which it does if a workspace is added at the top. Possibly similar bugs elsewhere in the code. Not caught because the testing system needs to be able to advance animations to complete a workspace switch up.
2. Toggling the option at runtime doesn't add or remove the topmost workspace. Not caught because there are no tests that change the options.

FluxTape:
> So that I don't forget, the current bugs are:
> 
>     1. Interactive move uses add_tile() but assumes the workspace idx doesn't change, which it does if a workspace is added at the top. Possibly similar bugs elsewhere in the code. Not caught because the testing system needs to be able to advance animations to complete a workspace switch up.
> 
>     2. Toggling the option at runtime doesn't add or remove the topmost workspace. Not caught because there are no tests that change the options.

To fix 2) it should be enough to add 
```rust
        if self.options.empty_workspace_above_first != options.empty_workspace_above_first
            && self.workspaces.len() > 1
        {
            if options.empty_workspace_above_first {
                self.add_workspace_top();
            } else {
                self.workspaces.remove(0);
                self.active_workspace_idx = self.active_workspace_idx.saturating_sub(1);
            }
        }
```
to `update_config(..)` in `monitor.rs`

YaLTeR:
Working on advancing animations in the test system in #825.

YaLTeR:
I've been running the randomized operations test a lot, and I've also implemented a proptest-state-machine version of the test that is a bit smarter in terms of generating valid sequences of operations in hopes that it will be more efficient at finding the issue 1 edge case. Unfortunately, I couldn't seem to get it to find it faster than the existing random operations test, and it adds quite a bit of code, so I kept it in a private branch for now. It did find one oversight in the AddWindowRightOf test implementation though which I just committed to main.

Both of these random tests manage to find the second output version of issue 1, but only after several minutes of runtime. Neither of the random tests has yet managed to find the single-output version of issue 1, which is slightly worrying. Makes me suspicious that there might be other problems that the current randomized tests have a hard time catching (probably due to the complexity and amount of operations).

Maybe it would be worth to also hook up cargo-fuzz with its coverage guidance. It uses a different Arbitrary trait though...

I'm not blocking this PR on this, just commenting what I've been spending time on. I'll next add a test for issue 2 and then the fix from your suggestion, and give it some more manual testing.

FluxTape:
For 2) you can take a look at [this branch](https://github.com/FluxTape/niri/tree/workspace-above-v3) for inspiration, which helped me find and fix some issues.

YaLTeR:
Thanks, I used that as a reference, but made a slightly simpler variation. Which immediately found an existing bug in update_config() :facepalm:

YaLTeR:
With the config fix, one more problem that is not caught by tests: when enabling ewaf during a workspace switch animation or gesture, it gets glitched because its indices are now one off.

Need to check if it's ok to update the indices, or better to just cancel the gesture/animation.

Also, probably want to preserve the first empty workspace when disabling ewaf in case it was focused.

YaLTeR (approved):

YaLTeR:
Note to self, remember to add it to the config wiki page

YaLTeR:
I'm adding some ad-hoc conditions to the randomized tests, and I caught another "workspace index changed underneath" issue:
1. Start a workspace switch gesture on the first empty workspace.
2. Open a window.
3. The workspace index jumps up, but the index stored in the gesture remains unchanged.

I'll see if maybe simply stopping the gesture in `add_workspace_top()` is the best solution for now. Outside of this PR, maybe the workspace switch anim/gesture should store a workspace ID instead? Though, it won't magically fix all problems either. Need better tests for this.

YaLTeR:
Another thing: need to make it so at startup with named workspaces, the named workspace is focused, rather than the first empty above named.

YaLTeR:
Haven't found any more issues, so let's merge this. Thanks!

PR #765, opened 2024-10-29
Hidden cursor fixes

YaLTeR:
cc @kchibisov who brought up the concern, does this behave as expected now?

Also cc @yzy-1, does this still work for your use-case?


kchibisov:
Seems to work fine.

yzy-1:
Works fine to me.

YaLTeR:
Thanks, that's good then

PR #771, opened 2024-11-01
xdg-desktop-portal: add basic support for access dialogs

cmeissl:
Accessing certain devices like cameras might require permissions. This is handled in xdg-desktop-portal by issuing an access dialog. The backend is expected to expose the [org.freedesktop.impl.portal.Access](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.Access.html) dbus interface.

## TODO:

- [x] Handling `options`, especially `choices` -> partially done, `grant_label` and `deny_label` should at least work
- [x] ~~Proper~~ UI (atm just copy of exit dialog with changed message) -> Uses `title`,`subtitle` and the optional `body`

### DIFF
--- a/src/dbus/xdg_deskop_portal_impl_access.rs
+++ b/src/dbus/xdg_deskop_portal_impl_access.rs
@@ -1,0 +1,107 @@
+use std::collections::HashMap;
+use std::ops::Not;
+
+use zbus::dbus_interface;
+use zbus::zvariant::{ObjectPath, OwnedValue, Value};
+
+use super::Start;
+use crate::ui::access_dialog::{AccessDialogOptions, AccessDialogRequest, AccessDialogResponse};
+
+pub struct AccessPortalImpl {
+    to_niri: calloop::channel::Sender<AccessDialogRequest>,
+}
+
+#[dbus_interface(name = "org.freedesktop.impl.portal.Access")]
+impl AccessPortalImpl {
+    /// AccessDialog method
+    #[allow(clippy::too_many_arguments)]
+    async fn access_dialog(
+        &self,
+        _handle: ObjectPath<'_>,
+        app_id: &str,
+        parent_window: &str,
+        title: &str,
+        subtitle: &str,
+        body: &str,
+        options: HashMap<&str, Value<'_>>,
+    ) -> zbus::fdo::Result<(u32, HashMap<String, OwnedValue>)> {
+        let options = AccessDialogOptions::from_dbus_options(options);
+        let (request, response_channel_receiver) = AccessDialogRequest::new(
+            app_id.to_string(),
+            parent_window
+                .is_empty()
+                .not()
+                .then(|| parent_window.to_string()),
+            title.to_string(),
+            subtitle.to_string(),
+            body.is_empty().not().then(|| body.to_string()),
+            options,
+        );
+
+        if let Err(err) = self.to_niri.send(request) {
+            tracing::warn!(?err, "failed to send access dialog request");
+            return Err(zbus::fdo::Error::Failed(format!(
+                "error creating access dialog: {err:?}"
+            )));
+        };
+
+        let result = HashMap::<String, OwnedValue>::new();
+        let response = match response_channel_receiver.recv().await {
+            Ok(AccessDialogResponse::Grant) => {
+                // FIXME: Add selected choices to the result
+                0
+            }
+            Ok(AccessDialogResponse::Deny) => 1,
+            Err(err) => {
+                tracing::warn!(?err, "failed to receive response for access dialog request");
+                return Ok((2, Default::default()));
+            }
+        };
+
+        return Ok((response, result));
+    }
+}
+
+impl AccessPortalImpl {
+    pub fn new(to_niri: calloop::channel::Sender<AccessDialogRequest>) -> Self {
+        Self { to_niri }
+    }
+}
+
+impl Start for AccessPortalImpl {
+    fn start(self) -> anyhow::Result<zbus::blocking::Connection> {
+        let conn = zbus::blocking::ConnectionBuilder::session()?
+            .name("org.freedesktop.impl.portal.desktop.niri")?
+            .serve_at("/org/freedesktop/portal/desktop", self)?
+            .build()?;
+        Ok(conn)
+    }
+}
+
+impl AccessDialogOptions {
+    fn from_dbus_options(options: HashMap<&str, Value<'_>>) -> Self {
+        let modal: bool = options
+            .get("modal")
+            .and_then(|option| option.downcast_ref())
+            .copied()
+            .unwrap_or(true);
+        let deny_label: Option<String> = options
+            .get("deny_label")
+            .and_then(|option| option.clone().downcast());
+        let grant_label: Option<String> = options
+            .get("grant_label")
+            .and_then(|option| option.clone().downcast());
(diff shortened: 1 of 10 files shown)


src/input/mod.rs
```diff
@@ -2991,13 +3000,17 @@ mod tests {
...
 
         // The key_code we pick is arbitrary, the only thing
         // that matters is that they are different between cases.
 
         let close_key_code = Keycode::from(close_keysym.raw() + 8u32);
-        let close_key_event = |suppr: &mut HashSet<Keycode>, mods: ModifiersState, pressed| {
+        let close_key_event = |suppr: &mut HashSet<Keycode>,
+                               mods: ModifiersState,
+                               pressed,
+                               access_dialog_ui: &AccessDialogUi| {
```
YaLTeR, line 3013: Hm, why can it capture screnshot_ui but not this?

src/ui/access_dialog_ui.rs
```diff
@@ -0,0 +1,179 @@
...
+    let cr = cairo::Context::new(&surface)?;
+    cr.set_source_rgb(0.1, 0.1, 0.1);
+    cr.paint()?;
+
+    cr.move_to(padding.into(), padding.into());
+    let layout = pangocairo::functions::create_layout(&cr);
+    layout.context().set_round_glyph_positions(false);
+    layout.set_font_description(Some(&font));
+    layout.set_alignment(Alignment::Center);
+    layout.set_markup(TEXT);
```
YaLTeR, line 153: Guess it should use the title/subtitle/body from the request.

src/input/mod.rs
cmeissl: Just a leftover, had it as mutable reference at some point.

src/ui/access_dialog_ui.rs
cmeissl: Yeah, UI is completely unfinished, just something to get started discussing what is needed. There are also custom options which do not seem optional. Pointer/Touch input is also needed I guess.

src/input/mod.rs
cmeissl: removed the unnecessary change

src/ui/access_dialog_ui.rs
cmeissl: Okay, it now uses `title`,`subtitle` and the optional `body`. In addition it *should* also  use the options for the grant and deny labels. I left out the choices for now.

src/dbus/xdg_deskop_portal_impl_access.rs
```diff
@@ -0,0 +1,150 @@
...
+            subtitle: subtitle.to_string(),
+            body: body.is_empty().not().then(|| body.to_string()),
+            options,
+
+            response_channel_sender,
+        };
+
+        if let Err(err) = self.to_niri.send(request) {
+            tracing::warn!(?err, "failed to send access dialog request");
+            return Ok((2, Default::default()));
```
YaLTeR, line 47: Niri-side errors should return an error from the portal. Something like:
```
return Err(fdo::Error::Failed(format!(
    "error creating session object: {err:?}"
)));
```

This is especially important for the access portal because an error return won't store the permission, whereas a successful return with nonzero value will store the denied permission: https://github.com/flatpak/xdg-desktop-portal/blob/bfb1c08718f91ae91b468ce309366120f5b8a162/src/camera.c#L141

src/dbus/xdg_deskop_portal_impl_access.rs
```diff
@@ -0,0 +1,150 @@
...
+        std::collections::HashMap<String, zbus::zvariant::OwnedValue>,
+    )> {
+        let (response_channel_sender, response_channel_receiver) = async_channel::bounded(1);
+
+        let options = AccessDialogOptions::from_options(options);
+        let request = AccessDialogRequest {
+            app_id: app_id.to_string(),
+            parent_window: parent_window
+                .is_empty()
+                .not()
```
YaLTeR, line 35: hah I didn't know this method was a thing, and on the `std::ops::Not` trait too

src/dbus/xdg_deskop_portal_impl_access.rs
```diff
@@ -0,0 +1,150 @@
...
+
+            response_channel_sender,
+        };
+
+        if let Err(err) = self.to_niri.send(request) {
+            tracing::warn!(?err, "failed to send access dialog request");
+            return Ok((2, Default::default()));
+        };
+
+        let result = std::collections::HashMap::<String, zbus::zvariant::OwnedValue>::new();
```
YaLTeR, line 50: Why not `use` the HashMap? Also zbus::zvariant::Something and tracing::warn

src/dbus/xdg_deskop_portal_impl_access.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn start(self) -> anyhow::Result<zbus::blocking::Connection> {
+        let conn = zbus::blocking::ConnectionBuilder::session()?
+            .name("org.freedesktop.impl.portal.desktop.niri")?
+            .serve_at("/org/freedesktop/portal/desktop", self)?
+            .build()?;
+        Ok(conn)
+    }
+}
+
+pub struct AccessDialogRequest {
```
YaLTeR, line 83: Could we maybe move these types into `src/ui/access_dialog.rs`? And don't gate that file under dbus. To avoid dbus-gating lots of statements in random places in the code.

src/ui/access_dialog_ui.rs
YaLTeR, line 1: I think just `access_dialog.rs`. And perhaps the type itself just `AccessDialog`. The screenshot ui is called ui because screenshot by itself is not a ui

src/dbus/xdg_deskop_portal_impl_access.rs
cmeissl: I am not sure I get what you mean? The hashmap is not needed atm, it would only be used for returning the choices on a grant, so I could return it. I left it here as a guidance for adding choices.

Not sure what you want me to do regarding `zbus::zvariant::Something and tracing::warn`

YaLTeR: I mean just import them instead of using fully qualified paths.

src/dbus/xdg_deskop_portal_impl_access.rs
cmeissl: Ah, okay, I interpreted the docs at https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html#org-freedesktop-portal-request-response as this should return 2 in this case and will not store a decision.

changed

src/dbus/xdg_deskop_portal_impl_access.rs
cmeissl: changed

src/dbus/xdg_deskop_portal_impl_access.rs
cmeissl: moved to `ui/access_dialog` and removed the gates

src/ui/access_dialog_ui.rs
cmeissl: deon

YaLTeR:
Guess this can be closed for now since the GTK portal works

cmeissl:
closed in favor of #774

PR #774, opened 2024-11-02
portal: prefer gtk for access portal

cmeissl:
using gnome for the access portal does not work,
so just override by directly using the gtk one


YaLTeR:
Thanks!

PR #777, opened 2024-11-02
flake: `libseat` renamed to `seatd`; add missing runtime dependencies

LoipesMas:
1. In newer nixpkgs, `libseat` alias for `seatd` got removed. This doesn't affect the current `flake.lock` version, but it will in the future.
2. Added `runtimeDependencies` and `autoPatchelfHook`. Without that, built `niri` binary wouldn't work (it would panic at startup). I think they were part of the flake some time ago, but got removed (I think in #687 ), I don't know why.

@getchoo @sodiboo you were involved with this flake, so maybe you have any comments. I just made it work for me ;]

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,7 +32,7 @@
           libclang,
           libdisplay-info,
           libinput,
-          libseat,
+          seatd,
           libxkbcommon,
           mesa,
           pango,
@@ -90,7 +90,7 @@
               libGL
               libdisplay-info
               libinput
-              libseat
+              seatd
               libxkbcommon
               mesa # libgbm
               pango
--- a/flake.nix
+++ b/flake.nix
@@ -123,7 +123,7 @@
 
             # Force linking with libEGL and libwayland-client
             # so they can be discovered by `dlopen()`
-            CARGO_BUILD_RUSTFLAGS = toString (
+            RUSTFLAGS = toString (
               map (arg: "-C link-arg=" + arg) [
                 "-Wl,--push-state,--no-as-needed"
                 "-lEGL"
@@ -208,7 +208,7 @@
               # in the package expression
               #
               # This should only be set with `CARGO_BUILD_RUSTFLAGS="$CARGO_BUILD_RUSTFLAGS -C your-flags"`
-              inherit (niri) CARGO_BUILD_RUSTFLAGS;
+              CARGO_BUILD_RUSTFLAGS = niri.RUSTFLAGS;
             };
           };
         }


sodiboo (changes requested):
> 1. In newer nixpkgs, libseat alias for seatd got removed. This doesn't affect the current flake.lock version, but it will in the future.

nice. i approve this first commit

> 2. Added runtimeDependencies and autoPatchelfHook. [...] I think they were part of the flake some time ago, but got removed (I think in #687), I don't know why.

The `runtimeDependencies` were removed in favor of force linking them with these cargo build flags, [introduced in that  PR](https://github.com/YaLTeR/niri/pull/687/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R208-R217)

https://github.com/YaLTeR/niri/blob/600cffb009cde7ed3e04b9f53365ad56cf2d6d6b/flake.nix#L124-L133

[According to getchoo](https://github.com/NixOS/nixpkgs/pull/347232#discussion_r1792200140), `runtimeDependencies` is mainly intended for prebuilt binaries we can't control build flags for, and link against all included libraries in the specified packages, rather than only linking to the relevant libraries as this approach does

> Without that, built niri binary wouldn't work (it would panic at startup).

In general, this is not true. The packages from `nixpkgs` and `niri-flake` both use the same approach without `runtimeDependencies`, and don't panic. But you are right that this package panics at startup. How curious.

This seems to stem from the fact that the package in this repo, and only this package, sets `CARGO_BUILD_RUSTFLAGS`. In niri-flake and nixpkgs, the variable that is set is `RUSTFLAGS`. indeed, if you correct this name, then the nix package no longer panics at startup. The shell variable does not need to be renamed, so the problem can be solved like so:

```diff
diff --git a/flake.nix b/flake.nix
index 0333e1e..f60d962 100644
--- a/flake.nix
+++ b/flake.nix
@@ -123,7 +123,7 @@
 
             # Force linking with libEGL and libwayland-client
             # so they can be discovered by `dlopen()`
-            CARGO_BUILD_RUSTFLAGS = toString (
+            RUSTFLAGS = toString (
               map (arg: "-C link-arg=" + arg) [
                 "-Wl,--push-state,--no-as-needed"
                 "-lEGL"
@@ -208,7 +208,7 @@
               # in the package expression
               #
               # This should only be set with `CARGO_BUILD_RUSTFLAGS="$CARGO_BUILD_RUSTFLAGS -C your-flags"`
-              inherit (niri) CARGO_BUILD_RUSTFLAGS;
+              CARGO_BUILD_RUSTFLAGS = niri.RUSTFLAGS;
             };
           };
         }
```

The above diff should replace the second commit in this PR.

LoipesMas:
@sodiboo it seems that you're right! I've replaced the second commit with your solution

sodiboo (approved):

YaLTeR:
ugh why does the update branch button create a merge commit

getchoo (approved):

LoipesMas:
> ugh why does the update branch button create a merge commit

I can rebase off main, if you want

sodiboo:
> ugh why does the update branch button create a merge commit

lol. "Merge branch 'main' into main". truly a commit

YaLTeR:
> I can rebase off main, if you want

Yeah, please do

sodiboo (approved):

LoipesMas:
> > I can rebase off main, if you want
> 
> Yeah, please do

Done

YaLTeR:
Thanks

PR #779, opened 2024-11-03
xdg: startup activation

cmeissl:
Okay, at first it looked like an easy task to just extend the logic to pass an activation token to process spawned through actions.

But it looks like the whole activation implementation misses some parts to actually make this work.
Actually it also looks like activation is not handled for launchers like fuzzel.

The reason is that `XdgActivationState::request_activation` might be called before the window is actually mapped
as part of the startup.

A quick look at sway reveals that it also handles unmapped windows, but in this case make sure that an token is only used once.

Draft because testing this will need the above to be fixed before.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1537,6 +1537,8 @@ pub struct DebugConfig {
     pub keep_laptop_panel_on_when_lid_is_closed: bool,
     #[knuffel(child)]
     pub disable_monitor_names: bool,
+    #[knuffel(child)]
+    pub strict_new_window_focus_policy: bool,
 }
 
 #[derive(knuffel::DecodeScalar, Debug, Clone, Copy, PartialEq, Eq)]
--- a/niri-visual-tests/src/cases/layout.rs
+++ b/niri-visual-tests/src/cases/layout.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 
 use niri::animation::Clock;
 use niri::layout::workspace::ColumnWidth;
-use niri::layout::{LayoutElement as _, Options};
+use niri::layout::{ActivateWindow, LayoutElement as _, Options};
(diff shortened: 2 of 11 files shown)


YaLTeR:
> The reason is that XdgActivationState::request_activation might be called before the window is actually mapped
as part of the startup.
>
> A quick look at sway reveals that it also handles unmapped windows, but in this case make sure that an token is only used once.

Hmm. Guess we can put a "focus on map" bool on an Unmapped?

cmeissl:
> > The reason is that XdgActivationState::request_activation might be called before the window is actually mapped
> > as part of the startup.
> > A quick look at sway reveals that it also handles unmapped windows, but in this case make sure that an token is only used once.
> 
> Hmm. Guess we can put a "focus on map" bool on an Unmapped?

yes, something like that, just hit this and did not have time to think about it.

YaLTeR (approved):

cmeissl:
Okay, so the second commit now allows to activate/focus windows in case a valid activation token has been passed while they were still un-mapped. While I believe this was the intended behavior from the beginning, it might still change how spawning and activation token work in niri today significantly.

YaLTeR:
I'll need to take a look myself where else to pass `activate` in the layout code to avoid the extra activation token data argument; rest looks good.

cmeissl:
> I'll need to take a look myself where else to pass `activate` in the layout code to avoid the extra activation token data argument; rest looks good.

Sure, I just wanted to avoid putting it into the mapped window as the activation data is no longer needed after mapping it.
An `Option<TokenData>` that is taken out on map also feels wrong imo.

YaLTeR:
Pushed the thing I was talking about + some fixes. Also added a `strict-new-window-focus-policy` debug flag which helps with testing the activation.

Should be good to merge.

PR #780, opened 2024-11-03
xdg: cleanup activation tokens

cmeissl:
valid tokens will stay around until explicitly cleaned-up. remove the token after it has been successfully used or we consider it timed out to prevent leaking the memory used by the activation tokens

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -110,7 +110,7 @@ use crate::dbus::gnome_shell_screenshot::{NiriToScreenshot, ScreenshotToNiri};
 #[cfg(feature = "xdp-gnome-screencast")]
 use crate::dbus::mutter_screen_cast::{self, ScreenCastToNiri};
 use crate::frame_clock::FrameClock;
-use crate::handlers::configure_lock_surface;
+use crate::handlers::{configure_lock_surface, XDG_ACTIVATION_TOKEN_TIMEOUT};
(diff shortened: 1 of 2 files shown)


src/niri.rs
```diff
@@ -1694,6 +1694,18 @@ impl Niri {
                 is_tty && !client.get_data::<ClientState>().unwrap().restricted
             });
         let activation_state = XdgActivationState::new::<State>(&display_handle);
+        event_loop
+            .insert_source(
+                Timer::from_duration(XDG_ACTIVATION_TOKEN_TIMEOUT),
+                |_, _, state| {
+                    state.niri.activation_state.retain_tokens(|_, token_data| {
+                        token_data.timestamp.elapsed() < XDG_ACTIVATION_TOKEN_TIMEOUT * 2
```
YaLTeR, line 1702: Is this `* 2` necessary? As far as I understand, if the `< timeout` condition fails here, the token is already unusable, even later this event loop iteration.

cmeissl: Something about giving it a grace period or something like this :shrug: 
Not actually needed atm, will remove it.

cmeissl: If this gets more complex it might be a good idea to add a shared function to test token/data validity and use it here and in `request_activation`, but not sure this will really work out.

YaLTeR:
Thanks

PR #789, opened 2024-11-08
feat: 'center-focused-column' add 'on-single' option

aeghn:
to center the column if it is the only one in a workspace.


YaLTeR:
Isn't it this? https://github.com/YaLTeR/niri/wiki/Configuration:-Layout#always-center-single-column

aeghn:
Sorry, I didn’t notice that option earlier. I will go ahead and close this PR, but I was wondering if you could clarify why this was separated out.

YaLTeR:
This way you can have never vs. on-overflow, but still always-center-single-column

PR #795, opened 2024-11-10
systemd: move toplevel to separate scopes

cmeissl:
Currently `niri` creates a separate scope per spawn action matching the name of the executable.
So for example launching the non flatpak version of `firefox` through `fuzzel` will result in this firefox instance
to run under a scope name `app-niri-fuzzel-<PID>.scope`.
This can be problematic in combination with `xdg-portal`, which will use the scope name as a fallback for the
non accessible app id.

To solve this, this PR tries to mimic what `systemd-sway` does, moving toplevels to systemd scopes matching their app id.
In addition to moving the process it also add slices like `systemd-sway` offers. This allows to override resource
limits per app id slice.

The code is more a PoC and needs a major cleanup. The worst part is finding out all child pids that should also be moved to the new scope.

### DIFF
--- a/src/utils/spawning.rs
+++ b/src/utils/spawning.rs
@@ -1,14 +1,20 @@
+use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::os::unix::process::CommandExt;
 use std::path::Path;
 use std::process::{Child, Command, Stdio};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::RwLock;
+use std::sync::{OnceLock, RwLock};
 use std::{io, thread};
 
+use anyhow::Context;
 use atomic::Atomic;
 use libc::{getrlimit, rlim_t, rlimit, setrlimit, RLIMIT_NOFILE};
 use niri_config::Environment;
+use smithay::reexports::wayland_server::{DisplayHandle, Resource};
+use smithay::wayland::compositor;
+use smithay::wayland::shell::xdg::{ToplevelSurface, XdgToplevelSurfaceData};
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -86,6 +86,7 @@ wayland-backend = "0.3.7"
 wayland-scanner = "0.31.5"
 xcursor = "0.3.8"
 zbus = { version = "~3.15.2", optional = true }
+zvariant = { version = "~3.15.2", optional = true }
```
YaLTeR, line 89: It's exported under zbus::zvariant::

src/utils/spawning.rs
```diff
@@ -382,6 +392,20 @@ mod systemd {
...
+        // systemd source.
+        for &c in name.as_bytes() {
+            if c.is_ascii_alphanumeric() || matches!(c, b':' | b'_' | b'.') {
+                slice_name.push(char::from(c));
+            } else {
+                let _ = write!(slice_name, "\\x{c:02x}");
+            }
+        }
+
+        let _ = write!(slice_name, ".slice");
```
YaLTeR, line 407: Why a slice and not a scope btw? All apps are in `app.slice`, and I only have the few pre-created slices on my system.
```
└─ systemctl --user | grep .slice
  -.slice
  app.slice
  background.slice
  session.slice
  user.slice
```

cmeissl: It should be both, the scope is named like `app-niri-<app_id>-{pid}.scope` and the slice is named `app-niri-<app_id>.slice`. This allows to override limits per `app_id`

PR #797, opened 2024-11-11
Unhide the pointer on scroll events

### LINKED ISSUE #796: Two-finger scrolling does not work while the cursor is hidden
I enabled the new options to automatically hide the cursor, both when typing and when the cursor is inactive. I noticed that when the cursor is inactive, two-finger scrolling on the touchpad doesn't have any effect. When I move the cursor to make it visible again, scrolling works again.

### System Information

* niri version: niri 0.1.10 (unknown commit) 

* Distro: NixOS

* GPU: AMD Radeon 780M

* CPU: AMD Ryzen 7 7840U

r-vdp:
Since we reset the surface under the pointer when we hide the pointer (see update_pointer_contents), scroll events don't work when the pointer is hidden.
So to make scrolling work, we make sure that we unhide the pointer when a scrolling event occurs.

Fixes #796

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1744,6 +1744,9 @@ impl State {
     fn on_pointer_axis<I: InputBackend>(&mut self, event: I::PointerAxisEvent) {
         let source = event.source();
 
+        // We received an event for the regular pointer, so show it now. This is also needed for
+        // update_pointer_contents() below to return the real contents, necessary for the pointer
+        // axis event to reach the window.
         self.niri.pointer_hidden = false;
         self.niri.tablet_cursor_location = None;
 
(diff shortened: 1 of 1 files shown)


src/input/mod.rs
```diff
@@ -1744,6 +1744,9 @@ impl State {
     fn on_pointer_axis<I: InputBackend>(&mut self, event: I::PointerAxisEvent) {
         let source = event.source();
 
+        self.niri.pointer_hidden = false;
```
YaLTeR, line 1747:
```suggestion
        // We received an event for the regular pointer, so show it now. This is also needed for
        // update_pointer_contents() below to return the real contents, necessary for the pointer
        // axis event to reach the window.
        self.niri.pointer_hidden = false;
```

YaLTeR:
Thanks, good catch.

PR #811, opened 2024-11-16
Add `focus-window-previous` action

zeroeightysix:
This action focuses the previously focused window. As a parallel to `focus-workspace-previous`, this works across outputs, or within the same workspace, and should come in handy for users that have scripts making use of `focus-window`.

- [x] what should happen when the previously focused window has since been closed?
- [x] todo: test this as daily driver and on a multi-monitor setup

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -36,7 +36,6 @@ use touch_move_grab::TouchMoveGrab;
 use self::move_grab::MoveGrab;
 use self::resize_grab::ResizeGrab;
 use self::spatial_movement_grab::SpatialMovementGrab;
-use crate::layout::LayoutElement as _;
 use crate::niri::State;
 use crate::ui::screenshot_ui::ScreenshotUi;
 use crate::utils::spawning::spawn;
@@ -641,9 +640,7 @@ impl State {
                 }
             }
             Action::FocusWindowPrevious => {
-                let window = self.niri.layout.active_window();
-                let window = window.and_then(|(m, _)| m.previous_focus_id());
-                if let Some(window) = window {
+                if let Some(window) = self.niri.previously_focused_window.clone() {
                     self.focus_window(&window);
                 }
             }
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -36,6 +36,7 @@ use touch_move_grab::TouchMoveGrab;
 use self::move_grab::MoveGrab;
 use self::resize_grab::ResizeGrab;
 use self::spatial_movement_grab::SpatialMovementGrab;
+use crate::layout::LayoutElement as _;
 use crate::niri::State;
 use crate::ui::screenshot_ui::ScreenshotUi;
 use crate::utils::spawning::spawn;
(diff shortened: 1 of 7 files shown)


src/niri.rs
```diff
@@ -891,20 +911,26 @@ impl State {
...
 
             // Tell the windows their new focus state for window rule purposes.
-            if let KeyboardFocus::Layout {
+            let previous_focus_id = if let KeyboardFocus::Layout {
                 surface: Some(surface),
             } = &self.niri.keyboard_focus
             {
                 if let Some((mapped, _)) = self.niri.layout.find_window_and_output_mut(surface) {
                     mapped.set_is_focused(false);
+                    Some(LayoutElement::id(mapped).clone())
```
YaLTeR, line 920:
```suggestion
                    Some(mapped.window().clone())
```

src/niri.rs
```diff
@@ -891,20 +911,26 @@ impl State {
             );
 
             // Tell the windows their new focus state for window rule purposes.
-            if let KeyboardFocus::Layout {
+            let previous_focus_id = if let KeyboardFocus::Layout {
```
YaLTeR, line 914: I think it'll be cleaner to have `let mut previous_focus_id = None;` and then set it to `Some(...)` inside the condition.

src/niri.rs
```diff
@@ -891,20 +911,26 @@ impl State {
...
+            } else {
+                None
+            };
             if let KeyboardFocus::Layout {
                 surface: Some(surface),
             } = &focus
             {
                 if let Some((mapped, _)) = self.niri.layout.find_window_and_output_mut(surface) {
                     mapped.set_is_focused(true);
+                    mapped.set_previous_focus_id(previous_focus_id)
```
YaLTeR, line 933: Doing it here, rather than in the layout, definitely sounds much simpler, but it means that in cases like "open keyboard-grabbing layer-shell, switch to a different layout window, close keyboard-grabbing layer-shell" the previous window will be lost. But I guess it's not really a big deal?

src/input/mod.rs
```diff
@@ -4,6 +4,14 @@ use std::collections::hash_map::Entry;
...
 use std::time::Duration;
 
+use self::move_grab::MoveGrab;
+use self::resize_grab::ResizeGrab;
+use self::spatial_movement_grab::SpatialMovementGrab;
+use crate::layout::LayoutElement;
+use crate::niri::State;
+use crate::ui::screenshot_ui::ScreenshotUi;
+use crate::utils::spawning::spawn;
+use crate::utils::{center, get_monotonic_time, ResizeEdge};
```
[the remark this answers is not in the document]
zeroeightysix: I think `cargo fmt` did this?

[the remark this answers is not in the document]
zeroeightysix: Nevermind, maybe it was rustrover. I'll put it back :shrug:

src/niri.rs
zeroeightysix: Hm, because `previous_focus_id` actually forms a graph of sorts (each window pointing to another window), maybe we can just follow the edges until we find a `None` or a window that still exists?

That would solve the question of "what to do when the previous window has closed", and also handles this case gracefully.

Is there an easy way to check if a `Window` still exists on any workspace?

YaLTeR: Well, you won't be able to follow previous focus id past a None or a window that no longer exists (since when a window closes, you lose the previous id stored on it). I don't think that is a big problem though.

Maybe for now it's sufficient to just skip setting the previous focus id if it is None? Then if you open and close layer-shell, the window won't lose its previous focus id. If you switch to a different window while focused on a layer-shell, you'll end up with an unexpected previous focus id, but I guess that doesn't happen very often?

zeroeightysix: > since when a window closes, you lose the previous id stored on it

Oh, right, window != mapped. I had that wrong in my mind.

I'm fine with your suggestion, this probably shouldn't cause too much unexpected behaviour. And it could be improved later on :shrug:

zeroeightysix:
Been using this for a few days now without issue, so it's ready for a final review :)

YaLTeR:
Did some minor cleanups

YaLTeR:
Hmm, what if we store the previous focused window just in the Niri state itself, rather than on each window? I think that should be simpler and should fix the layer-shell edge cases. I'll try doing it tomorrow when my brain is a bit clearer

YaLTeR:
Could you please try this branch again? I made the change above.

zeroeightysix:
Appears to be working as intended!

YaLTeR:
Thanks!

PR #825, opened 2024-11-23
Refactor animation timing to use lazy clocks

YaLTeR:
Instead of using `get_monotonic_time()`, creating animations now accepts the current time as argument. To track the current time, a `Clock` is added throughout the layout.

The main motivation is being able to advance the time in a controlled fashion in tests.

A secondary motivation is ensuring that animations that logically need to start at the same time do start at the exact same time (though, practically this doesn't seem to pose any sort of problem).

One complication is that now we need to not forget to update the current time before doing any action on the layout. In fact, the current version of the PR does forget to update the current time when a window is mapped, which results in cut-off window open animations.

### DIFF
--- a/niri-visual-tests/src/cases/gradient_angle.rs
+++ b/niri-visual-tests/src/cases/gradient_angle.rs
@@ -1,8 +1,6 @@
 use std::f32::consts::{FRAC_PI_2, PI};
-use std::sync::atomic::Ordering;
 use std::time::Duration;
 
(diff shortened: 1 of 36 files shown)


YaLTeR:
> One complication is that now we need to not forget to update the current time before doing any action on the layout.

Hm, this seems pretty error-prone... Perhaps I should make the "normal" (i.e. not test) clock fetch the current monotonic time after all.

Issue #808, opened 2024-11-16
Red glitch before lock screen shows up


hengtseChou:
I am trying to pick a locking utility for the niri setup. However, with swaylock and hyprlock, they glitch a red screen very briefly before proceeding to display the lock screen normally. I believe it is very similar to issue [#341](https://github.com/YaLTeR/niri/issues/341)?

I have tried out niri 0.1.9 on arch extra and 0.10.1 on arch extra-testing, and they both showing same glitch.

Thank you for this amazing project!

### System Information

* niri version: 0.10.1

* Distro: Arch 6.11.8-arch1-2

* GPU: Intel Iris Xe Graphics

* CPU:Intel i5-13500H

YaLTeR:
Hey, it's not a glitch, it's niri's locked session background. It shows up because niri doesn't wait for the lock screen to present a surface before drawing the locked session. This is tracked as part of #119.

With hyprlock specifically, there's an extra problem that it animates from transparency, so niri's red locked session background shows up during the animation. Sway seems to use a black background under lock surfaces so you don't notice it there. I'm not sure if the protocol mandates using a black background in this case.

lpnh:
My approach to this is to use a script calling [do-screen-transition](https://github.com/YaLTeR/niri/wiki/Configuration%3A-Key-Bindings#do-screen-transition) before the lock screen command:

```fish
#!/usr/bin/env fish

niri msg action do-screen-transition --delay-ms 300

swaylock -f
```

To bind a script in your niri config, using the `spawn` action:

```
binds {
    Mod+T { spawn "~/scripts/do-something.sh"; }
}
```

hengtseChou:
@lpnh Appreciate the workaround!

hengtseChou:
@YaLTeR Thanks for the clarification 😌

YaLTeR:
Ah yeah, the screen transition workaround, thanks!

salman-farooq-sh:
Can we hardcode a black background instead of the dark red one until https://github.com/YaLTeR/niri/issues/119 is done? @YaLTeR

YaLTeR:
Yeah, here: https://github.com/YaLTeR/niri/blob/8665003269d1fbe4efe3c477a71400392930cac9/src/niri.rs#L155

ypurpl:
[...]

PR #840, opened 2024-12-02
Change the momentary dark red screen before locking to black

salman-farooq-sh:


YaLTeR:
Uh, no, it's red for a reason. If your screen locker crashes, it hints you that the session is locked, and you don't have a random blackscreen.

A proper fix is to wait until the screen locker surface appears with a timeout before drawing it.

salman-farooq-sh:
@YaLTeR But https://github.com/YaLTeR/niri/issues/808#issuecomment-2512158069 ?

YaLTeR:
Yes, I meant that you can personally hardcode it in that place in your own build

salman-farooq-sh:
Umm, would everyone else also prefer having the red screen not flashing every time when locking at the cost of losing a hint that their locker has crashed? I mean we can definitely change this back to red once the proper fix is done. Hmmm, locker crashing can be tricky to debug indeed if it looks just the same as everything else black so maybe a much darker red or other color?

salman-farooq-sh:
How often do lockers crash? I don't think swaylock has ever crashed for me.

YaLTeR:
Does a screen transition not work for you?

salman-farooq-sh:
A screen transition also works but with that you can't use wallpapers in swaylock which are slow to load because then you either have to increase the transition time which makes the system feel less snappy or you are restricted to using light-weight wallpapers.

I can live with manually timing the screen transition for every wallpaper but putting the laptop in power-saving profile also affects the swaylock showing on screen time.

salman-farooq-sh:
@YaLTeR ping

YaLTeR:
Why? I explained why I want to keep the background red and solve the flash differently (by waiting until a surface appears before drawing the locked screen).

salman-farooq-sh:
> A screen transition also works but with that you can't use wallpapers in swaylock which are slow to load because then you either have to increase the transition time which makes the system feel less snappy or you are restricted to using light-weight wallpapers.
> 
> I can live with manually timing the screen transition for every wallpaper but putting the laptop in power-saving profile also affects the swaylock showing on screen time.

I meant w.r.t this, like... the screen is essentially stuck until the transition delay completes. Besides this I don't have any reason for this PR and yes you are right about the correct solution and this indeed is workaround-ish at best. Anyways, if you think this reason alone is not enough then you can close this PR. Thanks :)

salman-farooq-sh:
Especially when the laptop is in power-saving mode, the time taken to show the wallpaper by swaylock is a lot at least for me.

YaLTeR:
Could you check that happens on sway if you enable power saving and run swaylock with one of the heavy wallpapers? Do you see red or black for a while or frozen screen? 

Either way you can apply the patch locally until a proper fix is in place; it likely won't break.

salman-farooq-sh:
Nevermind, just tried hyprlock and it's blazing fast with its own transition I think so that solves that. Closing. Thanks.

YaLTeR:
Nice; still if you could check what happens on sway with swaylock that would be helpful

salman-farooq-sh:
Just checked sway with swaylock and it shows a pitch black screen before the wallpaper shows up. I am now actually having difficulty forming an opinion what color niri should have.

YaLTeR:
Thanks

Vortriz:
> Nevermind, just tried hyprlock and it's blazing fast with its own transition I think so that solves that. Closing. Thanks.

heyy can you elaborate on what setting did you in hyprlock to deal with that red splash screen?

salman-farooq-sh:
@Vortriz Sorry for the late reply but there was still a tiny amount of time the red splash showed up even with hyprlock. I just learned to live with it.

txkyel:
@Vortriz the suggestion to use `do-screen-transition` in https://github.com/YaLTeR/niri/issues/808#issuecomment-2480534649 works for me to hide the red splash screen. I imagine that the crossfade with `do-screen-transition` might interfere the hyprlock fade in animation but I can't really notice much a difference.

YaLTeR:
fwiw this has been fixed on main. Together with disabling the hyprlock fade it makes for no red flash

Vortriz:
yupp, `--no-fade-in` works. @txkyel cool trick nonetheless.

PR #841, opened 2024-12-02
nix: use bindgen hook over directly depending on `clang`

sodiboo:
https://github.com/NixOS/nixpkgs/pull/360711 helped me realize what `bindgenHook` actually does.

in fact, it is a simple thing that just sets `LIBCLANG_PATH` for us. this is meant to be used in conjunction with `bindgen`, hence `bindgenHook`.

image

so we don't need to set `LIBCLANG_PATH`.

no idea about `RUSTFLAGS` though. i tried to make it build properly without nix needing to pass it but i didn't get anywhere

bonus: updated flake lock, which was like 3 months out of date. there's not really a pressing need to do so but it's in general nice to have it be somewhat recentish. also means nix-based developers (hi that's me i work on niri at times from nix) get new nightly Rust yay :3

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nix-filter": {
       "locked": {
-        "lastModified": 1710156097,
-        "narHash": "sha256-1Wvk8UP7PXdf8bCCaEoMnOT1qe5/Duqgj+rL8sRQsSM=",
+        "lastModified": 1731533336,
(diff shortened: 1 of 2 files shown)


sodiboo:
cc @getchoo since you're apparently reviewing nix stuff for niri now

sodiboo:
what how i branched off like yesterday
image

sodiboo:
oh wait no. i guess it just means "this branch isn't based on the literal latest commit" it doesn't say anything about conflicts, and i guess that doesn't prevent merging? i feel like i don't usually see it though. hm. weird.

getchoo (approved):

PR #860, opened 2024-12-11
Bump MSRV to `1.80` and Upgrade `zbus` and `async-io`

bbb651:
Part of clickable screenshot notifications. I intended to do everything in one PR but I made a lot of dumb mistakes (e.g. not realizing the type `#[zbus::proxy]` generates is suffixed with `Proxy`) and had some trouble wrapping my head around the control flow of dbus (I tried to plug it in to `calloop` not realizing it's [impossible](https://github.com/rust-windowing/winit/issues/3592#issuecomment-2016802233) without a separate thread). I'll probably come back to it in a couple of days.


src/utils/spawning.rs
```diff
@@ -424,10 +424,9 @@ mod systemd {
 
         trace!("waiting for JobRemoved");
         for message in signals {
-            let body: (u32, OwnedObjectPath, &str, &str) =
-                message.body().context("error parsing signal")?;
+            let header = message.header();
```
YaLTeR, line 427: Is this right? (I'm not sure, it's just weird that body changed to header)

YaLTeR (commented):
> I intended to do everything in one PR

Upgrading zbus in a separate PR like this is good. I'll need to test it out before merging later.

src/utils/spawning.rs
bbb651: You're totally right, I was going through all of the boring changes and wasn't paying enough attention, I looked for a method returning `ObjectPath` not realizing this was deserializing and the path came from the data.

(eghh apparently `Resolve conversation` deletes your comment, some great UX github)

bbb651:
Do you have a MSRV policy? zbus 5 requires `1.80`, so if it's a problem I'll downgrade it to 4.

Not sure how accurate it is but I managed to generate this, it doesn't seem like there are a lot of distros on`1.77..=1.79`, they are either rolling/relatively fresh or very old:

YaLTeR:
> Do you have a MSRV policy? zbus 5 requires 1.80, so if it's a problem I'll downgrade it to 4.

Sort of. I try to stick to some Rust version for people that might be on not super up-to-date distros, and not bump it unnecessarily. This is one of the reasons why zbus was stuck on 3.

> Not sure how accurate it is but I managed to generate this, it doesn't seem like there are a lot of distros on1.77..=1.79, they are either rolling/relatively fresh or very old:

Thanks, this is useful. I checked Ubuntu package index, looks like 24.04 is on 1.80. So let's bump to 1.80 I guess.

YaLTeR:
I merged the MSRV bump into main, so now only the zbus commit is needed here.

bbb651:
I've been trying to rebase to remove these 2 commits for more than an hour I think I'll just open a new pr, the co-authored commits have the exact same content so from my admittedly fuzzy understanding of git reordering the commit to be after them should work, but the rebase causes conflicts and when resolving them it modified every commit after?..

YaLTeR:
You could likely get what you want with `git rebase -i` then manually dropping the two first commits. Well, I could do it for you too, whatever works :)

YaLTeR:
Pushed the update: 0113292cf6a77aea55739c7e779a891786aa35bd

Issue #122, opened 2024-01-22
Floating window layer


YaLTeR:
I'm thinking always on top, in global coordinate space, similar to the mouse cursor.

superiums:
i think this is thinkable. 

we need floating window in wm , because we need to see some window always fixed, instead of scroll away or disappeared.

add one floating layer , is one solution. but i think there is another:

- add a fix/sticky property.

window who has this  property, always display at the same position while scrolling.


and we could have these two type of fix/sticky:
- sticky in the workspace
- stikcy in the screen (all workspaces)

maybe this is more simple than floating layer in coding.

### using examples
- chat window, need to be fixed/sticky in all workspaces.
- dev doc window, with 5 editor windows, need to be sticky in workspace while focus editor window.
- video playing window,  need to be sticky in workspace/screen while doing some other work.

YaLTeR:
For sticky in workspace there's a curious suggestion in https://github.com/YaLTeR/niri/discussions/149. For globally sticky, that's how I imagine the floating layer to work in niri. Thanks for the examples, they will be good to keep in mind.

jgregoire:
I'm not sure if this is possible, but I would love to see two features:

1. New/independent floating windows default to centered in display, can be moved around. They stay in place when we pan the view left/right.
2. Child floating windows (like dialog popups) are centered on their parent window, and move with their parent when we pan the view left/right.

YaLTeR:
The 2. will only make sense for *modal* dialogs which block input to the parent window, however the *modal* state is not actually communicated in any of the stable Wayland protocols. It's in private gtk-shell and KDE, and there are MRs into wayland-protocols to make it into a public protocol.

KiaraGrouwstra:
one use-case for which i would like to make a window rule to make a window float is modal dialogs of keepassxc's browser extension.
when using the browser extension to select a credential, the native application will open a modal dialog asking the user to grant permission.
presently, using niri this dialog is opened next to keepassxc, as a new tiled window.
from a UX perspective, this is unfortunate since i do not even have any indication that such a window has opened.

i would prefer to see that addressed by allowing to make a window float from the window rules.
additionally, in hyprland i used a workspaces bar plugin that could visually mark workspaces wanting attention (which [they called](https://github.com/Alexays/Waybar/wiki/Module:-Hyprland#style) 'urgent'). i feel this was also useful to attend me to newly opened windows wanting attention as well (in case they lack focus, i.e. when no floating window rule has been created for them).

[...]

PR #871, opened 2024-12-15
Floating windows

YaLTeR:
So fed up with scrolling tiles. I'm converting niri into a floating WM.

Just kidding. This is a floating window layout on top of the scrollable-tiling layout. The floating layout does not scroll.

Most of the basics are done, though plenty of work remains; I'm focusing on this as time permits. No need to list every missing thing quite yet, but feel free to post if something is particularly annoying you so I can prioritize it.

Windows will auto-float if they have a parent (e.g. dialogs) or fixed size (e.g. splash screens). Otherwise, you can use the new `toggle-window-floating` bind or right click during an interactive move. You can also use the new `open-floating true/false` window rule to either force auto-float a window on open or disable auto-floating for a window.

Use `switch-focus-between-floating-and-tiling` (taking name suggestions) to switch between the two layouts. When focused on the floating layout, binds (like `focus-column-right`) will operate on the floating windows.

Resolves: #122


alicealysia:
toggle-floating-focus

YaLTeR:
Is that clear enough of a name? I feel like it might not be, especially to people new to standalone WMs.

alicealysia:
Hmmm. You're right, I think your name is much clearer. The only motivation on my end is brevity. Maybe I should finally try making a settings app.

20after4:
FWIW this is working well for me. 

I kind of expected the floating layer to be shared between workspaces. I'm not sure why I had that expectation and I don't have a good argument for why floating should be global as apposed to workspace specific.

quadbyte:
Working great here too.

What about "focus-next-layout" instead of "switch-focus-between-floating-and-tiling" ?

YaLTeR:
> FWIW this is working well for me.

Good; any missing things that you bump into more frequently than others?

> I kind of expected the floating layer to be shared between workspaces. I'm not sure why I had that expectation and I don't have a good argument for why floating should be global as apposed to workspace specific.

I'm planning to add that separately later for always-on-top/pinned picture-in-picture and such windows.

> What about "focus-next-layout" instead of "switch-focus-between-floating-and-tiling" 

Hm, might be easy to confuse with the keyboard layout binds.

20after4:
> Good; any missing things that you bump into more frequently than others?

The main thing I really miss with niri the traditional `alt-tab` behavior that will switch window focus in a MRU stack. That isn't really related to this PR though.

Only other thing I can think of is a suggestion for positioning multiple subsequent windows onto the floating layer:

It seems like the default placement of floating windows is to center the window on screen and multiple windows get stacked in the same position.  If I launch, for example, 2 calculator windows, using a window rule to make them start floating, then they are positioned exactly on top of each other. It might be nicer if each subsequent window had a slight offset down and to the right, so that they arrange in a "cascade" layout or something.  

Thinking it through, however, there are probably all kinds of edge cases if you start down the rabbit hole of trying to select optimal placement for various window types and sizes. Centered is probably the safest option and my suggestion might only be applicable to small windows like calculator.

YaLTeR:
> The main thing I really miss with niri the traditional alt-tab behavior that will switch window focus in a MRU stack. That isn't really related to this PR though.

Yeah that can be separate with a separate discussion like which windows it should include across monitors/workspaces/tiling/floating.

> It seems like the default placement of floating windows is to center the window on screen and multiple windows get stacked in the same position. If I launch, for example, 2 calculator windows, using a window rule to make them start floating, then they are positioned exactly on top of each other. It might be nicer if each subsequent window had a slight offset down and to the right, so that they arrange in a "cascade" layout or something.
>
> Thinking it through, however, there are probably all kinds of edge cases if you start down the rabbit hole of trying to select optimal placement for various window types and sizes. Centered is probably the safest option and my suggestion might only be applicable to small windows like calculator.

Yeah, I thought about it a bit and came to the conclusion that there are probably edge cases. Could do some more research into what Mutter does.

YaLTeR:
Alright, I finished everything I wanted, save for a few minor things that I'll do in the next few days. This PR is ready for general testing. Please report any bugs, issues, annoyances, or missing things. I'll consider them either for this PR or for further work.

wiki/Configuration:-Window-Rules.md
```diff
@@ -235,6 +256,21 @@ window-rule {
...
 
+#### `default-window-height`
+
+<sup>Since: next release</sup>
+
+Set the default height for the new window if it opens as floating.
+
+```kdl
+// Make Alacritty take 50% of the screen height on opening.
+window-rule {
```
phisch, line 267: Description above says it sets the height if it is opened as floating, but this rule doesn't apply to only floating windows.

YaLTeR: I actually intend to make it apply to all windows; currently it should only like half apply to scrolling ones I think

JaCoB1123:
I started using the branch yesterday and I'm already really happy with the feature. Some pop-ups that have been annoying me since switching to niri, are now behaving nicely. 👍 Thanks a lot 😀

What I'm missing is a way to set the default position for a popup. For example the Picture-in-Picture window of firefox always opens centered and I'd prefer it in the lower right corner. Ideally the position would still use the parent window's screen though (so no absolute coordinates across all my monitors).

Also thumbs up for the documentation. I found everything I needed and the examples are very relevant. 👍

YaLTeR:
@JaCoB1123 Thanks! Default floating position has actually been requested yesterday by @phisch too. The commits I just pushed have a new `default-floating-position` window rule for that. There's no absolute coordinates.

YaLTeR:
Thanks everyone, time to get more testing on main!

PR #872, opened 2024-12-15
layout/monitor: Add next monitor to focus/move to

julianschuler:
First of all, thank your for this awesome project!

This PR adds the option to focus/move to the next monitor.
It is mainly intended for cycling between multiple monitors using a single key and wraps around to the first monitor after the last one is reached.

Currently, the order follows the order of outputs given by Smithay, please let me know if a different order is desired instead.
Also let me know if I should add a similar option to focus/move to the previous monitor.


YaLTeR (commented):
Thanks, looks reasonable.

> Currently, the order follows the order of outputs given by Smithay, please let me know if a different order is desired instead.

That order is nondeterministic based on how the outputs appear or are plugged in. I think it's better to sort outputs by name like elsewhere in the code to get a stable order every time.

```
let outputs: Vec<_> = self.global_space.outputs().collect();
outputs.sort_unstable_by(|a, b| a.name.compare(&b.name));
```

> Also let me know if I should add a similar option to focus/move to the previous monitor.

Yeah, let's do that.

YaLTeR:
Updated, thanks

PR #885, opened 2024-12-21
Check for layer surface under cursor when clicking

calops:
Currently, windows don't get focus when clicking on them through a clickthrough layer. This PR fixes that by only considering layers under the cursor if we fall in an input area.

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -2408,7 +2408,19 @@ impl Niri {
 
         // Check if some layer-shell surface is on top.
         let layers = layer_map_for_output(output);
-        let layer_under = |layer| layers.layer_under(layer, pos_within_output).is_some();
+        let layer_under = |layer| {
+            layers
+                .layer_under(layer, pos_within_output)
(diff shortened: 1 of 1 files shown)


src/niri.rs
```diff
@@ -2408,7 +2408,18 @@ impl Niri {
 
         // Check if some layer-shell surface is on top.
         let layers = layer_map_for_output(output);
-        let layer_under = |layer| layers.layer_under(layer, pos_within_output).is_some();
+        let layer_under = |layer| {
+            layers
+                .layer_under(layer, pos_within_output)
+                .and_then(|layer| {
+                    let layer_pos_within_output = layers.layer_geometry(layer)?.loc.to_f64();
```
YaLTeR, line 2415: This can remain an unwrap(). It is guaranteed by the fact that we found it in `layer_under()`.

calops: :+1: updated

YaLTeR:
Thanks

PR #899, opened 2024-12-27
Add a window swap operation

rustn00b:
Swap the active window in the active column with a neighboring column's active window.

This PR adds a `swap-window-left` and `swap-window-right` action. They are used to swap the active window
in the active column with the active window in a neighboring column. When there is no neighboring column it
behaves just like the respective `expel-window-....` actions.

This is something quite simple but that I often found missing when using Niri. It was standard in Awesome and easy enough to script in Sway, but I was unable to create a decent niri-ipc based version.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -55,6 +55,7 @@ pub use self::monitor::MonitorRenderElement;
 use self::monitor::{Monitor, WorkspaceSwitch};
 use self::workspace::{OutputId, Workspace};
 use crate::animation::Clock;
+use crate::layout::scrolling::ScrollDirection;
 use crate::niri_render_elements;
 use crate::render_helpers::renderer::NiriRenderer;
 use crate::render_helpers::snapshot::RenderSnapshot;
@@ -2023,6 +2024,13 @@ impl<W: LayoutElement> Layout<W> {
             return;
         };
         monitor.expel_from_column();
+    }
+
+    pub fn swap_window_in_direction(&mut self, direction: ScrollDirection) {
+        let Some(monitor) = self.active_monitor() else {
+            return;
+        };
+        monitor.swap_window_in_direction(direction);
     }
 
     pub fn center_column(&mut self) {
@@ -4382,6 +4390,10 @@ mod tests {
         ]
     }
 
+    fn arbitrary_scroll_direction() -> impl Strategy<Value = ScrollDirection> {
+        prop_oneof![Just(ScrollDirection::Left), Just(ScrollDirection::Right)]
+    }
+
     #[derive(Debug, Clone, Copy, Arbitrary)]
     enum Op {
         AddOutput(#[proptest(strategy = "1..=5usize")] usize),
@@ -4462,6 +4474,9 @@ mod tests {
         },
         ConsumeWindowIntoColumn,
         ExpelWindowFromColumn,
+        SwapWindowInDirection(
+            #[proptest(strategy = "arbitrary_scroll_direction()")] ScrollDirection,
+        ),
         CenterColumn,
         CenterWindow {
             #[proptest(strategy = "proptest::option::of(1..=5usize)")]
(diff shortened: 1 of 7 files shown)


YaLTeR:
Interesting, what do you use this for?

rustn00b:
> Interesting, what do you use this for?

I usually have a column of things I'm working on in the context of some project or other (say 3-5 windows), and to the side a column with a single window (or maybe 2). The work happens in the single "master" window, and I swap the windows into this "master" position based on what needs doing.

What I like about this approach is that it keeps a set of windows available as a palette of sorts, while also affording
decent screen real estate to the current area of focus. But there might be smarter ways of going about it ;)

src/layout/scrolling.rs
```diff
@@ -1740,6 +1751,115 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
+        // recreated with `add_tile`
+        let source_removed = self.remove_tile_by_idx(
+            source_column_idx,
+            source_tile_idx,
+            transaction.clone(),
+            None,
+        );
+
+        // special case when the source column disappears after removing its last tile
+        let adjusted_target_column_idx =
```
YaLTeR, line 1809: I'd make `target_column_idx` mutable and `-= 1` it.

src/layout/mod.rs
```diff
@@ -1952,6 +1953,13 @@ impl<W: LayoutElement> Layout<W> {
         monitor.expel_from_column();
     }
 
+    pub fn swap_window_in_direction(&mut self, direction: ScrollDirection) {
```
YaLTeR, line 1956: Please add this to the `Op` test enum below so it can participate in randomized tests.

YaLTeR (changes requested):
Also don't forget to run `cargo +nightly fmt`.

src/layout/scrolling.rs
rustn00b: ah, but I still need the original `target_column_idx` further along. The `adjusted_target_column_idx` is only meaningful during the code window where the source column may have been removed and has not yet been recreated.
I added an explicit block to make this more obvious.

src/layout/scrolling.rs
```diff
@@ -1749,6 +1759,123 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         new_col.tiles[0].animate_move_from(offset);
     }
 
+    pub fn swap_window_in_direction(&mut self, direction: ScrollDirection) {
+        if self.columns.is_empty() {
+            return;
+        }
+
+        // if this is the first (resp. last column), then this operation is equivalent
+        // to an `consume_or_expel_window_left` (resp. `consume_or_expel_window_right`)
```
YaLTeR, line 1768: Should it be? Upon my initial testing, it feels weird that swap-window will preserve the number of tiles in a column, except when you do it towards an empty spot, at which point it'll break (you can't swap-window it back into a column).

Maybe in this case this should just do nothing?

src/layout/scrolling.rs
```diff
@@ -1749,6 +1759,123 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
+                    false,
+                );
+            }
+        }
+
+        // update the active tile in the modified columns
+        self.columns[source_column_idx].active_tile_idx = source_tile_idx;
+        self.columns[target_column_idx].active_tile_idx = target_tile_idx;
+
+        // Animations
```
YaLTeR, line 1870: The animation for the target tile is a little broken: it jumps vertically. This should be fixed.

YaLTeR (changes requested):
Rebased and fixed two small issues. Could you address these two remaining ones?

src/layout/scrolling.rs
rustn00b: Valid point, as I was putting this together I debated with myself over this one and went with the "swap with an empty column is an expel" approach. But I also found it weird that you couldn't reverse an unintended action by just swapping in the opposite direction. So I'll just make it a no-op instead.

YaLTeR:
Thanks!

PR #904, opened 2024-12-28
Allow workspace names to be changed dynamically

rustn00b:
A trivial PR that adds `set-workspace-name` and `unset-workspace-name` actions.

This is really nifty when combined with the [Waybar Niri workspace module](https://github.com/Alexays/Waybar/wiki/Module:-Niri#workspaces).

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -338,6 +338,14 @@ pub enum Action {
     MoveWorkspaceDown {},
     /// Move the focused workspace up.
     MoveWorkspaceUp {},
+    /// Set the name of the focused workspace.
+    SetWorkspaceName {
+        /// New name for the workspace.
+        #[cfg_attr(feature = "clap", arg())]
+        name: String,
+    },
+    /// Unset the name of te focused workspace.
+    UnsetWorkspaceName {},
     /// Focus the monitor to the left.
     FocusMonitorLeft {},
     /// Focus the monitor to the right.
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -3394,6 +3394,26 @@ impl<W: LayoutElement> Layout<W> {
         monitor.move_workspace_up();
     }
 
+    pub fn set_workspace_name(&mut self, name: String) {
+        // ignore request if the name is an integer
+        if name.parse::<usize>().is_ok() { return }
+
+        // also ignore the request if the name is already used by another workspace
+        if self.find_workspace_by_name(&name).is_some() { return }
+
+        let Some(monitor) = self.active_monitor() else {
+            return;
+        };
+        monitor.set_workspace_name(name);
+    }
+
+    pub fn unset_workspace_name(&mut self) {
+        let Some(monitor) = self.active_monitor() else {
+            return;
+        };
+        monitor.unset_workspace_name();
+    }
+
     pub fn start_open_animation_for_window(&mut self, window: &W::Id) {
         if let Some(InteractiveMoveState::Moving(move_)) = &self.interactive_move {
             if move_.tile.window().id() == window {
--- a/src/layout/monitor.rs
+++ b/src/layout/monitor.rs
@@ -658,6 +658,16 @@ impl<W: LayoutElement> Monitor<W> {
         }
     }
 
+    pub fn set_workspace_name(&mut self, name: String) {
+        let ws = &mut self.workspaces[self.active_workspace_idx];
+        ws.name.replace(name);
+    }
+
+    pub fn unset_workspace_name(&mut self) {
+        let ws = &mut self.workspaces[self.active_workspace_idx];
+        ws.name.take();
+    }
+
     pub fn consume_into_column(&mut self) {
         self.active_workspace().consume_into_column();
     }
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1221,6 +1221,8 @@ pub enum Action {
     MoveColumnToWorkspace(#[knuffel(argument)] WorkspaceReference),
     MoveWorkspaceDown,
     MoveWorkspaceUp,
+    SetWorkspaceName(#[knuffel(argument)] String),
+    UnsetWorkspaceName,
     FocusMonitorLeft,
     FocusMonitorRight,
     FocusMonitorDown,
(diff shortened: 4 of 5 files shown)


niri-ipc/src/lib.rs
```diff
@@ -338,6 +338,14 @@ pub enum Action {
     MoveWorkspaceDown {},
     /// Move the focused workspace up.
     MoveWorkspaceUp {},
+    /// Set the name of the focused workspace.
+    SetWorkspaceName {
+        /// New name for the workspace.
+        #[cfg_attr(feature = "clap", arg())]
+        name: String,
+    },
+    /// Unset the name of te focused workspace.
```
YaLTeR, line 347:
```suggestion
    /// Unset the name of the focused workspace.
```

niri-ipc/src/lib.rs
```diff
@@ -338,6 +338,14 @@ pub enum Action {
...
+    UnsetWorkspaceName {},
```
YaLTeR, line 348: Could you add a `WorkspaceReferenceArg` optional argument here? Like in `FocusWorkspace`.

src/layout/mod.rs
```diff
@@ -3394,6 +3394,26 @@ impl<W: LayoutElement> Layout<W> {
         monitor.move_workspace_up();
     }
 
+    pub fn set_workspace_name(&mut self, name: String) {
+        // ignore request if the name is an integer
+        if name.parse::<usize>().is_ok() { return }
```
YaLTeR, line 3399: Is this needed? I believe you can name workspaces as numbers in the config.

src/layout/mod.rs
```diff
@@ -3394,6 +3394,26 @@ impl<W: LayoutElement> Layout<W> {
...
+
+        // also ignore the request if the name is already used by another workspace
+        if self.find_workspace_by_name(&name).is_some() { return }
```
YaLTeR, line 3402: You'll need to run `cargo +nightly fmt`, the CI checks for that

src/layout/mod.rs
```diff
@@ -3394,6 +3394,26 @@ impl<W: LayoutElement> Layout<W> {
...
+
+        let Some(monitor) = self.active_monitor() else {
+            return;
+        };
+        monitor.set_workspace_name(name);
```
YaLTeR, line 3407: `self.active_workspace_mut()` and same below

src/layout/mod.rs
```diff
@@ -3394,6 +3394,26 @@ impl<W: LayoutElement> Layout<W> {
...
+    }
+
+    pub fn unset_workspace_name(&mut self) {
+        let Some(monitor) = self.active_monitor() else {
+            return;
+        };
+        monitor.unset_workspace_name();
```
YaLTeR, line 3414: Check `self.unname_workspace`. There's more logic that needs to run. Perhaps, extract it to `unname_workspace_by_id(id: WorkspaceId)` and call the logic from both places.

YaLTeR (changes requested):

src/layout/mod.rs
```diff
@@ -3394,6 +3394,26 @@ impl<W: LayoutElement> Layout<W> {
         monitor.move_workspace_up();
     }
 
+    pub fn set_workspace_name(&mut self, name: String) {
```
YaLTeR, line 3397: Also please add these two operations to the `Op` enum below so they can participate in the randomized tests.

src/layout/mod.rs
rustn00b: This was intended to avoid confusion with other auto-named workspaces. Having two different namespaces keeps things simple. I'm not sure how things are dealt with (possibly very well ;-) when a wkspc is automatically assigned a value that might already be used as an assigned name.

niri-ipc/src/lib.rs
rustn00b: This one is more than I anticipated. I'm trying to figure out how to lean on the existing functions to deal with the Layout/Monitor/Workspace. I'm probably looking at it the wrong way. Will keep at it though.

src/layout/mod.rs
rustn00b: Indeed, if the workspace name can be reset on _any_ workspace (as opposed to just the current one), then a named empty workspace that loses its name needs to be cleaned  up.

src/layout/mod.rs
rustn00b: The two tests I put in are kind of trivial, they may clear your bar for completeness/quality.

src/layout/mod.rs
```diff
@@ -3745,6 +3792,60 @@ impl<W: LayoutElement> Layout<W> {
...
+
+        let Some(ws) = self.active_workspace_mut() else {
+            return;
+        };
+
+        ws.name.replace(name);
+
+        let wsid = ws.id();
+
+        // ensure last workspace is "unnamed"
```
YaLTeR, line 3809: It's more fun than this: with `empty_workspace_above_first` option set, you need to do the same check for the *first* workspace.

Do you mind writing two unit tests which set the name of the last and the first workspace respectively? Don't need to check anything manually, just a list of operations and `check_ops()`, it should verify everything already.

src/layout/mod.rs
```diff
@@ -3745,6 +3792,60 @@ impl<W: LayoutElement> Layout<W> {
...
+                }
+            }
+            MonitorSet::NoOutputs { ref mut workspaces } => {
+                if workspaces.last().is_some_and(|last| last.id() == wsid) {
+                    workspaces.push(Workspace::new_no_outputs(
+                        self.clock.clone(),
+                        self.options.clone(),
+                    ));
+                }
+            }
```
YaLTeR, line 3832: `NoOutputs` workspaces must not have empty unnamed workspaces, so this is not needed.

src/layout/monitor.rs
```diff
@@ -383,12 +383,19 @@ impl<W: LayoutElement> Monitor<W> {
...
+            ws.name
                 .as_ref()
                 .map_or(false, |name| name.eq_ignore_ascii_case(workspace_name))
-            {
+        })
+    }
+
+    pub fn unname_workspace_with_filter<P>(&mut self, pred: P) -> bool
+    where
+        P: Fn(&Workspace<W>) -> bool,
```
YaLTeR, line 395: How about instead of accepting a generic predicate here and in `set_workspace_name()`, just always accept a `WorkspaceId`? I think it should make things simpler. And it's not like this can ever be called for multiple workspaces at once (and even then, it would be safer to call it for them one-by-one).

YaLTeR (changes requested):

src/layout/monitor.rs
rustn00b: Tried something, can't say I'm really convinced it's much better. Maybe I'm not looking at it the right way.

YaLTeR:
Pushed a few cleanups, should be good to merge now. Do you mind testing one more time?

rustn00b:
> Pushed a few cleanups, should be good to merge now. Do you mind testing one more time?

Will do!

YaLTeR:
One final change, made the workspace reference a non-positional option to make it less confusing what goes where (i wrote that commit message the opposite way, huh). Let's merge this

YaLTeR:
Thanks!

rustn00b:
> Thanks!

Pleasure! Beyond `cargo test`, I've been dog-fooding this without any issues for now.

PR #912, opened 2025-01-01
dbus: DisplayConfig: support monitor configuration

valpackett:
happy new year :tada: this here PR extends the org.gnome.Mutter.DisplayConfig implementation enough to make the GNOME display settings work

[niri-gnome-monitor-settings.webm](https://github.com/user-attachments/assets/e17cf955-e95c-4d34-ad93-94f88fe2d682)

(`XDG_CURRENT_DESKTOP=gnome gnome-control-center`)


src/dbus/mutter_display_config.rs
```diff
@@ -45,6 +82,137 @@ pub struct LogicalMonitor {
 
 #[dbus_interface(name = "org.gnome.Mutter.DisplayConfig")]
 impl DisplayConfig {
+    async fn get_resources(
```
YaLTeR, line 101: Is this method actually needed? The panel shows up as is for me, and I cannot actually find any calls to  `GetResources` in gnome-control-center: https://gitlab.gnome.org/search?search=GetResources&nav_source=navbar&project_id=660&group_id=8&search_code=true&repository_ref=main

src/dbus/mutter_display_config.rs
```diff
@@ -123,22 +291,11 @@ impl DisplayConfig {
...
-                    niri_ipc::Transform::Flipped180 => 6,
-                    niri_ipc::Transform::Flipped270 => 7,
-                };
-
                 let logical_monitor = LogicalMonitor {
                     x: logical.x,
                     y: logical.y,
                     scale: logical.scale,
-                    transform,
+                    transform: logical.transform as u32,
```
YaLTeR, line 312: I think it might be a good idea to leave this match as is, since it's converting into what is expected for the D-Bus API. And the `as u32` is whatever we happened to use in niri_ipc.

src/dbus/mutter_display_config.rs
```diff
@@ -308,7 +306,8 @@ impl DisplayConfig {
         // Sort by connector.
         monitors.sort_unstable_by(|a, b| a.0.names.0.cmp(&b.0.names.0));
 
-        let (monitors, logical_monitors) = monitors.into_iter().unzip();
+        let (monitors, logical_monitors): (Vec<_>, Vec<_>) = monitors.into_iter().unzip();
+        let logical_monitors = logical_monitors.into_iter().flatten().collect();
```
YaLTeR, line 326: Seems like at this point it's better to make two Vecs, then manually `for (a, b) in monitors.into_iter()` and push the items in.

src/dbus/mutter_display_config.rs
```diff
@@ -314,11 +330,110 @@ impl DisplayConfig {
...
+        }
+        for (_, output) in current_conf.iter() {
+            if let Some(msgs) = messages.remove(&output.name) {
+                for msg in msgs.into_iter() {
+                    self.send(&output.name, msg)?;
+                }
+            } else {
+                self.send(&output.name, OutputAction::Off)?;
+            }
+        }
```
YaLTeR, line 415: The config should be applied atomically, i.e. as one single message on the channel for all changes.

YaLTeR (changes requested):
Hey, thanks, this is pretty cool! Left some comments, also prepare to upgrade to zbus 5 since I'll be merging that PR soon.

src/dbus/mutter_display_config.rs
valpackett: The thing that calls it is [gnome-rr](https://github.com/GNOME/gnome-desktop/blob/cf8481db03400a80188b778cc230cd0a2d7d5347/libgnome-desktop/gnome-rr/gnome-rr-screen.c#L507) but I forgot where I encountered usage of that API… ah. Not for Settings, but [it's required for gsd-color](https://github.com/NixOS/nixpkgs/issues/123772#issuecomment-845005628). I'd like to get Night Light working, at least via an external client that would read g-s-d's temperature output and do `wlr-gamma-control-unstable-v1` based on it, but for that `gsd-color` must actually work

src/dbus/mutter_display_config.rs
valpackett: Sure.. didn't want to change the existing code too much but it would be better

src/dbus/mutter_display_config.rs
YaLTeR: Maybe in a separate PR then?

valpackett:
Fixed commit descriptions, improved style according to discussions above, rebased on top of the zbus 5 changes, added fractional scales :rocket:

niri-ipc/src/lib.rs
```diff
@@ -800,6 +800,22 @@ pub enum Transform {
     Flipped270,
 }
 
+impl Transform {
+    /// Return numeric identifier matching the `transform` enum of the `wl_output` interface.
+    pub fn to_wayland_id(self) -> u32 {
```
YaLTeR, line 805: This is not a "to wayland id", this is a "to mutter-specific private dbus interface id". So it should be left inside mutter_display_config.rs.

src/dbus/mod.rs
```diff
@@ -50,7 +50,20 @@ impl DBusServers {
...
 
         if is_session_instance || config.debug.dbus_interfaces_in_non_session_instances {
-            let display_config = DisplayConfig::new(backend.ipc_outputs());
+            let (to_niri, from_display_config) = calloop::channel::channel();
+            let display_config = DisplayConfig::new(to_niri, backend.ipc_outputs());
+            niri.event_loop
+                .insert_source(from_display_config, move |event, _, state| match event {
+                    calloop::channel::Event::Msg(messages) => {
+                        for (output, actions) in messages.into_iter() {
+                            for action in actions.into_iter() {
```
YaLTeR, line 59: I think these `.into_iter()` are unnecessary in a for loop?

src/dbus/mutter_display_config.rs
```diff
@@ -316,11 +332,102 @@ impl DisplayConfig {
...
+                    .values()
+                    .find(|o| o.name == connector)
+                    .is_none()
+                {
+                    return Err(zbus::fdo::Error::Failed(format!(
+                        "Connector '{}' not found",
+                        connector
+                    )));
+                }
+                let msgs = vec![
```
YaLTeR, line 357: In comparison, the wlr-output-management impl creates and sets a complete new output configuration from scratch. Wouldn't that also be more appropriate here?

As far as I understand, the functional difference is that it is applied all at once, without repositioning outputs every step. That said, I think your approach should result in the same outcome anyway? So I guess this code can stay as is.

valpackett: Oh, that's possible? Right, I'm not restricted to ipc related API now that I'm using the channels..

niri-ipc/src/lib.rs
valpackett: The actual numbers do match wayland.xml, not sure why you think mutter could ever switch to different numbers :roll_eyes:

YaLTeR: It couldn't. But this code was written specifically to convert according to the mutter xml, and isn't used anywhere else

src/dbus/mutter_display_config.rs
valpackett: Had to extract the find-or-create-config-by-name part from `apply_transient_output_config` but I think it does look better this way, creating the `niri_config::Output` from scratch.

Actually the outcome wasn't *exactly* the same: `apply_transient_output_config` was calling `reload_output_config` after every action…

YaLTeR:
Thanks, seems to work fine, and the new apply config looks good. I changed the night light property to `false` though because it's not implemented yet so the panel won't actually work.

PR #934, opened 2025-01-05
Set is-active-in-column to true for unmapped windows

TheZoq2:


wiki/Configuration:-Window-Rules.md
```diff
@@ -194,6 +194,8 @@ Matches the window that is the "active" window in its column.
 Contrary to `is-active`, there is always one `is-active-in-column` window in each column.
 It is the window that was last focused in the column, i.e. the one that will gain focus if this column is focused.
 
+<sup>Since: next release</sup> This rule will match true during the initial window opening.
```
YaLTeR, line 197:
```suggestion
<sup>Since: next release</sup> This rule will match `true` during the initial window opening.
```

PR #961, opened 2025-01-12
use standard padding syntax instead of implementing our own

sodiboo:
the padding of the two-digit-month can be implemented much more concisely using `std::fmt` syntax.


YaLTeR:
Right. Thanks

PR #963, opened 2025-01-12
Support `WAYLAND_SOCKET` in winit backend

bbb651:
I know of a single compositor that supports `WAYLAND_SOCKET` but not `WAYLAND_DISPLAY`: https://gitlab.freedesktop.org/mstoeckl/windowtolayer

This should also make niri more robust against accidentally setting `WAYLAND_SOCKET` when starting as a session, before programs could fail if they preffered `WAYLAND_SOCKET` over `WAYLAND_DISPLAY`

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -69,8 +69,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     if cli.session {
         // If we're starting as a session, assume that the intention is to start on a TTY. Remove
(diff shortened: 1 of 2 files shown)


YaLTeR:
Makes sense, thanks.

> https://gitlab.freedesktop.org/mstoeckl/windowtolayer

Huh, this is kinda like the current gtk4-layer-shell, but generic?

PR #970, opened 2025-01-13
make niri-session POSIX compatible

### LINKED ISSUE #969: niri-session fails to run if `/bin/sh` is not bash (dinit system)
```
$ niri-session
systemctl: not found
/usr/bin/niri-session: systemctl: not found
/usr/bin/niri-session: systemctl: not found
/usr/bin/niri-session: systemctl: not found
/usr/bin/niri-session: systemctl: not found
/usr/bin/niri-session: systemctl: not found
/usr/bin/niri-session: systemctl: not found
```

according to shellcheck:

```
Line 15:
if hash systemctl &> /dev/null; then
                  ^-- SC3020 (warning): In POSIX sh, &> is undefined.
 
Line 44:
elif hash dinitctl &> /dev/null; then
                   ^-- SC3020 (warning): In POSIX sh, &> is undefined.
 
Line 46:
    if ! pgrep -u $(id -u) dinit &> /dev/null; then
                  ^-- SC2046 (warning): Quote this to prevent word splitting.
                                 ^-- SC3020 (warning): In POSIX sh, &> is undefined.
 
Line 52:
    if dinitctl --user is-started niri &> /dev/null; then
                                       ^-- SC3020 (warning): In POSIX sh, &> is undefined.
```

changing the shebang of the script to `#!/bin/bash` works around the issue

### System Information
* niri version: niri 25.01 (unknown commit)
* Distro: Chimera Linux (still being packaged)
* GPU: N/A, but Intel Corporation Raptor Lake-P [UHD Graphics] (rev 04)
* CPU: N/A, but 13th Gen Intel(R) Core(TM) i5-1340P

z-erica:
&> redirection is undefined in POSIX sh (should fix #969)

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -5,14 +5,14 @@ if [ -n "$SHELL" ] &&
    ! (echo "$SHELL" | grep -q "false") &&
    ! (echo "$SHELL" | grep -q "nologin"); then
   if [ "$1" != '-l' ]; then
-    exec bash -c "exec -l '$SHELL' -c '$0 -l $*'"
+    exec sh -c "exec -l '$SHELL' -c '$0 -l $*'"
   else
     shift
   fi
 fi
 
 # Try to detect the service manager that is being used
-if hash systemctl &> /dev/null; then
+if hash systemctl >/dev/null 2>&1; then
     # Make sure there's no already running session.
     if systemctl --user -q is-active niri.service; then
       echo 'A niri session is already running.'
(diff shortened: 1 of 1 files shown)


YaLTeR:
As I mentioned in the issue, please leave /bin/sh and change the &> instead. It's supposed to be POSIX compatible

YaLTeR:
> sure thing! i also changed the bash invocation early on, since that was what made me assume it was intended to be a bash script

Hmm, so this is a good question. Looks like the `-l` argument to `exec` is also bash-specific:

> If the -l option is supplied, the shell places a dash at the beginning of the zeroth argument passed to command. This is what login(1) does.

```
$ dash
$ exec -l bash -c 'echo $0'
dash: 1: exec: -l: not found
```

So maybe it's better to leave `exec bash` on that line?

cc @kchibisov, but I think this invocation was cargo culted from other similar scripts

kchibisov:
The script pretty much does what the gnome shell wrapper does for the same thing, and was written the same way. There's generally no way to easy set the zero argument unless you use bash/zsh/etc.

You may try to pass the `-l` to shell directly though, if the goal is to use `sh`.

YaLTeR:
Let's just leave `exec bash` then for now

resources/niri-session
```diff
@@ -5,14 +5,14 @@ if [ -n "$SHELL" ] &&
    ! (echo "$SHELL" | grep -q "false") &&
    ! (echo "$SHELL" | grep -q "nologin"); then
   if [ "$1" != '-l' ]; then
-    exec bash -c "exec -l '$SHELL' -c '$0 -l $*'"
+    exec sh -c "exec -l '$SHELL' -c '$0 -l $*'"
```
YaLTeR, line 8:
```suggestion
    exec bash -c "exec -l '$SHELL' -c '$0 -l $*'"
```

YaLTeR:
Thanks

Issue #401, opened 2024-05-27
Per-output layout settings


YaLTeR:
It is sometimes useful to change some layout properties per output. For example, one might want a smaller default column width on a bigger monitor. Or always-centering mode restricted to a particular output (https://github.com/YaLTeR/niri/discussions/381).

The obvious way to represent it in the config is to just add the `layout` section to the outputs:

```
output "eDP-1" {
    layout {
        default-column-width { proportion 0.75; }

        border {
            on; // These need an extra "on" flag.
        }
    }
}
```

`niri-config`-wise, we need to implement merging for the `layout` structs (similar to window rules vs. resolved window rules), which will also be useful in the future for config `include`. Also, some way to deny the `on` flag in the global `layout` section without duplicating all of the structures involved.

Stugeh:
Love the project and I'm looking forward to it getting to the point I can replace Hvprland with this. I've begun work on this feature in a way that slightly deviates from the original idea in your description of the issue. Any feedback on the approach so far? I haven't had the chance to use Rust in a 'real' project before.

YaLTeR:
Hey,

> Creating new types for the Output layouts because it makes handling the
layout overriding much simpler down stream by allowing us to just ignore
fields with None instead of writing some custom solution to detect which
fields are/ aren't set by the user since the regular Layout struct
hierarchy has set deafaults for all sub fields in the hierarchy.

I suspect we may need to do the Option refactor either way for the ability to split a config into separate files. Might be worth looking into refactoring the Layout config itself to use Options. Similar to how window rules work, together with ResolvedWindowRules.

Stugeh:
> I suspect we may need to do the Option refactor either way for the ability to split a config into separate files. Might be worth looking into refactoring the Layout config itself

I'll have a look if I can reasonably fit into this feature. I suspect there might be enough work involved that it could be worth being a separate issue from this one or as a part of the multi file config issue.

tino376dev:
Before finding this issue, I looked for a "match output=my-display" window-rule option. Not sure if that would be an alternative solution (having no clue about the code-base).

kRHYME7:
>  Option refactor either way for the ability to split a config into separate files. 

Hi @YaLTeR ,I am pretty curious on how far are we to have a possibility of multi-file configuration?  I would like to incorporate NIRI with our [dotfiles](https://github.com/Hyde-Project/hyde) repo, but it is missing this feature.  

[...]

PR #973, opened 2025-01-13
WIP: per-output default column width

balsoft:
Add a way to set the default column width per-output.

This is useful when there are multiple outputs with wildly different
aspect ratios, such as one vertical and one horizontal; with this change,
one can set the default width to 1.0 on a vertical one and 0.5 on the
horizontal one, making niri much more usable.

I'm making this draft just to see if a change like this would be accepted;

If there's a chance, I'll add some tests, documentation, and maybe think about
the priorities of various defaults a bit more.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -372,6 +372,8 @@ pub struct Output {
     pub variable_refresh_rate: Option<Vrr>,
     #[knuffel(child, default = DEFAULT_BACKGROUND_COLOR)]
     pub background_color: Color,
+    #[knuffel(child)]
+    pub default_column_width: Option<DefaultPresetSize>,
 }
 
 impl Output {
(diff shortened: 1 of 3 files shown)


YaLTeR:
Hey, so I wanted to do this generally, by adding a layout {} section to the output: https://github.com/YaLTeR/niri/issues/401 It's quite a bit more work than just adding 1 setting unfortunately. But I don't want to add a config option that will be obviously obsoleted later.

I'm not sure what's the best way to proceed. You could of course change this PR to add this one option but inside `layout {}` so it would naturally extend into the full thing when we get to it. But also maybe we should just wait and do the full Layout config right away?

YaLTeR:
This has been a thing for a while https://niri-wm.github.io/niri/Configuration%3A-Outputs.html#layout-config-overrides

PR #975, opened 2025-01-14
add write-to-disk argument to screenshot actions

### LINKED ISSUE #916: Make Ctrl-C in screenshot UI only copy to clipboard without saving to file
Discussed in https://github.com/YaLTeR/niri/discussions/901

Probably also add an argument like `niri msg action screenshot-window --write-to-disk=false` and to the bind `{ screenshot-window write-to-disk=false; }`.

sornas:
Closes #916

I added a `write_to_disk: bool` parameter to most of the screenshot actions and a condition in `save_screenshot`. It was fairly straightforward but I would probably want to add a test. Where do you think that would go?

An alternative that I thought about at the end is to store the screenshot in the `tmp` directory instead if `!write_to_disk`, because my understanding is that in order to show an image preview in the notification we still need a file on disk _somewhere_. That would be very easy to implement since all of the parameter stuff would be the same.

Also, maybe we should rename `save_screenshot` since the name may be a bit misleading now.

Let me know what you think!


src/ui/screenshot_ui.rs
```diff
@@ -34,9 +34,11 @@ const FONT: &str = "sans 14px";
 const BORDER: i32 = 4;
 const TEXT_HIDE_P: &str =
     "Press <span face='mono' bgcolor='#2C2C2C'> Space </span> to save the screenshot.\n\
+     Press <span face='mono' bgcolor='#2C2C2C'> Ctr+C </span> to only copy the screenshot to the clipboard.\n\
```
YaLTeR, line 37: Don't think this is important enough to put here

src/niri.rs
```diff
@@ -4633,28 +4636,28 @@ impl Niri {
...
+                            warn!("error saving screenshot image: {err:?}");
+                        }
                     }
+                } else {
+                    debug!("not saving screenshot to disk");
                 }
-            } else {
-                debug!("not saving screenshot to disk");
             }
-
```
YaLTeR, line 4657: Blank line disappeared

niri-ipc/src/lib.rs
```diff
@@ -167,7 +167,11 @@ pub enum Action {
     /// Open the screenshot UI.
     Screenshot {},
     /// Screenshot the focused screen.
-    ScreenshotScreen {},
+    ScreenshotScreen {
+        /// Save the screenshot in addition to putting it in your clipboard.
```
YaLTeR, line 171:
```suggestion
        /// Write the screenshot to disk in addition to putting it in your clipboard.
        ///
        /// The screenshot is saved according to the `screenshot-path` config setting.
```
and same below

niri-ipc/src/lib.rs
```diff
@@ -167,7 +167,11 @@ pub enum Action {
...
+        #[cfg_attr(feature = "clap", arg(short, long, action = clap::ArgAction::Set, default_value_t = true))]
```
YaLTeR, line 172: How does this look on the CLI? Do you have to do `--write-to-disk=false`? Or is there a `--no-write-to-disk`? Could you show what the help looks like?

src/niri.rs
```diff
@@ -4633,28 +4636,28 @@ impl Niri {
...
-            if let Some(path) = path {
-                debug!("saving screenshot to {path:?}");
-
-                if let Some(parent) = path.parent() {
-                    if let Err(err) = std::fs::create_dir(parent) {
-                        if err.kind() != std::io::ErrorKind::AlreadyExists {
-                            warn!("error creating screenshot directory: {err:?}");
+            if write_to_disk {
+                if let Some(path) = path {
```
YaLTeR, line 4640: I think we want to print the debug message below in both cases, so something like: `if write_to_disk.then_some(path).flatten() {`

YaLTeR: Actually, even better idea: just do `path = None` if `!write_to_disk` at the top of this function, don't even call `make_screenshot_path()`

YaLTeR (changes requested):
Thanks, looks reasonable. Some small changes.

> It was fairly straightforward but I would probably want to add a test. Where do you think that would go?

Hm, not sure what you can usefully and reasonably easily test here. What did you have in mind?

> An alternative that I thought about at the end is to store the screenshot in the tmp directory instead if !write_to_disk, because my understanding is that in order to show an image preview in the notification we still need a file on disk somewhere.

There are actually some other ways to show an image: https://specifications.freedesktop.org/notification-spec/latest/icons-and-images.html But you'd need to check the actual notification daemons like mako to see what they support.

Anyhow I don't think we should save to /tmp, at least not in this PR. Maybe something for later after considering our options.

> Also, maybe we should rename save_screenshot since the name may be a bit misleading now.

Hmm, I guess, but I don't immediately have any good name suggestions so let's leave it as is.

sornas:
> > It was fairly straightforward but I would probably want to add a test. Where do you think that would go?
>
> Hm, not sure what you can usefully and reasonably easily test here. What did you have in mind?

No, you're right, I'm also having trouble figuring out what would be tested.

niri-ipc/src/lib.rs
YaLTeR: The second part of the comment would be nice to include too

niri-ipc/src/lib.rs
sornas: Apparently clap does not support automatic `--no-XYZ` flags (https://github.com/clap-rs/clap/issues/815) so we would have to do it manually. I can implement that if you want. As I said, I have no strong opinion.

YaLTeR: Eh, I suppose it's fine as is then

YaLTeR: Bikeshed: should the short flag be -d for disk perhaps rather than -w for write? Or would that be more confusing

sornas: `-d` also works. I think it's fine to change it later if another flag is added that fits better, especially for the short flag.

sornas:
That should be all review comments. `if write_to_disk.then(|| match ...).flatten()` is a bit more crowded than I would've liked but I think it's ok. At least it feels better than `let path = if write_to_disk { match ... } else { None };`.

YaLTeR:
Missed the: "The screenshot is saved according to the `screenshot-path` config setting."

sornas:
Sorry, I can't find where that would be. (I tried grepping for it.)

YaLTeR:
My suggested change from earlier: https://github.com/YaLTeR/niri/pull/975#discussion_r1914431559

image

sornas:
Oh! I misunderstood the diff. Sorry about that.

sornas:
```
$ cargo r -- msg action screenshot-window --help
     Running `target/debug/niri msg action screenshot-window --help`
Screenshot the focused window

Usage: niri msg action screenshot-window [OPTIONS]

Options:
      --id <ID>
          Id of the window to screenshot.
          
          If `None`, uses the focused window.

  -d, --write-to-disk <WRITE_TO_DISK>
          Write the screenshot to disk in addition to putting it in your clipboard.
          
          The screenshot is saved according to the `screenshot-path` config setting.
          
          [default: true]
          [possible values: true, false]

  -h, --help
          Print help (see a summary with '-h')
```

YaLTeR:
Thanks!

emiham:
Is there a reason why this flag is not available for the regular `screenshot` action?

sornas:
@emiham This was a while ago but I believe the `screenshot-screen` and `screenshot-window` actions take the screenshot immediately, so you want to be able to control `write-to-disk` with a flag, while the `screenshot` action merely _opens_ the screenshot UI where you can then choose if you want to write to disk or not with either Space (for `write-to-disk`) or Ctrl-C (for `write-to-disk false`)

Issue #861, opened 2024-12-12
RTL language (right justification) support


YaLTeR:
Also see discussion: https://github.com/YaLTeR/niri/discussions/658

RTL languages like Arabic expect the visual layout to be mirrored horizontally somewhat. In case of niri, there isn't much layout to mirror: mainly we need right justification where windows instead of spawning and tending to the left, spawn and tend to the right side of the screen.

This is generally useful outside RTL languages, so maybe it could be a setting on its own, but RTL languages should turn it on by default.

Our navigation keys (e.g. focus-column-left, focus-column-right) are for the most part directional so they shouldn't need any special changes. Though, I'm not sure about focus-column-first and focus-column-last. What do text editors do when you hit Home and End on RTL? Does Home go to the start of the line (on the right)?

For determining whether the current language is RTL, here's what GTK does: https://gitlab.gnome.org/GNOME/gtk/-/blob/72718b719387d15a29bf4524bc721eec490b357e/gtk/gtkmain.c#L774

bbb651:
I personally use English as my system language but my native tongue is RTL (Hebrew) so I can answer some of your questions.

> This is generally useful outside RTL languages, so maybe it could be a setting on its own, but RTL languages should turn it on by default.

Like people said in #658, I don't think automatically detecting it is a good idea. If niri itself eventually supports localization it makes sense to follow that, but imo detecting it based on the system will confuse users not aware of the feature's existence.

> What do text editors do when you hit Home and End on RTL? Does Home go to the start of the line (on the right)?

Yes - text editing will often includes mixed RTL and LTR, the way this works by default (i.e. html `dir="auto"`) is *each line* can have it's own text direction based on the text directionality  of the text that starts the line, and spans of text go the other way. In general, there's a distinction between *logical* directions and *physical* directions.

> Our navigation keys (e.g. focus-column-left, focus-column-right) are for the most part directional so they shouldn't need any special changes.

I would imagine they would need to be flipped because the representation of columns should remain the same, but now e.g. right means decreasing the index instead of increasing it.
I think it'll be useful to have variants of commands that do follow logical direction rather than physical direction (`*-next` and `*-previous`), to allow for example to bind column focus to <kbd>super</kbd> + <kbd>j</kbd>/<kbd>k</kbd> or <kbd>super</kbd> + <kbd>n</kbd>/<kbd>p</kbd> (although ultimately not that big of a deal since they can be flipped manually in the config).

YaLTeR:
[...]

PR #977, opened 2025-01-14
Add MRU window navigation actions

rustn00b:
This PR adds `focus-window-mru-previous` and `focus-window-mru-next` actions to
navigate windows in least-recently-used (respectively most-recently-used) order.

This is another quality-of-life modification that many users are probably
already accustomed to from other OSs/DEs/WMs using the Alt/Cmd/Mod-[Shift-]Tab
binding.

Whenever a window is focused, it records a timestamp that be used to sort
windows in MRU order. This timestamp is not updated immediately, but only after
a small delay (lock-in period) to ensure that the focus wasn't transferred to
another window in the meantime. This strategy avoids upsetting the MRU order
with focus events generated by intermediate windows when moving between two
non contiguous windows.

The lock-in delay can be configured using the `focus-lockin-ms` configuration
argument.

Calling either of the `focus-window-mru` actions starts an MRU window traversal
sequence if one isn't already in progress. When a sequence is in progress,
focus timestamps are no longer updated.

A traversal sequence ends when:
- either the `Mod` key is released, the focus then stays on the chosen window 
  and its timestamp is immediately refreshed,
- or if the `Escape` key is pressed, the focus returns to the window that
  initially had the focus when the sequence was started.

Ending the sequence when `Mod` is released works nicely under the assumption
that the `find-focus-mru` actions have a binding that uses `Mod` as the modifier
key. A more general solution would involve defining a binding that triggers on
the _release_ of an arbitrary modifier, combined with some sort of `mru-exit`
(?) action.


YaLTeR:
Some comments before looking in-depth.

First of all, MRU Alt-Tab is certainly one of the things I missed when testing out a blanket `open-floating true` workflow for a week. And you can't implement MRU well outside the compositor.

But the reasons I was so far reluctant to open an issue for this are:
- I'm used to it having a GUI switcher like in GNOME Shell. Especially in the case of MRU across workspaces, it helps when you can visually pick a window without everything moving around underneath you.
- It's unclear how to bind it with the current system because it is not a discrete action. You mention that it kind of assumes Mod currently; I'm not sure it's a good idea to have it bindable at all if it relies on that assumption to work properly. Besides, I'd expect it to be Alt-Tab rather than Mod. And pressing Shift-Tab instead of Tab should go backward, etc. Tbh I'd almost prefer to just hardcode it and not expose as an action for now, until we figure out a better way?

Also, GNOME Shell differentiates between Alt-Tab and Mod-Tab where one of them goes across windows of the same app, and the other one doesn't. Do we need this?

> This timestamp is not updated immediately, but only after
a small delay (lock-in period) to ensure that the focus wasn't transferred to
another window in the meantime. This strategy avoids upsetting the MRU order
with focus events generated by intermediate windows when moving between two
non contiguous windows.

How does this work elsewhere like Mutter/GNOME Shell? Does it also have a lock-in delay? Does it need to be configurable in niri?

rustn00b:
Something similar is available for Sway in `swayr` (with a lockin delay). The delay is nice to ignore focus updates resulting from navigation actions with no "semantic" meaning. For instance, if you have focus-follow-mouse active, the windows you move over while on your way to your intended target probably shouldn't cause your "mental" map of recent windows to diverge with the WM's version. Likewise, with scripts that perform multiple focus updates over IPC when these updates are just a means to designate the target of the following actions. For instance, I have one that resizes a bunch of columns at once by focus each one and then setting their width and then finally restores the focus to the window that originally had the focus when the script was called. As a user I don't need to be aware these focus events ever happened.

With respect to everything moving around on its own, I was kind of happy that this was happening on workspaces that had (too) many windows. Moving through them was a hassle, arguably I should just organize my windows better ;)
I have `hide-when-typing` active and 

I know that I personally missed this feature from swayr and have been really savoring having it back!

I agree with your point regarding the presumption that the next/prev bindings use the `Mod` key. A more generic approach would be to detect the (first?) modifier key for these MRU actions and end the MRU sequence when that modifier key is released. Of course for it to feel right for the user, the modifier would need to be the same for both actions (not an unreasonable assumption).

Hardcoding would mean appropriating the shortcut, which is also a problem if you have users that like things the way they are now! With the PR as it is there shouldn't be any user visible impact if they don't have bindings defined for the `focus-window-mru` actions.

Wrt the Alt/Mod-Tab distinction, iirc it allows navigation in a given application's windows or in all the windows (I haven't used Gnome in a while). I haven't given any thought to how the former is accomplished. Presumably some kind of app-id filter would be needed.

rustn00b:
> Does it need to be configurable in niri?

I played around with several values and settled on 500 millis but since this felt very arbitrary I made it configurable. Granted this would be an obscure setting that the majority of users would ignore except the few who think the default is terrible.

rustn00b:
Been mulling this over, and all I can come up with is a system that involves release key bindings, e.g.:
- 1 binding for `focus-window-mru-previous`, say `Mod-Tab`
- 1 binding for `focus-window-mru-next`,  say `Mod-Shift-Tab`
- 1 _release_ binding on `Mod` that cancels any active MRU traversal.

This would allow users to pick arbitrary modifier and key combos to trigger MRU traversal.

My other ideas (e.g. auto determine leader keys) look as if they lead to clunky designs. In swayr traversal can be configured to expire after given delay (in my experience this is not always very intuitive).

Any thoughts on whether it is worthwhile to start tinkering with a release binding mechanism to support this?

YaLTeR:
> I played around with several values and settled on 500 millis but since this felt very arbitrary I made it configurable.

Could you check what other similar programs or systems in programs use for this? I guess Mutter doesn't have any because it's only relevant for focus-follows-mouse and directional keynav, which it doesn't have. But for tiling WMs it should be relevant.

> Been mulling this over, and all I can come up with is a system that involves release key bindings

Even if you "just" implement release keybindings (I expect this to be quite involved and error-prone on its own), then you kinda just take up the whole Mod release keybinding for something that should honestly happen automatically. That doesn't sound very good. (Besides, Mod release should get bound to overview when that happens.)

> Hardcoding would mean appropriating the shortcut, which is also a problem if you have users that like things the way they are now!

Well yeah but the alternatives so far seem very clunky, making me think that hardcoding is the better solution, at least for now. Not sure

rustn00b:
Wrt the delays chosen for other similar solutions:
- swayr: 750ms "focus lockin_delay" by default [swayr-configuration](https://sr.ht/~tsdh/swayr/#swayr-configuration)
- mutter: seems to have had a 150ms at some point (not sure about
  the exact value bcs there are references stretching from 50ms to
  200ms in various forums and mailing lists), but the delay was applied
  _before_ focus was transferred to the target window (and this annoyed at
  least some users: https://gitlab.gnome.org/GNOME/mutter/-/issues/888). This
  delayed-focus system appears to still be present and can be toggled with the
  `focus-change-on-pointer-rest` gsetting in `org.gnome.mutter`.

As for jumping into the release key binding adventure just for this MRU feature,
it does indeed look like a tail-wagging-the-dog kind of situation ;)

Hard to argue against the simplicity of the hard-coded approach, so would you be
in favor of `Alt-[Shift-]tab` then? What would the mapping in
the winit case then be, `Modifiers::ISO_LEVEL3_SHIFT-[Shift-]tab`?

YaLTeR:
Alt-(Shift-)Tab should work, yeah. Actually, we can do it like this: if the user has anything bound to Alt-(Shift-)Tab, then it will call their bind, otherwise it will do the built-in Alt-Tab. We already do this for e.g. Mod+MouseLeft.

rustn00b:
> if the user has anything bound to Alt-(Shift-)Tab, then it will call their bind, otherwise it will do the built-in Alt-Tab.

With the last few commits, this should be the resulting behavior. I still have to adjust to the Alt vs Mod layout but otherwise I find quite comfortable to use.

bbb651:
An alternative for the lock-in delay could be user interaction with the window, i.e. any keyboard key or pointer button (not movement or wheel) that is actually sent to the application and not suppressed by the compositor. Maybe a combination of the two, whichever comes first?

Also I really dislike having `mru` in the name, it's non-obvious to the point of forcing people to look it up, a config should generally prioritize clarity over conciseness and the rest of niri is really good at that. I can't think of a good name, but even `focus-window-recent-previous` and `focus-window-recent-next` would be better, it's slightly broken English but it at least gives you a general idea. (I think just `focus-window-previous/next` isn't clear enough, and should be reserved for [logical direction actions](https://github.com/YaLTeR/niri/issues/861#issuecomment-2539982985)).

rustn00b:
>  any keyboard key or pointer button (not movement or wheel) that is actually sent to the application and not suppressed by the compositor

I'm not convinced though, I often flip back between code and doc windows - the code windows get interacted with plenty, but the doc windows mostly just gets stared at. For now, my impression is that the implemented behavior feels quite natural - but that's obviously as subjective as it gets!

> I can't think of a good name

Same here! Although MRU/LRU is quite common in projects I've dabbled in, I do concede that it may not be as generally understood as one could hope for.

>  a config should ...

If I'm not mistaken, the actions "MRU" actions should _not_ be available in the config. However they should be visible using IPC - not sure that makes any difference wrt your argument though.

Anyway I don't have any strong beliefs here, so I'm perfectly happy to go along with whatever the consensus is.

rustn00b:
> Maybe a combination of the two, whichever comes first?

This would work I think and shouldn't be too hard to achieve (I'm guessing). I can give it a shot time allowing.

rustn00b:
The latest commit implements the behavior suggested by @bb651 above.
It turns out that this change allows for a less aggressive lock-in timer while at the same time feeling responsive to user inputs. Thanks to @bb651 for the great suggestion!

I haven't changed the name of the actions, will wait for @YaLTeR  to weigh in on that one.

20after4:
Tested and works great for me! I don't see a big problem with the hard coded alt-tab since that is the default shortcut on pretty much every common (and not so common) OS or desktop environment.

Since niri already has a `focus-window-previous` action (which is what I was using before trying this PR)  and since you can override the hard coded mapping, there seems to be enough flexibility to be generally useful. 

My opinion is worth less than $0.02 but I think this should suit most people's needs fairly well.

src/input/mod.rs
```diff
@@ -2943,6 +3017,39 @@ fn find_bind(
...
+        action: Action::FocusWindowMruPrevious,
+        repeat: true,
+        cooldown: None,
+        allow_when_locked: false,
+        allow_inhibiting: true,
+    },
+    Bind {
+        key: Key {
+            trigger: Trigger::Keysym(Keysym::Tab),
+            modifiers: Modifiers::ALT.union(Modifiers::SHIFT),
```
YaLTeR, line 3043:
```suggestion
            modifiers: Modifiers::ALT | Modifiers::SHIFT,
```

src/niri.rs
```diff
@@ -5121,6 +5233,22 @@ impl Niri {
             .unwrap();
         self.pointer_inactivity_timer = Some(token);
     }
+
+    // Consume the active PendingLockinFocus, if any, and use the information
+    // it contains to update the (active) window's focus timestamp
+    pub fn lockin_focus(&mut self) {
```
YaLTeR, line 5239: Perhaps some better name than "lockin focus"? Something like "commit to MRU" idk. Especially considering that this is an option exposed in the config file: it should be clearer that this affects the alt-tab or most-recently-used windows.

src/niri.rs
```diff
@@ -5137,6 +5265,60 @@ impl ClientData for ClientState {
...
+            }
+        });
+
+        WindowMRU {
+            ids: bh
+                .into_sorted_vec()
+                .into_iter()
+                .map(|(_, id)| id)
+                .rev()
+                .collect(),
```
YaLTeR, line 5289: Instead of BinaryHeap maybe just use a regular vec and sort it? Or a pair of vecs that you can sort at once to avoid collecting into a new vec just to unpack the tuple.

src/niri.rs
```diff
@@ -5137,6 +5265,60 @@ impl ClientData for ClientState {
...
+            current: 0,
+        }
+    }
+
+    fn mru_with<F>(&mut self, niri: &Niri, f: F) -> Option<Window>
+    where
+        // function that yields the index of the next element, arguments
+        // are (current_index, number of elements in the list)
+        F: Fn(usize, usize) -> usize,
```
YaLTeR, line 5298: Kinda overcomplicated to do this with a function. Even just some kind of `reverse_order: bool` would work here

src/window/mapped.rs
```diff
@@ -96,6 +96,9 @@ pub struct Mapped {
     ///
     /// Used for double-resize-click tracking.
     last_interactive_resize_start: Cell<Option<(Duration, ResizeEdge)>>,
+
+    /// Most recent time the window had the focus
```
YaLTeR, line 100: Here and elsewhere doc-comments should end with a period

src/window/mapped.rs
```diff
@@ -108,7 +111,7 @@ niri_render_elements! {
 
 static MAPPED_ID_COUNTER: IdCounter = IdCounter::new();
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
```
YaLTeR, line 114: No, these are definitely not intended to be Ord/PartialOrd. If you replace BinaryHeap with Vec like I mentioned above, this will not be necessary.

YaLTeR (commented):
Sorry for taking a while to get to this. Haven't given this a test yet, but here's my initial batch of comments.

src/handlers/compositor.rs
```diff
@@ -272,7 +272,11 @@ impl CompositorHandler for State {
                     }
 
                     if was_active {
-                        self.maybe_warp_cursor_to_focus();
+                        if self.niri.window_mru.is_some() {
+                            self.focus_window_mru_previous();
```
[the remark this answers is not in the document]
rustn00b: Despite the `previous` in the name this would have been the next window in MRU order, so after renaming previous/next this probably lines up with expectations.
When I had traversal working with the Mod key, it was pretty simple to close windows as you moved through the list. This felt pretty natural. With the Alt leader key I'm not sure how you would go about doing this though.

src/input/mod.rs
```diff
@@ -127,6 +128,17 @@ impl State {
...
+        // If an event was forwarded to the focused window and it still has
+        // a running lockin_token, cancel the token and update the timestamp
+        // immediately
+        if self
+            .niri
+            .event_forwarded_to_focused_client
+            .swap(false, std::sync::atomic::Ordering::Relaxed)
+        {
+            self.niri.lockin_focus();
+        }
```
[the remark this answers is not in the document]
rustn00b: I initially wanted to handle this (at least for keyboard events) in `should_intercept_key` but calling `lockin_focus` requires a `mut ref` so that would have meant butchering `should_intercept_key`'s signature. Maybe wrongly, I was aiming to minimize changes on stuff I most likely don't have a good grasp.
 
As for possible delays, it's actually isn't an issue because the lockin timer will eventually fire. All that matters (at least for MRU) is ordering as opposed to accurate timestamping.

src/niri.rs
rustn00b: why not, what do you think of `mru-commit-ms` for the configuration and `pub fn mru_commit(&mut self)`?

src/input/mod.rs
rustn00b: That's what I tried initially, but since it is in a `const` context the more readable version unfortunately doesn't work and I had to go with the clunkier const-compatible version.

src/input/mod.rs
```diff
@@ -382,6 +393,48 @@ impl State {
                     }
                 }
 
+                // check if the released key is the mod key, if so and there was an active
+                // window-mru list: drop the list and update the current window's timestamp
+                if let Some(raw) = raw {
+                    if !pressed
+                        && matches!(raw, Keysym::Alt_L | Keysym::Alt_R)
```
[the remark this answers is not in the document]
rustn00b: Trouble is that at that point mods.alt is false since it is a key-release event.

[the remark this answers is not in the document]
YaLTeR: Yeah, hence `!mods.alt`

src/input/mod.rs
```diff
@@ -382,6 +393,48 @@ impl State {
...
+                            .and_then(|ws| ws.active_window_mut())
+                        {
+                            m.update_focus_timestamp(Instant::now());
+                        }
+                    }
+
+                    // If the ESC key was pressed with the Alt modifier and
+                    // there is an active window-mru, cancel the window-mru and
+                    // refocus the initial window (first in the list)
+                    if pressed && !this.niri.is_locked() && mods.alt && raw == Keysym::Escape {
```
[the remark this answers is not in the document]
rustn00b: Changed to ignore the `is_locked()` state. Anyway an active MRU traversal while locked doesn't make much sense anyway.

src/input/mod.rs
```diff
@@ -382,6 +393,48 @@ impl State {
...
+                            .niri
+                            .window_mru
+                            .take()
+                            .and_then(|wmru| wmru.ids.into_iter().next())
+                        {
+                            let window = this.niri.layout.windows().find(|(_, m)| m.id() == id);
+                            let window = window.map(|(_, m)| m.window.clone());
+                            if let Some(window) = window {
+                                this.focus_window(&window);
+                                return FilterResult::Intercept(None);
```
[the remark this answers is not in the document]
rustn00b: I guess the behavior is similar. However if by adding Esc to the `suppressed_keys`, you also mean its effect should be handled in `should_intercept_key`, we once again hit the needing-a-mut-ref on niri problem.

src/niri.rs
rustn00b: For now I switched this to a collect into `Vec<(Instant, MappedId)>`, that gets sorted and then collected again into a simple `Vec<MappedId>`. Not fancy but does the job.

src/input/mod.rs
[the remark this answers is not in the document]
rustn00b: Yes of course! I brain-farted here.

src/input/mod.rs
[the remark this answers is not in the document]
rustn00b: I'm trying out a variant of `should_intercept_key` that avoids delaying the MRU lockin (or commit if the we go with that nomenclature) until the next event loop. This should get rid of the `event_forwarded_to_focused_client` bool.

src/handlers/compositor.rs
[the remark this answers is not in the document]
YaLTeR: Ah, this is only during an active MRU traversal. Then it makes sense, yeah. I thought it was unconditional.

src/input/mod.rs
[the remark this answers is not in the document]
YaLTeR: I mean that we're handing an Esc key press here, so the subsequent Esc key release should not make it to the focused window. I believe adding it to suppressed_keys will do this.

src/input/mod.rs
YaLTeR: Ah. Guess const traits can't arrive soon enough.

src/input/mod.rs
```diff
@@ -2173,10 +2168,8 @@ impl State {
         }
 
         // The event is getting forwarded to a client, consider that the
-        // focus should be locked-in
-        self.niri
-            .event_forwarded_to_focused_client
-            .store(true, std::sync::atomic::Ordering::Relaxed);
+        // focus should be locked-in.
+        self.niri.lockin_focus();
```
YaLTeR, line 2172: Guess this should also happen on touch taps and tablet touches and possibly other stuff?

src/input/mod.rs
```diff
@@ -430,22 +418,29 @@ impl State {
...
+                        modified,
+                        raw,
+                        pressed,
+                        *mods,
+                        &this.niri.screenshot_ui,
+                        this.niri.config.borrow().input.disable_power_key_handling,
+                        is_inhibiting_shortcuts,
+                    )
+                };
+                if matches!(intercept_result, FilterResult::Forward) {
```
YaLTeR, line 437: This is probably ok but fwiw you don't have to handle this inside. If `keyboard.input()` returns `None`, that means the closure returned `FilterResult::Forward`.

src/niri.rs
YaLTeR: Makes sense I guess.

src/niri.rs
```diff
@@ -359,6 +360,14 @@ pub struct Niri {
...
     #[cfg(feature = "xdp-gnome-screencast")]
     pub mapped_cast_output: HashMap<Window, Output>,
+
+    // List of windows for MRU prev/next traversal.
+    // Only defined when there is an active traversal in progress
+    pub window_mru: Option<WindowMRU>,
+
+    pending_focus_lockin: Option<PendingFocusLockin>,
+
+    pub event_forwarded_to_focused_client: AtomicBool,
```
YaLTeR, line 370: Also remove it from here

src/niri.rs
```diff
@@ -5186,6 +5314,52 @@ impl ClientData for ClientState {
...
+                if let Some(window) = niri.find_window_by_id(*id) {
+                    return Some(window);
+                }
+                self.ids.remove(self.current);
+            }
+        }
+        None
+    }
+
+    fn mru_next(&mut self, niri: &Niri) -> Option<Window> {
```
YaLTeR, line 5354: Maybe call this "advance" and "advance_back"?

niri-ipc/src/lib.rs
```diff
@@ -217,6 +217,10 @@ pub enum Action {
     },
     /// Focus the previously focused window.
     FocusWindowPrevious {},
+    /// Focus the next window in most-recently-used order.
+    FocusWindowMruNext {},
+    /// Focus the previous window in most-recently-used order.
+    FocusWindowMruPrevious {},
```
YaLTeR, line 223: As a data point, GNOME calls this "Switch Applications". Then if you manually dig in dconf, the other direction is "switch-applications-backward" with a comment of "Reverse switch applications".

src/niri.rs
rustn00b: Coulda sworn having done that. oops

src/niri.rs
rustn00b: yes, just kept `advance` and removed `mru_next` and `mru_previous` seeing as they were trivial.

src/input/mod.rs
rustn00b: you mean something like `match ...input(...) { None => { mru_commit(); return}, Some(None) => return, _ => ()}`?

YaLTeR: Right now there's an `if let Some(Some(bind)) = keyboard.input(...)` and it has an `} else {` branch, I think you can put this line into that else branch?

YaLTeR: Actually not exactly since then it will catch `Some(None)` too which is not what we want.. Idk maybe there's a cleaner way here

src/input/mod.rs
```diff
@@ -382,6 +393,48 @@ impl State {
                     }
                 }
 
+                // check if the released key is the mod key, if so and there was an active
+                // window-mru list: drop the list and update the current window's timestamp
+                if let Some(raw) = raw {
```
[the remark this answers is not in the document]
rustn00b: A bit tricky to do things inside `should_intercept_key` bcs handling the triggered action requires a ref to `State`.

src/input/mod.rs
rustn00b: Indeed. I've sprinkled `mru_commit()` in interactions that are forwarded to the client App (to the extent I understood what was going on). However tbh I mostly can't test out if it's really working as intended.

src/input/mod.rs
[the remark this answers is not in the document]
rustn00b: Ah understood, I hadn't groked that bit. Fixed.

YaLTeR:
Is there a reason why this PR is still marked as Draft? Something still missing?

rustn00b:
It's working fine for me, however I can't really speak to the changes for touch/tablets. I'll "ready for review" it then.

YaLTeR:
Gonna give it some testing.

First thing I noticed: at compositor startup, when I have auto-started windows on different workspaces, MRU doesn't go through all of the windows until I focus them the first time. And similarly, if a window is never focused on spawn (e.g. a dialog for an unfocused window, or with `open-focused false`), then it is not added to the MRU, although it should (at least to the end).

For mru-commit-ms config, probably put it into a new section (like the new `clipboard`), but need to think of a name. Or maybe just don't have it in the config at all for this PR?

YaLTeR:
It's also weird that as you're navigating the MRU, it will scroll around and mess up all your workspaces and monitors, and won't restore them back when you pick a window or cancel. This wouldn't be a problem if MRU used a dialog to select windows instead of jumping to them. But that's a chunk of work by itself so idk.

rustn00b:
> MRU doesn't go through all of the windows until I focus them the first time

Fair comment, I'll add in something to include windows without a focus timestamp in the MRU list. This shouldn't be any trouble.

>  Or maybe just don't have it in the config at all for this PR?

I guess so, I don't really see anyone tinkering with this setting much. A dedicated section for only this would feel a bit bloated. So I'll just set the default 750ms as a static value and comment out the config related parts.

> it will scroll around and mess up all your workspaces and monitors

Yup, especially with tabbing! Improving this  was actually up next on my todo-wishlist but I haven't gotten actually started anything.

bbb651:
I think it would be cool to use the MRU order when closing a window, currently (in 25.02, haven't tested this PR but I don't see changes related to it) it seems to have logic to switch back to the previous window if you open a window then close it without switching focus in between, but for other cases it focuses the window to the right. This could be an option, not sure what the default should be.

YaLTeR:
Yeah, the current logic more or less follows the behavior you'd find in tab bars (think Firefox tabs or some such). Full MRU would be annoying (think switching workspace, then immediately closing a window, you don't want it to switch back). Maybe limited MRU within a workspace. But idk, I feel that the current logic feels fairly good

jbms:
It would be great to expose the MRU timestamps of each window over IPC --- then the actual switching logic can be implemented externally with greater flexibility, e.g. separate commands for all windows or just within workspace, showing a text list, etc.

rustn00b:
> It would be great to expose the MRU timestamps of each window over IPC --- then the actual switching logic can be implemented externally with greater flexibility, e.g. separate commands for all windows or just within workspace, showing a text list, etc.

The PR currently just walks the full list of windows, but I did imagine making this more flexible by basically adding a filter to the window list traversal when the MRU list is created, some examples of pre-defined filters I was planning on adding would be per workspace and per app_id.
Not quite the flexibility you describe, but it should still cover quite a few use cases. As for exporting the timestamp over IPC that should be fairly straightforward to implement.

RubenAstudillo:
I have been reading this discussion. I just want to say thanks to @rustn00b and Yalter for the great job on this. This branch has been my daily driver for like 2 months now and it works great. I did this via the nix flake :-). 

 I don't say this to put any kind of pressure, you guys are doing fine IMHO. I just want tell this is being tested in the wild and works great.

rustn00b:
> It's also weird that as you're navigating the MRU, it will scroll around and mess up all your workspaces and monitors, and won't restore them back when you pick a window or cancel. This wouldn't be a problem if MRU used a dialog to select windows instead of jumping to them. But that's a chunk of work by itself so idk.

In an attempt to improve UX, I put [this variant](https://github.com/rustn00b/niri/tree/window-mru-ui) together. It's a bit rough because it's missing animations, so not really PR-ready material, but I'm kind of enjoying using it. Beyond that, I'm not exactly confident that it's making proper use of either Niri  or Smithay.

However it should hopefully address the "scroll around and mess up all your workspaces" comment.

Anyway, if there's interest, I can submit the PR and see where things go from there :)

rustn00b:
Superseded by  #1704.

PR #991, opened 2025-01-16
add `toggle-window-rule-opacity` action

Faervan:
Add an action that toggles the opacity of the active window (or the window with the given id).

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -578,6 +578,18 @@ pub enum Action {
         )]
         y: PositionChange,
     },
+    /// Toggle the opacity of a window.
+    #[cfg_attr(
+        feature = "clap",
+        clap(about = "Toggle the opacity of the focused window")
+    )]
+    ToggleWindowRuleOpacity {
+        /// Id of the window.
+        ///
+        /// If `None`, uses the focused window.
+        #[cfg_attr(feature = "clap", arg(long))]
+        id: Option<u64>,
+    },
 }
 
 /// Change in window or column size.
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -190,6 +190,7 @@ pub trait LayoutElement {
     fn set_active_in_column(&mut self, active: bool);
     fn set_floating(&mut self, floating: bool);
     fn set_bounds(&self, bounds: Size<i32, Logical>);
+    fn is_ignoring_opacity_window_rule(&self) -> bool;
 
     fn configure_intent(&self) -> ConfigureIntent;
     fn send_pending_configure(&mut self);
@@ -4346,6 +4347,10 @@ mod tests {
         }
 
         fn set_bounds(&self, _bounds: Size<i32, Logical>) {}
+
+        fn is_ignoring_opacity_window_rule(&self) -> bool {
+            false
+        }
 
         fn configure_intent(&self) -> ConfigureIntent {
             ConfigureIntent::CanSend
--- a/src/window/mapped.rs
+++ b/src/window/mapped.rs
@@ -69,6 +69,9 @@ pub struct Mapped {
     /// Whether this window is floating.
     is_floating: bool,
 
+    /// Whether this window should ignore opacity set through window rules.
+    ignore_opacity_window_rule: bool,
+
     /// Buffer to draw instead of the window when it should be blocked out.
     block_out_buffer: RefCell<SolidColorBuffer>,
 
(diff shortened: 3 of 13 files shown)


phisch:
Could possibly be done if something like this was implemented: https://github.com/YaLTeR/niri/discussions/1001

Faervan:
> Could possibly be done if something like this was implemented: #1001

well as I understand #1001 it would only allow to toggle a config block on or off. The point of this pr is to enable a per-window toggle of the opacity. The global opacity toggle can already be accomplished by making a custom script using `sed` (as YaLTeR pointed out on matrix) or smth similar, but a per-window toggle without this pr would require setting config rules for the focused window using the app-id and title, which would be a really hacky and ugly workaround I imagine.
This pr is a more elegant and 100% working solution to archive this.

niri-ipc/src/lib.rs
```diff
@@ -578,6 +578,14 @@ pub enum Action {
         )]
         y: PositionChange,
     },
+    /// Toggles the opacity of the focused window
```
YaLTeR, line 581:
```suggestion
    /// Toggle the opacity of a window.
    #[cfg_attr(feature = "clap", clap(about = "Toggle the opacity of the focused window"))]
```

src/layout/mod.rs
```diff
@@ -204,6 +204,9 @@ pub trait LayoutElement {
     /// This *will* switch immediately after a [`LayoutElement::request_fullscreen()`] call.
     fn is_pending_fullscreen(&self) -> bool;
 
+    fn is_forced_opaque(&self) -> bool;
+    fn toggle_forced_opaqueness(&mut self);
+
```
YaLTeR, line 209: Move these up to right above configure_intent()

src/window/mapped.rs
```diff
@@ -69,6 +69,9 @@ pub struct Mapped {
     /// Whether this window is floating.
     is_floating: bool,
 
+    /// Whether this window is forced to be opaque
```
YaLTeR, line 72:
```suggestion
    /// Whether this window is forced to be opaque.
```

wiki/Configuration:-Window-Rules.md
```diff
@@ -491,6 +491,8 @@ This is applied on top of the window's own opacity, so semitransparent windows w
 
 Opacity is applied to every surface of the window individually, so subsurfaces and pop-up menus will show window content behind them.
 
+Opacity can be toggle on or off for a window using the [toggle-window-opacity](./Configuration:-Key-Bindings.md#toggle-window-opacity) action.
```
YaLTeR, line 494:
```suggestion
Opacity can be toggled on or off for a window using the [toggle-window-opacity](./Configuration:-Key-Bindings.md) action.
```
Unfortunately, our current wiki github action does not support linking to headers directly when it removes the `.md` file extension.

wiki/Configuration:-Key-Bindings.md
```diff
@@ -253,3 +253,13 @@ Or, in scripts:
 ```shell
 niri msg action do-screen-transition --delay-ms 100
 ```
+
+### `toggle-window-opacity`
+
+Toggle the opacity of the active window. This has only an effect if the window's opacity is already set to be semitransparent.
```
YaLTeR, line 259:
```suggestion
Toggle the opacity window rule of the focused window.
This only has an effect if the window's opacity window rule is already set to semitransparent.
```

niri-ipc/src/lib.rs
```diff
@@ -578,6 +578,14 @@ pub enum Action {
...
+    ToggleWindowOpacity {
```
YaLTeR, line 582: So I'm thinking, this is a bit confusing, because windows can have their own client-side opacity, and this action doesn't (and shouldn't) toggle that. Maybe it's better to call it `ToggleWindowRuleOpacity` or something?

YaLTeR (changes requested):

YaLTeR:
Did some small fixes, particularly when the opacity window rule becomes 1, the ignore flag is reset to avoid surprises later on. Let's merge this.

YaLTeR:
Thanks!

Faervan:
thank you for merging this :)

YaLTeR:
Apparently I overwrote the commit author with myself during that squash and never noticed. Sorry for this, I did not intend to do it. I'll tag you in the release notes of course.

Faervan:
thanks, no problem :)

PR #1007, opened 2025-01-18
Added move-workspace-to-index and move-workspace-to-monitor IPC actions

Kirottu:


Kirottu:
I must say I have genuinely no idea what happened to the previous PR, but GitHub has decided that it's just... not a part of the equation anymore after some problems with getting rebase done correctly. All the previous comments should have now been addressed. I am deeply sorry for the confusion this must be causing... well everyone.

YaLTeR:
7 copilot buttons and none could save them

src/input/mod.rs
```diff
@@ -1127,6 +1127,21 @@ impl State {
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
+            Action::MoveWorkspaceToIndex(new_idx) => {
+                self.niri.layout.move_workspace_to_idx(None, new_idx);
```
YaLTeR, line 1131: All new actions need a `queue_redraw_all()`

src/input/mod.rs
```diff
@@ -1127,6 +1127,21 @@ impl State {
...
+            }
+            Action::MoveWorkspaceToIndexByRef {
+                new_idx: index,
```
YaLTeR, line 1134: Sounds like this doesn't need to be renamed

src/input/mod.rs
```diff
@@ -1139,6 +1154,7 @@ impl State {
             Action::UnsetWorkSpaceNameByRef(reference) => {
                 self.niri.layout.unset_workspace_name(Some(reference));
             }
+
```
YaLTeR, line 1157:
```suggestion
```

src/input/mod.rs
```diff
@@ -1461,6 +1477,30 @@ impl State {
                     }
                 }
             }
+            Action::MoveWorkspaceToMonitor(new_output) => {
```
YaLTeR, line 1480: These two need to warp pointer, but only when the new output is different from the current workspace output. Might be convenient to return a bool from the layout on whether the workspace moved to a new output.

src/layout/mod.rs
```diff
@@ -4602,6 +4679,18 @@ mod tests {
...
+        MoveWorkspaceToIndex {
+            #[proptest(strategy = "proptest::option::of(1..=5usize)")]
+            ws_name: Option<usize>,
+            #[proptest(strategy = "0..=4usize")]
+            target_idx: usize,
+        },
+        MoveWorkspaceToMonitor {
+            #[proptest(strategy = "proptest::option::of(1..=5usize)")]
+            ws_name: Option<usize>,
+            #[proptest(strategy = "0..5usize")]
```
YaLTeR, line 4691: Our output names here are `1..=5usize`, see Op::AddOutput etc.

src/layout/mod.rs
```diff
@@ -5174,6 +5263,80 @@ mod tests {
...
+                                .find_map(|(i, ws)| {
+                                    if ws.name == Some(format!("ws{ws_name}")) {
+                                        Some(i)
+                                    } else {
+                                        None
+                                    }
+                                })
+                                .map(|i| (i, monitor.output.clone()))
+                        })
+                        .unwrap();
```
YaLTeR, line 5290: Doesn't this fail when the workspace doesn't exist? These are randomized tests so they shouldn't panic if the randomly generated ws name doesn't exist.

YaLTeR: Also don't forget to actually run these tests: https://github.com/YaLTeR/niri/wiki/Developing-niri#running-tests

YaLTeR (changes requested):

src/input/mod.rs
Kirottu: What do you mean by that? There is now a `MoveWorkspaceToIndex` and a `MoveWorkspaceToIndexByRef`. The same way as I did for the ToMonitor actions.

YaLTeR: I mean that instead of `new_idx: index` just have `new_idx`

Kirottu: Ah yea that makes sense, that's done now.

YaLTeR (changes requested):
Rebased. Noticed two issues:
1. Command like `move-workspace-to-index 0 --reference 2` doesn't work because in this case `find_output_and_workspace_index()` returns a None output.
2. Command like `move-workspace-to-index 0 --reference browser` focuses the workspace after moving it, even if it wasn't focused before.

Kirottu:
Those should be fixed:
1. In case a numbered workspace reference is used, the code now assumes the active monitor
2. The active workspace index will now be adjusted through a few conditions as seen in the code

src/layout/monitor.rs
```diff
@@ -890,7 +890,17 @@ impl<W: LayoutElement> Monitor<W> {
...
 
         let previous_workspace_id = self.previous_workspace_id;
-        self.activate_workspace(new_idx);
+
+        // Only refocus the workspace if it was already focused
+        if self.active_workspace_idx == old_idx {
+            self.activate_workspace(new_idx);
+        // If the workspace order was switched so that the current workspace moved down the workspace stack, focus correctly
+        } else if new_idx <= self.active_workspace_idx && old_idx > self.active_workspace_idx {
+            self.activate_workspace(self.active_workspace_idx + 1);
```
YaLTeR, line 899: I think this should just set self.active_workspace_idx directly. Because updating previous_workspace_id as done in activate_workspace() doesn't make sense in these two cases.

src/input/mod.rs
```diff
@@ -1497,6 +1509,38 @@ impl State {
...
+                        self.move_cursor_to_output(&new_output);
+                    }
+                }
+            }
+            Action::MoveWorkspaceToMonitorByRef {
+                output_name,
+                reference,
+            } => {
+                if let Some((Some(output), old_idx)) =
+                    self.niri.find_output_and_workspace_index(reference)
```
YaLTeR, line 1526: Same thing here, by-index addressing will return None output.

src/input/mod.rs
```diff
@@ -1497,6 +1509,38 @@ impl State {
...
+                        if new_output != output {
+                            self.niri.layout.move_workspace_to_output_by_id(
+                                old_idx,
+                                output,
+                                new_output.clone(),
+                            );
+                            // Cursor warp already calls `queue_redraw_all`
+                            if !self.maybe_warp_cursor_to_focus_centered() {
+                                self.move_cursor_to_output(&new_output);
+                            }
```
YaLTeR, line 1539: This should also only activate and warp if the previous output was the active output and the workspace was the active workspace.

Kirottu:
Those problems should also be done, also fixed a bit of an issue where the active workspace on the monitor that had a workspace moved off it would be wrong if the workspace that was moved was not the active workspace.

src/layout/monitor.rs
```diff
@@ -889,21 +889,16 @@ impl<W: LayoutElement> Monitor<W> {
...
         // If the workspace order was switched so that the current workspace moved down the workspace stack, focus correctly
         } else if new_idx <= self.active_workspace_idx && old_idx > self.active_workspace_idx {
-            self.activate_workspace(self.active_workspace_idx + 1);
+            self.active_workspace_idx += 1;
         } else if new_idx >= self.active_workspace_idx && old_idx < self.active_workspace_idx {
-            self.activate_workspace(self.active_workspace_idx - 1);
+            self.active_workspace_idx = self.active_workspace_idx.saturating_sub(1);
         }
 
-        self.workspace_switch = None;
```
YaLTeR: I believe this `= None` is still needed here

Kirottu: Fixed now

src/layout/mod.rs
```diff
@@ -3884,11 +3898,20 @@ impl<W: LayoutElement> Layout<W> {
         // the last empty workspace, then insert before.
         let target_ws_idx = min(target.active_workspace_idx + 1, target.workspaces.len() - 1);
         target.workspaces.insert(target_ws_idx, ws);
-        target.active_workspace_idx = target_ws_idx;
-        target.workspace_switch = None;
```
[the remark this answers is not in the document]
Kirottu: Alright, reordered it to mirror the original order at which the operations were done.

YaLTeR:
Had to fix some more issues but should be good now I think.

YaLTeR:
Thanks

PR #1011, opened 2025-01-19
Add setting to disable primary selection

notpeelz:
I've been using a custom build for a while where I commented out the primary selection code.
There's probably other users that don't want the primary selection, so I made it into a configurable feature.

I wasn't sure how to implement this (especially with live-reload involved, which I've intentionally ignored).
Feedback welcome.

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -475,6 +475,7 @@ impl SecurityContextHandler for State {
                 let data = Arc::new(ClientState {
                     compositor_state: Default::default(),
                     can_view_decoration_globals: config.prefer_no_csd,
+                    primary_selection_disabled: config.clipboard.disable_primary,
                     restricted: true,
                     credentials_unknown: false,
                 });
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -53,6 +53,8 @@ pub struct Config {
     ]
     pub screenshot_path: Option<String>,
     #[knuffel(child, default)]
+    pub clipboard: Clipboard,
+    #[knuffel(child, default)]
     pub hotkey_overlay: HotkeyOverlay,
     #[knuffel(child, default)]
     pub animations: Animations,
@@ -783,6 +785,12 @@ pub struct Struts {
 pub struct HotkeyOverlay {
     #[knuffel(child)]
     pub skip_at_startup: bool,
+}
+
+#[derive(knuffel::Decode, Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct Clipboard {
+    #[knuffel(child)]
+    pub disable_primary: bool,
 }
 
 #[derive(knuffel::Decode, Debug, Clone, PartialEq)]
@@ -3432,6 +3440,10 @@ mod tests {
 
             screenshot-path "~/Screenshots/screenshot.png"
 
+            clipboard {
+                disable-primary
+            }
+
             hotkey-overlay {
                 skip-at-startup
             }
@@ -3686,6 +3698,9 @@ mod tests {
                     hide_after_inactive_ms: Some(3000),
                 },
                 screenshot_path: Some(String::from("~/Screenshots/screenshot.png")),
+                clipboard: Clipboard {
+                    disable_primary: true,
+                },
                 hotkey_overlay: HotkeyOverlay {
                     skip_at_startup: true,
                 },
--- a/src/dbus/mutter_service_channel.rs
+++ b/src/dbus/mutter_service_channel.rs
@@ -28,6 +28,7 @@ impl ServiceChannel {
             compositor_state: Default::default(),
             // Would be nice to thread config here but for now it's fine.
             can_view_decoration_globals: false,
+            primary_selection_disabled: false,
             restricted: false,
             // FIXME: maybe you can get the PID from D-Bus somehow?
             credentials_unknown: true,
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -1810,7 +1810,13 @@ impl Niri {
         let idle_notifier_state = IdleNotifierState::new(&display_handle, event_loop.clone());
         let idle_inhibit_manager_state = IdleInhibitManagerState::new::<State>(&display_handle);
         let data_device_state = DataDeviceState::new::<State>(&display_handle);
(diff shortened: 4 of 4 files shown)


YaLTeR:
Looks reasonable. For a disabling mechanism, I think it should work better to hide it per-client, kinda like xdg-decoration that is hidden depending on the `prefer-no-csd` option. It stores a bool on `ClientState`.

notpeelz:
I was able to hide primary-selection on a per-client basis by manually implementing `delegate_primary_selection!()`:

<details>
  <summary>diff</summary>

```diff
diff --git a/src/handlers/mod.rs b/src/handlers/mod.rs
index d258975e..270030f7 100644
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -384,10 +384,56 @@ delegate_data_device!(State);
 
 impl PrimarySelectionHandler for State {
     fn primary_selection_state(&self) -> &PrimarySelectionState {
-        self.niri.primary_selection_state.as_ref().unwrap()
+        &self.niri.primary_selection_state
     }
 }
-delegate_primary_selection!(State);
+
+impl GlobalDispatch<zwp_primary_selection_device_manager_v1::ZwpPrimarySelectionDeviceManagerV1, ()>
+    for State
+{
+    fn bind(
+        state: &mut Self,
+        dhandle: &DisplayHandle,
+        client: &Client,
+        resource: New<zwp_primary_selection_device_manager_v1::ZwpPrimarySelectionDeviceManagerV1>,
+        global_data: &(),
+        data_init: &mut DataInit<'_, Self>,
+    ) {
+        <PrimarySelectionState as GlobalDispatch<
+            zwp_primary_selection_device_manager_v1::ZwpPrimarySelectionDeviceManagerV1,
+            (),
+            Self,
+        >>::bind(state, dhandle, client, resource, global_data, data_init)
+    }
+
+    fn can_view(client: Client, global_data: &()) -> bool {
+        if client
+            .get_data::<ClientState>()
+            .unwrap()
+            .primary_selection_disabled
+        {
+            return false;
+        }
+
+        <PrimarySelectionState as GlobalDispatch<
+            zwp_primary_selection_device_manager_v1::ZwpPrimarySelectionDeviceManagerV1,
+            (),
+            Self,
+        >>::can_view(client, global_data)
+    }
+}
+
+delegate_dispatch!(State: [
+    zwp_primary_selection_device_manager_v1::ZwpPrimarySelectionDeviceManagerV1: ()
+] => selection::primary_selection::PrimarySelectionState);
+
+delegate_dispatch!(State: [
+    zwp_primary_selection_device_v1::ZwpPrimarySelectionDeviceV1: PrimaryDeviceUserData
+] => selection::primary_selection::PrimarySelectionState);
+
+delegate_dispatch!(State: [
+    zwp_primary_selection_source_v1::ZwpPrimarySelectionSourceV1: PrimarySourceUserData
+] => selection::primary_selection::PrimarySelectionState);
 
 impl DataControlHandler for State {
     fn data_control_state(&self) -> &DataControlState {
```
  
</details>

With these changes, middle-click pasting is disabled (I tested within the same client and across clients) and the protocol no longer appears in waycheck. However `wl-copy -p`/`wl-paste -p` still works.
I think I would have to modify the [ZwlrDataControlManagerV1 `bind()` logic](https://github.com/Smithay/smithay/blob/953959e6069b3e14dba96fdaa46c65990c21d5c9/src/wayland/selection/wlr_data_control/mod.rs#L158-L163), but I don't know if I can do that cleanly since so much of the logic is encapsulated (and `DataControlManagerUserData#primary` is a private field).

Is there any easier way?

YaLTeR:
Yes, this kind of filtering in `can_view` is exactly what other globals in Smithay do (the ones which accept the filter as an argument). So that is correct.

> I think I would have to modify the [ZwlrDataControlManagerV1 bind() logic](https://github.com/Smithay/smithay/blob/953959e6069b3e14dba96fdaa46c65990c21d5c9/src/wayland/selection/wlr_data_control/mod.rs#L158-L163)

Is there a reason why you also need to disable data-control? Actually, what's your use-case for disabling the primary selection in general?

notpeelz:
> Is there a reason why you also need to disable data-control? 

I don't want to disable it entirely. I'm trying to prevent it from modifying/reading the primary selection, otherwise `wl-copy -p`/`wl-paste -p` still works.
I would expect it to inform me that the primary selection is disabled/unsupported.
For example, the current PR code causes `wl-copy -p hello` to print this message: `The compositor does not seem to support primary selection`.

> Actually, what's your use-case for disabling the primary selection in general?

I don't use it and it's annoying when I accidentally trigger it.
On Firefox and chromium-based programs, I use the autoscroll feature (activated via middle-click).
For chromium-based programs, I need to enable it with a launch parameter (` --enable-blink-features=MiddleClickAutoscroll`), but for some reason it conflicts with the primary selection handling logic, which triggers both the autoscroll and pasting from the primary selection. Disabling primary selection at the compositor level fixes that.

notpeelz:
I had to make some changes in smithay to make it configurable per-client. You can test it with my branch here: <https://github.com/notpeelz/smithay/tree/feat-primary-selection-filter>
If that looks good to you, I'll open a PR.

src/handlers/mod.rs
```diff
@@ -116,7 +116,13 @@ impl SeatHandler for State {
         let dh = &self.niri.display_handle;
         let client = focused.and_then(|s| dh.get_client(s.id()).ok());
         set_data_device_focus(dh, seat, client.clone());
-        set_primary_focus(dh, seat, client);
+        let client_state = client.as_ref().and_then(|c| c.get_data::<ClientState>());
+        if !client_state
+            .map(|c| c.primary_selection_disabled)
+            .unwrap_or(false)
```
YaLTeR, line 122: Is this check necessary? When it's disabled, the client won't have a global anyway so this should be inert?

YaLTeR (commented):
The Smithay branch and the changes here make sense to me. Though when PRing, you might want to run it by kchibisov who last reworked the Smithay clipboard handling.

notpeelz:
The Smithay PR has been merged. Waiting for the dependabot PR to rebase my changes.

YaLTeR:
You should be able to rebase now

YaLTeR:
Thanks

Issue #955, opened 2025-01-11
One-to-many swap-window-left/right changes the view position when columns have different width


YaLTeR:
Some cases might be tricky to fix since windows don't change size immediately. We don't want to do any waiting for the window to change size before changing the view position. Also check how consume/expel/move handle this.

PR #1014, opened 2025-01-19
Fix animations for swap-window actions

rustn00b:
Avoid calling add_tile/remove_tile that can trigger unwanted animations in neighboring tiles/columns. When the swapped tiles have a different size, tiles following the ones swapped in their respective columns will move slightly. Ideally they should stay put, but the general behavior is still improved.

This PR is should address #955.


src/layout/scrolling.rs
```diff
@@ -1807,80 +1806,30 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
+            let (a, b) = self.columns.split_at_mut(target_column_idx);
+            (&mut a[source_column_idx], &mut b[0])
+        } else {
+            let (a, b) = self.columns.split_at_mut(source_column_idx);
+            (&mut b[0], &mut a[target_column_idx])
+        };
+        std::mem::swap(
+            &mut source_col.tiles[source_tile_idx],
+            &mut target_col.tiles[target_tile_idx],
         );
```
YaLTeR, line 1820: If you run the randomized tests for a longer period as described [here on the wiki](https://github.com/YaLTeR/niri/wiki/Developing-niri#running-tests), they will tell you that you forgot to also swap the corresponding cached tile data:
```
Test failed: assertion `left == right` failed: tile data must be up to date
```

YaLTeR (changes requested):
I like the swap approach, it is what I had in mind for redoing this function. However, as is, it does not completely fix the animations.

> tiles following the ones swapped in their respective columns will move slightly

They move quite a bit actually:

https://github.com/user-attachments/assets/f8883635-bdf6-4954-b0a4-af5c439fb75c

If you'd like to continue to try to fix this, please make sure that all animations work properly with no big jumps, including when you trigger the action in the middle of a different animation, or trigger the action repeatedly quickly. Otherwise there isn't really a point since you're replacing a minor issue in current main with a much more noticeable animation issue.

The animation slowdown config helps a lot with this:
```
animations {
    slowdown 4.0
}
```

src/layout/scrolling.rs
rustn00b: Terribly sorry about this one, I had mistakenly ticked it off on my todo list and hadn't noticed it was actually missing. My bad.

rustn00b:
First off, thanks for the pro-tips on the slowdown, it sure beats the squinting I was doing to figure out what was amiss. After some tinkering around I have something that's - just - about right. Tiles that aren't involved in the animation stay put, and those that are involved in the swap resize as they move. Combining move actions while the tiles are in motion  (e.g. move-window-up or down in their destination column) is smooth. But... I'm still seeing a  glitch at the beginning of the animation that I haven't been able to figure out.
Suggestions welcome!

src/layout/tile.rs
```diff
@@ -106,6 +106,10 @@ pub struct Tile<W: LayoutElement> {
 
     /// Configurable properties of the layout.
     pub(super) options: Rc<Options>,
+
+    /// Marker to indicate that the resize animation should not propagate
+    /// to other tiles below in the same column
+    pub(super) inhibit_sibling_move_on_resize: Option<()>,
```
YaLTeR, line 112: Hmm. I really don't want this kind of special case bools if I can avoid it. They make the logic very tricky to follow and pave the way to all sorts of weird visual issues.

Especially considering that logically, this should not be necessary if the right animations are triggered in the right order. Imagine swapping a bigger tile with a smaller tile like in my video above.

1. Since the bigger tile is bigger, it should start an animation for all tiles below the new spot to move down to accommodate it.
2. We ask the bigger tile to resize down.
3. The bigger tile commits a new, smaller buffer. All tiles below the new spot start an animation to move back into the original position. Since they (ideally) barely moved by this point, there's no jump.

The real problem might (not sure) actually be in the resize animation code. I know it's possible to make things jump if you resize up and down more or less at once, i.e. hold Mod and press + and - at the same time. Although, if you wait even a tiny bit then it will work fine without jumps.

rustn00b: The latest commit (68555c2) deals with the glitch mentioned above.

I would also prefer not to have the `inhibit_sibling_move_on_resize` boolean, but `scrolling.rs::update_window()` starts a move animation for all tiles following the swapped tile (by an offset based on the animation snapshot, so with the size of the tile before the swap). This behavior makes perfect sense when a tile is resized in a column, but in this instance starts an animation that assumes the tile being swapped in was already in the column. 

Some mechanism to inhibit this behavior is needed, I'm just not sure where it should live.

rustn00b:
Oddly the CI fails on a test that runs ok locally with 1a3434d0. The command used to run the tests:
```sh
env RUN_SLOW_TESTS=1 PROPTEST_CASES=200000 \
    PROPTEST_MAX_GLOBAL_REJECTS=200000 RUST_BACKTRACE=1 \
    cargo test --release layout::tests::random_operations_dont_panic
```
Animation behavior looks ok, even when "stacking" animations, e.g. resizing/moving during swap.

YaLTeR:
Well, it's a randomized test, maybe you just didn't hit that case. Viewing the full output reveals one failing case:
```
2025-02-10T21:16:11.2180009Z thread 'layout::tests::random_operations_dont_panic' panicked at src/layout/tests.rs:3304:1:
2025-02-10T21:16:11.2180119Z Test failed: assertion `left == right` failed
2025-02-10T21:16:11.2180197Z   left: true
2025-02-10T21:16:11.2180270Z  right: false.
2025-02-10T21:16:11.2180367Z minimal failing input: ops = [
2025-02-10T21:16:11.2180439Z     AddWindow {
2025-02-10T21:16:11.2180533Z         params: TestWindowParams {
2025-02-10T21:16:11.2180611Z             id: 1,
2025-02-10T21:16:11.2180694Z             parent_id: None,
2025-02-10T21:16:11.2180781Z             is_floating: true,
2025-02-10T21:16:11.2180917Z             bbox: Rectangle<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2180991Z                 x: 0,
2025-02-10T21:16:11.2181065Z                 y: 0,
2025-02-10T21:16:11.2181140Z                 width: 1,
2025-02-10T21:16:11.2181220Z                 height: 1,
2025-02-10T21:16:11.2181291Z             },
2025-02-10T21:16:11.2181369Z             min_max_size: (
2025-02-10T21:16:11.2181492Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2181568Z                     w: 0,
2025-02-10T21:16:11.2181643Z                     h: 0,
2025-02-10T21:16:11.2181712Z                 },
2025-02-10T21:16:11.2181823Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2181894Z                     w: 0,
2025-02-10T21:16:11.2181965Z                     h: 0,
2025-02-10T21:16:11.2182043Z                 },
2025-02-10T21:16:11.2182112Z             ),
2025-02-10T21:16:11.2182186Z         },
2025-02-10T21:16:11.2182253Z     },
2025-02-10T21:16:11.2182334Z     AddWindowNextTo {
2025-02-10T21:16:11.2182432Z         params: TestWindowParams {
2025-02-10T21:16:11.2182503Z             id: 5,
2025-02-10T21:16:11.2182586Z             parent_id: None,
2025-02-10T21:16:11.2182667Z             is_floating: false,
2025-02-10T21:16:11.2182798Z             bbox: Rectangle<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2182873Z                 x: 0,
2025-02-10T21:16:11.2182943Z                 y: 0,
2025-02-10T21:16:11.2183022Z                 width: 1,
2025-02-10T21:16:11.2183096Z                 height: 1,
2025-02-10T21:16:11.2183169Z             },
2025-02-10T21:16:11.2183250Z             min_max_size: (
2025-02-10T21:16:11.2183354Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2183431Z                     w: 0,
2025-02-10T21:16:11.2183502Z                     h: 0,
2025-02-10T21:16:11.2183632Z                 },
2025-02-10T21:16:11.2183737Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2183810Z                     w: 0,
2025-02-10T21:16:11.2183886Z                     h: 0,
2025-02-10T21:16:11.2183953Z                 },
2025-02-10T21:16:11.2184027Z             ),
2025-02-10T21:16:11.2184096Z         },
2025-02-10T21:16:11.2184171Z         next_to_id: 1,
2025-02-10T21:16:11.2184284Z     },
2025-02-10T21:16:11.2184363Z     AddWindowNextTo {
2025-02-10T21:16:11.2184459Z         params: TestWindowParams {
2025-02-10T21:16:11.2184528Z             id: 2,
2025-02-10T21:16:11.2184649Z             parent_id: None,
2025-02-10T21:16:11.2184728Z             is_floating: false,
2025-02-10T21:16:11.2184857Z             bbox: Rectangle<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2184933Z                 x: 0,
2025-02-10T21:16:11.2185004Z                 y: 0,
2025-02-10T21:16:11.2185177Z                 width: 1,
2025-02-10T21:16:11.2185256Z                 height: 1,
2025-02-10T21:16:11.2185326Z             },
2025-02-10T21:16:11.2185409Z             min_max_size: (
2025-02-10T21:16:11.2185571Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2185651Z                     w: 0,
2025-02-10T21:16:11.2185722Z                     h: 0,
2025-02-10T21:16:11.2185792Z                 },
2025-02-10T21:16:11.2185899Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2185973Z                     w: 0,
2025-02-10T21:16:11.2186050Z                     h: 0,
2025-02-10T21:16:11.2186119Z                 },
2025-02-10T21:16:11.2186195Z             ),
2025-02-10T21:16:11.2186262Z         },
2025-02-10T21:16:11.2186338Z         next_to_id: 1,
2025-02-10T21:16:11.2186414Z     },
2025-02-10T21:16:11.2186488Z     AddWindow {
2025-02-10T21:16:11.2186579Z         params: TestWindowParams {
2025-02-10T21:16:11.2186650Z             id: 3,
2025-02-10T21:16:11.2186728Z             parent_id: None,
2025-02-10T21:16:11.2186816Z             is_floating: false,
2025-02-10T21:16:11.2186941Z             bbox: Rectangle<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2187018Z                 x: 0,
2025-02-10T21:16:11.2187088Z                 y: 0,
2025-02-10T21:16:11.2187162Z                 width: 1,
2025-02-10T21:16:11.2187241Z                 height: 1,
2025-02-10T21:16:11.2187311Z             },
2025-02-10T21:16:11.2187394Z             min_max_size: (
2025-02-10T21:16:11.2187500Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2187575Z                     w: 0,
2025-02-10T21:16:11.2187647Z                     h: 0,
2025-02-10T21:16:11.2187718Z                 },
2025-02-10T21:16:11.2187828Z                 Size<smithay::utils::geometry::Logical> {
2025-02-10T21:16:11.2187899Z                     w: 0,
2025-02-10T21:16:11.2187976Z                     h: 0,
2025-02-10T21:16:11.2188046Z                 },
2025-02-10T21:16:11.2188115Z             ),
2025-02-10T21:16:11.2188189Z         },
2025-02-10T21:16:11.2188259Z     },
2025-02-10T21:16:11.2188345Z     UnnameWorkspace {
2025-02-10T21:16:11.2188417Z         ws_name: 1,
2025-02-10T21:16:11.2188485Z     },
2025-02-10T21:16:11.2188565Z     AddOutput(
2025-02-10T21:16:11.2188633Z         1,
2025-02-10T21:16:11.2188704Z     ),
2025-02-10T21:16:11.2188794Z     FocusWindowOrWorkspaceDown,
2025-02-10T21:16:11.2188887Z     FocusWindowOrWorkspaceUp,
2025-02-10T21:16:11.2188987Z     ConsumeOrExpelWindowLeft {
2025-02-10T21:16:11.2189058Z         id: None,
2025-02-10T21:16:11.2189129Z     },
2025-02-10T21:16:11.2189219Z     ConsumeOrExpelWindowLeft {
2025-02-10T21:16:11.2189298Z         id: Some(
2025-02-10T21:16:11.2189368Z             2,
2025-02-10T21:16:11.2189435Z         ),
2025-02-10T21:16:11.2189508Z     },
2025-02-10T21:16:11.2189586Z     FullscreenWindow(
2025-02-10T21:16:11.2189657Z         3,
2025-02-10T21:16:11.2189724Z     ),
2025-02-10T21:16:11.2189809Z     SwapWindowInDirection(
2025-02-10T21:16:11.2189884Z         Left,
2025-02-10T21:16:11.2190017Z     ),
2025-02-10T21:16:11.2190105Z ], options = Options {
2025-02-10T21:16:11.2190178Z     gaps: 0.0,
2025-02-10T21:16:11.2190255Z     struts: Struts {
2025-02-10T21:16:11.2190338Z         left: FloatOrInt(
2025-02-10T21:16:11.2190410Z             0.0,
2025-02-10T21:16:11.2190482Z         ),
2025-02-10T21:16:11.2190560Z         right: FloatOrInt(
2025-02-10T21:16:11.2190628Z             0.0,
2025-02-10T21:16:11.2190699Z         ),
2025-02-10T21:16:11.2190837Z         top: FloatOrInt(
2025-02-10T21:16:11.2190909Z             0.0,
2025-02-10T21:16:11.2190976Z         ),
2025-02-10T21:16:11.2191053Z         bottom: FloatOrInt(
2025-02-10T21:16:11.2191197Z             0.0,
2025-02-10T21:16:11.2191263Z         ),
2025-02-10T21:16:11.2191333Z     },
2025-02-10T21:16:11.2191414Z     focus_ring: FocusRing {
2025-02-10T21:16:11.2191486Z         off: false,
2025-02-10T21:16:11.2191569Z         width: FloatOrInt(
2025-02-10T21:16:11.2191637Z             0.0,
2025-02-10T21:16:11.2191708Z         ),
2025-02-10T21:16:11.2191792Z         active_color: Color {
2025-02-10T21:16:11.2191866Z             r: 0.49803922,
2025-02-10T21:16:11.2191943Z             g: 0.78431374,
2025-02-10T21:16:11.2192053Z             b: 1.0,
2025-02-10T21:16:11.2192130Z             a: 1.0,
2025-02-10T21:16:11.2192198Z         },
2025-02-10T21:16:11.2192285Z         inactive_color: Color {
2025-02-10T21:16:11.2192359Z             r: 0.3137255,
2025-02-10T21:16:11.2192432Z             g: 0.3137255,
2025-02-10T21:16:11.2192508Z             b: 0.3137255,
2025-02-10T21:16:11.2192578Z             a: 1.0,
2025-02-10T21:16:11.2192650Z         },
2025-02-10T21:16:11.2192736Z         active_gradient: None,
2025-02-10T21:16:11.2192824Z         inactive_gradient: None,
2025-02-10T21:16:11.2192898Z     },
2025-02-10T21:16:11.2192977Z     border: Border {
2025-02-10T21:16:11.2193052Z         off: false,
2025-02-10T21:16:11.2193134Z         width: FloatOrInt(
2025-02-10T21:16:11.2193204Z             0.0,
2025-02-10T21:16:11.2193283Z         ),
2025-02-10T21:16:11.2193366Z         active_color: Color {
2025-02-10T21:16:11.2193441Z             r: 1.0,
2025-02-10T21:16:11.2193517Z             g: 0.78431374,
2025-02-10T21:16:11.2193595Z             b: 0.49803922,
2025-02-10T21:16:11.2193668Z             a: 1.0,
2025-02-10T21:16:11.2193735Z         },
2025-02-10T21:16:11.2193825Z         inactive_color: Color {
2025-02-10T21:16:11.2193899Z             r: 0.3137255,
2025-02-10T21:16:11.2193972Z             g: 0.3137255,
2025-02-10T21:16:11.2194051Z             b: 0.3137255,
2025-02-10T21:16:11.2194119Z             a: 1.0,
2025-02-10T21:16:11.2194192Z         },
2025-02-10T21:16:11.2194277Z         active_gradient: None,
2025-02-10T21:16:11.2194373Z         inactive_gradient: None,
2025-02-10T21:16:11.2194440Z     },
2025-02-10T21:16:11.2194516Z     shadow: Shadow {
2025-02-10T21:16:11.2194593Z         on: false,
2025-02-10T21:16:11.2194674Z         offset: ShadowOffset {
2025-02-10T21:16:11.2194754Z             x: FloatOrInt(
2025-02-10T21:16:11.2194826Z                 0.0,
2025-02-10T21:16:11.2194897Z             ),
2025-02-10T21:16:11.2194976Z             y: FloatOrInt(
2025-02-10T21:16:11.2195046Z                 5.0,
2025-02-10T21:16:11.2195218Z             ),
2025-02-10T21:16:11.2195288Z         },
2025-02-10T21:16:11.2195370Z         softness: FloatOrInt(
2025-02-10T21:16:11.2195444Z             0.0,
2025-02-10T21:16:11.2195512Z         ),
2025-02-10T21:16:11.2195600Z         spread: FloatOrInt(
2025-02-10T21:16:11.2195671Z             5.0,
2025-02-10T21:16:11.2195736Z         ),
2025-02-10T21:16:11.2195831Z         draw_behind_window: false,
2025-02-10T21:16:11.2195906Z         color: Color {
2025-02-10T21:16:11.2195984Z             r: 0.0,
2025-02-10T21:16:11.2196055Z             g: 0.0,
2025-02-10T21:16:11.2196124Z             b: 0.0,
2025-02-10T21:16:11.2196200Z             a: 0.4392157,
2025-02-10T21:16:11.2196273Z         },
2025-02-10T21:16:11.2196357Z         inactive_color: None,
2025-02-10T21:16:11.2196425Z     },
2025-02-10T21:16:11.2196586Z     tab_indicator: TabIndicator {
2025-02-10T21:16:11.2196666Z         off: false,
2025-02-10T21:16:11.2196753Z         hide_when_single_tab: false,
2025-02-10T21:16:11.2196845Z         place_within_column: false,
2025-02-10T21:16:11.2196925Z         gap: FloatOrInt(
2025-02-10T21:16:11.2196999Z             0.0,
2025-02-10T21:16:11.2197065Z         ),
2025-02-10T21:16:11.2197143Z         width: FloatOrInt(
2025-02-10T21:16:11.2197216Z             0.0,
2025-02-10T21:16:11.2197340Z         ),
2025-02-10T21:16:11.2197433Z         length: TabIndicatorLength {
2025-02-10T21:16:11.2197524Z             total_proportion: Some(
2025-02-10T21:16:11.2197645Z                 0.0,
2025-02-10T21:16:11.2197718Z             ),
2025-02-10T21:16:11.2197790Z         },
2025-02-10T21:16:11.2197872Z         position: Left,
2025-02-10T21:16:11.2197967Z         gaps_between_tabs: FloatOrInt(
2025-02-10T21:16:11.2198035Z             0.0,
2025-02-10T21:16:11.2198106Z         ),
2025-02-10T21:16:11.2198185Z         active_color: None,
2025-02-10T21:16:11.2198273Z         inactive_color: None,
2025-02-10T21:16:11.2198354Z         active_gradient: None,
2025-02-10T21:16:11.2198491Z         inactive_gradient: None,
2025-02-10T21:16:11.2198564Z     },
2025-02-10T21:16:11.2198646Z     insert_hint: InsertHint {
2025-02-10T21:16:11.2198722Z         off: false,
2025-02-10T21:16:11.2198795Z         color: Color {
2025-02-10T21:16:11.2198872Z             r: 0.49803922,
2025-02-10T21:16:11.2198945Z             g: 0.78431374,
2025-02-10T21:16:11.2199016Z             b: 1.0,
2025-02-10T21:16:11.2199092Z             a: 0.5019608,
2025-02-10T21:16:11.2199161Z         },
2025-02-10T21:16:11.2199244Z         gradient: None,
2025-02-10T21:16:11.2199311Z     },
2025-02-10T21:16:11.2199398Z     center_focused_column: Never,
2025-02-10T21:16:11.2199495Z     always_center_single_column: false,
2025-02-10T21:16:11.2199590Z     empty_workspace_above_first: false,
2025-02-10T21:16:11.2199684Z     default_column_display: Normal,
2025-02-10T21:16:11.2199768Z     preset_column_widths: [
2025-02-10T21:16:11.2199841Z         Proportion(
2025-02-10T21:16:11.2199919Z             0.3333333333333333,
2025-02-10T21:16:11.2199988Z         ),
2025-02-10T21:16:11.2200067Z         Proportion(
2025-02-10T21:16:11.2200137Z             0.5,
2025-02-10T21:16:11.2200205Z         ),
2025-02-10T21:16:11.2200284Z         Proportion(
2025-02-10T21:16:11.2200356Z             0.6666666666666666,
2025-02-10T21:16:11.2200432Z         ),
2025-02-10T21:16:11.2200500Z     ],
2025-02-10T21:16:11.2200583Z     default_column_width: None,
2025-02-10T21:16:11.2200672Z     preset_window_heights: [
2025-02-10T21:16:11.2200745Z         Proportion(
2025-02-10T21:16:11.2200824Z             0.3333333333333333,
2025-02-10T21:16:11.2200892Z         ),
2025-02-10T21:16:11.2200970Z         Proportion(
2025-02-10T21:16:11.2201039Z             0.5,
2025-02-10T21:16:11.2201107Z         ),
2025-02-10T21:16:11.2201185Z         Proportion(
2025-02-10T21:16:11.2201257Z             0.6666666666666666,
2025-02-10T21:16:11.2201330Z         ),
2025-02-10T21:16:11.2201397Z     ],
2025-02-10T21:16:11.2201481Z     animations: Animations {
2025-02-10T21:16:11.2201557Z         off: false,
2025-02-10T21:16:11.2201634Z         slowdown: 1.0,
2025-02-10T21:16:11.2201746Z         workspace_switch: WorkspaceSwitchAnim(
2025-02-10T21:16:11.2201821Z             Animation {
2025-02-10T21:16:11.2201898Z                 off: false,
2025-02-10T21:16:11.2201983Z                 kind: Spring(
2025-02-10T21:16:11.2202067Z                     SpringParams {
2025-02-10T21:16:11.2202161Z                         damping_ratio: 1.0,
2025-02-10T21:16:11.2202245Z                         stiffness: 1000,
2025-02-10T21:16:11.2202329Z                         epsilon: 0.0001,
2025-02-10T21:16:11.2202405Z                     },
2025-02-10T21:16:11.2202473Z                 ),
2025-02-10T21:16:11.2202547Z             },
2025-02-10T21:16:11.2202615Z         ),
2025-02-10T21:16:11.2202710Z         window_open: WindowOpenAnim {
2025-02-10T21:16:11.2202842Z             anim: Animation {
2025-02-10T21:16:11.2202917Z                 off: false,
2025-02-10T21:16:11.2202998Z                 kind: Easing(
2025-02-10T21:16:11.2203080Z                     EasingParams {
2025-02-10T21:16:11.2203169Z                         duration_ms: 150,
2025-02-10T21:16:11.2203257Z                         curve: EaseOutExpo,
2025-02-10T21:16:11.2203327Z                     },
2025-02-10T21:16:11.2203402Z                 ),
2025-02-10T21:16:11.2203511Z             },
2025-02-10T21:16:11.2203597Z             custom_shader: None,
2025-02-10T21:16:11.2203665Z         },
2025-02-10T21:16:11.2203754Z         window_close: WindowCloseAnim {
2025-02-10T21:16:11.2203877Z             anim: Animation {
2025-02-10T21:16:11.2203950Z                 off: false,
2025-02-10T21:16:11.2204028Z                 kind: Easing(
2025-02-10T21:16:11.2204106Z                     EasingParams {
2025-02-10T21:16:11.2204188Z                         duration_ms: 150,
2025-02-10T21:16:11.2204278Z                         curve: EaseOutQuad,
2025-02-10T21:16:11.2204347Z                     },
2025-02-10T21:16:11.2204419Z                 ),
2025-02-10T21:16:11.2204490Z             },
2025-02-10T21:16:11.2204612Z             custom_shader: None,
2025-02-10T21:16:11.2204682Z         },
2025-02-10T21:16:11.2204825Z         horizontal_view_movement: HorizontalViewMovementAnim(
2025-02-10T21:16:11.2204904Z             Animation {
2025-02-10T21:16:11.2204982Z                 off: false,
2025-02-10T21:16:11.2205065Z                 kind: Spring(
2025-02-10T21:16:11.2205238Z                     SpringParams {
2025-02-10T21:16:11.2205325Z                         damping_ratio: 1.0,
2025-02-10T21:16:11.2205415Z                         stiffness: 800,
2025-02-10T21:16:11.2205496Z                         epsilon: 0.0001,
2025-02-10T21:16:11.2205570Z                     },
2025-02-10T21:16:11.2205638Z                 ),
2025-02-10T21:16:11.2205707Z             },
2025-02-10T21:16:11.2205779Z         ),
2025-02-10T21:16:11.2205880Z         window_movement: WindowMovementAnim(
2025-02-10T21:16:11.2205959Z             Animation {
2025-02-10T21:16:11.2206032Z                 off: false,
2025-02-10T21:16:11.2206115Z                 kind: Spring(
2025-02-10T21:16:11.2206194Z                     SpringParams {
2025-02-10T21:16:11.2206278Z                         damping_ratio: 1.0,
2025-02-10T21:16:11.2206364Z                         stiffness: 800,
2025-02-10T21:16:11.2206445Z                         epsilon: 0.0001,
2025-02-10T21:16:11.2206520Z                     },
2025-02-10T21:16:11.2206588Z                 ),
2025-02-10T21:16:11.2206657Z             },
2025-02-10T21:16:11.2206730Z         ),
2025-02-10T21:16:11.2206820Z         window_resize: WindowResizeAnim {
2025-02-10T21:16:11.2206903Z             anim: Animation {
2025-02-10T21:16:11.2206978Z                 off: false,
2025-02-10T21:16:11.2207056Z                 kind: Spring(
2025-02-10T21:16:11.2207140Z                     SpringParams {
2025-02-10T21:16:11.2207228Z                         damping_ratio: 1.0,
2025-02-10T21:16:11.2207312Z                         stiffness: 800,
2025-02-10T21:16:11.2207395Z                         epsilon: 0.0001,
2025-02-10T21:16:11.2207470Z                     },
2025-02-10T21:16:11.2207538Z                 ),
2025-02-10T21:16:11.2207608Z             },
2025-02-10T21:16:11.2207693Z             custom_shader: None,
2025-02-10T21:16:11.2207762Z         },
2025-02-10T21:16:11.2207945Z         config_notification_open_close: ConfigNotificationOpenCloseAnim(
2025-02-10T21:16:11.2208022Z             Animation {
2025-02-10T21:16:11.2208095Z                 off: false,
2025-02-10T21:16:11.2208181Z                 kind: Spring(
2025-02-10T21:16:11.2208260Z                     SpringParams {
2025-02-10T21:16:11.2208351Z                         damping_ratio: 0.6,
2025-02-10T21:16:11.2208432Z                         stiffness: 1000,
2025-02-10T21:16:11.2208512Z                         epsilon: 0.001,
2025-02-10T21:16:11.2208586Z                     },
2025-02-10T21:16:11.2208723Z                 ),
2025-02-10T21:16:11.2208798Z             },
2025-02-10T21:16:11.2208867Z         ),
2025-02-10T21:16:11.2208985Z         screenshot_ui_open: ScreenshotUiOpenAnim(
2025-02-10T21:16:11.2209061Z             Animation {
2025-02-10T21:16:11.2209137Z                 off: false,
2025-02-10T21:16:11.2209221Z                 kind: Easing(
2025-02-10T21:16:11.2209300Z                     EasingParams {
2025-02-10T21:16:11.2209451Z                         duration_ms: 200,
2025-02-10T21:16:11.2209536Z                         curve: EaseOutQuad,
2025-02-10T21:16:11.2209607Z                     },
2025-02-10T21:16:11.2209734Z                 ),
2025-02-10T21:16:11.2209804Z             },
2025-02-10T21:16:11.2209879Z         ),
2025-02-10T21:16:11.2209947Z     },
2025-02-10T21:16:11.2210042Z     disable_resize_throttling: false,
2025-02-10T21:16:11.2210137Z     disable_transactions: false,
2025-02-10T21:16:11.2210220Z }, post_options = None
```

rustn00b:
I've been wrestling with this for longer than I care to admit. The latest commit
works well for all the non-test-case situations I have tried on it. However
as of 25.01, the test cases did fail. As far as I was able to make out, the test
cases expect that following an action, TileData (size cache) has been updated
to match the size reported by a Tile. But in the case of the Tile swap action,
Tile Data is intentionally *not* updated because the cached value is precisely
the size that the moved tiles are supposed to have at the end of their animation. The cached
size is then used to calculate column widths, and therefore column offsets
and ultimately the view position. Not updating the cache makes all these
calculations yield the same result as prior to the swap action, which turns out
to be exactly what's needed visually (no columns jumping around). Trying to update the cache
(to satisfy the test case) and then "compensating" for the update only led to
more complexity and other misbehavior, so I reverted these changes. Of course
this is most likely a case of "your holding it wrong" ;)

In an interesting development, pulling 25.02 looked as if it had fixed things!? The
`random_operations_dont_panic` test isn't complaining with the following proptest
settings: `PROPTEST_CASES=200000 PROPTEST_MAX_GLOBAL_REJECTS=200000`. Bumping
that up to 400000 yielded no failures either (I only let it run for an hour or
so). The CI run however did fail :/

YaLTeR:
The way this works is proptest generates random actions from that big Op enum, and when it finds a failure, it writes the random seed into proptest_regressions. When we add a new action in to the Op enum, I think what happens is the seed no longer generates the same sequence of actions, so the test may go back to passing until you hit that bad case again.

Anyhow, regarding the animation fix: I'm planning to look into improving the resize animation code at some point to fix other issues I found, and I believe doing it properly will also help fix the swap-window problems without the workaround field. (I'm not rushing to do it, but I'll get there at some point, if that's ok.)

rustn00b:
Yup, I had looked into proptest from when I started working on failing test cases for this PR. I kept my test seeds to measure progress. As you mention, messing with the Ops or strategies voids reproducibility of seeds (but reverting to the original config makes them once again behave as expected). I hadn't realized 25.02 had modified the Ops so I initially thought a happy side-effect of the update was to have fixed this PR.

I'm curious to see how you change the animation code and whether it either solves the failing TC for this PR, or somehow makes the path to resolving it clearer to me.

Fwiw f595cdc has been working nicely (despite the failing TC).

rustn00b:
I picked this one up again because I think it may turn out to be needed for https://github.com/YaLTeR/niri/pull/1704. It is behaving as intended, namely the test cases run to completion, and the visual glitch (where a window would sometimes flash when swapped) is resolved.

YaLTeR:
Hey, I finally got to this problem. In https://github.com/YaLTeR/niri/pull/2229 I fixed the window resize cancel animation jump along with a metric ton of tests. It is a proper fix instead of the `inhibit_sibling_move_on_resize`.

I haven't investigated the swap-window animation issue yet, but my hope is that on top of that PR it can be fixed without hacks.

rustn00b:
Nice to see this PR getting some attention again! The `inhibit_sibling_move_on_resize` hack has actually been gone for a while. The way things have since changed is with the introduction of a `prefer_expected_size` boolean on Tile. When it is set, calculations for the positions of columns and other tiles consider that the marked Tile already has its (final) expected size. When two Tiles are swapped, their expected size is set to the current size of their peer. This produces the behavior I was initially aiming for: other tiles are unaffected and the the swapped Tiles just glide into the vacated slot. However, if another action (with an effect on the Tile size) is processed, then all the other Tiles move to adapt to the change.
Besides this, there was a 1 frame glitch that was distracting (also since addressed in this PR).
And finally, the handling of the "data" cache was changed to ensure invariants were preserved (and therefore allow the TCs to complete).

Without this PR a Tile sitting below a swapped Tile seems to jump slightly during the animation, when one would instead expect it to stay still.

YaLTeR:
Could you please enumerate all glitchy animation cases that you know of without this PR? I still want to take a stab at this without adding extra fields to Tile so I need a list of things to check.

If you feel like it, you can even write these cases in form of animation tests like the ones I added in the last PR. I'll probably end up having to do that anyway.

rustn00b:
Here's a list of the differences in behavior I've noticed between this PR and 25c1c04 (main). Unless otherwise stated the behavior described is the one from 25c1c04:

- with the layout  illustrated below,
  ```
  +---+ +---+
  | A | |   |
  +---+ |   |
  +---+ | C |
  | B | |   |
  +---+ +---+ 
  ```
  Swapping A and C makes B jump slightly.

- With a layout similar to the one in the previous bullet, if A has
  a width smaller than C's, and assuming that A is on the compete left of the current view, then swapping A and C will make the view shift (as if it had to accommodate the original width of C in the "C + B" column). Expected behavior is for the column to stay still.

- swapping two windows in one direction and immediately swapping back into the opposite direction shows expected behavior for the focused window, however the other swapped window jumps back to its starting position instead of animating from the position it had reached due to the first swap.

- With the following layout:
  ```
  +---+ +---+
  | A | | C |
  +---+ +---+
  +---+ +---+
  | B | | D |
  +---+ +---+ 
  ```
  Swapping B and D makes the gap between A and C change during the animation where it should stay constant.

I did not notice the 1 frame flash that was previously present. 

I tried to look around in the test cases to see where tests for these behaviors could fit but couldn't really figure it out. Layout tests seem animation-agnostic, so I guess niri-visual-tests would be more appropriate but I was a bit stumped as to how it should be done.

YaLTeR:
Thanks!

> I tried to look around in the test cases to see where tests for these behaviors could fit but couldn't really figure it out.

These new ones: https://github.com/YaLTeR/niri/blob/main/src/layout/tests/animations.rs

These are snapshots, meaning you run `cargo insta test --accept -- animations` and it fills in the correct values automatically.

YaLTeR:
I think I came up with one case which may appear to require `prefer_expected_size`. Consider this layout, assume it covers the whole screen. B is focused.
```
┌───────────────┐ (1)
│┌───────┐ ┌───┐│
││   A   │ │ C ││
│└───────┘ └───┘│
│┌───────┐ ┌───┐│
││   B   │ │ D ││
│└───────┘ └───┘│
└───────────────┘
```
Swap B and D.
```
┌───────────────┐ (2)
│┌───────┐ ┌───┐│   
││   A   │ │ C ││   
│└───────┘ └───┘│   
│┌───┐     ┌────┼──┐
││ D │     │   B│  │
│└───┘     └────┼──┘
└───────────────┘   
```
The windows will take some time to resize, meanwhile the view offset logic starts to move the (right) newly focused column to be fully in view.
```
┌───────────────┐ (3)
┼────┐┌───┐     │
│A   ││ C │     │
┼────┘└───┘     │
┼┐    ┌───────┐ │
││    │   B   │ │
┼┘    └───────┘ │
└───────────────┘
```
B and D finally resize, but the focused column is the right one, which is already (animating to) fully in view, so as far as the layout is concerned, no further view movement is needed, so the final state stays partially moved like this.
```
┌───────────────┐ (4)
┼────┐┌───┐     │
│A   ││ C │     │
┼────┘└───┘     │
┼────┐┌───┐     │
│D   ││ B │     │
┼────┘└───┘     │
└───────────────┘
```
This is not expected, the expected state is that all columns stay in view like initially, but with B and D swapped places.

----

You propose to solve this with `prefer_expected_size`. This way, at (2), the view offset logic will look at the (small) expected size rather than the current size, and won't move the columns just yet since the expected sizes do fit. Then, once B updates, if it assumed the expected size then nothing will happen, and if it did not comply, only then the view offset logic will (correctly) start moving the column to the left to fit in view.

Now, adding `prefer_expected_size` isn't the only way so solve this (and it's not my favorite because it feels like a hack). Here are two more to consider:
1. Adding transactions for more actions like consume, expel, swap that will first wait for the windows to resize to the new sizes (up to a certain time limit), and only then start animating the resizes, movement and recompute the view offset.

    This already works for column resizes for example (where it is admittedly simpler), note how the view offset doesn't start animating until wleird-slow-ack-configure finally resizes:

    https://github.com/user-attachments/assets/9dccebe5-3ef2-49c8-9a19-007acd7b93ee
2. Making the "put focused column in view" logic (and possibly other logic) *always* consider the expected size rather than the current size. The expected size became a thing only with the introduction of floating windows to niri, so perhaps it's time to reevaluate some older actions and figure out if they need to use it too.

Now, option 1 is likely *quite involved*, but it's also something that we want to do anyway in the long run, since it fixes flickering when animations are disabled.

Option 2 should be easier, but needs careful consideration. I'm not prepared to say that it's the right thing to do. For example, it will cause moving a wide wleird-slow-ack-configure into a narrow right-flushed column to peek outside the screen until it finally decides to resize. Actually, since expected size doesn't have a timeout (unlike transactions), if a client never acks and commits a new size (which shouldn't happen but we want to be robust against buggy clients), then the window will stay peaking out of the screen. This makes me think that option 2 is the wrong choice, and the transactions approach is ultimately the way to go.

----

This is one specific case that I think is currently unsolvable without either option 1 or 2 (or something like `prefer_expected_size` which I'd like to avoid). Other animation jank with swap-window may be fixable as is (I have some ideas).

Anyway, did this change turn out to be required for MRU?

> I picked this one up again because I think it may turn out to be needed for https://github.com/YaLTeR/niri/pull/1704.

Because if not then that's good as I'd rather look at MRU than keep looking into this for the time being (since as I mentioned, option 1 may be quite involved to do, even if we ultimately want it anyway at some point).

rustn00b:
Thanks for the detailed description! 

> because it feels like a hack

I went with the `prefer_expected_size` because I thought it was _less_ hackish ¯\_(ツ)_/¯. Though I did have to try several things that all didn't quite work, I settled on the approach because I thought it really expressed the intent: for this particular tile don't consider transient size when positioning other tiles, and instead use the size the tile will ultimately have. This was what looked appropriate wrt the swap action.

As for doing this with transactions (resize, then move), I guess that would work out wrt to shuffling other tiles around, but wouldn't it look odd for it to happen in two stages? Here I really must admit being rather clueless about how animation transactions work, so I'm likely completely wrong.

> Anyway, did this change turn out to be required for MRU?

Nope. I had considered using a _resize_ animation for the thumbnail selection animation for a time and thought I would need the `prefer_expected_size` behavior to avoid upsetting other tiles. But I just opted to inhibit rendering of the selected tile (with another boolean that marks the tile to be skipped in regular rendering, sry....), and the thumbnail for the tile is rendered as part of the MRU UI.

ccjmne:
I came here to look into the issue that @YaLTeR brought up regarding to "too eagerly" panning the view: it also occurs when simply consuming a larger tile:

```
(1)    ┌───────────────┐
       │┌─────┐ ┏━━━━━┓│ ┌─────────────┐
       ││     │ ┃     ┃│ │             │
       ││  A  │ ┃  B  ┃│ │      C      │
       ││     │ ┃     ┃│ │             │
       │└─────┘ ┗━━━━━┛│ └─────────────┘
       └───────────────┘

(2)    ┌───────────────┐
       │┌─────┐ ┏━━━━━┓│
       ││     │ ┃  B  ┃│
       ││  A  │ ┗━━━━━┛│
       ││     │ ┌──────┼──────┐
       ││     │ │      C      │
       │└─────┘ └──────┼──────┘
       └───────────────┘

(3)    ┌───────────────┐
┌─────┐│┏━━━━━┓        │
│     ││┃  B  ┃        │
│  A  ││┗━━━━━┛        │
│     ││┌─────────────┐│
│     │││      C      ││
└─────┘│└─────────────┘│
       └───────────────┘

(4)    ┌───────────────┐
┌─────┐│┏━━━━━┓        │
│     ││┃  B  ┃        │
│  A  ││┗━━━━━┛        │
│     ││┌─────┐        │
│     │││  C  │        │
└─────┘│└─────┘        │
       └───────────────┘
```

The `B` tile (with the thicker outline) is the one focused before executing `consume-window-into-column`.

Curiously, I cannot reproduce that with the same consistency across all applications—all native Wayland. It possibly has to do with the speed at which a client window can handle the resize:
- I can't *at all* reproduce this issue with A, B, C being instances of `ghostty`,
- I can *sometimes* reproduce this issue with them being all `nautilus`, and
- I can *always* reproduce it with a triplet of `google-chrome` windows.

Is this the right place to have that conversation?

I just checked out, compiled and ran this fork (at `14fbd5d`) but the issue with the "overly eager panning" is still very much present. Is the current implementation (with `prefer_expected_size`) not supposed to address that?
I intend to get somewhat more familiar with the code base "one of these days", but figured I could just ask outright as well.

YaLTeR:
It's a similar problem yes. Would also be solved with transactions (but it is not an easy thing to add here, will need careful consideration and work).

PR #1017, opened 2025-01-19
IPC: Accept multiple actions from stdin

bbb651:
This needs some testing, `io::stdin().lines()` behaves a little weird with EOF, notably pressing <kbd>Ctrl</kbd> + <kbd>D</kbd> twice (apparently that's a unix thing, you need to hit it twice when not at the start of a line), you still need to hit it an additional time for some reason.. with `echo -e "one\ntwo" | niri msg action` it works as expected (by the way, it seems like `NIRI_SOCKET` isn't replaced correctly in the winit backend).

I made the serialization prefer keeping 1-length vectors as single values to maximize compatibility. The amount of tests for `one_or_many` is a bit excessive (I thought a visitor was needed at first, and then got the order of the enum variants wrong, so they were helpful), should I only keep the derive ones?

Also I'm not sure what should happen if you pass no actions, currently the cli prints a warning and does nothing but it's allowed in the ipc, should niri respond with an error instead?

Also, this currently splits on spaces, meaning it handles quoting and multiple spaces incorrectly, should this use e.g. `shlex` instead?

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -43,6 +43,8 @@ use serde::{Deserialize, Serialize};
 pub mod socket;
 pub mod state;
 
+mod utils;
+
 /// Request from client to niri.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
@@ -63,8 +65,8 @@ pub enum Request {
     FocusedOutput,
     /// Request information about the focused window.
     FocusedWindow,
-    /// Perform an action.
-    Action(Action),
+    /// Perform actions.
+    Action(#[serde(with = "utils::one_or_many")] Vec<Action>),
     /// Change output configuration temporarily.
     ///
     /// The configuration is changed temporarily and not saved into the config file. If the output
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -74,9 +74,13 @@ pub enum Msg {
     FocusedWindow,
     /// Perform an action.
     Action {
+        // NOTE: This is only optional because of clap_derive limitations and will never be `None`.
+        // If an action is not provided it reads actions from stdin and turns to [`Msg::Actions`].
         #[command(subcommand)]
-        action: Action,
+        action: Option<Action>,
     },
+    #[clap(skip)]
+    Actions { actions: Vec<Action> },
     /// Change output configuration temporarily.
     ///
     /// The configuration is changed temporarily and not saved into the config file. If the output
--- a/niri-ipc/src/utils.rs
+++ b/niri-ipc/src/utils.rs
@@ -30,55 +30,57 @@ pub(crate) mod one_or_many {
 
     #[cfg(test)]
     mod tests {
+        use std::fmt::Debug;
+
         use serde_json::de::SliceRead;
         use serde_json::{Deserializer, Serializer, Value};
 
         use super::*;
 
+        fn test_serialize<T: Serialize>(value: &Vec<T>, expected: &str) {
+            let mut bytes = Vec::new();
+            let mut serializer = Serializer::new(&mut bytes);
+            serialize(value, &mut serializer).expect("failed to serialize");
+            assert_eq!(String::from_utf8_lossy(&bytes), expected);
+        }
+
+        fn test_deserialize<'de, T>(value: &'de str, expected: &Vec<T>)
+        where
+            T: Deserialize<'de> + Debug + PartialEq,
+        {
+            let mut deserailier = Deserializer::new(SliceRead::new(value.as_bytes()));
+            let result: Vec<T> = deserialize(&mut deserailier).expect("failed to deserialize");
+            assert_eq!(&result, expected);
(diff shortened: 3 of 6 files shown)


niri-ipc/src/lib.rs
```diff
@@ -63,8 +65,8 @@ pub enum Request {
     FocusedOutput,
     /// Request information about the focused window.
     FocusedWindow,
-    /// Perform an action.
-    Action(Action),
+    /// Perform actions.
+    Action(#[serde(with = "utils::one_or_many")] Vec<Action>),
```
YaLTeR, line 69: Couldn't we just add a separate `Actions` variant and avoid the whole custom parsing?

src/cli.rs
```diff
@@ -74,9 +74,13 @@ pub enum Msg {
     FocusedWindow,
     /// Perform an action.
     Action {
+        // NOTE: This is only optional because of clap_derive limitations and will never be `None`.
+        // If an action is not provided it reads actions from stdin and turns to [`Msg::Actions`].
```
YaLTeR, line 78: Similarly, could this remain as non-optional, with a separate `niri msg actions` that does the stdin reading?

YaLTeR (commented):
> Also, this currently splits on spaces, meaning it handles quoting and multiple spaces incorrectly, should this use e.g. shlex instead?

I understand the problem but I feel like more complex lexing is a worse trade-off. Especially when actions are generated programmatically from an existing shell, the last thing you want is some weird extra splitting.

niri-ipc/src/lib.rs
bbb651: Yeah that makes sense, it's simpler and also more ergonomic for `niri-ipc`. The "compactness" of the json representation doesn't really matter.

src/cli.rs
bbb651: To me this feels a bit weird to me, because it is the same subcommand in terms of functionality, just with different input, unlike how `niri msg outputs` and `niri msg output` are different.
Most commands that takes a positional arguments fallback to stdin if it's missing, e.g. `cat`, `jq`, `nvim`, etc. so it feels logical to follow the same convention. Granted this is a bit different since it takes it normally takes a single literal action and not a file of actions (and a flag like `-i -` feels redundant since it's mutually exclusive with a literal action and the majority of the time you'd use stdin). It might be a bit confusing to have `niri msg action` not print the help message though...

niri-ipc/src/lib.rs
bbb651: I wish rust allowed somehow matching over both a `T` and `Vec<T>` as `&[T]`, the code in the match arm ends up duplicated and it's awkward to break into a helper since it's async and environment capturing, it's just short enough that it's probably fine..?
```rs
        Request::Action(action) => {
            let (tx, rx) = async_channel::bounded(1);

            ctx.event_loop.insert_idle(move |state| {
                // Make sure some logic like workspace clean-up has a chance to run before doing
                // actions.
                state.niri.advance_animations();
                state.do_action(action.into(), false);
                let _ = tx.send_blocking(());
            });

            // Wait until the action has been processed before returning. This is important for a
            // few actions, for instance for DoScreenTransition this wait ensures that the screen
            // contents were sampled into the texture.
            let _ = rx.recv().await;
            Response::Handled
        }
        Request::Actions(actions) => {
            if actions.is_empty() {
                return Ok(Response::Handled);
            }

            let (tx, rx) = async_channel::bounded(1);

            ctx.event_loop.insert_idle(move |state| {
                state.niri.advance_animations();
                for action in actions.into_iter().map(niri_config::Action::from) {
                    state.do_action(action, false);
                }
                let _ = tx.send_blocking(());
            });

            let _ = rx.recv().await;
            Response::Handled
        }
```

src/cli.rs
YaLTeR: > It might be a bit confusing to have niri msg action not print the help message though...

Yeah, I want that to keep printing help.

Maybe then something more descriptive like `niri msg multiple-actions` or `niri msg actions-from-stdin` or something?

bbb651: > Yeah, I want that to keep printing help.

Another option is [checking if it's a terminal](https://doc.rust-lang.org/beta/std/io/trait.IsTerminal.html#tymethod.is_terminal) and printing the help message only in that case, then the help message can show a `echo -e '...' | niri msg action` example

niri-ipc/src/lib.rs
YaLTeR: I guess you could try something like `async do_actions(ctx, actions: impl Iterator<Item=Action>)`?

src/cli.rs
YaLTeR: Nah, I think an explicit separate subcommand is better. Also I guess if we pick `multiple-actions` or `do-multiple-actions`, then the IPC enum variant should also be renamed to `MultipleActions` for consistency.

YaLTeR: Or maybe ActionBatch?

YaLTeR: Or maybe ActionBatch?

PR #1026, opened 2025-01-20
Add `scroll-factor` Window Rule

### LINKED ISSUE #909: Scroll Factor Window Rule
I think it could be very useful if you can configure the scroll-factor as window-rule. 

In my case, in the browser and apps with electron, I would prefer to have the scroll factor at 0.1. but in helix, zed and so on, I would prefer to have the scroll-factor disabled

bbb651:
Example window rule:
```kdl
window-rule {
	match app-id=r#"^wev$"#
	scroll-factor 4
}
```
Note this multiples the scroll factor rather than override it like the issue suggested. This way you can normalize all the window scroll factors once, and it'll scale like you'd expect when changing your input scroll factor and when using multiple input devices with different scroll factors. You can "reset" a scroll factor by setting it to the input device's scroll factor inverse.

Fixes #909.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2086,6 +2086,8 @@ impl State {
     }
 
     fn on_pointer_axis<I: InputBackend>(&mut self, event: I::PointerAxisEvent) {
+        let pointer = &self.niri.seat.get_pointer().unwrap();
+
         let source = event.source();
 
         // We received an event for the regular pointer, so show it now. This is also needed for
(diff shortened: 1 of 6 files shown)


src/input/mod.rs
```diff
@@ -2171,7 +2172,12 @@ impl State {
             AxisSource::Finger => self.niri.config.borrow().input.touchpad.scroll_factor,
             _ => None,
         };
-        let scroll_factor = scroll_factor.map(|x| x.0).unwrap_or(1.);
+        let window_scroll_factor = self
+            .niri
+            .window_under_cursor()
+            .and_then(|window| window.rules().scroll_factor);
```
YaLTeR, line 2178: So this probably works in most cases, but ideally you should call `update_pointer_contents()` from below first, then get `pointer.current_focus()` and then find the window for that WlSurface. This is because pointer grabs will modify the pointer focus under the hood.

YaLTeR (commented):
Otherwise looks good

YaLTeR:
Rebased and made it work in popups too.

YaLTeR:
Thanks. Guess it would make sense to add as a layer rule too

PR #1031, opened 2025-01-21
`niri msg pick-window`

### LINKED ISSUE #589: niri msg pick-window
A new command, `niri msg pick-window`, that lets you pick a window with your mouse, then prints information about it. Kinda like `niri msg focused-window` but with a mouse picker.

Will also be useful for IPC in the future.

bbb651:
https://github.com/user-attachments/assets/94144688-caf6-4ffb-a467-139c3e5a538d

Originally I tried to indicate picking was occurring by turning the cursor into a crosshair, in the same way as the screenshot UI [in the seat handler](https://github.com/YaLTeR/niri/blob/ae2ea53c9598a58de4a78d56f150915c3610157d/src/handlers/mod.rs#L95) but that means we're forgetting what the original cursor was, I think doing it [here](https://github.com/YaLTeR/niri/blob/b01b8afa8c8f9070300243050d9790e38fd19145/src/niri.rs#L2904) will work but still feels a bit hacky, so I removed it for now.

Fixes #589.


src/input/mod.rs
```diff
@@ -1833,6 +1841,13 @@ impl State {
             self.niri.pointer_hidden = false;
             self.niri.tablet_cursor_location = None;
 
+            if let Some(mapped_id) = self.niri.window_under_cursor().map(|mapped| mapped.id()) {
```
bbb651, line 1844: Calling `.window_under_cursor()` twice is not ideal, I ran into an annoying lifetime issue because `mapped` borrows from `self`, so I can't clear `self.pick_window` inside the if... now that I think about it, it *always* calls `.window_under_cursor()` twice not just when `self.pick_window` is active

bbb651: I guess this works?
```diff
diff --git a/src/input/mod.rs b/src/input/mod.rs
index 00292de2..6bba9f71 100644
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1841,11 +1841,12 @@ impl State {
             self.niri.pointer_hidden = false;
             self.niri.tablet_cursor_location = None;

-            if let Some(mapped_id) = self.niri.window_under_cursor().map(|mapped| mapped.id()) {
-                if let Some(tx) = self.niri.pick_window.take() {
+            if let Some(tx) = self.niri.pick_window.take() {
+                if let Some(mapped_id) = self.niri.window_under_cursor().map(|mapped| mapped.id()) {
                     let _ = tx.send_blocking(Some(mapped_id));
                     return;
                 }
+                self.niri.pick_window = Some(tx);
             }

             if let Some(mapped) = self.niri.window_under_cursor() {
```

niri-ipc/src/lib.rs
```diff
@@ -129,6 +131,10 @@ pub enum Response {
     FocusedOutput(Option<Output>),
     /// Information about the focused window.
     FocusedWindow(Option<Window>),
+    /// Information about the picked window.
+    ///
+    /// `None` if the request was cancelled by the user.
```
YaLTeR, line 136: I guess there's other ways to get None, like clicking on empty space or a layer surface or whatever. But I also guess this can be through of as cancelled?

What I'm getting at is maybe this clarification in the comment isn't needed altogether since it kinda makes it read like only explicit cancelling with Escape will return None?

src/cli.rs
```diff
@@ -72,6 +72,8 @@ pub enum Msg {
     FocusedOutput,
     /// Print information about the focused window.
     FocusedWindow,
+    /// Request picking a window.
```
YaLTeR, line 75:
```suggestion
    /// Pick a window with the mouse and print information about it.
```

src/input/mod.rs
```diff
@@ -361,6 +361,14 @@ impl State {
                     }
                 }
 
+                if pressed && raw == Some(Keysym::Escape) {
+                    this.niri.pick_window.retain(|tx| !tx.is_closed());
+                    if let Some(tx) = this.niri.pick_window.pop_front() {
+                        let _ = tx.send_blocking(None);
+                        return FilterResult::Intercept(None);
```
YaLTeR, line 368: Should this add Escape to `suppressed_keys`?

src/input/mod.rs
```diff
@@ -1833,6 +1841,15 @@ impl State {
             self.niri.pointer_hidden = false;
             self.niri.tablet_cursor_location = None;
 
+            self.niri.pick_window.retain(|tx| !tx.is_closed());
+            if !self.niri.pick_window.is_empty() {
+                if let Some(mapped_id) = self.niri.window_under_cursor().map(|mapped| mapped.id()) {
```
YaLTeR, line 1846: Could probably fetch `window_under_cursor()` above to reuse it also in the interactive move condition right below and not fetch twice

src/ipc/client.rs
```diff
@@ -252,6 +253,23 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+            if json {
+                let window = serde_json::to_string(&window).context("error formatting response")?;
+                println!("{window}");
+                return Ok(());
+            }
+
+            if let Some(window) = window {
+                print_window(&window);
+            } else {
+                println!("Picking window was cancelled.");
```
YaLTeR, line 270:
```suggestion
                println!("No window selected.");
```

src/niri.rs
```diff
@@ -332,6 +333,8 @@ pub struct Niri {
     pub hotkey_overlay: HotkeyOverlay,
     pub exit_confirm_dialog: Option<ExitConfirmDialog>,
 
+    pub pick_window: VecDeque<async_channel::Sender<Option<MappedId>>>,
```
YaLTeR, line 336: Since this is a modal operation, I think we should only have one in progress. Just cancel the existing one upon getting a new request.

Then also `error getting picked window info` will no longer be an error, but a regular cancellation.

YaLTeR (commented):
> Originally I tried to indicate picking was occurring by turning the cursor into a crosshair, in the same way as the screenshot UI [in the seat handler](https://github.com/YaLTeR/niri/blob/ae2ea53c9598a58de4a78d56f150915c3610157d/src/handlers/mod.rs#L95) but that means we're forgetting what the original cursor was, I think doing it [here](https://github.com/YaLTeR/niri/blob/b01b8afa8c8f9070300243050d9790e38fd19145/src/niri.rs#L2904) will work but still feels a bit hacky, so I removed it for now.

I would say that having a crosshair cursor is more important than remembering the original cursor. So if there's no easy way to restore the original cursor, just make it crosshair with a FIXME comment.

I think some visual indication of what window you're about to pick would also be good. But I guess that can be done separately.

> WIP: Currently the ipc server gets stuck waiting for niri to return the
picked window, and if the ipc client closes it doesn't realize that until
niri returns the picked window leading to "ghost" window picking.

Hmm, yeah, maybe some kind of `future::select()` can be used to also check for the client disconnecting?

bbb651:
> Hmm, yeah, maybe some kind of `future::select()` can be used to also check for the client disconnecting?

`futures_util` has one, the problem is that I don't know if it's possible to get a future of when the socket disconnects. I think constantly doing `is_close`/reads/peeks in a `future_util::poll_fn` might work, perhaps with some delay.

YaLTeR:
On the other hand, maybe it's not that big a deal? Picking is a quick operation either way

YaLTeR:
Added some minor fixes.

However, I'm now realizing that this should be a pointer grab. Because you don't want the mouse events to go to windows while you're picking. See how spatial movement grab works for example. It also shows how to change and restore the cursor without the condition in the cursor rendering code.

bbb651:
What should happen when there's already a grab when a request to pick a window is invoked? Should it instantly return `None`, or should it queue to up until the current grab is done?

YaLTeR:
I think it should cancel the ongoing grab. This is stuff like popup grabs, or ongoing DnD.

> I'm also not sure what to do about start data, it seems to not be optional and there's no sensible value it could be because it wasn't started by a pointer event..

Get the current location I guess.

bbb651:
I don't like that the cancellation logic is now very hidden in the ungrab handler and the code makes the assumption that `self.pick_window` is some iff there's an active pick window grab without checking it or encoding it in the type system..
Ideally `self.pick_window` would be something like `Option<(Sender<...>, PickWindowGrab)>` but smithay needs ownership of it, would wrapping it in a `Mutex<RefCell<_>>` work?

YaLTeR:
Did some minor fixes and some testing, seems to work fine now. Good to merge.

> I don't like that the cancellation logic is now very hidden in the ungrab handler and the code makes the assumption that self.pick_window is some iff there's an active pick window grab without checking it or encoding it in the type system..

Once we start using keyboard grabs too (I need to experiment with them since I haven't used them before), I think we can just make this a grab on both mouse and keyboard and then move the pick_window field from Niri to inside the grab start data.

YaLTeR:
I thought I was gonna add touch support real quick but TouchGrabStartData requires a touch slot, so I'll need to ask in the Smithay room what to do about it

PR #1045, opened 2025-01-25
Extend switch-layout action to accept layout index

ezemtsov:
_**Problem:**_
Cycling over keyboard layouts becomes uncomfortable with **three or more xkb layouts**.

_**Proposal:**_
Deterministic switching to a specific layout enables "muscle memory" and makes the experience of working with multiple languages much better overall.

_**Solution:**_
Extending `niri msg action switch-layout` to accept numbers for O(1) switch.
Later a user can assign this command to create a dedicated keyboard shortcut per layout.

P.S. @YaLTeR Thanks a lot for your effort, niri is great!

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -668,6 +668,8 @@ pub enum LayoutSwitchTarget {
     Next,
     /// The previous configured layout.
     Prev,
+    /// The specific layout by index.
+    Index(u8),
 }
 
 /// Output actions that niri can perform.
@@ -1174,7 +1176,10 @@ impl FromStr for LayoutSwitchTarget {
         match s {
             "next" => Ok(Self::Next),
             "prev" => Ok(Self::Prev),
-            _ => Err(r#"invalid layout action, can be "next" or "prev""#),
+            other => match other.parse() {
+                Ok(layout) => Ok(Self::Index(layout)),
+                _ => Err(r#"invalid layout action, can be "next", "prev" or a layout index"#),
+            },
         }
     }
 }
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -17,7 +17,7 @@ use smithay::backend::input::{
     TabletToolTipState, TouchEvent,
 };
 use smithay::backend::libinput::LibinputInputBackend;
-use smithay::input::keyboard::{keysyms, FilterResult, Keysym, ModifiersState};
+use smithay::input::keyboard::{keysyms, FilterResult, Keysym, Layout, ModifiersState};
 use smithay::input::pointer::{
     AxisFrame, ButtonEvent, CursorIcon, CursorImageStatus, Focus, GestureHoldBeginEvent,
     GestureHoldEndEvent, GesturePinchBeginEvent, GesturePinchEndEvent, GesturePinchUpdateEvent,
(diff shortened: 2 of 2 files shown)


niri-ipc/src/lib.rs
```diff
@@ -668,6 +668,8 @@ pub enum LayoutSwitchTarget {
     Next,
     /// The previous configured layout.
     Prev,
+    /// The specific layout by index.
+    ByIndex(u32),
```
YaLTeR, line 672:
```suggestion
    Index(u8),
```
Name consistent with WorkspaceReferenceArg, type consistent with KeyboardLayouts::current_idx

niri-ipc/src/lib.rs
```diff
@@ -1174,7 +1176,11 @@ impl FromStr for LayoutSwitchTarget {
         match s {
             "next" => Ok(Self::Next),
             "prev" => Ok(Self::Prev),
-            _ => Err(r#"invalid layout action, can be "next" or "prev""#),
+            other =>
+                match other.parse::<u32>() {
```
YaLTeR, line 1180:
```suggestion
                match other.parse() {
```
I think this should be able to infer itself?

src/input/mod.rs
```diff
@@ -691,6 +691,7 @@ impl State {
                 keyboard.with_xkb_state(self, |mut state| match action {
                     LayoutSwitchTarget::Next => state.cycle_next_layout(),
                     LayoutSwitchTarget::Prev => state.cycle_prev_layout(),
+                    LayoutSwitchTarget::ByIndex(layout) => state.set_layout(Layout(layout)),
```
YaLTeR, line 694: Needs to check if the index is valid

YaLTeR (changes requested):
Looks sensible, just some small changes

niri-ipc/src/lib.rs
ezemtsov: Done

niri-ipc/src/lib.rs
ezemtsov: Done

src/input/mod.rs
ezemtsov: Was thinking of that myself, but not sure what should be the correct behavior on failure?

**smithay** currently resolves this by cycling through existing indexes instead of failing.

For example if a user has 3 layouts configured (`us,no,ru`), running `niri msg action switch-layout 3` would just resolve to `us` instead of failing.

YaLTeR: I think just ignoring a request with out-of-bounds layout would be fine.

ezemtsov: done

src/input/mod.rs
```diff
@@ -691,6 +691,14 @@ impl State {
                 keyboard.with_xkb_state(self, |mut state| match action {
                     LayoutSwitchTarget::Next => state.cycle_next_layout(),
                     LayoutSwitchTarget::Prev => state.cycle_prev_layout(),
+                    LayoutSwitchTarget::Index(layout) => {
+                        let num_layouts = state.xkb().lock().unwrap().layouts().count();
+                        if usize::from(layout) > (num_layouts - 1) {
```
YaLTeR, line 696:
```suggestion
                        if usize::from(layout) >= num_layouts {
```

YaLTeR:
Thanks

PR #1052, opened 2025-01-26
adjust horizontal view movement gestures snap points for center-focused-column "on-overflow"

pranaless:
adjust snap points for view offset gestures to account for any centering induced by center-focused-column "on-overflow", making them consistent with other navigation methods.

in the implementation of `ScrollingSpace::view_offset_gesture_end`, unless center-focused-column is "always", every column has 2 snap points: one that snaps its left edge to the left side of the screen and one that snaps its right edge to the right side of the screen. note however that when center-focused-column is "on-overflow", snapping the left edge of the column to the left edge of the screen is invalid if together with the column to the right it overflows the screen, and similarly for snapping its right edge. for snap points where such overflow would happen, we instead adjust them to position the column in the center of the screen. so a column with overflow on both sides will always be centered and a column with overflow on only one side can be either centered or snapped to the opposite side.

### DIFF
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -2569,7 +2569,7 @@ impl<W: LayoutElement> ScrollingSpace<W> {
                                 && adj_col_w
                                     .filter(|adj_col_w| {
                                         center_on_overflow
-                                            && adj_col_w + gaps + col_w > working_area_width
+                                            && adj_col_w + 3.0 * gaps + col_w > working_area_width
                                     })
                                     .is_some()
                         };
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -2536,42 +2536,85 @@ impl<W: LayoutElement> ScrollingSpace<W> {
                 col_x += col_w + self.options.gaps;
             }
         } else {
+            let center_on_overflow = matches!(
+                self.options.center_focused_column,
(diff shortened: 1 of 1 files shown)


src/layout/scrolling.rs
```diff
@@ -2536,42 +2536,85 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
+                            col_x - left_strut
+                        } else {
+                            col_x - (self.working_area.size.w - col_w) / 2. - left_strut
+                        };
+                        let is_overflowing = |adj_col_w: Option<f64>| {
+                            center_on_overflow
+                                && adj_col_w
+                                    .filter(|adj_col_w| {
+                                        center_on_overflow
+                                            && adj_col_w + gaps + col_w > working_area_width
```
YaLTeR, line 2572: Shouldn't this include gaps 3 times? To the very left, between windows, to the very right.

YaLTeR (commented):
This is... a much more intuitive solution design-wise to this problem. Somehow I didn't think about this.

src/layout/scrolling.rs
pranaless: yep, ig we thought working_area_width already accounted for outer gaps. fixed

YaLTeR:
Thanks!

PR #1062, opened 2025-01-28
add option to set xkb config from file

m4rch3n1ng:
as far as i can tell, it is not possible to set a keymap from a string in the constructor, as [`KeyboardHandle::set_keymap_from_string`](https://smithay.github.io/smithay/smithay/input/keyboard/struct.KeyboardHandle.html#method.set_keymap_from_string) needs a `State`, and `add_keyboard` only takes an `XkbConfig`. this means that i have to initialize the KeyboardHandle with the XkbConfig, before immediately changing that after creating the State struct.

when the config is reloaded, it first looks if the xkb file config option is set and then tries to read the file, compile and set the keymap, falling back to the XkbConfig if any of that fails.

a config option like this also exists in other compositors like sway (see [sway-input](https://man.archlinux.org/man/sway-input.5#KEYBOARD_CONFIGURATION)) and hyprland, so i think it makes sense to add that as a config option to niri too.


src/niri.rs
```diff
@@ -590,6 +590,9 @@ impl State {
 
         // Initialize some IPC server state.
         state.ipc_keyboard_layouts_changed();
+        // Load the xkb_file config option,
+        // if set by the user.
+        state.load_xkb_file();
```
YaLTeR, line 595: Should be above `ipc_keyboard_layouts_changed()` I guess?

src/niri.rs
```diff
@@ -1083,6 +1086,42 @@ impl State {
...
+    fn set_xkb_file(&mut self, xkb_file: Option<String>) -> bool {
+        let Some(xkb_file) = xkb_file else {
+            return false;
+        };
+
+        let xkb_file = PathBuf::from(xkb_file);
+        let xkb_file = match expand_home(xkb_file.as_ref()) {
+            Ok(expanded) => expanded.unwrap_or(xkb_file),
+            Err(err) => {
+                warn!("failed to expand ~: {err:?}");
```
YaLTeR, line 1101: I think it might be nicer to return `anyhow::Result<()>` and use `.context()` for these messages, and then print the warning from the caller. Many other places in the code also do it this way.

If you pass `xkb_file: String` instead of `Option<String>` then you also get rid of the only case where you would want to return `Ok(false)`, so you can use `Result<()>` instead of `Result<bool>`.

wiki/Configuration:-Input.md
```diff
@@ -112,6 +113,18 @@ input {
...
+
+```kdl
+input {
+    keyboard {
+        xkb {
+            file "~/.config/keymap.xkb"
+        }
+    }
+}
+```
```
YaLTeR, line 126:
```suggestion
> [!TIP]
>
> <sup>Since: next release</sup>
>
> Alternatively, you can directly set a path to a .xkb file containing an xkb keymap.
> This overrides all other xkb settings.
> 
> ```kdl
> input {
>     keyboard {
>         xkb {
>             file "~/.config/keymap.xkb"
>         }
>     }
> }
> ```
```

resources/default-config.kdl
```diff
@@ -15,6 +15,9 @@ input {
             // For example:
             // layout "us,ru"
             // options "grp:win_space_toggle,compose:ralt,ctrl:nocaps"
+
+            // You can also specify a file from where a keymap will be loaded.
+            // file "~/.config/keymap.xkb"
```
YaLTeR, line 20:
```suggestion
```
I think we can leave it out of the default config, kind of a niche thing it feels.

YaLTeR (changes requested):
Makes sense, yeah.

src/niri.rs
m4rch3n1ng: i wanted to avoid doing duplicating the `keyboard.set_xkb_config` code because i had to nest twice, but i think i found a pretty good solution with setting a bool to `false` if it was able to successfully load the xkb file, and only doing `set_xkb_config` then.

src/niri.rs
m4rch3n1ng: no idea why i didn't thing to do that in the first place, good catch

YaLTeR:
Thanks!

PR #1068, opened 2025-01-29
winit: Set cursor instead of rendering it when possible

bbb651:
Before both cursors were visible.

The alternative would be to always hide the winit cursor, which is useful when testing changes for cursor rendering but a worse experience in normal use.


YaLTeR:
> Before both cursors were visible.

This is kind of intended. Makes it very easy to debug hotspot and other input issues. Like just recently when mouse input got inverted in #630.

Issue #1075, opened 2025-02-01
hide-when-typing causes first-person mouselook jumps in Wine in native Wayland mode


hallettj:
I've set up Wine 10.0 with native Wayland support, and it's great! Native Wayland mode seems to avoid issues I've been having running in Gamescope that come up in Niri under NixOS specifically. For example I've had the issue described in #1037 but with Wine games, and it has come and gone at different times. Recently I've also been seeing visual artifacts.

From my testing so far It looks like running in native Wayland avoids ~~both of those problems~~ the graphical artifacts. But I did run into another issue that turned out to be easy to work around:

When in a first-person game (Overwatch), pressing a key while simultaneously moving the mouse causes my character's crosshair to jump down a few dozen pixels or so. I also saw a little flickering of the cursor in game menus under the same conditions which reminded me that I had `hide-when-typing` enabled. Turning off that setting fixed the problem.

I've only seen this issue in Wine running in native Wayland mode. So it's entirely possible this is really a Wine bug. I wanted to get this information somewhere searchable because it took me a while to put two and two together.

### System Information

* niri version: niri stable v25.01 (commit e05bc269e678ecf828b96ae79c991c13b00b38a5)

* Distro: NixOS 24.11 with Wine 10.0 package from NixOS Unstable

* GPU: AMD Radeon RX 7800 XT

* CPU: AMD Ryzen 7 5800X

YaLTeR:
I wonder, does this happen in other compositors with `hide-when-typing` like functionality?

hallettj:
> I wonder, does this happen in other compositors with `hide-when-typing` like functionality?

Oh, good question! I'm too used to thinking of these slick features as Niri-exclusive. Yes, Sway has exactly the same problem when I use this setting:

```
seat "*" {
  hide_cursor when-typing enable
}
```

I'd totally understand if you want to close this bug and leave it up to Wine to fix their stuff.

I noticed that I spoke too soon about the lag issue. As in #1037 I'm still getting lag corresponding with mouse movement if I play a game long enough.

YaLTeR:
I think in this case this might be on wine to fix, yeah. So I'll close this, at least for now.

As for the lag issue, perhaps you could try an older calloop version? I think it should be drop-in to change it (with a `=0.14.0` version in Cargo.toml maybe).

hallettj:
I've tested downgrading calloop, and I'm replying in #1037

YaLTeR:
[...]

PR #1076, opened 2025-02-01
document interaction between hide-when-typing and wine wayland

hallettj:
There is a specific interaction between `hide-when-typing` and Wine running in native Wayland mode. In first person games pressing a key while simultaneously moving the mouse causes the character's crosshair to jump down a few dozen pixels or so until pretty quickly you are looking straight down at your character's feet. There is also slight cursor flickering in game menus.

I filed an issue, #1075. I thought it might be helpful to follow up with a documentation PR.

### DIFF
--- a/wiki/Configuration:-Miscellaneous.md
+++ b/wiki/Configuration:-Miscellaneous.md
@@ -119,6 +119,9 @@ cursor {
 
 If set, hides the cursor when pressing a key on the keyboard.
 
+This setting might interfere with games running in Wine in native Wayland mode that use mouselook, such as first-person games.
+If your character's point of view jumps down when you press a key and move the mouse simultaneously try disabling this setting.
+
 ```kdl
 cursor {
     hide-when-typing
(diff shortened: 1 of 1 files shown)


wiki/Configuration:-Miscellaneous.md
```diff
@@ -119,6 +119,9 @@ cursor {
 
 If set, hides the cursor when pressing a key on the keyboard.
 
+This setting might interfere with games running in Wine in native Wayland mode that use mouselook, such as first-person games.
+If your character's point of view jumps down when you press a key and move the mouse simultaneously try disabling this setting.
```
YaLTeR, line 123:
```suggestion
> [!NOTE]
> This setting might interfere with games running in Wine in native Wayland mode that use mouselook, such as first-person games.
> If your character's point of view jumps down when you press a key and move the mouse simultaneously, try disabling this setting.
```

YaLTeR:
Thanks

Issue #933, opened 2025-01-05
Tabbed mode for columns


YaLTeR:
Add a toggle that switches a column from a vertical stack of windows to a vertical set of tabs. See https://github.com/YaLTeR/niri/discussions/876 for design concepts. I think I prefer tab indicators on the left to tab indicators at the top, but that's an aside.

The cool detail of this idea is that it completely preserves keyboard navigation with no new necessary binds. You still switch up and down between windows in a column, still consume/expel, they are just rendered as tabs instead of as a stack.

YaLTeR:
I have a draft implementation ready at https://github.com/YaLTeR/niri/pull/1085, looking for feedback.

PR #1085, opened 2025-02-04
Tabbed columns

YaLTeR:
Tabbed display mode for columns. Implements https://github.com/YaLTeR/niri/issues/933 / https://github.com/YaLTeR/niri/discussions/876.

This is a WIP draft meant to gather feedback and figure out how exactly this should work. I know animations are broken, I'll deal with them later.

UX-wise, there's a single new bind:
```
Mod+W { toggle-column-tabbed-display; }
```
It toggles the focused column between normal and tabbed display mode. Everything else works the same way. Tabbed columns can be fullscreen with multiple windows, unlike normal columns.

The tab indicator is vertical and isn't included in the column size. I'm considering adding settings for both of those so you can have a more traditional horizontal tab bar at the top, maybe also with window titles as another setting.

By default, the tab indicator occupies 50% of the window height and has a gap. After going through several defaults here, this setup seems to be the only one that doesn't look broken for any kind of window: arbitrary shape, CSD with rounded corners, no-CSD with square or rounded corners. The requirement for the default is that it looks decent with everything.

You can override the active/inactive colors with layout settings and window rules. If you don't, then the tab indicator will try to sample colors and gradients from the window's border or focus ring, whichever is enabled:

I'm mainly looking for UX and design feedback here:
- Are there better defaults?
    - Keep in mind they should look decent with: CSD rounded corners, arbitrary window shape, SSD rect corners, SSD rounded corners.
- Do the settings make sense?
- What behavior do you want that is currently missing? (E.g. "include tab indicator in column size"; still looking for a clear name for this one.)
    - I got feedback that it should be possible to have a tab bar with readable window titles to tell apart several tabbed PDF files. My current idea to achieve this is adding settings for: "draw tab indicator at the top", "include it in column size", "draw titles in tab indicator".
    - That said, I'd like to avoid a million settings that implement half of CSS.


VaclavC:
I would like to have indicator on the top, with text titles, you have it covered by this proposal. And I would like it arranged from left, not centered, if possible.

Also, will it work with focus ring disabled?

YaLTeR:
> Also, will it work with focus ring disabled?

Yes. It only fetches default colors from these, otherwise it's a separate thing.

YaLTeR:
> And I would like it arranged from left, not centered, if possible.

Do you mean, like, left-aligned *and* not full-length? Because I imagine you'd just make it full-length (you already can with `length total-proportion=1.0`), at which point it will be the same as left-aligned since it matches the window size.

VaclavC:
I should read the proposal more thoroughly. You take proportion of available size and split it between tab indicators then. I had indicator sizes based on text content width in my mind. With that it would be nice to have it left aligned. But it should follow rounded corners setting then ... It is more complicated than it looks.

YaLTeR:
> I had indicator sizes based on text content width in my mind.

Usually one way or another, you want tabs to be equal length. (See tabs in browsers etc., they are all the same length.)

VaclavC:
Yes, that's true. Let forget that. But I would like to have it on 100% width then.

emenel:
The main design issue that jumps out at me is that there is no way to see what the tabbed windows are without switching to them. If I had a number of different applications stacked in a column, then switch to tabs, if I don't remember what they were, and in what order, it would be hard to use the tabs efficiently.

YaLTeR:
New stuff:
* layout `default-column-display "tabbed"`
* `hide-when-single-tab`
* `position "top"`/right/bottom
* gap can go negative

YaLTeR:
Pushed (to main but useful for tabs): focus-window-top/bottom/down-or-top/up-or-bottom and focus-window-in-column [index].

druskus20:
Is there anything like `focus-window-down-or-first` / `focus-window-up-or-last`? 

I believe there is:   `focus-window-down-or-column-left` /  `focus-window-down-or-column-right` which does not seem quite the same.

Thinking about cycling through tabbed columns

druskus20:
A big enough gap overflows over the next window. Maybe this is okay. 

image

YaLTeR:
It's intended, yes. This needs the aforementioned "include tab indicator in column size" setting

YaLTeR:
Updates:
- some animation fixes though not all
- interactive move into a tabbed column now works correctly
- `default-column-display` window rule

Supreeeme:
Is there supposed to be some kind of animation for a window moving into a tabbed column? I don't actually know what that should look like, but as it stands they appear to just pop out of existence into the column.

YaLTeR:
I haven't done anims yet

gkleen:
Tabs work great for my workflow – thanks a lot!

> The tab indicator […] isn't included in the column size. I'm considering adding [a] setting[…] for [that] […]

:+1: for that :)

I have very narrow gaps and borders and my current solution of having a semi-transparent indicator on top of the window contents works but doesn't really appeal to me – feels wrong to cover parts of the window and to have a tradeoff wrt to visibility of the indicator and usability of the window itself.

YaLTeR:
Alright, big change: added `place-within-column`, the aforementioned "include in column size" setting. This one affects quite a few places, so please test that it doesn't break.

As a bonus, added `gaps-between-tabs`.

> To get ahead on bikeshedding: how about reduce-column-width below layout » tab-indicator?

Sorry, missed this edit (edits don't send emails unlike new comments). I think `place-within-column` is more intuitive for the setting you might look for?

the-fuckin-nobody:
Just tried it and works great for me. Some additions I would love UI wise are:
1. ability to make the tab indicators rounded.
2. Something like this (a frame like thing for tab indicators maybe) would look cool but idk if  that's achievable or that's the goal or not?
image

YaLTeR:
> Something like this (a frame like thing for tab indicators maybe) would look cool but idk if that's achievable or that's the goal or not?

This might be a bit too much into the implementing half of css territory

the-fuckin-nobody:
> > Something like this (a frame like thing for tab indicators maybe) would look cool but idk if that's achievable or that's the goal or not?
> 
> This might be a bit too much into the implementing half of css territory

Understandable, thanks.

blepabyte:
Not strictly tabs-related, but it might fit some workflows to have newly spawned windows be placed into the same column/tab stack; that way, related windows naturally accumulate in larger tab stacks.

For example:
- Opening a new browser window via Ctrl+N from within Firefox (append below/after active window in same column/stack)
- Anything opened from an external launcher like wofi (open in new column to the right, as usual)

YaLTeR:
The problem with this is that you can't generally tell these two cases apart.

YaLTeR:
Did some work on the window animations around tabbed columns. Decided to go for a fade for this.

Please test and tell if anything looks broken or should be improved.

the-fuckin-nobody:
Working fine for me, but I have some doubts:
1. shouldn't the windows when tabbed , retain their original size (currently windows when tabbed, change their size to the window whose direction they are consumed)?
2. Also , wouldn't it be nice if the tab indicator renders after the window has completely rendered ( I have default-column-display set as "tabbed") ? 

If `2.` gets implemented, addition pop up animations for tabbed indicators might also be a good idea if thats not too much of a burden on you.

YaLTeR:
> shouldn't the windows when tabbed , retain their original size (currently windows when tabbed, change their size to the window whose direction they are consumed)?

When windows are consumed (or generally added) into a column, they change their size to the column's size. This had always been the case.

> Also , wouldn't it be nice if the tab indicator renders after the window has completely rendered ( I have default-column-display set as "tabbed") ?

I'll need to think of something here, yeah.

the-fuckin-nobody:
> When windows are consumed (or generally added) into a column, they change their size to the column's size. This had always been the case.

Yeah, I see.

Also, shouldn't windows retain their original size (width mainly, prior to being added to a column), if they are expelled from a column ??

YaLTeR:
I don't think so, no. It would be hidden state, and one that you can't easily access if you leave only one window in a column

the-fuckin-nobody:
One more thing before going to bed, I'd like to add is that, there should be a way to style the tabbed indicator for inactive windows too:
image

From the screenshot, it seems really confusing which tabbed indicators belongs to which window unless you look very carefully.

YaLTeR:
Yeah, would be good. But I'm not sure what to call it, and which colors to use by default.

the-fuckin-nobody:
I think, the whole approach of having tabbed indicators in left or right side of the windows is somewhat flawed as having tabbed indicators at the top or bottom would eliminate such confusions. Also, since the the scrolling layout is horizontal, it makes much sense to just have the indicator at top or bottom.


Nonetheless, here are my suggestions:
> But I'm not sure what to call it

Maybe `unfocused-color` and `unfocused-gradient`?

> which colors to use by default.

Maybe make the tab indicators for unfocused windows, a bit transparent (20 - 30%) ??

YaLTeR:
> I think, the whole approach of having tabbed indicators in left or right side of the windows is somewhat flawed as having tabbed indicators at the top or bottom would eliminate such confusions. Also, since the the scrolling layout is horizontal, it makes much sense to just have the indicator at top or bottom.

You can set `position "top"` or bottom. I think it makes sense to have it on the side by default, considering that the windows are logically vertically arranged (and the binds reflect that), and that the horizontal space is infinite so to say.

YaLTeR:
Pushed a fix to tab indicator gradient area matching the window, plus a fix to tab indicator sliding around vertically weirdly when toggling tabbed mode from a non-topmost tile.

the-fuckin-nobody:
> You can set `position "top"` or bottom. I think it makes sense to have it on the side by default, considering that the windows are logically vertically arranged (and the binds reflect that), and that the horizontal space is infinite so to say.

Yeah makes sense now, thanks.

the-fuckin-nobody:
> Pushed a fix to tab indicator gradient area matching the window, plus a fix to tab indicator sliding around vertically weirdly when toggling tabbed mode from a non-topmost tile.

Definitely looks better than before.

Hey, I was thinking maybe we can shrink the tabbed indicator's size of unfocused windows (idk maybe 50% smaller than its original size? )? So it looks distinguishable for unfocused windows.

YaLTeR:
Updates:
- Fixed #1092.
- Made tab indicators clickable to switch tab. Scrolling is postponed because it needs more refactors.
- Added dimming (50% transparency) for tab indicators on inactive columns.
- Added an animation for tab indicator appearing.

With this I'm about ready to merge this PR to get more testing on the main branch. There's no text titles, but they can come later. If there's something else simple I might add it, but I won't block on anything complex.

YaLTeR:
Thanks everyone for testing!

the-fuckin-nobody:
> Also it seems you forgot to activate your Linux

I'm too broke to afford linux lol.

PR #1089, opened 2025-02-05
Initialize tracing_subscriber earlier

notpeelz:
Otherwise the `warn!()` message in the `NOTIFY_SOCKET` block won't print.


YaLTeR:
Good catch. I'm sure I tested it when I added it and it worked though..

PR #1090, opened 2025-02-05
Configurable mod key

notpeelz:
Addresses #197 and (indirectly) #682

1. At first, I had two settings: `mod-key` and `nested-mod-key`. I tried preserving the existing code structure, but unfortunately it made the code overly complicated.
2. I dropped the `nested-mod-key` setting in 40df1356dae07c6689ad9c125f5cb3a8a518c7e5 as an attempt to simplify the logic.
3. In order to restore the "alternate mod key" behavior for the winit backend, I added an environment variable to override the mod key without altering the config file: c87287b436b3d9b8bf1baa69016f90e079f7dd3e

Question: I didn't know where to add the setting. Is `input { mod-key "super"; }` fine?

Feedback welcome :)

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -100,6 +100,10 @@ pub struct Input {
     pub focus_follows_mouse: Option<FocusFollowsMouse>,
     #[knuffel(child)]
     pub workspace_auto_back_and_forth: bool,
+    #[knuffel(child, unwrap(argument, str))]
+    pub mod_key: Option<ModKey>,
+    #[knuffel(child, unwrap(argument, str))]
+    pub mod_key_nested: Option<ModKey>,
 }
 
 #[derive(knuffel::Decode, Debug, PartialEq, Eq)]
@@ -367,6 +371,29 @@ pub struct FocusFollowsMouse {
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Percent(pub f64);
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum ModKey {
+    Ctrl,
+    Shift,
+    Alt,
+    Super,
+    IsoLevel3Shift,
+    IsoLevel5Shift,
+}
+
+impl ModKey {
+    pub fn to_modifiers(&self) -> Modifiers {
+        match self {
+            ModKey::Ctrl => Modifiers::CTRL,
+            ModKey::Shift => Modifiers::SHIFT,
+            ModKey::Alt => Modifiers::ALT,
+            ModKey::Super => Modifiers::SUPER,
+            ModKey::IsoLevel3Shift => Modifiers::ISO_LEVEL3_SHIFT,
+            ModKey::IsoLevel5Shift => Modifiers::ISO_LEVEL5_SHIFT,
+        }
+    }
+}
 
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Outputs(pub Vec<Output>);
@@ -3427,6 +3454,22 @@ where
     }
 }
 
+impl FromStr for ModKey {
+    type Err = miette::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match &*s.to_ascii_lowercase() {
+            "ctrl" | "control" => Ok(Self::Ctrl),
+            "shift" => Ok(Self::Shift),
+            "alt" => Ok(Self::Alt),
+            "super" | "win" => Ok(Self::Super),
+            "iso_level3_shift" | "mod5" => Ok(Self::IsoLevel3Shift),
+            "iso_level5_shift" | "mod3" => Ok(Self::IsoLevel5Shift),
+            _ => Err(miette!("invalid Mod key: {s}")),
+        }
+    }
+}
+
 impl FromStr for Key {
     type Err = miette::Error;
 
@@ -3664,6 +3707,9 @@ mod tests {
                 warp-mouse-to-focus
                 focus-follows-mouse
                 workspace-auto-back-and-forth
+
+                mod-key "Mod5"
+                mod-key-nested "Super"
             }
 
             output "eDP-1" {
@@ -3984,6 +4030,12 @@ mod tests {
                     },
                 ),
                 workspace_auto_back_and_forth: true,
+                mod_key: Some(
+                    IsoLevel3Shift,
+                ),
+                mod_key_nested: Some(
+                    Super,
+                ),
             },
             outputs: Outputs(
                 [
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 
 use calloop::timer::{TimeoutAction, Timer};
 use input::event::gesture::GestureEventCoordinates as _;
-use niri_config::{Action, Bind, Binds, Key, Modifiers, SwitchBinds, Trigger};
+use niri_config::{Action, Bind, Binds, Key, ModKey, Modifiers, SwitchBinds, Trigger};
 use niri_ipc::LayoutSwitchTarget;
 use smithay::backend::input::{
     AbsolutePositionEvent, Axis, AxisSource, ButtonState, Device, DeviceCapability, Event,
@@ -59,12 +59,6 @@ pub mod touch_resize_grab;
 use backend_ext::{NiriInputBackend as InputBackend, NiriInputDevice as _};
 
 pub const DOUBLE_CLICK_TIME: Duration = Duration::from_millis(400);
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum CompositorMod {
-    Super,
-    Alt,
-}
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct TabletData {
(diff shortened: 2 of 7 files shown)


niri-config/src/lib.rs
```diff
@@ -3022,13 +3086,28 @@ where
...
+                        continue;
+                    };
+
+                    let mut key = bind.key;
+                    if key.modifiers.contains(Modifiers::COMPOSITOR) {
+                        key.modifiers.remove(Modifiers::COMPOSITOR);
+                        key.modifiers.insert(mod_key.0);
+                    }
+
+                    if seen_keys.insert(key) {
```
YaLTeR, line 3110: Why all this new logic? It works fine now (even with mod key being different between host and nested), so why do this with configurable mod key?

src/main.rs
```diff
@@ -93,14 +94,28 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            Ok(x) => Some(x),
+            Err(err) => {
+                warn!("invalid value for NIRI_OVERRIDE_MOD: {err}");
+                None
+            }
+        })
+    };
+
+    let config_opts = ConfigOpts { mod_override };
+
```
YaLTeR, line 110: I don't think this is necessary at all? Just leave a `mod-key` setting. For nested mod key, you could have a `mod-key-nested` like you say you originally had, and default it to a simple condition: if host mod key is alt then use super nested mod, otherwise use alt nested mod.

niri-config/src/lib.rs
notpeelz: While the current setup works, there's an edge-case bug I'd like to address. For example, with a config like this:

```
input {
  mod-key "super"
}

binds {
  Mod+Z { spawn "kitty"; }
  Super+Z { show-hotkey-overlay; }
}
```

The parser can't detect conflicts between Mod+Z and Super+Z unless it knows what `Mod` maps to.

Here's the approach I propose:

1. Parse the config once (without checking for conflicting binds).
2. After parsing, validate the binds with the the `mod-key` it just extracted from the config.
3. If duplicates exist, re-parse with `ModKey` passed via the context to generate a `DecodeError` (leveraging knuffel's error formatting).

This way, we avoid custom error handling and ensure conflicts are reported accurately by the parser.
What do you think?

src/main.rs
notpeelz: > if host mod key is alt then use super nested mod, otherwise use alt nested mod.

I'm not sure I understand what you mean. Can you clarify?

From my perspective, exposing the "nested mod" as a setting doesn't seem necessary since the winit backend is primarily for development and testing.

Adding a `mod-key-nested` setting would require passing a `mod_key` field/parameter to several methods and structs, which I'd prefer to avoid. It also complicates the [`HotkeyOverlay` update logic](https://github.com/notpeelz/niri/commit/af1fdbe8d881cb80602af5f65d4ae1ccabca9b2c#diff-7d6e974fc702a349992658e868735716105361af2179217551bfe6d09be4ad9dR72-R75) (and any other structs that depend on the mod key), which I find a bit cumbersome.

In 40df1356dae07c6689ad9c125f5cb3a8a518c7e5, you can see how removing the backend dependency simplifies the logic. More importantly, this keeps `Config` as the single source of truth.

When hacking on niri, I usually run `NIRI_CONFIG=$PWD/config.kdl cargo run` and set `mod-key` to something that doesn't conflict with my primary compositor. For convenience, I added `NIRI_OVERRIDE_MOD` as an alternative.

niri-config/src/lib.rs
notpeelz: > and it shouldn't

Why not? In this scenario, `Mod+Z` and `Super+Z` clearly conflict, so I'd expect it to report an error.

> You don't want your config to suddenly break when you open nested for example.

You can use `NIRI_OVERRIDE_MOD` for developing/testing niri. Do you think that's good enough?

src/main.rs
YaLTeR: Running `niri` should *just work* for most common host compositor setups. No environment variables, no custom config. Most commonly, the host mod key will be Super, so nested niri should default to Alt. If you override your host niri key to Alt, then it likely makes sense to default to Super for nested niri.

You can leave it non-configurable, but it will still need to change to Alt on its own when running on nested. People who first-try niri will not set any environment variables or change any config.

> When hacking on niri, I usually run NIRI_CONFIG=$PWD/config.kdl cargo run and set mod-key to something that doesn't conflict with my primary compositor.

When hacking on niri I don't set any environment variables and don't use a custom config. It should not be required.

niri-config/src/lib.rs
YaLTeR: > You can use NIRI_OVERRIDE_MOD for developing/testing niri. Do you think that's good enough?

See my comment in the other thread on why I don't think that's a good solution.

src/main.rs
notpeelz: > Running niri should just work for most common host compositor setups. No environment variables, no custom config.

Gotcha, I'll keep that in mind for my next iteration of this PR.

> Most commonly, the host mod key will be Super, so nested niri should default to Alt. If you override your host niri key to Alt, then it likely makes sense to default to Super for nested niri.

My use case may be a bit uncommon. I have caps-lock mapped to Mod5/ISO_Level3_Shift, which I use as my `mod-key`. I guess it should map to Alt in this scenario?

> You can leave it non-configurable, but it will still need to change to Alt on its own when running on nested. People who first-try niri will not set any environment variables or change any config.

Got it.

niri-config/src/lib.rs
notpeelz: What about the conflicts question? My current implementation allows multiple modifiers in `mod-key`:

```
input {
  mod-key "ctrl+shift"
}

binds {
  Mod+Z { spawn "kitty"; }
  // this conflicts with the bind above
  Ctrl+Shift+Z { show-hotkey-overlay; }
}
```

Should I simplify the implementation and only allow a single modifier for `mod-key`?

YaLTeR: I think you should allow only single if that's fine. And I think I would leave conflict resolution out, at least for this PR.

notpeelz: Sounds good. I'll take another crack at it later.

notpeelz:
I rebased my changes and got rid of multi-modifier support for `ModKey`.

Questions:
- I don't remember why I renamed the type from `CompositorMod` to `ModKey`. What do you prefer?
- I changed the displayed names in hotkey_overlay for `ISO_LevelX_Shift` to `Mod5` and `Mod3`. `ISO_LevelX_Shift` is just way too long. Is that okay?

niri-config/src/lib.rs
```diff
@@ -3364,6 +3391,52 @@ where
     }
 }
 
+impl<S: knuffel::traits::ErrorSpan> knuffel::DecodeScalar<S> for ModKey {
```
YaLTeR, line 3394: Why do you need a custom DecodeScalar? knuffel can use FromStr on its own with `unwrap(argument, str)`

src/input/mod.rs
```diff
@@ -2913,6 +2891,7 @@ impl State {
         // We're using touch, hide the pointer.
         self.niri.pointer_hidden = true;
     }
+
```
YaLTeR, line 2894: unrelated change

src/niri.rs
```diff
@@ -591,13 +591,15 @@ impl State {
             Backend::Tty(tty)
         };
 
+        let mod_key = get_mod_key(&*config.borrow(), &backend);
```
YaLTeR, line 594: Doesn't seem like you need to pass this in? Both config and backend are passed in, so you can get it inside.

src/niri.rs
```diff
@@ -1227,14 +1229,17 @@ impl State {
...
+            || config.binds != old_config.binds
+        {
+            self.niri.mods_with_mouse_binds = mods_with_mouse_binds(new_mod_key, &config.binds);
+            self.niri.mods_with_wheel_binds = mods_with_wheel_binds(new_mod_key, &config.binds);
             self.niri.mods_with_finger_scroll_binds =
-                mods_with_finger_scroll_binds(self.backend.mod_key(), &config.binds);
+                mods_with_finger_scroll_binds(new_mod_key, &config.binds);
+            self.niri
+                .hotkey_overlay
+                .on_hotkey_config_updated(new_mod_key);
```
YaLTeR, line 1242: Probably don't move this statement to avoid unnecessary diff

src/niri.rs
```diff
@@ -1813,6 +1818,17 @@ impl State {
...
+pub fn get_mod_key(config: &Config, backend: &Backend) -> ModKey {
+    match backend {
+        Backend::Winit(_) => config
+            .input
+            .mod_key_nested
+            .or(config.input.mod_key)
+            .unwrap_or(ModKey::Alt),
+        Backend::Tty(_) | Backend::Headless(_) => config.input.mod_key.unwrap_or(ModKey::Super),
+    }
+}
```
YaLTeR, line 1830: I think this can remain a function on the Backend then?

YaLTeR (changes requested):
Looks much better now.

> I don't remember why I renamed the type from CompositorMod to ModKey. What do you prefer?

While I don't much like the extra diff, ModKey matches the new input setting and it also brought some multiple lines back to one, so let's leave it.

> I changed the displayed names in hotkey_overlay for ISO_LevelX_Shift to Mod5 and Mod3. ISO_LevelX_Shift is just way too long. Is that okay?

That's fine I guess.

notpeelz:
Should be good now. Let me know if there's anything else that needs changing.

YaLTeR (approved):
Did some cosmetic changes and wrote docs, good to go now.

YaLTeR:
Thanks!

PR #1094, opened 2025-02-06
add option to not move cursor when focussing a window via ipc

prog-r-amer:
Hi, I added a flag for not moving the cursor when focussing a window via ipc. Seems like I was the only one interested in that, but it was a personal pet peeve of mine.
https://github.com/YaLTeR/niri/discussions/935

Thanks for the amazing compositor.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -4188,9 +4188,9 @@ mod tests {
                             trigger: Trigger::Keysym(Keysym::_1),
                             modifiers: Modifiers::COMPOSITOR | Modifiers::SHIFT,
                         },
-                        action: Action::FocusWorkspace(WorkspaceReference::Name(
+                        action: Action::FocusWorkspace(FocusWorkspaceWithReference{reference: WorkspaceReference::Name(
                             "workspace-1".to_string(),
-                        ), false),
+                        ), no_mouse_warp: false}),
                         repeat: true,
                         cooldown: None,
                         allow_when_locked: false,
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1434,7 +1434,10 @@ pub enum Action {
     #[knuffel(skip)]
     FullscreenWindowById(u64),
     #[knuffel(skip)]
-    FocusWindow(u64),
+    FocusWindow {
+        id: u64,
+        no_mouse_warp: bool,
+    },
     FocusWindowInColumn(#[knuffel(argument)] u8),
     FocusWindowPrevious,
     FocusColumnLeft,
@@ -1485,9 +1488,9 @@ pub enum Action {
     CenterWindow,
     #[knuffel(skip)]
     CenterWindowById(u64),
-    FocusWorkspaceDown,
-    FocusWorkspaceUp,
-    FocusWorkspace(#[knuffel(argument)] WorkspaceReference),
+    FocusWorkspaceDown(FocusWorkspaceArgument),
+    FocusWorkspaceUp(FocusWorkspaceArgument),
+    FocusWorkspace(FocusWorkspaceWithReference),
     FocusWorkspacePrevious,
     MoveWindowToWorkspaceDown,
     MoveWindowToWorkspaceUp,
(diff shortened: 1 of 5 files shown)


niri-ipc/src/lib.rs
```diff
@@ -214,6 +214,9 @@ pub enum Action {
         /// Id of the window to focus.
         #[cfg_attr(feature = "clap", arg(long))]
         id: u64,
+        /// Dont move the cursor when focussing the window.
+        #[cfg_attr(feature = "clap", arg(long = "dont-move-cursor", action = clap::ArgAction::SetFalse))]
+        move_cursor: bool,
```
YaLTeR, line 219: I'm thinking to make it something like `no_mouse_warp: bool` (especially since it's "warp mouse to focus" in the config).

src/input/mod.rs
```diff
@@ -674,11 +674,15 @@ impl State {
                     self.niri.queue_redraw_all();
                 }
             }
-            Action::FocusWindow(id) => {
+            Action::FocusWindow(id, move_cursor) => {
```
YaLTeR, line 677: Do you also want to add it to FocusWorkspace, FocusWorkspaceDown, FocusWorkspaceUp? Same argument and works the same way. Others were asking for these

src/niri.rs
```diff
@@ -765,6 +765,13 @@ impl State {
         self.niri.queue_redraw_all();
     }
 
+    /// Focus a specific window without moving the cursor.
+    pub fn focus_window_without_moving_cursor(&mut self, window: &Window) {
```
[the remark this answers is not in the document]
prog-r-amer: sure, I just thought maybe there's more and more arguments coming and it's better to just duplicate since I didn't know what to do with the maybe warp cursor. I guess not moving cursor should override maybe warp to cursor.

niri-ipc/src/lib.rs
prog-r-amer: Sure, I just thought maybe the negation in the argument is confusing, since we then have a no_mouse_warp that is by default false.

niri-config/src/lib.rs
```diff
@@ -1322,7 +1322,7 @@ pub enum Action {
     #[knuffel(skip)]
     FullscreenWindowById(u64),
     #[knuffel(skip)]
-    FocusWindow(u64),
+    FocusWindow(u64, bool),
```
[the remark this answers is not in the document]
prog-r-amer: done

src/input/mod.rs
prog-r-amer: yes, but I had a problem. I didn't manage to do the focus workspace the same as the others cause it had the ```#[knuffel(argument)] WorkspaceReference``` in it.

prog-r-amer:
git didn't do what I wanted...

niri-config/src/lib.rs
```diff
@@ -1485,9 +1488,19 @@ pub enum Action {
     CenterWindow,
     #[knuffel(skip)]
     CenterWindowById(u64),
-    FocusWorkspaceDown,
-    FocusWorkspaceUp,
-    FocusWorkspace(#[knuffel(argument)] WorkspaceReference),
+    #[knuffel(skip)]
```
YaLTeR, line 1491: They shouldn't be skipped, otherwise you cannot put them in the config.

niri-ipc/src/lib.rs
```diff
@@ -214,6 +214,9 @@ pub enum Action {
         /// Id of the window to focus.
         #[cfg_attr(feature = "clap", arg(long))]
         id: u64,
+        /// Dont move the cursor when focussing the window.
+        #[cfg_attr(feature = "clap", arg(long = "no_mouse_warp", action = clap::ArgAction::SetTrue))]
```
YaLTeR, line 218: Why spell out the name in `long = "no_mouse_warp"`? Also, is `SetTrue` needed even for bools?

niri-ipc/src/lib.rs
```diff
@@ -214,6 +214,9 @@ pub enum Action {
         /// Id of the window to focus.
         #[cfg_attr(feature = "clap", arg(long))]
         id: u64,
+        /// Dont move the cursor when focussing the window.
```
YaLTeR, line 217:
```suggestion
        /// Disable moving the cursor to the newly focused window or output.
```
Same change in other places

prog-r-amer:
@YaLTeR sorry I am super stuck.
image

The reason I commented out the needed knuffel thing is because I wanted to change it to a named struct so there are no weird bools flying around. It seems I have to implement something for the knuffle derive to work but I don't have any clue how to do it.

Sorry I am new to the macro stuff. Maybe I am on a completely wrong track.

YaLTeR:
> It seems I have to implement something for the knuffle derive to work

No, you should be able to just skip the field itself (rather than the whole enum variant).

Issue #426, opened 2024-06-05
Restore window to its column upon unfullscreening


thepragmaticmero:
Entering fullscreen (Win+Shift+F) expels consumed window, I suppose that it's not intended behaviour. Valve pls fix.

1. Consumed window
![Screenshot2024-06-05 09-10-32](https://github.com/YaLTeR/niri/assets/169491353/032247a2-309c-424f-8450-b18c347de7ac)

2. Entering fullscreen window
![Screenshot2024-06-05 09-11-51](https://github.com/YaLTeR/niri/assets/169491353/a872ac08-73f7-4564-b830-8f96e0f87793)

3. Exiting fullscreen expels consumed window
![Screenshot2024-06-05 09-11-59](https://github.com/YaLTeR/niri/assets/169491353/062cab7d-4445-492c-9811-a58d17a46b01)

### System Information

* niri version: 0.1.6

* Distro: Fedora 40 Immutable (copr)

* GPU: Intel Graphics

* CPU: N2840 Intel

YaLTeR:
> Entering fullscreen (Win+Shift+F) expels consumed window, I suppose that it's not intended behaviour.

It is in fact intended. Try switching away from a fullscreen column: it's a regular column just like any other, and that's a feature.

The real issue is that it would be nice to restore the window back to where it was upon unfullscreening.

thepragmaticmero:
Oh.... yeah obviously Lol, for horizontal size keeping I would had [Win+F] not [Win+Shift+F] heh, funny you... but yes, I was talking about expelling it after exiting fullscreen... ye buddy. I like your hair.

gvolpe:
> The real issue is that it would be nice to restore the window back to where it was upon unfullscreening.

Following up on this, assuming this ticket can be treated as a feature request, I'd love to have a way to achieve this 🙏🏽 

Did anyone manage to do this programatically at least? On my use case, it would be as simple as running `niri msg action consume-or-expel-window-left` on the focused window upon exiting fullscreen, but I couldn't find any information about fullscreen on the IPC event stream, this is all I see:

```console
Workspace 1: active window changed to Some(80)
Window focus changed: Some(80)
Window layouts changed: [(80, WindowLayout { pos_in_scrolling_layout: Some((3, 1)), tile_size: (831.0, 664.0), window_size: (831, 664), tile_pos_in_workspace_view: None, window_offset_in_tile: (0.0, 0.0) })]
[...]

PR #1102, opened 2025-02-08
Keep layout on fullscreen

rehanzo:
Instead of the current approach of taking the tile out into its own column, this PR basically moves the bottom tile in a column to the top before fullscreening. When unfullscreening, it moves it back down to where it was.

Not sure if this would be the proper way of going about it. I just noticed that attempting fullscreening on tiles anywhere else but the top one would extend downwards, but not up.

This seems to work well, but also does seem a bit hacky. If there might be a more proper way, any pointers are appreciated.

Closes https://github.com/YaLTeR/niri/issues/426

### DIFF
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -156,6 +156,9 @@ pub struct Column<W: LayoutElement> {
     /// Index of the currently active tile.
     active_tile_idx: usize,
 
+    /// Index of the active tile before fullscreening.
+    pre_fs_tile_idx: usize,
+
     /// Desired width of this column.
     ///
     /// If the column is full-width or full-screened, this is the width that should be restored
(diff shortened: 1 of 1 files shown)


YaLTeR:
It is intended that fullscreen expels a window into its own column. I didn't want to change this. For tabbed columns in #1085 I made it so you can keep several windows in a column even with fullscreen, but for normal columns I kept the current behavior.

The way I want to try solving #426 is similar to the current view offset restoration on closing or unfullscreening. Additionally, it would remember if the fullscreen window needs to go back into the previous column into some spot.

Though, if that turns out to be too cumbersome to implement (mainly resetting this flag on changes to the layout), then I'll reconsider the approach from this PR (or rather from the tabs PR) where you keep the fullscreen window in the column.

rehanzo:
Ah I see, I missed your comment on the parent issue itself. Thanks for the info.

I'll close then.

PR #1088, opened 2025-02-05
input: add touchpad `drag-lock` setting

zzzsyyy:
add the `drag-lock` setting for touchpad

PR #1107, opened 2025-02-09
Allow disabling tap-and-drag

alexdavid:
Similar to https://github.com/YaLTeR/niri/pull/1088, this adds a new touchpad `drag` configuration option that configures tap-and-drag behavior.

Currently tap-and-drag is always enabled when the `tap` setting is enabled, but other compositors allow setting this separately.

### DIFF
--- a/wiki/Configuration:-Input.md
+++ b/wiki/Configuration:-Input.md
@@ -30,6 +30,7 @@ input {
         tap
         // dwt
         // dwtp
+        // drag false
         // drag-lock
         natural-scroll
         // accel-speed 0.2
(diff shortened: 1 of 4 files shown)


YaLTeR:
Hm, so does this mean that when this PR is merged, everyone will suddenly have tap-and-drag disabled? That's not good.

If tap-and-drag is on by default, then I guess it makes more sense to have an inverse `no-drag` flag? Or an explicit `drag false` where unset fetches and uses the default libinput value?

YaLTeR:
It needs to reset to default value if you comment it out while running niri.

alexdavid:
Ah right, I suppose we could go with

```rust
device.config_tap_set_drag_enabled(c.drag.unwrap_or(false))
```

But let me know if you prefer the `no-drag` or fetching the default value from libinput (is this default configurable somewhere in libinput?)

YaLTeR:
As far as I can tell from libinput code, tap-and-drag is always enabled by default. So I guess we could go with `no-drag`. Though, that reads a bit weird, so maybe it's better to go with `drag: Option<bool>`, at which point we might as well fetch the default from libinput when it's unset.

alexdavid:
That makes sense, I'll look into fetching the default from libinput this weekend. 🙂

wiki/Configuration:-Input.md
```diff
@@ -186,6 +187,7 @@ Settings specific to `touchpad`s:
 - `tap`: tap-to-click.
 - `dwt`: disable-when-typing.
 - `dwtp`: disable-when-trackpointing.
+- `drag`: can be `true` or `false`, controls if tap-to-drag is enabled.
```
YaLTeR, line 190:
```suggestion
- `drag`: can be `true` or `false`, controls if tap-and-drag is enabled.
```

YaLTeR:
Thanks

Issue #856, opened 2024-12-10
Add example to config describing how to power off monitors after locking


Hofer-Julian:
With the current default behavior, `Mod + Alt + L` opens `swaylock`, but doesn't power off monitors. The config should include a commented out example showing how to add the power-off after logging out.

BRS5672023:
Actually I think this might be the desired behavior.. One can add a rule in swayidle so that if swaylock is running, then after some time the screen is turned off. For example, I'm using the following swayidle config:
```
timeout 900 'pidof swaylock || swaylock -f'
timeout 960 'niri msg action power-off-monitors'
timeout 60 'pidof swaylock && niri msg action power-off-monitors'
before-sleep 'pidof swaylock || swaylock -f'
```

YaLTeR:
Oh, that's an interesting way to do it.

wzbfyb:
I configured it like this, I guess this could be added to the example config?

`    Super+Alt+L hotkey-overlay-title="Lock the Screen: swaylock" { spawn-sh "niri msg action power-off-monitors ; swaylock";}
`

tomasciccola:
Yeah, one can use niri's own ipc mechanism to turn monitors off. Buut, if niri supported `wlr-output-power-managment-v1`, one could do that in a compositor agnostic way

wzbfyb:
sure, but I do not quite get why you need to do that in a compositor agnostic way while using a specific compositor.
anyhow I think this example is good enough for the documentation and saves people from googling around.

tomasciccola:
because that way, you can have scripts that don't depend on a compositor. You can play/try other compositors and a lot of stuff will just keep working.
f.e. I have a script to focus an open window or launch an app (using fuzzel). Since I discover [wlrctl](https://git.sr.ht/~brocellous/wlrctl), I replace the part of the script for the window selection from using `niri msg` to using `wlrctl`. 
In the end its just preference, but I like not depending on compositors IPCs mechanisms

Issue #108, opened 2024-01-19
Turn off monitors after inactivity


YaLTeR:
Can be solved together with #104 because the Smithay PR linked there implements an idle timer perfect for this case.

ThatOneCalculator:
Now that it's implemented in smithay, can this be looked at again?

YaLTeR:
The reason this stalled is because right after that Smithay PR was merged, I implemented the idle protocol in niri, which means you can use tools like `swayidle` to do this.

This issue is still open because doing it in niri will allow a nicer fade out animation, but it also needs some config design. Where to put the setting and how to call it? What else do we need, like running a command when the screens power off (`swaylock` for instance)?

ThatOneCalculator:
My naïve thought would be


```
idle {
    timeout "300" // 300 seconds, could be "5m" instead 
    run-after { spawn "swaylockniri"; } // optional
}
```

YaLTeR:
What about the power off monitors part? :p Idle is a general term to run a command after idle, not necessarily turn off monitors.

ThatOneCalculator:
And that's why I shouldn't write GitHub comments right before going to bed 😆

ThatOneCalculator:
Although this could work...


```
idle {
    timeout 300
    monitor-off // this could be omitted for a non-monitor-off idle tast
    run-after { spawn "swaylockniri"; }
}

**Comment by YaLTeR** [comment]
I would somewhat prefer a spelling that emphasizes the turning off monitors part, since generic idle functionality can be done with swayidle

**Comment by ThatOneCalculator** [comment]
Maybe it could be a part of `output`?

```
output "DP-1" {
    // Other monitor settings
    poweroff-timeout 300 // 300 seconds
}
```

**Comment by YaLTeR** [comment]
Nah, they all turn off at the same time.

**Comment by zefr0x** [comment]
```kdl
inactivity {
timeout 900
monitor-off
}
inactivity {
timeout 7200
device-suspend
// or for move flexibility:
run-after { spawn "systemctl suspend"; }
}
```

We cloud also do:
```kdl
inactivity "900" {
monitor-off
}
inactivity "7200" {
device-suspend
// or for move flexibility:
run-after { spawn "systemctl suspend"; }
}
```

egrieco:
[...]

PR #1108, opened 2025-02-09
Implement the wlr-output-power-management protocol

agausmann:
Related to https://github.com/YaLTeR/niri/issues/108, https://github.com/YaLTeR/niri/issues/856

This protocol allows external programs (e.g. wlopm) to control and query
the output power state (on/off).

In the current implementation, when the client sends a
`set_mode` request, the handler calls `activate_monitors()` or
`deactivate_monitors()` to set that state for all monitors. This
is probably simpler than adding full support for per-output power
management, and is good enough for the most common use cases like idle
screensavers.

The `output_power.mode` event is now emitted by `activate_monitors()`
and `deactivate_monitors()` when the state changes. A bit of a rewrite
was needed because the tty backend tries to activate monitors on resume,
but previously it wasn't calling `activate_monitors()` because it is
unable to provide a reference to the `Backend` that wraps it. I worked
around this by adding a second method that does not take a backend
parameter, instead assuming that the backend will take care of calling
itself.


src/protocols/output_power_management.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    // todo: set output power per-monitor. Currently, niri is written to power
+    // all monitors at the same time.
+    fn get_output_power_mode(&mut self, output: &Output) -> Mode;
+    fn set_output_power_mode(&mut self, output: &Output, mode: Mode);
+}
+
+pub struct OutputPowerManagementManagerState {
+    // Active controls only. Failed ones are removed.
+    output_powers: HashMap<Output, ZwlrOutputPowerV1>,
```
agausmann, line 67: Something felt off about this, and I think I figured it out.

I based this module off of `gamma_control`. After reading the docs for it, I found out it offers _exclusive_ access to each output, so only one client can create a gamma control for an output. I'm pretty sure that's not true for output power; it should allow multiple clients to connect (operating more like the `output_management` module.

agausmann:
While the implementation may be simpler, the behavior of all-off / all-on might be confusing/unexpected for users. Even if it is well-documented, it's not what they would expect from a protocol that seems to provide per-output power management.
Because of this, we might want to skip this PR and go straight to per-output power management.

YaLTeR:
Sure, that makes sense to me

YaLTeR:
> A bit of a rewrite was needed because the tty backend tries to activate monitors on resume, but previously it wasn't calling activate_monitors() because it is unable to provide a reference to the Backend that wraps itself. I worked
around this by adding a second method that does not take a backend parameter, instead assuming that the backend will take care of calling itself.

Perhaps you don't need to change `activate_monitors()` at all, but instead add a call into `refresh_ipc_outputs()` that just checks the current `monitors_active` and if it doesn't match what OutputPowerManagementState thinks it is, sends the events? Other protocols also do it this way, see `output_management_state.notify_changes()` or `foreign_toplevel::refresh()`.

src/protocols/output_power_management.rs
```diff
@@ -0,0 +1,238 @@
...
+
+pub struct OutputPowerManagerGlobalData {
+    filter: Box<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
+}
+
+pub trait OutputPowerManagementHandler {
+    fn output_power_manager_state(&mut self) -> &mut OutputPowerManagementManagerState;
+
+    // todo: set output power per-monitor. Currently, niri is written to power
+    // all monitors at the same time.
```
YaLTeR, line 60: These callbacks on their own support it, so this comment is not needed here?

src/protocols/output_power_management.rs
```diff
@@ -0,0 +1,238 @@
...
+                if let Some(output) = Output::from_resource(&output) {
+                    // We borrow state in the middle.
+                    #[allow(clippy::map_entry)]
+                    if !state
+                        .output_power_manager_state()
+                        .output_powers
+                        .contains_key(&output)
+                    {
+                        let zwlr_output_power = data_init.init(id, OutputPowerState {});
+                        zwlr_output_power.mode(state.get_output_power_mode(&output).into());
```
YaLTeR, line 154: So yeah as I wrote in the other comment, the way this should work is `OutputPowerManagerState` should itself store the current power mode (to its knowledge) and send that straight to clients. Then there should be a `refresh()` call that tells it the current up-to-date power mode from niri. If it changes, it will send that update to clients.

src/protocols/output_power_management.rs
```diff
@@ -0,0 +1,238 @@
...
+                            .output_power_manager_state()
+                            .output_powers
+                            .insert(output, zwlr_output_power);
+                        return;
+                    }
+                }
+
+                // Output not found,
+                // or output power instance already exists
+                data_init.init(id, OutputPowerState {}).failed();
```
YaLTeR, line 165: I'd put this data_init at the top of the block so that you cannot miss it

src/protocols/output_power_management.rs
```diff
@@ -0,0 +1,238 @@
...
+    ) {
+        match request {
+            zwlr_output_power_v1::Request::SetMode { mode } => {
+                let output_powers = &mut state.output_power_manager_state().output_powers;
+                let Some((output, _)) = output_powers.iter().find(|(_, x)| *x == resource) else {
+                    return;
+                };
+                let output = output.clone();
+
+                // todo: handle invalid enum values
```
YaLTeR, line 196: Send a protocol error or something, see how other impls do it

src/protocols/output_power_management.rs
agausmann: Whoops, yeah

agausmann:
I've decided to retract this, this change is not important to me anymore, and I can't devote enough time to it. Right now I'm using this helper script when I need a command that works in all of my compositors:

```sh
[ -e "$NIRI_SOCKET" ] && niri msg action power-$onoff-monitors \
    || wlopm --$onoff '*'
```

If the socket file doesn't exist (or if it does for some reason and niri communication fails), fallback to wlopm.

Issue #816, opened 2024-11-19
` hide-after-inactive-ms` makes Waybar tootlips disappear


remche:
I'm a Niri and Waybar user.
When using the new `hide-after-inactive-ms` or `hide-when-typing` feature, the Waybar tooltip disappear with the cursor. I opened an [issue](https://github.com/Alexays/Waybar/issues/3779) on Waybar side but after our discusson on Matrix room, let's open one here too.

### System Information

* niri version: niri 0.1.10-1

* Distro: Arch Linux

YaLTeR:
Need to check what sway does. Probably want to keep the pointer focus intact until it changes, rather than immediately sending pointer leave.

tv42:
This seems to happen with many activate-on-hover things.

Timing (`hide-after-inactive-ms`) and keyboard (`hide-when-typing`) seem to behave the same.

Affected:

- Chrome and Firefox link hover
- VSCode/VSCodium tooltips

Not affected:

- Zed (using its own GUI toolkit: https://www.gpui.rs/)

tv42:
Quick glance says maybe the only thing Sway is doing is making the cursor invisible (empty image): https://github.com/swaywm/sway/pull/1979/files

YaLTeR:
Given what the problem is, I would expect it to happen on any client that shows a tooltip, so if Zed isn't affected I'd say this might be a bug in Zed.

Also the fix should be relatively straightforward, just I'm in the middle of another refactor rn (and also thought of some more edge cases). I'll get to it before the next niri release.

augustebaum:
Related PR in swayfx: https://github.com/WillPower3309/swayfx/pull/266

tv42:
Thank you! I can confirm Niri v25.05 hide pointer works great.

PR #1109, opened 2025-02-10
Add `keep-focus-when-hidden` option

rehanzo:
This option helps with situations where you want to do something like read a tooltip and have the inactive timer set. It can hide while you're reading, so you have to keep it moving.

From the comments, I understand this behaviour is expected so that tooltips and whatnot get removed when using touch input. So, I've created this option.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -761,6 +761,8 @@ pub struct Cursor {
     pub xcursor_size: u8,
     #[knuffel(child)]
     pub hide_when_typing: bool,
+    #[knuffel(child)]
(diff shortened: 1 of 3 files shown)


YaLTeR:
This is not how I wanted to fix it (see https://github.com/YaLTeR/niri/issues/816#issuecomment-2486009371). There should be no option; just when the cursor is hidden on a timeout specifically (and probably when it's hidden on key press too, but not on touch), it should remain "entered" on a surface until the next time contents_under change, at which point it should leave.

YaLTeR:
This has been fixed for several releases

PR #1122, opened 2025-02-11
feat: support for setting tablet calibration matrix; this allows for rotating tablet inputs

JohnDowson:


JohnDowson:
Ideally I'd want https://wayland.freedesktop.org/libinput/doc/latest/tablet-support.html#tablet-pressure-range
but it does not look like that functionality is exposed in libinput bindings at the moment.

JohnDowson:
Smithay issue.
https://github.com/Smithay/input.rs/issues/78

src/input/mod.rs
```diff
@@ -3449,6 +3449,19 @@ pub fn apply_libinput_settings(config: &niri_config::Input, device: &mut input::
         } else {
             input::SendEventsMode::ENABLED
         });
+
+        #[rustfmt::skip]
+        const MATRIX: [f32; 6] = [
+            1., 0., 0.,
+            0., 1., 0.,
+        ];
```
YaLTeR, line 3457: Instead of hardcoding the default matrix, you should get it from [`libinput_device_config_calibration_get_default_matrix()`](https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#ga3d9f1b9be10e804e170c4ea455bd1f1b) since it can be overridden in udev.

niri-config/src/lib.rs
```diff
@@ -3539,6 +3541,8 @@ mod tests {
 
                 tablet {
                     map-to-output "eDP-1"
+                    calibration-matrix 1.0 0.0 0.0 \
+                                       0.0 1.0 0.0
```
YaLTeR, line 3545: Do you mind changing this to something like 1 2 3 4 5 6 so we verify that we get the values in expected places.

YaLTeR (changes requested):

src/input/mod.rs
JohnDowson: I think the rust binding for this returns an option, how should we deal with that?

YaLTeR (approved):

YaLTeR:
Thanks

PR #1129, opened 2025-02-15
Slightly clearer wording in Tabs.md

w-jablonski:
When reading first words of the Tabs.md page, I was like: what are "display windows"? 
Replacing the word "display" with "present" makes it quicker to understand, as the word "display" is often a noun.

Also in the same line, the word "switch" could be replaced by "ask" to also arguably improve readability. 

Thanks

### DIFF
--- a/wiki/Tabs.md
+++ b/wiki/Tabs.md
@@ -2,7 +2,7 @@
 
 <sup>Since: next release</sup>
 
-You can switch a column to display windows as tabs, rather than as vertical tiles.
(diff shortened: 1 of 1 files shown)


YaLTeR:
Maybe "show" would be even better instead of present/display?

w-jablonski:
Maybe, hard to say. "Show" is more common, while also less official. "Present" is more "colorful" and literary thus appropriate for a written wiki. "Present" also reads more melodic imo. Ultimately it depends on the style you're choosing for your wiki, I tried to match the official style...

PR #1130, opened 2025-02-15
niri-ipc: implement MultiSocket wrapper

ein-shved:
Each time, when using `Socket::send` we have to reopen it. This patch adds `MultiSocket` class which wraps `Socket` and allows you to reuse single object with many `send` calls. This is handy when you are implementing complex scenario.

I am not sure, is this really have to be inside mainstream, but each time, I implements my new scenario, I have to create such [wrapper](https://github.com/ein-shved/niri-single-output/blob/master/src/lib.rs#L96) by myself

### DIFF
--- a/niri-ipc/src/socket.rs
+++ b/niri-ipc/src/socket.rs
@@ -4,7 +4,7 @@ use std::env;
 use std::io::{self, BufRead, BufReader, Write};
 use std::net::Shutdown;
 use std::os::unix::net::UnixStream;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use crate::{Event, Reply, Request};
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
Hm, yeah, also not sure this should be inside niri-ipc. I haven't quite decided how much niri-ipc is supposed to be used outside niri itself. You can use the types of course, but otherwise it's more or less "helpers that are used inside niri itself with little consideration for the outside world" currently.

titaniumtraveler:
> Hm, yeah, also not sure this should be inside niri-ipc. I haven't quite decided how much niri-ipc is supposed to be used outside niri itself. You can use the types of course, but otherwise it's more or less "helpers that are used inside niri itself with little consideration for the outside world" currently.

Would an alternative solution be to rework the way `niri-ipc` works?

My ideal workflow would be if I could just have *one* long-lived socket that I can get `EventStream` events from and still send requests and receive requests through.

I'm currently porting a workspace manager tool that I use on `hyprland` to `niri`.

Mostly because the `niri` IPC is **much** friendlier IMO and having a compositor that I actually would trust myself to read code from and edit/PR in is an incredible boon to me!

ein-shved:
> RPC asynchronous

Sounds very complex. Is it worth it?

YaLTeR:
> Would an alternative solution be to rework the way niri-ipc works?

Thing is, niri-ipc is not just some helper crate that happens to live in the niri repo. It is used internally in niri and so is structured primarily for the purposes of internal use. (This is also why it's versioned the way it is.) So I don't think reworking it inside the niri repo is a good idea (would be better to fork to another crate for any large reworks).

> My ideal workflow would be if I could just have one long-lived socket that I can get EventStream events from and still send requests and receive requests through.

It is intentionally one socket per call to avoid the problem of interleaving responses from different requests. Is it really that problematic to open a fresh socket to call something?

titaniumtraveler:
And again. I wouldn't want anyone else to have to work on this.
I'm pretty confident I can implement it myself and just wanted to ask how you would stand to this idea.

titaniumtraveler:
@YaLTeR I have implemented what I described in #1153.

It was actually surprisingly easy and was mostly just a refactor of how the client messages are handled.

One thing that is missing are tests to actually verify the IPC behavior around connecting, disconnecting and under what conditions it blocks:

My implementation is `future_util::future::select()`ing on receiving an `EventStream` event and writing it to the socket and receiving a `Request`, processing it and writing the `Response` to the socket. But it might be helpful to actually test that.

[`future_util::future::select()`]: https://docs.rs/futures-util/0.3.31/futures_util/future/fn.select.html

PR #1139, opened 2025-02-17
add kdl v2 support

zkat:
This adds v2 support with fallback to v1 for legacy configs. It might be a bit bumpy at first but I'm hoping it works ok. Probably worth treating as experimental for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1760,17 +1760,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "is-terminal"
-version = "0.4.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e19b23d53f35ce9f56aebc7d1bb4e6ac1e9c0db7ac85c8d1760c04379edced37"
-dependencies = [
- "hermit-abi 0.4.0",
- "libc",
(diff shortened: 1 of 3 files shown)


YaLTeR:
Thanks, looks quite reasonable. If you don't mind, I'll postpone looking at this to after the next release (planned for this weekend).

One thing that I wanted to look into for KDL 2 is to remove/change a few select things in the config by checking or editing the document before passing it on to knuffel. (So that the KDLv1 config remains exactly as it was, but the KDLv2 config can remove some deprecated stuff.) But that's for me to think about and poke around with.

zkat:
Yeah! There might still be stuff to fix on the kdl-rs side but I won't be able to properly debug stuff until I get back on my Linux machine later this week. The v2 parsing is definitely not quite passing tests yet and I don't know if it's because I just messed up porting the document over. I know the default config parses fine in v2 though.

Either way I just wanted to put this up to start the conversation and also give you a sense of the impact it might have

YaLTeR:
Wrote up a discussion about things I want to look into changing in KDLv2: https://github.com/YaLTeR/niri/discussions/1142

PR #1149, opened 2025-02-19
Niri-IPC: introduce WindowAreaChanged

aeghn:
Response to #624 .

I have observed that the author has recently been developing many fascinating features, such as floating windows, shadow effects, and tab indicators. I am deeply grateful for the author's diligent work in these areas. After personal experimentation, I found these features not only fascinating but also stable, without any performance degradation.

Compared to other features, this one is relatively minor, yet in the relevant discussions, I have not found anyone submitting a pull request here, so I decided to give it a try myself.

In general, I have introduced a new message type called `WindowAreaChanged`, which includes coordinates, dimensions, and whether it is a floating window.

When dealing with scrolling layouts, I found that it is not possible to use relative screen coordinates directly, as other window managers do. Therefore, I believe that floating windows and tiled windows should adopt two different coordinate systems(for `x`), hence the introduction of the `is_floating` configuration in `WindowArea`.

During the traversal of scrolling layouts, the system calculates the sum of xy coordinates, while floating windows directly use the method `tiles_with_offsets`.

The effect after using the `niri msg -j event-stream` command is as follows:

```
{"WindowAreaChanged":{"workspace_id":8,"window_id":65,"area":{"x":676,"y":-306,"w":947,"h":1022,"is_floating":true}}}
```

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -925,6 +925,22 @@ pub enum Transform {
     Flipped270,
 }
 
+/// Window Area
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct WindowArea {
+    /// The left top x of window
+    pub x: i32,
+    /// The left top y of window
+    pub y: i32,
+    /// The width of window
+    pub w: i32,
+    /// The height of window
+    pub h: i32,
+    /// Window floating state
+    pub is_floating: bool,
+}
+
 /// Toplevel window.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
@@ -1096,6 +1112,15 @@ pub enum Event {
         /// This configuration completely replaces the previous configuration. I.e. if any windows
         /// are missing from here, then they were closed.
         windows: Vec<Window>,
+    },
+    /// The window location has changed.
+    WindowAreaChanged {
+        /// Id of the workspace
+        workspace_id: Option<u64>,
+        /// Unique id of this window.
+        window_id: u64,
+        /// The new window area configuration.
+        area: Option<WindowArea>,
     },
     /// A new toplevel window was opened, or an existing toplevel window changed.
     WindowOpenedOrChanged {
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -9,7 +9,7 @@
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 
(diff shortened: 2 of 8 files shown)


YaLTeR:
Pinging @calops who was working on the same feature.

I'm not sure we wrote down the design anywhere, but my idea was to add new optional fields to `Window` for position within workspace (floating coordinates) and position within the scrolling layout (column index and index within column). This allows for adding more variants in the future (e.g. position within output for sticky windows) and allows setting multiple sets of coordinates (e.g. position within workspace for tiled windows too).

One important consideration is that positions and sizes should be reported for both the window (excluding borders) and for the tile (including borders).

aeghn:
I have some confusion, do you mean reporting the window position and other attributes simultaneously in the IPC? Or only making this change within the internal structure?
I have tried to put these coordinates into the existing window messages, but in this case, when we move the floating window, a large number of position messages will be reported, at this time I think other window attributes are redundant, so should we only output these attributes when using the command `niri msg windows`, and output messages independently in the IPC?

YaLTeR:
The events can stay. I'm saying that the data should be laid out a bit differently, and there should be more data.

ficcdaf:
> my idea was to add new optional fields to Window for position within workspace (floating coordinates) and position within the scrolling layout (column index and index within column).

This would be amazing. I've been trying to implement a "column indicator" waybar module that shows how many columns are in the active workspace and which one is currently focused -- and a feature like this is exactly what I'm looking for.

oliverpool:
Somehow related feature request: the ability to get the current scroll focus (relative to the workspace, not the viewport)
- left
- width
- right

This would enable creating a workspace "scrollbar" indicating the area of the workspace currently focused (like a website scrollbar).

I would only be interested in knowing the above information only for the scrolled window, however some other tools might want to know the width of each column (in which case I can compute the overall width - I would still need to get the current scroll position somehow).

ficcdaf:
That's an excellent idea!

YaLTeR:
FWIW there's a new PR: https://github.com/YaLTeR/niri/pull/1265

> the ability to get the current scroll focus (relative to the workspace, not the viewport)

Should probably just expose the scroll position instead

yrkv:
You're specifically referring to exposing scroll position for each workspace, right? (in `niri_ipc::Workspace`)

I imagine the changes for that are very similar to the window ones.

It might have an issue of spamming events for changing the view in some cases, if the scroll position is moved gradually via touchpad.

YaLTeR:
I think it could use the same logic as in Niri::refresh_pointer_contents() where it doesn't do it during transitions.

oliverpool:
https://github.com/YaLTeR/niri/pull/1265#issuecomment-2724621758 is exactly what I had in mind, so my usecase would be covered by that PR. Thanks!

aeghn:
> FWIW there's a new PR: https://github.com/YaLTeR/niri/pull/1265

I will close this.

PR #1153, opened 2025-02-20
Long living socket

titaniumtraveler:
Closes #1130

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1,8 +1,23 @@
 //! Types for communicating with niri via IPC.
 //!
-//! After connecting to the niri socket, you can send a single [`Request`] and receive a single
-//! [`Reply`], which is a `Result` wrapping a [`Response`]. If you requested an event stream, you
-//! can keep reading [`Event`]s from the socket after the response.
+//! After connecting to the niri socket, you can send [`Request`]s. Niri will process them one by
+//! one, in order, and to each request it will respond with a single [`Reply`], which is a `Result`
+//! wrapping a [`Response`].
+//!
+//! If you send a [`Request::EventStream`], niri will *stop* reading subsequent [`Request`]s, and
+//! will start continuously writing compositor [`Event`]s to the socket. If you'd like to read an
+//! event stream and write more requests at the same time, you need to use two IPC sockets.
+//!
+//! <div class="warning">
+//!
+//! Requests are *always* processed separately. Time passes between requests, even when sending
+//! multiple requests to the socket at once. For example, sending [`Request::Workspaces`] and
+//! [`Request::Windows`] together may not return consistent results (e.g. a window may open on a
+//! new workspace in-between the two responses). This goes for actions too: sending
+//! [`Action::FocusWindow`] and <code>[Action::CloseWindow] { id: None }</code> together may close
+//! the wrong window because a different window got focused in-between these requests.
+//!
+//! </div>
 //!
 //! You can use the [`socket::Socket`] helper if you're fine with blocking communication. However,
 //! it is a fairly simple helper, so if you need async, or if you're using a different language,
@@ -12,7 +27,9 @@
 //! 2. Connect to the socket and write a JSON-formatted [`Request`] on a single line. You can follow
 //!    up with a line break and a flush, or just flush and shutdown the write end of the socket.
 //! 3. Niri will respond with a single line JSON-formatted [`Reply`].
-//! 4. If you requested an event stream, niri will keep responding with JSON-formatted [`Event`]s,
+//! 4. You can keep writing [`Request`]s, each on a single line, and read [`Reply`]s, also each on a
+//!    separate line.
+//! 5. After you request an event stream, niri will keep responding with JSON-formatted [`Event`]s,
 //!    on a single line each.
 //!
 //! ## Backwards compatibility
--- a/niri-ipc/src/socket.rs
+++ b/niri-ipc/src/socket.rs
@@ -16,7 +16,7 @@ pub const SOCKET_PATH_ENV: &str = "NIRI_SOCKET";
 /// This struct is used to communicate with the niri IPC server. It handles the socket connection
 /// and serialization/deserialization of messages.
 pub struct Socket {
-    stream: UnixStream,
+    stream: BufReader<UnixStream>,
 }
 
 impl Socket {
@@ -37,6 +37,7 @@ impl Socket {
     /// Connects to the niri IPC socket at the given path.
     pub fn connect_to(path: impl AsRef<Path>) -> io::Result<Self> {
         let stream = UnixStream::connect(path.as_ref())?;
+        let stream = BufReader::new(stream);
         Ok(Self { stream })
     }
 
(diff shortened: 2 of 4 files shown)


YaLTeR:
Thanks, but as I mentioned in the issue, one socket per call is intentional. I don't think the benefits of multiple calls per socket connection (I'm still not sure what benefits there are exactly? Slightly simpler code in the clients?) outweigh the drawbacks of possible interleaving and the server-side code complexity. I'm looking at all of these futures and I cannot for example tell right away if it's possible to buffer input and output in just the right way to produce interleaving output bytes due to unfortunate IO scheduling.

titaniumtraveler:
> Thanks, but as I mentioned in the issue, one socket per call is intentional. I don't think the benefits of multiple calls per socket connection (I'm still not sure what benefits there are exactly? Slightly simpler code in the clients?) outweigh the drawbacks of possible interleaving and the server-side code complexity. I'm looking at all of these futures and I cannot for example tell right away if it's possible to buffer input and output in just the right way to produce interleaving output bytes due to unfortunate IO scheduling.

> I'm still not sure what benefits there are exactly? Slightly simpler code in the clients?

Simpler client code is definitely the main factor, but also opening and closing new socket connections all the time is *somewhat* costly. Not *very* costly, but somewhat costly.

Also opening a socket takes *time*. Not much, but if you need to send lets say 20 RPC calls to reorder some windows that would definitely add up.

> I'm looking at all of these futures and I cannot for example tell right away if it's possible to buffer input and output in just the right way to produce interleaving output bytes due to unfortunate IO scheduling.

No, it is impossible to interleave output messages.
In both the `EventStream` and the `Request` case, it *does* poll the input at the same time (the `EventStream` on the stream if it is there and the `Request` until there is a newline or an EOF), but when writing bytes, both cases use `write_json_reply()`, which uses `write_all().await`, which means they stop the task until the **complete** reply is written.

The only point of where futures interleave is the `futures_util::select()` function here

https://github.com/YaLTeR/niri/blob/9931a6263fab8afc417dcc31f0b9357a19a44b81/src/ipc/server.rs#L233-L237

and that would in the worst case read some bytes from the socket, `Poll::Pending`, receive and write an `EventStream` event, receive the rest of the line and send the `Reply`.

Interleaving could happen when the `select()` function would include a future to `write()` to the output buffer and allow partial writes, but that is not the case here.

YaLTeR:
> if you need to send lets say 20 RPC calls to reorder some windows

This will be better addressed by #1017

titaniumtraveler:
> > if you need to send lets say 20 RPC calls to reorder some windows
> 
> This will be better addressed by #1017

Yeah, but its only part of why I want it.

Sending bulk actions is *nice*, but still does not allow for one *ongoing* connection.

YaLTeR:
Yeah but that's about the only argument so far that could reasonably justify this code complexity?

titaniumtraveler:
> Yeah but that's about the only argument so far that could reasonably justify this code complexity?

I do think it does. It really makes creating clients **much** easier.
For example when writing something bigger with the `niri-ipc` it is pretty useful to open an socket with
```
socat STDIO $NIRISOCKET
```
to test something out. It is much less setup to poke into `niri`s state and to check whether things actually work as expected. And having a repl to explore different actions just feels nice.

Also I think while this PR definitely increases the complexity of *some* code it makes some of the code also *more* understandable.

For example every request is handled in `process_request()`. (Including `Request::EventStream`.)

I have a hard time finding a good single argument why an ongoing socket is better.
More like a bunch of nebulous benefits that the flexibility of it can bring that IMO compound.

YaLTeR:
> For example when writing something bigger with the `niri-ipc` it is pretty useful to open an socket with
> 
> ```
> socat STDIO $NIRISOCKET
> ```
>
> to test something out. It is much less setup to poke into niris state and to check whether things actually work as expected. And having a repl to explore different actions just feels nice.

You can use the shell: `while socat STDIO "$NIRI_SOCKET"; do true; done`

titaniumtraveler:
> > For example when writing something bigger with the `niri-ipc` it is pretty useful to open an socket with
> > ```
> > socat STDIO $NIRISOCKET
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to test something out. It is much less setup to poke into niris state and to check whether things actually work as expected. And having a repl to explore different actions just feels nice.
> 
> You can use the shell: `while socat STDIO "$NIRI_SOCKET"; do true; done`

Yeah, but it does not support `EventStream` ...

titaniumtraveler:
> I'd just open the event stream in a second terminal. It would be even better as the spammy events are separated from my repl.

Fair. **Still** having a single point of entry would be better for my purposes.

Asking the kernel for a new 4K buffer + socket connection on each action to be executed just feels incredibly wasteful.

Every other IPC like thing also has one long running socket rather than reopening the socket every time.

See [`mpd`](https://mpd.readthedocs.io/en/stable/protocol.html#protocol-overview), or [`tmux`](https://github.com/tmux/tmux/wiki/Control-Mode).

YaLTeR:
> Asking the kernel for a new 4K buffer + socket connection on each action to be executed just feels incredibly wasteful.

What kind of system are you running which can both run niri and where a socket connection "every now and then" (considering batch actions should soon become a thing) is "incredibly wasteful"?

> See [`mpd`](https://mpd.readthedocs.io/en/stable/protocol.html#protocol-overview), or [`tmux`](https://github.com/tmux/tmux/wiki/Control-Mode).

I believe the tmux socket is also supposed to be used over SSH, so it makes perfect sense for it to have a long-lived connection.

---

I'm willing to entertain a limited version of a long-living niri connection which does not multiplex with event stream. I think that should be much simpler to implement (something like adding a loop somewhere around handling requests, and no `select`?).

However, even with this approach I see one potential problem. It gives you the impression that multiple requests submitted at once will be handled atomically one after another. And *most of the time* it will probably be true. So people might start relying on this in their scripts. But as soon as you cross a message buffer size boundary, or the request needs an idle to handle, it suddenly stops being atomic.

This sounds like an unnecessary source of bugs to me, especially considering that people might try batching actions this way instead of the batched action request, and get not what they want. So yeah, I'm really not quite that even this limited form of a long-lived socket pulls its weight.

titaniumtraveler:
> > Asking the kernel for a new 4K buffer + socket connection on each action to be executed just feels incredibly wasteful.
> 
> What kind of system are you running which can both run niri and where a socket connection "every now and then" (considering batch actions should soon become a thing) is "incredibly wasteful"?

Admittedly that is *mostly* perfectionism on my part.
More programming aesthetics than an actual functional requirement.

> > See [`mpd`](https://mpd.readthedocs.io/en/stable/protocol.html#protocol-overview), or [`tmux`](https://github.com/tmux/tmux/wiki/Control-Mode).
> 
> I believe the tmux socket is also supposed to be used over SSH, so it makes perfect sense for it to have a long-lived connection.

Not *directly*, but you can definitely bridge the `tmux` socket over `ssh` with `socat`.

> I'm willing to entertain a limited version of a long-living niri connection which does not multiplex with event stream. I think that should be much simpler to implement (something like adding a loop somewhere around handling requests, and no `select`?).

Yeah, definitely much simpler. If you want, I could write that in ~20 minutes.
It is basically the same code as this PR, but instead of the `select()`, when given an `"EventStream"`, it creates a new task, gives that task the `writer` and event stream receiver, and returns from the outer task.

> However, even with this approach I see one potential problem. It gives you the impression that multiple requests submitted at once will be handled atomically one after another. And _most of the time_ it will probably be true. So people might start relying on this in their scripts. But as soon as you cross a message buffer size boundary, or the request needs an idle to handle, it suddenly stops being atomic.
> 
> This sounds like an unnecessary source of bugs to me, especially considering that people might try batching actions this way instead of the batched action request, and get not what they want. So yeah, I'm really not quite that even this limited form of a long-lived socket pulls its weight.

titaniumtraveler:
@YaLTeR I have trimmed the PR down to reply to the client requests line-by-line, until the user sends a `"EventStream"` at which point the writer is given to the `event_stream_handler` and the original client handler returns.

(Though I have moved the `event_stream_handler` to a inline `async`-block considering it is only 5 lines.)

YaLTeR:
> Buffer size shouldn't have an influence on that.
All messages are strictly processed in-order.

What I meant was: say you request Windows and Workspaces. If they fit within one buffer size, they will be read at once and processed one after another without any await yielding, which means that nothing has a chance to modify the IPC state in-between, making the results atomic. If they don't fit within one buffer size, then an await can yield between processing Windows and processing Workspaces, and something else can change the IPC state in the middle, resulting in non-atomic results.

Having to reconnect to a socket between requests makes it very obvious that the responses are never atomic. Though, this kind of problem also exists in the event stream, where it is an intentional design choice and is documented.

titaniumtraveler:
> > Buffer size shouldn't have an influence on that.
> > All messages are strictly processed in-order.
> 
> What I meant was: say you request Windows and Workspaces. If they fit within one buffer size, they will be read at once and processed one after another without any await yielding, which means that nothing has a chance to modify the IPC state in-between, making the results atomic. If they don't fit within one buffer size, then an await can yield between processing Windows and processing Workspaces, and something else can change the IPC state in the middle, resulting in non-atomic results.
> 
> Having to reconnect to a socket between requests makes it very obvious that the responses are never atomic. Though, this kind of problem also exists in the event stream, where it is an intentional design choice and is documented.

Ah, now I understand what you mean. (I guess I worked with too much `async` code that I **always** assume that you can be interrupted between two messages) 

I will update the `niri-ipc` documentation to reflect that.

I think when the action batching PR lands (#1017), it might also useful to explicitly link that, as way to ensure the atomic execution of a list of action.

niri-ipc/src/lib.rs
```diff
@@ -1,19 +1,35 @@
...
-//! You can use the [`socket::Socket`] helper if you're fine with blocking communication. However,
-//! it is a fairly simple helper, so if you need async, or if you're using a different language,
-//! you are encouraged to communicate with the socket manually.
+//! <div class="warning">
+//!
+//! Note that even if you send multiple [`Request`]s at once, the `niri` event loop might still
+//! interupt the execution of these messages between executing the current and the next [`Request`].
+//! That means that if you send multiple messages at once, the compositor state might change, which
+//! could cause unexpected results.
+//!
```
titaniumtraveler, line 13: When #1017 is implemented, something like

> If you **need** to ensure the atomicity of a list of [`Request`]s, use `<link-to-batching-action>`.

should be added here.

titaniumtraveler, line 13: When #1017 is implemented, something like

```suggestion
//!
//! If you **need** to ensure the atomicity of a list of [`Request`]s, use `<link-to-batching-action>`.
//!
```

should be added here.

YaLTeR:
I'm taking a look at this now.

Any reason why you changed the input buffer from `String` to `Vec<u8>`? I'll change back if no

YaLTeR:
In general I'll undo a bunch of leftover refactors from the more-general impl. To keep the diff small.

YaLTeR:
Alright, I finished with my changes. Also added multiple request support to `Socket`. Should be good to merge now. 
Want to give it another test?

YaLTeR:
Let's get it some testing on main. Thanks!

PR #1157, opened 2025-02-20
Implement focus-monitor, move-window-to-monitor and move-column-to-monitor

annikahannig:
I noticed that there was no option to bind moving between monitor. I am often switching between three monitors and being able to access those directly would neatly fit my existing muscle memory. :)

This patch implements:

 * focus-monitor
 * move-window-to-monitor
 * move-column-to-monitor

Thank you so much for building niri!

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -63,6 +63,8 @@ pub enum Request {
     FocusedOutput,
     /// Request information about the focused window.
     FocusedWindow,
+    /// Request picking a window and get its information.
+    PickWindow,
     /// Perform an action.
     Action(Action),
     /// Change output configuration temporarily.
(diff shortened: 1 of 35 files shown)


niri-ipc/src/lib.rs
```diff
@@ -457,6 +467,16 @@ pub enum Action {
     MoveWindowToMonitorPrevious {},
     /// Move the focused window to the next monitor.
     MoveWindowToMonitorNext {},
+    /// Move the focused window to a specific monitor.
+    #[cfg_attr(
+        feature = "clap",
+        clap(about = "Move the focused window to a specific monitor")
+    )]
+    MoveWindowToMonitor {
+        /// The target output name.
```
YaLTeR, line 476: Do you mind also adding an `--id` argument here like in other Window actions? (at least most ones)

YaLTeR (commented):
Looks reasonable. Let's merge this shortly after the next release.

niri-ipc/src/lib.rs
annikahannig: Looks doable! :)

annikahannig:
Hmm looks like my branch is now broken. I will start from scratch and issue a new PR.

PR #1165, opened 2025-02-24
WIP: Implement `hyprland-global-shortcuts-v1`

bbb651:
I made a simple client for testing: https://github.com/bbb651/global-shortcuts-test (and tested that it works correctly in hyprland)

I haven't tested it yet but it should work with the `GlobalShortcuts` portals with `~/.config/xdg-desktop-portal/portals.conf`:
```ini
[preferred]
default=gnome
org.freedesktop.impl.portal.GlobalShortcuts=hyprland
```

Unresolved questions:
- What happens with there are other implementations of the `GlobalShortcuts` portal? Wlroots stopped adding protocols, ext seems likely to happen at some point, I think they'll be pretty similar (they'll both be following the portal pretty closely) so don't think it'll be hard to support both of them (and maybe remove the hyprland one at some point if clients aren't using it?). Gnome seems more problematic assuming it'll be DBus based, I think there's still value in having supporting a wayland protocol even if we support the Gnome one for layer-shell use cases so I don't think this should be a blocker.
- How does it interact with binds? I'm currently made a `Action::TriggerGlobalShortcut` action that immediately presses and releases the shortcut, but this doesn't match the capabilities of the protocol (and the `GlobalShortcuts` portal). Once we have release binds we could have separate `Action::PressGlobalShortcut` & `Action::ReleaseGlobalShortcut` actions, which is the most flexible but is a bit verbose. I like the idea of "hold actions" that are continuous instead of discrete and are bound to both an entire bind press and release, this works for other continuous actions like resizing and translating the view, but I'm not sure about a design that plays nice with both multiple actions per bind and release binds.
- Representing the shortcuts as `HashMap<(String, String), HyprlandGlobalShortcut>` is a bit awkward in some places, the key reference becomes `&(String, String)` and in `niri msg global-shortcuts` I think it makes more sense to print them by app id. Hyprland seems to join them with `:` and store and accepts them like that, this doesn't work correctly for app ids containing `:` but they are technically invalid. Either way I still prefer separate arguments in the config and cli.   

Also currently ipc doesn't quite work (and I've just realized I've broken handling of duplicate shortcuts by creating the `HyprlandGlobalShortcutV1` object even when it errors later to try to workaround lifetimes..), the protocol implementation is a bit messy and clone strings a lot and [CI fails](https://github.com/bbb651/niri/actions/runs/13488048470/job/37681792570) (something is really wrong with my rust/cargo today, rust-analyzer repeatedly failed to run, it seemed to constantly be doing full rebuilds and it apparently didn't recompile ipc client...).

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1623,6 +1623,7 @@ pub enum Action {
     ToggleWindowRuleOpacity,
     #[knuffel(skip)]
     ToggleWindowRuleOpacityById(u64),
+    TriggerGlobalShortcut(#[knuffel(argument)] String, #[knuffel(argument)] String),
 }
 
 impl From<niri_ipc::Action> for Action {
@@ -1854,6 +1855,9 @@ impl From<niri_ipc::Action> for Action {
             niri_ipc::Action::ToggleWindowRuleOpacity { id: None } => Self::ToggleWindowRuleOpacity,
             niri_ipc::Action::ToggleWindowRuleOpacity { id: Some(id) } => {
                 Self::ToggleWindowRuleOpacityById(id)
+            }
+            niri_ipc::Action::TriggerGlobalShortcut { app_id, id } => {
+                Self::TriggerGlobalShortcut(app_id, id)
             }
         }
     }
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -59,6 +59,8 @@ pub enum Request {
     Layers,
     /// Request information about the configured keyboard layouts.
     KeyboardLayouts,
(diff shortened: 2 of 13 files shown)


YaLTeR:
Haven't looked at the code yet, but: why the Hyprland portal impl + protocol? It would make more sense to support the Mutter API to get it working through the GNOME portal, considering we already use it for screencast. Unless the Mutter API for global shortcuts is somehow much harder to implement in niri?

IVKLD:
So, are global shortcuts available in Niri or will they be added in the future?

Sempyos:
They aren't available currently,

PR #1168, opened 2025-02-24
Floating windows default proportional position

Mandarancio:
Hi, 
I added the possibility to have the default position of floating windows expressed as a proportion of the working space size.
This will allow to better match the windows that have size proportional to the working area in particular in setup with multiple outputs.

https://github.com/user-attachments/assets/0759a6e0-67a4-43c3-b119-d0bc925ef148

Thank you very much for this amazing project!

Cheers!

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1336,14 +1336,16 @@ pub struct TabIndicatorRule {
     pub inactive_gradient: Option<Gradient>,
 }
 
-#[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
+#[derive(knuffel::Decode, Debug, Default, Clone, Copy, PartialEq)]
 pub struct FloatingPosition {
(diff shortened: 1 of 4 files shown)


Mandarancio:
I am not sure, it is true that with new `bottom`, `top`, `left` and `right` anchor options you could have centered windows on the side of the workspace, however you could not specify custom position: e.g. 5% from the left.

In the example the window has proportional width of 80% and it is placed 10% from the top left corner, so it is centered, however it could be as well not centered.

I do not know if I am explaining my self well.

YaLTeR:
It is not the goal of niri to support every possible configuration, so until there's some convincing use case for precise proportional positioning, I would prefer to not add it (the extra complexity is not worth it). I'd be fine with adding top/bottom/left/right for relative-to though.

Mandarancio:
Agree, close in favor of #1169

PR #1169, opened 2025-02-24
Added top, left, bottom and right floating windows alignement

Mandarancio:
As discussed in #1168 instead of adding the relative position I added 4 new options for the alignment of floating windows to the side of the screen.

https://github.com/user-attachments/assets/601a7041-96a4-4b3f-b071-2c2c0cbc983e


YaLTeR (changes requested):

Mandarancio:
@YaLTeR : thank you for the review. Updated with your suggestions!

YaLTeR:
Thanks!

Issue #1181, opened 2025-02-27
Wiki: Wezterm Issues


lanastara:
It seems wezterm fixed the 2 issues mentioned in the wiki (even tho one is still technically open but I removed the windowrule workaround in my config and it seems to work fine).

If it is confirmed that those issues don't happen anymore the entry in the wiki should be updated.

YaLTeR:
Want to create a pull request?

lanastara:
Sorry I usually use gitlab so I didn't know the wiki here is also a repo. I'll make a pr ^^

I'd really like some other people to try though since we all know "Works on my machine"

PR #1182, opened 2025-02-27
wiki: remove wezterm issues

lanastara:
Related: #1181


lanastara:
It's in the nightly corp for fedora

YaLTeR:
The current wezterm flatpak still requires them, so I think it's a bit early to merge this.

lanastara:
Do you want me to update the wiki text to just mention that on current nightly it seems fixed or should I just keep an eye on the wezterm releases and notify you when the changes are in stable?

YaLTeR:
Could do both

YaLTeR:
Thanks

PR #1196, opened 2025-03-02
Implement focus-monitor, move-window-to-monitor and move-column-to-monitor

annikahannig:
New PR, since due to some unknown reason my old fork exploded in merge issues. :see_no_evil: 

----

I noticed that there was no option to bind moving between monitor. I am often switching between three monitors and being able to access those directly would neatly fit my existing muscle memory. :)

This patch implements:

    focus-monitor
    move-window-to-monitor
    move-column-to-monitor

Thank you so much for building niri!

----

This patch now also includes `move-window-to-monitor` by `id`.

Thanks again for your work and sorry for the inconvenience!

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1576,6 +1576,11 @@ pub enum Action {
     MoveWindowToMonitorPrevious,
     MoveWindowToMonitorNext,
     MoveWindowToMonitor(#[knuffel(argument)] String),
+    #[knuffel(skip)]
+    MoveWindowToMonitorById {
+        id: u64,
+        output: String,
+    },
     MoveColumnToMonitorLeft,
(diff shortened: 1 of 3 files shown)


YaLTeR:
Did some minor fixes. Thanks for cleanly separated commits!

Let's merge this. I have some minor behavioral questions left, like:
- If the target monitor is already focused, then the mouse probably shouldn't warp. I think some other actions will do this too in some cases, so I'll figure it out when I finally get to refactoring that code a bit.
- For move-window-to-monitor by id, if the window is already on the target monitor but on an unfocused workspace, it will be moved to the focused workspace. I'm not sure if it should do that. But I don't have a strong opinion, and it seems a bit involved to check, so let's leave it as is for now.

PR #1221, opened 2025-03-06
feat: support color picker functionality

nnyyxxxx:


YaLTeR:
Haven't looked in detail, but I think this should include the D-Bus call for xdg-desktop-portal-gnome too (as bbb detailed in the issue). Actually, do we even need the niri IPC version in this case?

nnyyxxxx:
> Haven't looked in detail, but I think this should include the D-Bus call for xdg-desktop-portal-gnome too (as bbb detailed in the issue). Actually, do we even need the niri IPC version in this case?

yeah im still working on it, uhm i'm not sure if we need the ipc version i was just following what you said

YaLTeR:
The simple reason that I am very busy and haven't gotten to this yet. I haven't forgotten about it, don't worry

src/dbus/gnome_shell_screenshot.rs
```diff
@@ -39,6 +43,9 @@ impl Screenshot {
             Ok(NiriToScreenshot::ScreenshotResult(None)) => {
                 return Err(fdo::Error::Failed("internal error".to_owned()));
             }
+            Ok(NiriToScreenshot::ColorResult(_)) => {
+                return Err(fdo::Error::Failed("unexpected color result".to_owned()));
+            }
```
YaLTeR, line 48: So this is actually problematic because it can happen if screenshot and pick color are started at once, then the wrong task is woken up. Pick color should use a different channel instead, perhaps it could simply use the same channel used by regular niri IPC?

niri-ipc/src/lib.rs
```diff
@@ -133,6 +135,8 @@ pub enum Response {
     FocusedWindow(Option<Window>),
     /// Information about the picked window.
     PickedWindow(Option<Window>),
+    /// Information about the picked color as RGBA bytes.
+    PickedColor(Option<[u8; 4]>),
```
YaLTeR, line 139: The GNOME D-Bus interface has the right idea here to use floats, because with 10-bit color depth a u8 is no longer enough. I would also make this an `Option<SomeNewStruct>` because when HDR happens we will need to signal some extra metadata like the color space, and we want to do that without breaking backwards compatibility.

src/input/pick_color_grab.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    fn pick_color_at_point(location: Point<f64, Logical>, data: &mut State) -> Option<[u8; 4]> {
+        let (output, pos_within_output) = data.niri.output_under(location)?;
+
+        data.backend
+            .with_primary_renderer(|renderer| {
+                let scale = Scale::from(output.current_scale().fractional_scale());
+                let physical_pos: Point<i32, Physical> =
+                    pos_within_output.to_physical_precise_round(scale);
```
YaLTeR, line 48: This might need to be a `floor` with a FIXME comment currently because I suspect that at the right/bottom edge of monitors it will return one past.

src/input/pick_color_grab.rs
```diff
@@ -0,0 +1,223 @@
...
+                    }),
+                ) {
+                    Ok(pixels) => pixels,
+                    Err(_) => return None,
+                };
+
+                if pixels.len() == 4 {
+                    Some([pixels[0], pixels[1], pixels[2], pixels[3]])
+                } else {
+                    None
```
YaLTeR, line 79: If this can never happen then print an `error!()`

src/input/pick_color_grab.rs
```diff
@@ -0,0 +1,223 @@
...
+};
+use smithay::input::SeatHandler;
+use smithay::utils::{Logical, Physical, Point, Scale};
+
+use crate::niri::State;
+use crate::render_helpers::render_to_vec;
+
+pub struct PickColorGrab {
+    start_data: PointerGrabStartData<State>,
+    temporary_location: Option<Point<f64, Logical>>,
```
YaLTeR, line 19: Why is this needed?

src/ipc/client.rs
```diff
@@ -270,6 +271,30 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+
+            if json {
+                let color = serde_json::to_string(&color).context("error formatting response")?;
+                println!("{color}");
+                return Ok(());
+            }
+
+            if let Some(color) = color {
+                println!(
+                    "Picked color: rgba({}, {}, {}, {})",
```
YaLTeR, line 287: I believe alpha in rgba() is a float between 0 and 1?

YaLTeR (changes requested):

YaLTeR (approved):
I removed alpha from the niri IPC color too (it cannot ever be non-1 currently anyway), and did some fixes and simplifications. Good to merge now.

YaLTeR:
Thanks!

YaLTeR:
Apparently the D-Bus method had a wrong return type so it didn't work... Fixed it now

PR #1226, opened 2025-03-08
Add Shell completions

titaniumtraveler:
Simple setup to generate shell completions using `clap_complete`.
That is probably not the best way to *distribute* the shell completions, but it
is at good enough for development purposes.

There is probably a smart way to generate and package the completions as part of
the release workflow, but I haven't looked into what the best way to do that is
yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ bitflags.workspace = true
 bytemuck = { version = "1.22.0", features = ["derive"] }
 calloop = { version = "0.14.2", features = ["executor", "futures-io"] }
 clap = { workspace = true, features = ["string"] }
+clap_complete = "4.5.45"
 directories = "6.0.0"
 drm-ffi = "0.9.0"
 fastrand = "2.3.0"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,6 +2,7 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 
 use clap::{Parser, Subcommand};
+use clap_complete::Shell;
 use niri_ipc::{Action, OutputAction};
 
 use crate::utils::version;
@@ -54,6 +55,8 @@ pub enum Sub {
     },
     /// Cause a panic to check if the backtraces are good.
     Panic,
+    /// Generate completions for [`Shell`]
+    Completions { shell: Shell },
 }
 
 #[derive(Subcommand)]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -55,7 +55,7 @@ pub enum Sub {
     },
     /// Cause a panic to check if the backtraces are good.
     Panic,
-    /// Generate completions for [`Shell`]
+    /// Generate shell completions.
     Completions { shell: Shell },
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -630,6 +630,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -91,6 +91,7 @@ wayland-backend = "0.3.8"
 wayland-scanner = "0.31.6"
 xcursor = "0.3.8"
 zbus = { version = "5.5.0", optional = true }
+clap_complete = "4.5.45"
```
YaLTeR, line 94: Insert into a sorted position pls

src/cli.rs
```diff
@@ -54,6 +55,8 @@ pub enum Sub {
     },
     /// Cause a panic to check if the backtraces are good.
     Panic,
+    /// Generate completions for [`Shell`]
```
YaLTeR, line 58:
```suggestion
    /// Generate shell completions.
```

src/cli.rs
```diff
@@ -54,6 +55,8 @@ pub enum Sub {
...
+    Completions { shell: Shell },
```
YaLTeR, line 59: Is this how the subcommand is usually called in other tools? Can you give some examples? What about dynamic completions (in the future)?

Cargo.toml
titaniumtraveler: Hmm, would have thought that `cargo add` inserts dependencies sorted if possible ...

Fixed.

src/cli.rs
titaniumtraveler: That commit should probably then just squashed together.

Cargo.toml
YaLTeR: It does; we just have some sorting mistake in there somewhere I think

src/cli.rs
titaniumtraveler: > What about dynamic completions (in the future)?

I haven't looked too deep into dynamic completions, but looking at the docs it seems to be basically come down to having a subcommand that forwards `Vec<OsString>` to [`clap_complete::engine::complete()`].
When that gets stabilized it wouldn't be *too* hard to add support for it without breaking the subcommand.

> Is this how the subcommand is usually called in other tools? Can you give some examples?

It is at least how `rustup` is doing it *shrug*
image

But in general it is also how *I* usually add completions to the tooling I write.
It makes developing CLIs pretty comfortable, especially since I have a shell function to automatically create an alias called `niri` (or whatever the binary target of a crate is called) and then automatically recompile the crate and re-source the completions when I run it.

I also recently created a `cargo` PR so that custom cargo subcommands can get completions as well. (At least in bash. `zsh` apparently already had support for it!)

[`clap_complete::engine::complete()`]: https://docs.rs/clap_complete/latest/clap_complete/engine/fn.complete.html

YaLTeR:
Thanks.

I noticed some small issues:
- Actions don't complete options:

  image
- Actions keep completing multiple times:

  image
- `niri msg output` completes filenames in addition to commands:

  image

I wonder if those are fixable without too much effort.

titaniumtraveler:
> Thanks.
> 
> I noticed some small issues:
> 
>     * Actions don't complete options:
>       image
> 
>     * Actions keep completing multiple times:
>       image
> 
>     * `niri msg output` completes filenames in addition to commands:
>       image
> 
> 
> I wonder if those are fixable without too much effort.

I will look into it.
Though it might also depend on the specific shell you use.
(Hence the want of the `clap` team to move to rust native completions)

PR #1230, opened 2025-03-08
Added missing "Since: 25.02" to clipboard config documentation

LunarEclipse363:
Noticed that this is in the 25.02 changelog but the wiki doesn't mention when it was added.


YaLTeR:
Thanks. Missed it

PR #1238, opened 2025-03-09
Allow negative shadow spread + documentation improvements

LunarEclipse363:
The documentation part is things I ran into when reading that part of the wiki for the first time.

As for the shadow spread, this is something that you can do with CSS and I was missing for mako (the shadows seem too big for its notifications).

I hope that I did this correctly, I spent a while figuring out what can and can't be a negative number.
I'm not familiar with the codebase so I'm not sure if there are any other requirements/assumptions I may have missed.

Here is a snippet of the updated code with some extra comments showing my thought process:
```rust
let spread = ceil(self.config.spread.0);
// I'm assuming Point can be negative
let offset = offset - Point::from((spread, spread));

let win_radius = radius.fit_to(win_size.w as f32, win_size.h as f32);

// Size must be positive according to smithay documentation
let box_size = if spread >= 0. {
    win_size + Size::from((spread, spread)).upscale(2.)
} else {
    win_size - Size::from((-spread, -spread)).upscale(2.)
};
// I added a check to `CornerRadius::expanded_by` to ensure corner radii won't become negative with a negative argument
let radius = win_radius.expanded_by(spread as f32);

let shader_size = box_size + Size::from((width, width)).upscale(2.);
```

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -653,7 +653,7 @@ pub struct Shadow {
     #[knuffel(child, unwrap(argument), default = Self::default().softness)]
     pub softness: FloatOrInt<0, 1024>,
     #[knuffel(child, unwrap(argument), default = Self::default().spread)]
-    pub spread: FloatOrInt<0, 1024>,
+    pub spread: FloatOrInt<-1024, 1024>,
     #[knuffel(child, unwrap(argument), default = Self::default().draw_behind_window)]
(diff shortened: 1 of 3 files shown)


YaLTeR:
Thanks

YaLTeR:
Also please give it a try again to make sure that it works (I didn't change any logic so it should not have broken but still)

PR #1244, opened 2025-03-12
feat(cli): add `niri validate --json` to print report as json

titaniumtraveler:
This can be used to more or less trivially integrate the errors that `niri
validate` returns into your editor.

For example I wrote script that converts the json output into neovim
diagnostics on-save. (I haven't committed + pushed it yet, but when I do I will make
sure to add the link here)

image

One thing to note is that this PR also moves the `tracing` output from `stdout`
to `stderr`, (I **really** don't understand why that's not the default of `tracing_subscriber::fmt()`?!),
which *technically* might be a breaking change if there are any tools out there
trying to parse those logs.

But I think doing so is **definitely** the right thing to do *regardless* of
this PR, considering that `niri msg --json` prints its output to `stdout` as
well, which would cause the logging and the json output to interleave if both
are present. (AFAIK there is no instance of that *currently* happening that I
know of, but I could imagine some code adding a log statement somewhere and
with that causing the logging to the json output to interleave.

### DIFF
--- a/niri-ipc/src/diagnostic.rs
+++ b/niri-ipc/src/diagnostic.rs
@@ -1,0 +1,76 @@
+//! Config parsing diagnostics.
+//!
+//! The types in this module are based on `Report` and `Diagnostic` from [miette], which is what
+//! niri's config parsing uses.
+//!
+//! [miette]: https://crates.io/crates/miette
+
+use serde::Serialize;
+
+// Doc-comments in this file are losely based on ones from miette, licensed under Apache-2.0.
+//
+// https://github.com/zkat/miette
+
+/// Diagnostic from parsing a file.
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct Diagnostic {
+    /// Printable message.
+    pub message: String,
+    /// Diagnostic severity.
+    pub severity: Severity,
+    /// URL to visit for a more detailed explanation.
+    pub url: Option<String>,
+    /// Additional help about this diagnostic.
+    pub help: Option<String>,
+    /// Name of the file where the diagnostic occurred.
+    pub filename: String,
+    /// Labels to apply to this diagnostic's file.
+    pub labels: Vec<Label>,
+    /// Additional related diagnostics.
+    pub related: Vec<Diagnostic>,
+}
+
+/// Diagnostic severity.
+#[derive(Serialize)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub enum Severity {
+    /// Just some help.
+    Advice,
+    /// Warning to take note of.
+    Warning,
+    /// Critical failure.
+    Error,
+}
+
+/// Labeled [`Span`].
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct Label {
+    /// Label string.
+    pub label: String,
+    /// Span the label applies to.
+    pub span: Span,
+}
+
+/// Span within a source file.
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct Span {
+    /// The 0-based starting byte offset.
+    pub offset: usize,
+    /// Number of bytes this span spans.
+    pub length: usize,
+    pub start: Option<LineSpan>,
+    pub end: Option<LineSpan>,
+}
+
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct LineSpan {
+    /// The 0-indexed line into the file.
+    pub line: usize,
+    /// The 0-indexed column into the file, relative to `line`.
+    pub col: usize,
+}
--- a/niri-ipc/src/diagnostic.rs
+++ b/niri-ipc/src/diagnostic.rs
@@ -63,9 +63,9 @@ pub struct Span {
     /// Number of bytes this span spans.
     pub length: usize,
     /// Starting line position of this span
-    pub start: Option<LinePosition>,
+    pub start: LinePosition,
     /// Ending line position of this span
-    pub end: Option<LinePosition>,
+    pub end: LinePosition,
 }
 
 /// Position in a document in terms of line number + character offset.
--- a/niri-ipc/src/diagnostic.rs
+++ b/niri-ipc/src/diagnostic.rs
@@ -62,15 +62,21 @@ pub struct Span {
     pub offset: usize,
     /// Number of bytes this span spans.
     pub length: usize,
-    pub start: Option<LineSpan>,
-    pub end: Option<LineSpan>,
+    /// Starting line position of this span
+    pub start: Option<LinePosition>,
+    /// Ending line position of this span
+    pub end: Option<LinePosition>,
 }
 
+/// Position in a document in terms of line number + character offset.
+///
+/// Note that this counts line numbers as declared by the KDL specification:
+/// <https://kdl.dev/spec/#section-3.18>
 #[derive(Serialize)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
-pub struct LineSpan {
-    /// The 0-indexed line into the file.
+pub struct LinePosition {
+    /// The 0-indexed line position in the file.
     pub line: usize,
-    /// The 0-indexed column into the file, relative to `line`.
-    pub col: usize,
+    /// The 0-indexed character offset into the file, relative to `line`.
+    pub character: usize,
 }
--- a/niri-ipc/src/diagnostic.rs
+++ b/niri-ipc/src/diagnostic.rs
@@ -24,7 +24,7 @@ pub struct Diagnostic {
     /// Additional help about this diagnostic.
     pub help: Option<String>,
     /// Name of the file where the diagnostic occurred.
-    pub filename: String,
+    pub filename: Option<String>,
     /// Labels to apply to this diagnostic's file.
     pub labels: Vec<Label>,
     /// Additional related diagnostics.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -52,6 +52,9 @@ pub enum Sub {
         /// command line argument takes precedence.
         #[arg(short, long)]
         config: Option<PathBuf>,
+        /// Format config errors as JSON to stdout.
+        #[arg(short, long)]
+        json: bool,
     },
     /// Cause a panic to check if the backtraces are good.
     Panic,
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,7 +3,7 @@ extern crate tracing;
 
 use std::fmt::Write as _;
 use std::fs::{self, File};
-use std::io::{self, Write};
+use std::io::{self, BufWriter, Write};
 use std::os::fd::FromRawFd;
 use std::path::{Path, PathBuf};
 use std::process::Command;
@@ -22,7 +22,7 @@ use niri::utils::spawning::{
 };
 use niri::utils::watcher::Watcher;
 use niri::utils::{cause_panic, version, IS_SYSTEMD_SERVICE};
-use niri_config::Config;
+use niri_config::{ipc_diagnostic, Config};
 use niri_ipc::socket::SOCKET_PATH_ENV;
 use portable_atomic::Ordering;
 use sd_notify::NotifyState;
@@ -94,13 +94,24 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Handle subcommands.
     if let Some(subcommand) = cli.subcommand {
         match subcommand {
-            Sub::Validate { config } => {
+            Sub::Validate { config, json } => {
                 tracy_client::Client::start();
 
                 let (path, _, _) = config_path(config);
-                Config::load(&path)?;
-                info!("config is valid");
-                return Ok(());
+                let Err(report) = Config::load(&path) else {
+                    // Print to stderr because:
+                    // - the human-readable error also goes to stderr
+                    // - JSON output must be empty
+                    eprintln!("The config is valid.");
+                    return Ok(());
+                };
+
+                if json {
+                    let diagnostic = ipc_diagnostic::convert_to_ipc(&*report);
+                    serde_json::to_writer(BufWriter::new(io::stdout()), &diagnostic)?;
+                }
+
+                return Err(report.into());
             }
             Sub::Msg { msg, json } => {
                 handle_msg(msg, json)?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -107,7 +107,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 };
 
                 if json {
-                    let diagnostic = ipc_diagnostic::convert_to_ipc(&*report);
+                    let diagnostic = ipc_diagnostic::convert_to_ipc(&*report)?;
                     serde_json::to_writer(BufWriter::new(io::stdout()), &diagnostic)?;
                 }
 
--- a/niri-config/src/ipc_diagnostic.rs
+++ b/niri-config/src/ipc_diagnostic.rs
@@ -1,5 +1,5 @@
 use miette::{LabeledSpan, SourceCode, SourceSpan, SpanContents};
-use niri_ipc::diagnostic::{Diagnostic, Label, LineSpan, Severity, Span};
+use niri_ipc::diagnostic::{Diagnostic, Label, LinePosition, Severity, Span};
 
 pub fn convert_to_ipc(diagnostic: &dyn miette::Diagnostic) -> Diagnostic {
     diagnostic_to_ipc(diagnostic, None)
@@ -78,11 +78,11 @@ fn convert_severity(value: miette::Severity) -> Severity {
 fn line_span_from_label<S: SourceCode + ?Sized>(
     code: Option<&S>,
     span: &SourceSpan,
-) -> Option<LineSpan> {
+) -> Option<LinePosition> {
     let code = code?;
     let content = code.read_span(span, 0, 0).ok()?;
-    Some(LineSpan {
+    Some(LinePosition {
         line: content.line(),
-        col: content.column(),
+        character: content.column(),
     })
 }
--- a/niri-config/src/ipc_diagnostic.rs
+++ b/niri-config/src/ipc_diagnostic.rs
@@ -36,7 +36,6 @@ fn diagnostic_to_ipc(
                     .and_then(SpanContents::name)
                     .map(ToOwned::to_owned)
             })
-            .unwrap_or_default()
         },
         labels: diagnostic
             .labels()
--- a/niri-config/src/ipc_diagnostic.rs
+++ b/niri-config/src/ipc_diagnostic.rs
@@ -1,7 +1,7 @@
-use miette::{LabeledSpan, SourceCode, SourceSpan, SpanContents};
+use miette::{miette, Context, LabeledSpan, SourceCode, SourceSpan, SpanContents};
 use niri_ipc::diagnostic::{Diagnostic, Label, LinePosition, Severity, Span};
 
-pub fn convert_to_ipc(diagnostic: &dyn miette::Diagnostic) -> Diagnostic {
+pub fn convert_to_ipc(diagnostic: &dyn miette::Diagnostic) -> Result<Diagnostic, miette::Error> {
     diagnostic_to_ipc(diagnostic, None)
 }
 
@@ -9,9 +9,9 @@ pub fn convert_to_ipc(diagnostic: &dyn miette::Diagnostic) -> Diagnostic {
 fn diagnostic_to_ipc(
     diagnostic: &dyn miette::Diagnostic,
     parent_src: Option<&dyn SourceCode>,
-) -> Diagnostic {
+) -> miette::Result<Diagnostic> {
     let src = diagnostic.source_code().or(parent_src);
-    Diagnostic {
+    Ok(Diagnostic {
         message: diagnostic.to_string(),
         severity: diagnostic
             .severity()
@@ -39,32 +39,34 @@ fn diagnostic_to_ipc(
         },
         labels: diagnostic
             .labels()
-            .map(|iter| {
-                iter.map(|label| Label {
+            .into_iter()
+            .flatten()
+            .map(|label| -> miette::Result<Label> {
+                Ok(Label {
                     label: label.label().map(ToOwned::to_owned).unwrap_or_default(),
                     span: Span {
                         offset: label.offset(),
                         length: label.len(),
-                        start: line_span_from_label(src, label.inner()),
+                        start: line_position_from_label(src, label.inner())
+                            .context("start position")?,
                         // Because miette doesn't just give us the ending line + column for
                         // some reason.
-                        end: line_span_from_label(src, &{
+                        end: line_position_from_label(src, &{
                             let label = label.inner();
                             SourceSpan::from(label.len() + label.offset())
-                        }),
+                        })
+                        .context("end position")?,
                     },
                 })
-                .collect()
             })
-            .unwrap_or_default(),
+            .collect::<miette::Result<_>>()?,
         related: diagnostic
             .related()
-            .map(|iter| {
-                iter.map(|diagnostic| diagnostic_to_ipc(diagnostic, src))
-                    .collect()
-            })
-            .unwrap_or_default(),
-    }
+            .into_iter()
+            .flatten()
+            .map(|diagnostic| diagnostic_to_ipc(diagnostic, src))
+            .collect::<miette::Result<_>>()?,
+    })
 }
 
 fn convert_severity(value: miette::Severity) -> Severity {
@@ -75,13 +77,13 @@ fn convert_severity(value: miette::Severity) -> Severity {
     }
 }
 
-fn line_span_from_label<S: SourceCode + ?Sized>(
+fn line_position_from_label<S: SourceCode + ?Sized>(
     code: Option<&S>,
     span: &SourceSpan,
-) -> Option<LinePosition> {
-    let code = code?;
-    let content = code.read_span(span, 0, 0).ok()?;
-    Some(LinePosition {
+) -> miette::Result<LinePosition> {
+    let code = code.ok_or_else(|| miette!("missing source code"))?;
+    let content = code.read_span(span, 0, 0)?;
+    Ok(LinePosition {
         line: content.line(),
         character: content.column(),
     })
--- a/niri-config/src/ipc_diagnostic.rs
+++ b/niri-config/src/ipc_diagnostic.rs
@@ -1,0 +1,89 @@
+use miette::{LabeledSpan, SourceCode, SourceSpan, SpanContents};
+use niri_ipc::diagnostic::{Diagnostic, Label, LineSpan, Severity, Span};
+
+pub fn convert_to_ipc(diagnostic: &dyn miette::Diagnostic) -> Diagnostic {
+    diagnostic_to_ipc(diagnostic, None)
+}
+
+/// Implementation based on [`miette::JSONReportHandler::render_report()`].
+fn diagnostic_to_ipc(
+    diagnostic: &dyn miette::Diagnostic,
+    parent_src: Option<&dyn SourceCode>,
+) -> Diagnostic {
+    let src = diagnostic.source_code().or(parent_src);
+    Diagnostic {
+        message: diagnostic.to_string(),
+        severity: diagnostic
+            .severity()
+            .map(convert_severity)
+            .unwrap_or(Severity::Error),
+        url: diagnostic.url().as_ref().map(ToString::to_string),
+        help: diagnostic.help().as_ref().map(ToString::to_string),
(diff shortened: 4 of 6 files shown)


titaniumtraveler:
That should be it now I think.
Though if I find a nice solution to package the diagnostic thing, I might add it to the wiki later.

titaniumtraveler:
I've now packaged my solution as editor independent [`efm-langserver`](https://github.com/mattn/efm-langserver) configuration.
<https://github.com/titaniumtraveler/efm-langserver-niri-validate>

src/cli.rs
```diff
@@ -52,6 +52,9 @@ pub enum Sub {
         /// command line argument takes precedence.
         #[arg(short, long)]
         config: Option<PathBuf>,
+        /// Format output as JSON.
```
YaLTeR, line 55: I guess this is more like, format errors as JSON?

niri-config/src/json_report.rs
```diff
@@ -0,0 +1,149 @@
...
+        JsonReport::from_diagnostic(AsRef::<dyn Diagnostic + Send + Sync>::as_ref(value), None)
+    }
+}
+
+impl JsonReport {
+    /// Implementation based on [miette::JSONReportHandler::render_report()].
+    fn from_diagnostic<D: Diagnostic + ?Sized>(
+        diagnostic: &D,
+        parent_src: Option<&dyn SourceCode>,
+    ) -> Self {
```
YaLTeR, line 27: I think we can simplify this to something like `fn from_report(report: &miette::Report)`

niri-config/src/json_report.rs
```diff
@@ -0,0 +1,149 @@
...
+use serde::Serialize;
+
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct JsonReport {
+    pub message: String,
+    pub severity: Severity,
+    pub url: String,
+    pub help: String,
+    pub filename: String,
```
YaLTeR, line 11: Maybe these should be Options?

niri-config/src/json_report.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub labels: Vec<Label>,
+    pub related: Vec<JsonReport>,
```
YaLTeR, line 13: When is this filled?

niri-config/src/json_report.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct LineSpan {
+    /// 0-indexed line into the file
+    line: usize,
+    /// 0-indexed column into the file
+    col: usize,
+}
```
YaLTeR, line 138: Do we need this?

niri-config/Cargo.toml
```diff
@@ -14,10 +14,15 @@ knuffel = "3.2.0"
...
 regex = "1.11.1"
+schemars = { workspace = true, optional = true }
+serde.workspace = true
 smithay = { workspace = true, features = ["backend_libinput"] }
 tracing.workspace = true
 tracy-client.workspace = true
 
+[features]
+json-schema = ["dep:schemars"]
+
```
YaLTeR, line 25: You know, I think let's put this JsonReport into niri-ipc to avoid these extra direct deps in niri-config. It's not used there now, but if there ever appears a niri msg validate or something, it'll move there anyway.

YaLTeR: But, leave the from miette::Report in niri-config, also to avoid deps.

src/main.rs
```diff
@@ -93,13 +94,18 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
                 let (path, _, _) = config_path(config);
-                Config::load(&path)?;
-                info!("config is valid");
-                return Ok(());
+                let Err(report) = Config::load(&path) else {
+                    info!("config is valid");
+                    return Ok(());
+                };
+                if json {
+                    let json = JsonReport::from(&report);
```
YaLTeR, line 105: Can parsing produce multiple errors? Would it be able to in the future, like with kdlv2?

src/cli.rs
titaniumtraveler: Fair, wording was never my strong suit lol

niri-config/src/json_report.rs
titaniumtraveler: Nope, `from_diagnostic()` is (and **has** to be) recursive.
(Though you are right that I should refactor it to just take `&dyn Diagnostic` *always*)

niri-config/src/json_report.rs
titaniumtraveler: `filename` is definitely better as `String`. (And should now actually never be empty because of <https://github.com/YaLTeR/niri/pull/1244/commits/c6629b1f6e038290392baf1d9d96ded982cc48c4>)
But yeah, `url` and `help` are definitely more *optional*.

niri-config/src/json_report.rs
titaniumtraveler: Short answer is when the `miette::Report` fills them.
Long answer is that the `miette::Report` is collected from types that implement [`Diagnostic`](https://docs.rs/miette/5.10.0/miette/trait.Diagnostic.html)s, which then can decide whatever information they want to provide.

`help` and `url` are not used by `knuffel` at the moment, but *could* be in the future.
(The whole "knuffel not letting us set a `miette::Report` as error" is something that I kinda want to look into further)

titaniumtraveler: `labels` are provided every time there is something that references a position in the `knuffel` parsing.
`related` are for nested errors.

niri-config/src/json_report.rs
titaniumtraveler: Yes! Most tools need line and col instead of the byte offsets to generate diagnostics.
Regardless how stupid that is. Especially because they already have the buffer.

For example for nvim there doesn't seem to be *any* trivial way to translate byte-offset to line + column.

And for `efm-langserver` (which is what I used to build a *cross-editor* solution for all of this **requires** line+column).

niri-config/Cargo.toml
titaniumtraveler: Yeah, good idea.

src/main.rs
titaniumtraveler: If it does, they are nested in [`miette::Diagnostic::related()`](https://docs.rs/miette/5.10.0/miette/trait.Diagnostic.html#method.related).

Technically we *already* return multiple errors. For example colliding keybindings don't stop the parsing of the config and only emit an error, which then are later returned as one `Report`.

(And in the worst case we could return multiple json lines ...)

titaniumtraveler: I think the best way to show it, is an example:
image

Contains

image

So the top level error is always the "error loading config" thingy

Originally with the `JSONReportHandler` there also was the `causes` field that contains the `error parsing` and `error parsing KDL` lines, but the impl of that uses some private type of `miette` and considering I don't display the `causes` anywhere I really didn't want to invest the time to figure out how to do that with out those types.

YaLTeR: I see, thanks.

niri-config/Cargo.toml
YaLTeR: Does it? The whole point of making our own type is that it doesn't strictly depend on miette.

titaniumtraveler: *Using* `JsonReport` doesn't, but constructing it does.
The only reason why we can construct `JsonReport` in `src/main.rs` is because rust doesn't require a type to be available to be used.

`Config::load()` returns `miette::Report`, which then is converted to `JsonReport`.
(Even though `miette::Report` isn't imported in that scope)

YaLTeR: Can't you have `pub fn convert_to_json(miette::Report) -> niri_ipc::JsonReport {}` in niri-config?

titaniumtraveler: Yes. Thought I'm not sure what that gets us.
It mostly means we can't use methods as part of the implementation and `schemars` is part of the build *anyways*. Like the build artifact is shared between `niri-ipc` and `niri-config`.

I feel like that is a whole lot of refactoring for not much benefit.
Also it splits the implementation weirdly between crate boundaries, which makes the code less readable.

titaniumtraveler: (I guess *one* solution would be to split it off into a micro crate, but pretty sure that wouldn't be worth it)

YaLTeR: > It mostly means we can't use methods as part of the implementation

I'm not sure I understand. What methods? The function will be in niri-config, and it calls miette methods.

If you mean the From impls, I don't think they are needed? Like, you can just normal functions instead. And it's imo not a big deal that they are freestanding functions in niri-config instead of methods on the types.

> and schemars is part of the build anyways. Like the build artifact is shared between niri-ipc and niri-config.

I know, but it's one less crate with a feature flag to remember. And you can get the entire schema from niri-ipc instead of having to deal with two crates with feature flags.

> Also it splits the implementation weirdly between crate boundaries

Not really? The entire implementation is in niri-config, and niri-ipc just has the type. Imagine we update miette or move from miette to something else, then the impl would change but the type would remain the same (for backwards compat). Which is coincidentally the whole point of doing our own type. So I don't think there's any intention mismatch here?

titaniumtraveler: I will just do the refactor and then see how that looks like *shrug*

titaniumtraveler:
@YaLTeR Moved the types to `niri-ipc::json_report` now.

One thing that is definitely still missing are the docs and there I'm not sure how detailed they should be. (Especially for the fields because the user will probably never actually have to interact with them directly)

I will look at how they look when exported with `schemars` considering that is likely the context where they are most useful.

niri-ipc/src/diagnostic.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Printable message.
+    pub message: String,
+    /// Diagnostic severity.
+    pub severity: Severity,
+    /// URL to visit for a more detailed explanation.
+    pub url: Option<String>,
+    /// Additional help about this diagnostic.
+    pub help: Option<String>,
+    /// Name of the file where the diagnostic occurred.
+    pub filename: String,
```
YaLTeR, line 27: How does this work exactly? A filename on its own without any path doesn't seem very helpful. Is it `~/.config/niri/config.kdl` or `/etc/niri/config.kdl`?

How will this work with includes in the future?

What if the config file name has invalid UTF-8?

Should this be optional? What if in the future there's a command to validate a config snippet without writing it to a file, which seems like a logical extension to your language server idea?

niri-ipc/src/diagnostic.rs
```diff
@@ -0,0 +1,76 @@
...
+/// Span within a source file.
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct Span {
+    /// The 0-based starting byte offset.
+    pub offset: usize,
+    /// Number of bytes this span spans.
+    pub length: usize,
+    pub start: Option<LineSpan>,
+    pub end: Option<LineSpan>,
```
YaLTeR, line 66: I am still very much not convinced that niri is the right place to compute this.

- I really don't like that this is fallible. We just parsed the config, what do you mean we can't figure out the line numbers?
- Miette's own JSONReport doesn't have this information.
- The editor seems like a better suited place for computing the lines. It probably has the info already cached. Whereas in niri if you go inside `SourceCode::read_span()` impls, it will read byte by byte and count the newlines. Twice for every single label. That's a lot of work that the editor had literally already done.
- If you need to return this from the language server, then arguably it should be computed on the language server side, which is an expected long-running process that has persistent cache for all the files it's working with. Once again it can compute the lines much faster there.

YaLTeR (commented):
I pushed some docs and cosmetic changes. Now, unfortunately I'm back with questions about line and column numbers.

YaLTeR:
The CI fail found a bug on main, I'll fix it there.

YaLTeR:
The stderr logging commit is now on main

niri-ipc/src/diagnostic.rs
titaniumtraveler: > How does this work exactly? A filename on its own without any path doesn't seem very helpful. Is it `~/.config/niri/config.kdl` or `/etc/niri/config.kdl`?

The filename is **exactly** the filename is set by the [`miette::SourceCode`](https://docs.rs/miette/latest/miette/trait.SourceCode.html). In `niri`s case it is what is set by the parse: https://github.com/YaLTeR/niri/blob/4f46374e21ee57849b6debddc77ba8c532d2e2b6/niri-config/src/lib.rs#L2152-L2157

> How will this work with includes in the future?

It will work as long as the config name is set as part of the parsing.
 
> What if the config file name has invalid UTF-8?

`miette` doesn't allow that in the first place. (Which is honestly pretty weird to me, but I guess it is what we are working with *shrug*) <https://docs.rs/miette/latest/miette/trait.SpanContents.html#method.name>

> Should this be optional? What if in the future there's a command to validate a config snippet without writing it to a file, which seems like a logical extension to your language server idea?

Yeah, you are probably correct that it should be optional, but provided when ever possible. For example when feeding the file via stdin (or any similar non-file), the user should be able to **set** a filename that is associated with the output.

As for the language server protocol in specific, all diagnostics are associated with a "resource", which are identified by an URI.
See
- <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic>
- <https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri>

niri-ipc/src/diagnostic.rs
titaniumtraveler: > I am still very much not convinced that niri is the right place to compute this.
> 
>     * I really don't like that this is fallible. We just parsed the config, what do you mean we can't figure out the line numbers?

It being fallible is not really something that happens in *our* cases.
It is more about some implementations of the `miette` traits that might not give us the information.

>     * Miette's own JSONReport doesn't have this information.

But the visual output *has*. (Else it wouldn't be able to reference the code)

>     * The editor seems like a better suited place for computing the lines. It probably has the info already cached. Whereas in niri if you go inside `SourceCode::read_span()` impls, it will read byte by byte and count the newlines. Twice for every single label. That's a lot of work that the editor had literally already done.
> 
>     * If you need to return this from the language server, then arguably it should be computed on the language server side, which is an expected long-running process that has persistent cache for all the files it's working with. Once again it can compute the lines much faster there.

I *generally* agree with you there, but for example [`efm-langserver`](https://github.com/mattn/efm-langserver) does **not** give you the option to give it the byte offsets.
It basically comes down to the question: Do we rather **require** smarter clients, or do we support dumber clients as well, but do a bit more work.

I guess the best compromise would be to make the line spans optional.

titaniumtraveler:
I also updated <https://github.com/titaniumtraveler/efm-langserver-niri-validate> to make sure it still runs as expected.

niri-ipc/src/diagnostic.rs
YaLTeR: > It being fallible is not really something that happens in our cases.
> It is more about some implementations of the miette traits that might not give us the information.

Well, we shouldn't push our implementation weirdness to IPC clients. If we're expected to be able to provide column and line info for anything we parse (and when wouldn't we?), then it should not be optional.

> I guess the best compromise would be to make the line spans optional.

Hmm, not a big fan of that. Just a weird extra CLI flag. If we provide line spans then let's always provide them.

> I generally agree with you there, but for example [efm-langserver](https://github.com/mattn/efm-langserver) does not give you the option to give it the byte offsets.
It basically comes down to the question: Do we rather require smarter clients, or do we support dumber clients as well, but do a bit more work.

I also have this question: is there a defined way to handle e.g. \r\n vs. \n for line counting? If yes, then where is it defined? Maybe there are multiple competing standards? E.g. LSP does one thing, but something else does another thing? This is FWIW another argument against doing this inside niri: efm-langserver knows that it's targeting LSP where this is presumably defined, so it can adjust as necessary, but niri doesn't know that it's targeting LSP.

And a sidenote, is the end LineSpan even needed? https://github.com/mattn/efm-langserver/issues/287 seems like it's unused.

titaniumtraveler: > > It being fallible is not really something that happens in our cases.
> > It is more about some implementations of the miette traits that might not give us the information.
> 
> Well, we shouldn't push our implementation weirdness to IPC clients. If we're expected to be able to provide column and line info for anything we parse (and when wouldn't we?), then it should not be optional.

I mean what happens when we actually don't in some case? Be it because of a bug, or any other reason. I think it should be fine to write IPC clients assuming it is never null. (That's an assumption that I wrote mine with) The alternatives are either to return some sentinel values like `(0,0),(0,0)` or panic and I think just returning the explicitly `null` values is the best option of that.

> > I guess the best compromise would be to make the line spans optional.
> 
> Hmm, not a big fan of that. Just a weird extra CLI flag. If we provide line spans then let's always provide them.

I don't really care either way. As long as there is a way to get them it is fine by me.

> > I generally agree with you there, but for example [efm-langserver](https://github.com/mattn/efm-langserver) does not give you the option to give it the byte offsets.
> > It basically comes down to the question: Do we rather require smarter clients, or do we support dumber clients as well, but do a bit more work.
> 
> I also have this question: is there a defined way to handle e.g. \r\n vs. \n for line counting? If yes, then where is it defined? Maybe there are multiple competing standards? E.g. LSP does one thing, but something else does another thing? This is FWIW another argument against doing this inside niri: efm-langserver knows that it's targeting LSP where this is presumably defined, so it can adjust as necessary, but niri doesn't know that it's targeting LSP.

I didn't even knew that `\r\n` line endings are an option for `knuffel`, but I assume it should work correctly anyway.

> And a sidenote, is the end LineSpan even needed? [mattn/efm-langserver#287](https://github.com/mattn/efm-langserver/issues/287) seems like it's unused.

It definitely is supported. It is even documented on the parsing library that`efm-langserver` uses.

YaLTeR: > what happens when we actually don't in some case?

Can return an error from the IPC.

> I didn't even knew that \r\n line endings are an option for knuffel, but I assume it should work correctly anyway.

This is not a question about the parser. This is a question about the consumer.

titaniumtraveler: > > what happens when we actually don't in some case?
> 
> Can return an error from the IPC.

Fair. I would usually prefer at least giving back *some* information rather than erroring out completely, but considering that the case should never actually happen, returning an error should be fine too.

> > I didn't even knew that \r\n line endings are an option for knuffel, but I assume it should work correctly anyway.
> 
> This is not a question about the parser. This is a question about the consumer.

Shouldn't matter either way. If the consumer decides to store the file with `\n\r` and then sends a `\n` document to `validate`, I would expect it to be able to translate back as well.

YaLTeR: What I mean is: say you're editing your actual config.kdl. It happens to have \r\n or whatever. Then niri returns LineSpan lines assuming \r\n is two newlines, but the editor thinks that they are computed assuming \r\n is one newline. Or vice versa or something

titaniumtraveler: > What I mean is: say you're editing your actual config.kdl. It happens to have \r\n or whatever. Then niri returns LineSpan lines assuming \r\n is two newlines, but the editor thinks that they are computed assuming \r\n is one newline. Or vice versa or something

That shouldn't be a problem. I don't know how it is with line endings in LSP, I'm not even sure if non `\n` line-endings are allowed for in LSP (I would have to look it up), but considering how LSP handles different file encodings it shouldn't be a problem.
For example the LSP-Server might instruct the LSP-Client to send the files in UTF-16, even if the *actual* file is in UTF-8. So I'm pretty sure that problem is solved on the protocol level.

YaLTeR: Exactly, it instructs the *LSP client*. But niri is not an LSP client, it's completely on its own and reads the file in whatever other way.

titaniumtraveler: Okay, I can imagine two scenarios:

1. `niri-validate` and the editor both parse the same physical file. In that case the LSP-Server (i.e. the thing that uses the json output to generate diagnostics) forces the LSP-Client (i.e. the editor) to use the same representation as the actual file, which means both end up with the right diagnostics.

2. `niri-validate` takes in the file from stdin and the LSP-Server asks the client to do some kind of transformation on the file. That transformed file is given to `niri-validate`, which then returns the transformed diagnostics to the LSP-Server, which gives them to the LSP-Client (i.e. the editor), which transforms them back.

In both cases the location data will match up.

YaLTeR: > In that case the LSP-Server (i.e. the thing that uses the json output to generate diagnostics) forces the LSP-Client (i.e. the editor) to use the same representation as the actual file

In this case, how does the LSP-Server know what representation niri used to compute its line spans? I guess at the very least we need to document what niri does.

niri-ipc/src/diagnostic.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct LineSpan {
```
YaLTeR, line 71: In an unrelated comment, if it's a Line"Span" then it should include both start and end in the same struct, rather than split into two.

niri-ipc/src/diagnostic.rs
titaniumtraveler: I mean we kind of did already.
We state that the config file is in KDL, which implies we use KDL's definition of newline: <https://kdl.dev/spec/#section-3.17>
At most we could link to that section.

niri-ipc/src/diagnostic.rs
titaniumtraveler: Actually I would prefer to keep it two separate structs instead of having to duplicate the fields to `start_line, start_col, end_line, end_col`. Having them in a single format makes the handling easier.

It is incidentally also what LSP is doing:
image

titaniumtraveler: I will rename it though. [`Range`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range) seems like a good way to name it.

titaniumtraveler: Done

niri-ipc/src/diagnostic.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// Span within a source file.
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct Span {
+    /// The 0-based starting byte offset.
+    pub offset: usize,
+    /// Number of bytes this span spans.
+    pub length: usize,
+    /// Starting line position of this span
```
YaLTeR, line 65:
```suggestion
    /// Starting line position of this span.
```

niri-ipc/src/diagnostic.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub start: LinePosition,
+    /// Ending line position of this span
```
YaLTeR, line 67:
```suggestion
    /// Ending line position of this span.
```

niri-ipc/src/diagnostic.rs
```diff
@@ -0,0 +1,82 @@
...
+///
+/// Note that this counts line numbers as declared by the KDL specification:
+/// <https://kdl.dev/spec/#section-3.18>
+#[derive(Serialize)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct LinePosition {
+    /// The 0-indexed line position in the file.
+    pub line: usize,
+    /// The 0-indexed character offset into the file, relative to `line`.
+    pub character: usize,
```
YaLTeR, line 81: "Character" is loosely defined in computers. Actually, this brings up another question: what is this supposed to mean? Bytes? UTF-8 codepoints? Grapheme clusters? And where is it defined, what this is supposed to mean?

titaniumtraveler: I'm pretty sure we should use the definition of character that LSP uses.
Let me read up on it real quick.

titaniumtraveler: According to this it is code points. (In whatever encoding client and server agree on) https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification#:~:text=So%20in%20a%20string%20of%20the%20form%20a%f0%90%90%80b%20the%20character%20offset%20of%20the%20character%20a%20is%200%2c%20the%20character%20offset%20of%20%f0%90%90%80%20is%201%20and%20the%20character%20offset%20of%20b%20is%203%20since%20%f0%90%90%80%20is%20represented%20using%20two%20code%20units%20in%20UTF%2d16.%0a

YaLTeR: Yeah, LSP rather infamously uses UTF-16 code units, which is awkward for most languages...

YaLTeR: Maybe efm-langserver does some conversion?

titaniumtraveler: > Yeah, LSP rather infamously uses UTF-16 code units, which is awkward for most languages...

(If you read the link it doesn't *anymore*! Or at least not necessarily!)

titaniumtraveler: > Maybe efm-langserver does some conversion?

Not sure.
I would assume it forwards the exact positions given, but I could look into the code if it's important.

titaniumtraveler: Also note that at least *currently* the script I wrote doesn't access the file via stdin or something, but gets the path of the file *directly*.

YaLTeR: > I would assume it forwards the exact positions given, but I could look into the code if it's important.

Well yeah it's kinda important. For ASCII it doesn't matter, but we might very well end up with some complex characters in the diagnostic.

titaniumtraveler: > > I would assume it forwards the exact positions given, but I could look into the code if it's important.
> 
> Well yeah it's kinda important. For ASCII it doesn't matter, but we might very well end up with some complex characters in the diagnostic.

Looking at https://github.com/mattn/efm-langserver/blob/bad115b56bd96f517e25d66ca660d5f194af629e/langserver/handler.go#L508-L578 seems like it pretty much returns the untranslated diagnostics.

YaLTeR: Makes me wonder if different linters even agree how they count columns

YaLTeR: FWIW if there's no clear solution I'd err on the side of UTF-8 code points, sounds the least cursed?

YaLTeR: I think that's ~the same as Rust `char`

titaniumtraveler: Yep.
And I mean just saying ASCII for now and then later fixing any bugs that might come up should be fine.
No need deciding on some detail that might be better to change later if we don't need to.

YaLTeR: We can't say ASCII because KDL is UTF-8. How does the current code count?

titaniumtraveler: Whatever `knuffel` does. IMO it doesn't really matter until it actually does.

YaLTeR: I just don't want to find out that I need to break back compat in our JSON output :)

YaLTeR: Could you check the source? It's usually easy, just go-to-def a few times right into miette

titaniumtraveler: *Technically* it depends on the specific impl of `SourceCode`, but the reference impl seems to use bytes.
(Not 100% sure about that tho)
https://docs.rs/miette/5.10.0/src/miette/source_impls.rs.html#13-98

YaLTeR: Seems like it is bytes. Not sure bytes are expected somewhere like here though, could you try if it looks right with your langserver?

```
/-щщщщ {}; asdf
```

Also not sure if that end is correct, being on the second line, rather than at the end of the same line?

image

titaniumtraveler: (I haven't tested it), but I'm pretty sure it would return the wrong col.

Pretty fucking annoying that `miette` uses *bytes* not utf-8 code points as it arguably should!

I will probably at some point look into writing an actual language server for `niri` (or see if the KDL one can be adapted to use the `niri` schema + some extra diagnostics).

In the mean time I would say we should probably do one of the following things:
- Merge the PR as-is with the *half* broken character offsets (with docs pointing out the problems and potential future changes)
  Even if only ASCII is supported correctly, having diagnostics directly in the editor would be a pretty big improvement IMO
- Remove the `line`/`character` part and only keep the byte offsets.
  With the right software that can still be very useful!
- Not merge the PR at all.
  Having `json` readable diagnostics are *too* useful for that IMO

YaLTeR: I am still kinda leaning on keeping only the byte offsets because they are correct and obvious. We could also keep something similar to line/character but with byte offset on the line instead of "character offset" (basically rename "character" to something with the word bytes I guess). This leaves path forward to add something else in the future if/when it's convenient.

PR #1253, opened 2025-03-13
Add focus-column (by index) and move-column-to-index actions.

Duncaen:
I'd like to add `focus-column` to focus a specific column by index and `move-column-to-index` to move the focused column to the specified index.

I use those actions and `focus-window-previous` to navigate quickly between the first 9 columns as my most active windows tend to stay within that range. This also works pretty well with tiled/split columns since they remember the last focused window which is usually the one I'd want to navigate to.

The index's start at 1, since that is what other actions used, the downside is that a contributor confuse the `Scrolling::focus_column` with `ScrollingSpace::activate_column` and `ScrollingSpace::move_column_to_index` with `ScrollingSpace::move_to`. Maybe it would be good to name those functions in a way that its clear that they are supposed to be called from user actions and are indexed by 1.  

```
binds {
    Mod+Tab { focus-window-previous; }
    Mod+1 { focus-column 1; }
    Mod+2 { focus-column 2; }
    Mod+3 { focus-column 3; }
    Mod+4 { focus-column 4; }
    Mod+5 { focus-column 5; }
    Mod+6 { focus-column 6; }
    Mod+7 { focus-column 7; }
    Mod+8 { focus-column 8; }
    Mod+9 { focus-column 9; }
    Mod+Shift+1 { move-column-to-index 1; }
    Mod+Shift+2 { move-column-to-index 2; }
    Mod+Shift+3 { move-column-to-index 3; }
    Mod+Shift+4 { move-column-to-index 4; }
    Mod+Shift+5 { move-column-to-index 5; }
    Mod+Shift+6 { move-column-to-index 6; }
    Mod+Shift+7 { move-column-to-index 7; }
    Mod+Shift+8 { move-column-to-index 8; }
    Mod+Shift+9 { move-column-to-index 9; }
}
```


YaLTeR:
Looks like you need one more is_empty() check.

src/layout/scrolling.rs
```diff
@@ -1491,6 +1491,13 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         self.activate_column(self.columns.len() - 1);
     }
 
+    pub fn focus_column(&mut self, index: usize) {
+        if self.columns.is_empty() {
+            return;
+        }
```
YaLTeR, line 1497:
```suggestion
        }

```

src/layout/scrolling.rs
```diff
@@ -1575,6 +1582,13 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         self.columns[self.active_column_idx].focus_bottom()
     }
 
+    pub fn move_column_to_index(&mut self, index: usize) {
+        if self.columns.is_empty() {
+            return;
+        }
```
YaLTeR, line 1588:
```suggestion
        }

```

YaLTeR (approved):

YaLTeR:
Thanks

PR #1258, opened 2025-03-13
Add option to warp-mouse-to-focus to always center

TyberiusPrime:
For my mental model, it's important that the mouse always warps to the center of the focused window.  Niri right now only centers X or Y, but not both, depending on where the mouse was.

This PR adds an option to wrap-mouse-to-focus called 'force-center' which enables centering both, always.

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -14,8 +14,8 @@ use _server_decoration::server::org_kde_kwin_server_decoration_manager::Mode as 
 use anyhow::{bail, ensure, Context};
 use calloop::futures::Scheduler;
 use niri_config::{
-    Config, FloatOrInt, Key, Modifiers, OutputName, PreviewRender, TrackLayout, WorkspaceReference,
-    DEFAULT_BACKGROUND_COLOR,
+    Config, FloatOrInt, Key, Modifiers, OutputName, PreviewRender, TrackLayout,
+    WarpMouseToFocusMode, WorkspaceReference, DEFAULT_BACKGROUND_COLOR,
 };
 use smithay::backend::allocator::Fourcc;
 use smithay::backend::input::Keycode;
@@ -510,6 +510,8 @@ struct SurfaceFrameThrottlingState {
 pub enum CenterCoords {
     Separately,
     Both,
+    // Force centering even if the cursor is already in the rectangle.
+    BothAlways,
 }
 
 #[derive(Clone, PartialEq, Eq)]
@@ -753,6 +755,7 @@ impl State {
                     center_f64(rect)
                 }
             }
+            CenterCoords::BothAlways => center_f64(rect),
         };
 
         self.move_cursor(p);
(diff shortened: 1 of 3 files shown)


src/niri.rs
```diff
@@ -791,15 +793,29 @@ impl State {
...
 
     pub fn maybe_warp_cursor_to_focus_centered(&mut self) -> bool {
-        if !self.niri.config.borrow().input.warp_mouse_to_focus {
+        if self
+            .niri
+            .config
+            .borrow()
+            .input
+            .warp_mouse_to_focus
+            .is_none()
```
YaLTeR, line 817: Shouldn't this also use BothForce? Since as you say you want it to work even if it's within the rectangle already?

wiki/Configuration:-Input.md
```diff
@@ -240,11 +240,13 @@ input {
...
 Makes the mouse warp to newly focused windows.
 
-X and Y coordinates are computed separately, i.e. if moving the mouse only horizontally is enough to put it inside the newly focused window, then it will move only horizontally.
+By default, X and Y coordinates are computed separately, i.e. if moving the mouse only horizontally is enough to put it inside the newly focused window, then it will move only horizontally. 
+If force-center is set, the mouse will always warp to the center of the focused window. 
 
 ```kdl
 input {
     warp-mouse-to-focus
+    // or warp-mouse-to-focus { force-center; }
```
YaLTeR, line 249: Instead of this, it should be a knuffel property, so `warp-mouse-to-focus force-center=true`

YaLTeR (commented):
Some name bikeshedding. This property does two separate things:
1. Center instead of separate coordinate
2. Always warp even if already within rectangle

Do you need/intended to do both of these things or just 1? If not, then let's leave just 1 and maybe rephrase the property name a little. If yes, then I am somewhat thinking that maybe it should be two properties, or at least some clear future path in case people start asking for 1 without 2.

src/niri.rs
TyberiusPrime: That's the function that get's called when you're focusing a window on a different output, isn't it? I figured that the cursor would then always be outside of the rect, but I might be biased because I subconsciously assumed 'focus-follows-mouse' as well.

TyberiusPrime:
> Some name bikeshedding. This property does two separate things:
> 
>     1. Center instead of separate coordinate
> 
>     2. Always warp even if already within rectangle
> 
Hm, I believe it's a bit of a tri-state.

1. if you're doing them separately, force has no effect (which would it change, X or Y?)
2. Both, but only if outside of the window before ('unforced')
3. Both, but always center on focus change, independent of where the mouse was.

Right now this PR only does 1 & 3.

How should we present this to the user?

YaLTeR:
Do you need the mouse to warp even if it's already within the target window?

TyberiusPrime:
> Do you need the mouse to warp even if it's already within the target window?

Yes, that's what I want, my brain is weird :).

I changed it to work like CenterFocusedColumn, 
having four states.

- off  (default)
- separately (current behavior if enable)
- both 
- both-always.

wiki/Configuration:-Input.md
```diff
@@ -240,11 +240,14 @@ input {
 
 Makes the mouse warp to newly focused windows.
 
-X and Y coordinates are computed separately, i.e. if moving the mouse only horizontally is enough to put it inside the newly focused window, then it will move only horizontally.
+ - "off": do not warp (default)
+ - "separately": warp X and Y coordinates to the window's center separately, i.e. if moving the mouse only horizontally is enough to put it inside the newly focused window, then it will move only horizontally and vice versa. Only warps if the mouse was outside of the newly focused window.
+ - "both":  warp both X and Y coordinates to the window's center. Only warps if the mouse was outside of the newly focused window.
+ - "both-always": warp both X and Y coordinates to the window's center. Always warps, even if the mouse was already inside the newly focused window.
```
YaLTeR, line 246: I was trying to come up with better names because "both" is kinda not obvious what it means just looking at the config, but I can't think of anything else rn

wiki/Configuration:-Input.md
```diff
@@ -240,11 +240,14 @@ input {
 
 Makes the mouse warp to newly focused windows.
 
-X and Y coordinates are computed separately, i.e. if moving the mouse only horizontally is enough to put it inside the newly focused window, then it will move only horizontally.
+ - "off": do not warp (default)
```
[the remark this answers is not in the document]
YaLTeR: Also remove the "separately" variant (since it's the default). Which would mean either `Option<Option<WarpMouseToFocus>>` (idk if knuffel will understand this), or `Option<WarpMouseToFocus>` where `WarpMouseToFocus` itself contains one `Option<>` field.

wiki/Configuration:-Input.md
TyberiusPrime: Maybe x-or-y, x-and-y, x-and-y-always?

YaLTeR: How about: `xy` and `xy-always`

bbb651: `never` is more consistent than `off` because of `center-focused-column "never"` (there is a lot more `off`s but as node names not as arguments), and reads better with never/always.
I think `both-always` could just be `always`, then it reads `warp-mouse-to-focus=always`? I expect warping to be to the center, so to me `separately` is the exception and it makes `both-always` more confusing.
And maybe `separate` > `separately`?

YaLTeR: There is not supposed to be any "off". "off" is just omitting the line entirely from the config. There's also not supposed to be "separately" because just writing the line by itself is already "separately". So the question is about "both" and "both-always".

TyberiusPrime:
Once more, with passion.

- leave off -> off.
- `wrap-mouse-to-focus` -> previous behaviour, x-or-y
- `wrap-mouse-to-focus mode = "center-xy"` -> center x and y, if outside the window
- `wrap-mouse-to-focus mode = "center-xy-always"` -> center x and y, no matter what.

I went with 'center-xy', so we'd have 'room' if anybody ever wants to do non-center-behavior ('top-left'/'bottom-right' maybe?).

YaLTeR:
Thanks

Issue #1260, opened 2025-03-13
Workspace binded numbers


maikelthedev:
this page https://github.com/YaLTeR/niri/wiki/Configuration:-Named-Workspaces

Doesn't mention how to put the workspaces in order. 

As discussed in Matrix, the order is by appearance on the config file and this **needs a restart** for it to be enforced. 

I think the file should be amended to reflect this and avoid other people's headaches trying to figure this out. Simply that:

> the order is implicit by the order of workspaces assigned in the config. 
> to enforce them Niri needs to be restarted, a reload of the file won't reassigned the order .

YaLTeR:
The way it works is that named workspaces are always added in the order of the config file, but if they already exist, then they are not reordered. So if you're in the process of editing your configuration, you may end up with them in a different order. You can move them around manually with move-workspace-up/down. If you feel like it needs clarification, do you mind submitting a pull request?

maikelthedev:
I don't mind it at all. Doing it shortly.

PR #1261, opened 2025-03-13
Clarified numbered workspaces

maikelthedev:
Regarding this: https://github.com/YaLTeR/niri/issues/1260


YaLTeR:
As we discussed on Matrix, this change isn't really correct, and the misunderstanding likely stems from unfamiliarity with the niri's dynamic workspace system. I think I should instead write a separate page about dynamic workspaces with clarifications to how numbers work and whatnot.

maikelthedev:
I agree. And that other clarification would help too.

PR #1262, opened 2025-03-13
Clarified how to make it run in Nixos

maikelthedev:
There seems to be an excess of information on how to run a flake but not on how to basically get it up and running in configuration.nix on Nixos when it is mostly programs.niri.enable = true; and that's it. 

So I modified a page in Getting Started to make this clearer. Including how to fix the Electron problem.

### DIFF
--- a/wiki/Getting-Started.md
+++ b/wiki/Getting-Started.md
@@ -66,6 +66,25 @@ When this happens, you usually see a black screen when trying to start niri from
 
 Also, on Intel graphics, you may need a workaround described [here](https://nixos.wiki/wiki/Intel_Graphics).
 
+To run Niri on NixOS is as simple as editing your configuration.nix to add
+
+```nix
+{
+    programs.niri.enable = true;
+}
+```
+
+For your Electron-based apps to not fail it is highly recommendable you also add on that file
+
+```
+{
+  environment = {
+    sessionVariables.NIXOS_OZONE_WL = "1";
+    sessionVariables.ELECTRON_OZONE_PLATFORM_HINT = "auto";
+  };
+}
+```
+
 ### Virtual Machines
 
 To run niri in a VM, make sure to enable 3D acceleration.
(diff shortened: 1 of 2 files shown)


YaLTeR:
Hm, I'm not sure I want this on Getting Started. I don't have any other distro-specific installation instructions there.

dbeley:
For NixOS you have https://github.com/sodiboo/niri-flake

maikelthedev:
> For NixOS you have https://github.com/sodiboo/niri-flake which is very complete to install and configure niri

except it doesn't say the most basic part: how to make it part of the session manager

```nix 
programs.niri.enabled = true;
```

takes care of that. I didn't even have to do anything of the instructions on that flake repo. It goes way over-the-top while ignores the very basics of being able to log into it.

Brisingr05:
> except it doesn't say the most basic part: how to make it part of the session manager

[It does](https://github.com/sodiboo/niri-flake/blob/main/docs.md#programsnirienable). Quite early on.
> ```nix
> programs.niri.enabled = true;
> ```
It's `enable`, not `enabled`.
> I didn't even have to do anything of the instructions on that flake repo. It goes way over-the-top while ignores the very basics of being able to log into it.

Not sure what you mean by "over-the-top". Detailed docs are a good thing IMO. The `niri-flake` repo existed long before the niri NixOS module was added to Nixpkgs, and was the default method of installing niri on NixOS for a long time. Also, using the flake is currently the only way to configure niri using Home Manager unless you create your own module options.

wiki/Getting-Started.md
```diff
@@ -66,6 +66,25 @@ When this happens, you usually see a black screen when trying to start niri from
 
 Also, on Intel graphics, you may need a workaround described [here](https://nixos.wiki/wiki/Intel_Graphics).
 
+To run Niri on NixOS is as simple as editing your configuration.nix to add
```
abmantis, line 69: For people using NixOS, looking for an option in https://search.nixos.org/options should be the first thing when installing some new software, as that usually takes care of a few extras in addition to installing the package.

I am not sure if it makes sense to add this here, unless it makes sense to also add instructions for other distros.

PR #1265, opened 2025-03-14
(WIP) niri-ipc: Include window positions within the workspace columns in Window state and output

yrkv:
This is a WIP draft to extend the IPC and the behavior of `niri msg -j windows` to support listing where each window is in terms of column/row indices, as well as (client-side) compute the geometry of all windows. This allows for development of much more interactive/useful status bars for niri.

- Add three new fields into `niri-ipc::Window`:
    - `tile_coordinates` -- Location of the window's tile within a workspace in terms of indices.
    - `tile_size` -- Size of the tile.
    - ~`geometry` -- [x,y,width,height] of the tile relative to the workspace's top-left corner, in pixels.~
- Add two new window events into the event stream to maintain the above state:
    - One for when a collection of tiles/windows are resized at once.
    - Another for when a collection of tiles/windows are shifted over by some number of columns.

Current limitations of this draft:
As of yet, this does not expose pos/size for both tile and window, only for the tile.
Additionally, there is no consideration of tabbed windows yet.

Discussion of the `geometry` field:
(edit: agreement has been reached to not include it, don't bother reading this paragraph)
Window state `geometry` is never actually set on the niri server side, it's computed and filled in client-side for `niri msg -j windows`. This is pretty awkward, but it seems better IMO than leaving it to every single user to implement, especially since many uses of this will be in environments such as shell scripts where doing the necessary math would suck. It might be better to make a new `niri msg window-geometries` request or something for this in particular, and remove it from Window state. It's difficult to actually set the window geometry in the server because it would require spamming the event stream a bit (but it's not a completely unreasonable amount, so this would also be feasible).

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1156,12 +1156,16 @@ pub struct Window {
     /// Whether this window requests your attention.
     pub is_urgent: bool,
     /// Position- and size-related properties of the window.
-    pub location: WindowLayout,
+    pub layout: WindowLayout,
 }
 
 /// Position- and size-related properties of a [`Window`].
 ///
 /// Optional properties will be unset for some windows, do not rely on them always being present.
+///
+/// All sizes and positions are in *logical pixels* unless stated otherwise. Logical sizes may be
+/// fractional. For example, at 1.25 monitor scale, a 2-physical-pixel-wide window border is 1.6
+/// logical pixels wide.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
 pub struct WindowLayout {
@@ -1170,9 +1174,12 @@ pub struct WindowLayout {
     pub pos_in_scrolling_layout: Option<(usize, usize)>,
     /// Size of the tile this window is in, including decorations like borders.
     pub tile_size: (f64, f64),
-    /// Size of the window itself.
+    /// Size of the window's visual geometry itself.
     ///
-    /// Note that Wayland windows can only be integer sized.
+    /// Does not include niri decorations like borders.
+    ///
+    /// Currently, Wayland toplevel windows can only be integer-sized in logical pixels, even
+    /// though it doesn't necessarily align to physical pixels.
     pub window_size: (i32, i32),
     /// Tile position within the current view of the workspace.
     ///
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1155,6 +1155,35 @@ pub struct Window {
     pub is_floating: bool,
     /// Whether this window requests your attention.
     pub is_urgent: bool,
+    /// Position- and size-related properties of the window.
+    pub location: WindowLayout,
+}
+
+/// Position- and size-related properties of a [`Window`].
+///
+/// Optional properties will be unset for some windows, do not rely on them always being present.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct WindowLayout {
+    /// Location of a tiled window within a workspace in terms of (column index, tile index in
+    /// column).
+    pub pos_in_scrolling_layout: Option<(usize, usize)>,
+    /// Size of the tile this window is in, including decorations like borders.
+    pub tile_size: (f64, f64),
+    /// Size of the window itself.
+    ///
+    /// Note that Wayland windows can only be integer sized.
+    pub window_size: (i32, i32),
+    /// Tile position within the current view of the workspace.
+    ///
+    /// This is the same "workspace view" as in gradients' `relative-to` in the niri config.
+    pub tile_pos_in_workspace_view: Option<(f64, f64)>,
+    /// Window position within the current view of the workspace.
+    ///
+    /// Unlike `tile_pos_in_workspace_view`, this is the position of the window contents, excluding
+    /// borders and other decorations. You probably want to use `tile_pos_in_workspace_view`
+    /// instead, since borders are a part of the window, as far as the user is concerned.
+    pub window_pos_in_workspace_view: Option<(f64, f64)>,
 }
 
 /// Output configuration change result.
@@ -1330,6 +1359,11 @@ pub enum Event {
         id: u64,
         /// The new urgency state of the window.
         urgent: bool,
+    },
+    /// The layout of one or more windows has changed.
+    WindowLayoutsChanged {
+        /// Pairs consisting of a window id and new layout information for the window.
+        changes: Vec<(u64, WindowLayout)>,
     },
     /// The configured keyboard layouts have changed.
     KeyboardLayoutsChanged {
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1178,8 +1178,11 @@ pub struct Window {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
 pub struct WindowLayout {
-    /// Location of a tiled window within a workspace in terms of (column index, tile index in
-    /// column).
+    /// Location of a tiled window within a workspace: (column index, tile index in column).
+    ///
+    /// The indices are 1-based, i.e. the leftmost column is at index 1 and the topmost tile in a
+    /// column is at index 1. This is consistent with [`Action::FocusColumn`] and
+    /// [`Action::FocusWindowInColumn`].
     pub pos_in_scrolling_layout: Option<(usize, usize)>,
     /// Size of the tile this window is in, including decorations like borders.
     pub tile_size: (f64, f64),
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1161,11 +1161,20 @@ pub struct Window {
 
 /// Position- and size-related properties of a [`Window`].
 ///
-/// Optional properties will be unset for some windows, do not rely on them always being present.
+/// Optional properties will be unset for some windows, do not rely on them being present. Whether
+/// some optional properties are present or absent for certain window types may change across niri
+/// releases.
 ///
 /// All sizes and positions are in *logical pixels* unless stated otherwise. Logical sizes may be
 /// fractional. For example, at 1.25 monitor scale, a 2-physical-pixel-wide window border is 1.6
 /// logical pixels wide.
+///
+/// This struct contains positions and sizes both for full tiles ([`Self::tile_size`],
+/// [`Self::tile_pos_in_workspace_view`]) and the window geometry ([`Self::window_size`],
+/// [`Self::window_offset_in_tile`]). For visual displays, use the tile properties, as they
+/// correspond to what the user visually considers "window". The window properties on the other
+/// hand are mainly useful when you need to know the underlying Wayland window sizes, e.g. for
+/// application debugging.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
 pub struct WindowLayout {
@@ -1185,12 +1194,12 @@ pub struct WindowLayout {
     ///
     /// This is the same "workspace view" as in gradients' `relative-to` in the niri config.
     pub tile_pos_in_workspace_view: Option<(f64, f64)>,
-    /// Window position within the current view of the workspace.
+    /// Location of the window's visual geometry within its tile.
     ///
-    /// Unlike `tile_pos_in_workspace_view`, this is the position of the window contents, excluding
-    /// borders and other decorations. You probably want to use `tile_pos_in_workspace_view`
-    /// instead, since borders are a part of the window, as far as the user is concerned.
-    pub window_pos_in_workspace_view: Option<(f64, f64)>,
+    /// This includes things like border sizes. For fullscreened fixed-size windows this includes
+    /// the distance from the corner of the black backdrop to the corner of the (centered) window
+    /// contents.
+    pub window_offset_in_tile: (f64, f64),
 }
 
 /// Output configuration change result.
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -151,6 +151,15 @@ impl EventStreamStatePart for WindowsState {
             Event::WindowsChanged { windows } => {
                 self.windows = windows.into_iter().map(|win| (win.id, win)).collect();
             }
+            Event::WindowLayoutsChanged { changes } => {
+                for (id, update) in changes {
+                    let w = self
+                        .windows
+                        .get_mut(&id)
+                        .expect("changed window was missing from the map");
+                    w.location = update;
+                }
+            }
             Event::WindowOpenedOrChanged { window } => {
                 let (id, is_focused) = match self.windows.entry(window.id) {
                     Entry::Occupied(mut entry) => {
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -151,15 +151,6 @@ impl EventStreamStatePart for WindowsState {
             Event::WindowsChanged { windows } => {
                 self.windows = windows.into_iter().map(|win| (win.id, win)).collect();
             }
-            Event::WindowLayoutsChanged { changes } => {
-                for (id, update) in changes {
-                    let w = self
-                        .windows
-                        .get_mut(&id)
-                        .expect("changed window was missing from the map");
-                    w.location = update;
-                }
-            }
             Event::WindowOpenedOrChanged { window } => {
                 let (id, is_focused) = match self.windows.entry(window.id) {
                     Entry::Occupied(mut entry) => {
@@ -196,6 +187,13 @@ impl EventStreamStatePart for WindowsState {
                         win.is_urgent = urgent;
                         break;
                     }
+                }
+            }
+            Event::WindowLayoutsChanged { changes } => {
+                for (id, update) in changes {
+                    let win = self.windows.get_mut(&id);
+                    let win = win.expect("changed window was missing from the map");
+                    win.layout = update;
                 }
             }
             event => return Some(event),
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -447,6 +447,9 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
                     Event::WindowUrgencyChanged { id, urgent } => {
                         println!("Window {id}: urgency changed to {urgent}");
                     }
+                    Event::WindowLayoutsChanged { changes } => {
+                        println!("Window layouts changed: {changes:?}");
+                    }
                     Event::KeyboardLayoutsChanged { keyboard_layouts } => {
                         println!("Keyboard layouts changed: {keyboard_layouts:?}");
                     }
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -7,7 +7,7 @@ use niri_config::OutputName;
 use niri_ipc::socket::Socket;
 use niri_ipc::{
     Event, KeyboardLayouts, LogicalOutput, Mode, Output, OutputConfigChanged, Overview, Request,
-    Response, Transform, Window,
+    Response, Transform, Window, WindowLayout,
 };
 use serde_json::json;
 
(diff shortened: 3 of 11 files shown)


yrkv:
Also, here's a gif of a prototype [eww](https://github.com/elkowar/eww) widget that uses this functionality, to demonstrate what it can be used for.

niri-ipc/src/lib.rs
```diff
@@ -996,6 +996,19 @@ pub struct Window {
...
+    pub tile_coordinates: Option<(usize, usize)>,
+    /// Size of the tile this window is in, if applicable.
+    ///
+    /// Unset for floating windows and for windows being dragged with a mouse.
+    pub tile_size: Option<(f64, f64)>,
+    /// Location of the tile relative to the top-left corner of the workspace scrollbox itself.
+    /// [x,y,width,height].
+    ///
+    /// This is NOT set within niri state, but gets computed client-side within `niri msg windows`
+    pub geometry: Option<[f64; 4]>,
```
YaLTeR, line 1011: I don't think this should be included; the clients can compute on their own. We can provide helpers and docs. (Bars will want to connect to the socket directly rather than niri msg event-stream anyway)

niri-ipc/src/lib.rs
```diff
@@ -996,6 +996,19 @@ pub struct Window {
     ///
     /// If the window isn't floating then it is in the tiling layout.
     pub is_floating: bool,
+    /// Location of the window within a workspace in terms of (column index, tile index in column).
+    ///
+    /// Unset for floating windows.
+    pub tile_coordinates: Option<(usize, usize)>,
```
YaLTeR, line 1002:
```suggestion
    pub tile_pos_in_scrolling_layout: Option<(usize, usize)>,
```
Idk? Kind of a mouthful still but I want to communicate this.

niri-ipc/src/lib.rs
```diff
@@ -996,6 +996,19 @@ pub struct Window {
...
+    /// Size of the tile this window is in, if applicable.
+    ///
+    /// Unset for floating windows and for windows being dragged with a mouse.
```
YaLTeR, line 1005: I don't see any problems with setting this for those windows?

niri-ipc/src/lib.rs
```diff
@@ -1156,6 +1169,26 @@ pub enum Event {
...
         id: Option<u64>,
     },
+    /// Resize multiple windows at once. Some operations can cause many resizes at once, and they
+    /// are grouped together to avoid spamming events.
+    ///
+    /// Example: resizing a window in a column with many windows resizes each window.
+    BatchResizeTiles {
+        /// Pairs consisting of an id and the new size of each window.
+        id_size_pairs: Vec<(u64, (f64, f64))>,
+    },
```
YaLTeR, line 1179:
```
WindowPositionsAndSizesChanged {
    changes: Vec<(u64, PositionAndSize)>,
}
```
I think like this, and put all position and size related properties into a new struct, maybe? Also inline that struct into Window instead of individual fields. What to call it though? Geometry comes to mind, but it's also a Wayland-specific different concept so...

niri-ipc/src/lib.rs
```diff
@@ -1156,6 +1169,26 @@ pub enum Event {
...
+    /// One or more windows' colums were changed together. For example, opening a window to the
+    /// left of a group of windows would cause all of them to shift over by one column.
+    ///
+    /// NOTE: Do not emit this in conjunction with setting the value any other way,
+    ///       as it could result in an invalid column value if applied on top of an updated value.
+    BatchShiftColumns {
```
YaLTeR, line 1185: This is not needed with change like above

niri-ipc/src/lib.rs
```diff
@@ -996,6 +996,19 @@ pub struct Window {
...
     /// If the window isn't floating then it is in the tiling layout.
     pub is_floating: bool,
+    /// Location of the window within a workspace in terms of (column index, tile index in column).
+    ///
+    /// Unset for floating windows.
+    pub tile_coordinates: Option<(usize, usize)>,
+    /// Size of the tile this window is in, if applicable.
+    ///
+    /// Unset for floating windows and for windows being dragged with a mouse.
+    pub tile_size: Option<(f64, f64)>,
```
YaLTeR, line 1006: Then:
- `tile_pos_in_workspace_view: Option<(f64, f64)>` set for floating windows ("workspace view" is also used for gradients relative-to in the config)
- `window_pos_in_workspace_view: Option<(f64, f64)>` same but windows themselves (can omit for now)
- `window_size: Option<(i32, i32)>` on Wayland windows themselves can only be integer sized

niri-ipc/src/lib.rs
yrkv: I agree with putting that in helpers, so I removed it completely :+1:

niri-ipc/src/lib.rs
yrkv: It wasn't the most convenient at the time but I've since added it.

niri-ipc/src/lib.rs
yrkv: I'll add the rest of the changes to the Window location state once everything else is confirmed

YaLTeR:
also gonna cc @calops for feedback on this

niri-ipc/src/state.rs
```diff
@@ -132,6 +132,14 @@ impl EventStreamStatePart for WindowsState {
             Event::WindowsChanged { windows } => {
                 self.windows = windows.into_iter().map(|win| (win.id, win)).collect();
             }
+            Event::WindowsLocationsChanged { changes } => {
+                for (id, update) in changes {
+                    let Some(w) = self.windows.get_mut(&id) else {
+                        continue; // should never be reached
+                    };
```
YaLTeR, line 139: This should just .expect() it since niri should emit events properly.

niri-ipc/src/lib.rs
```diff
@@ -1038,6 +1038,20 @@ pub struct Window {
     ///
     /// If the window isn't floating then it is in the tiling layout.
     pub is_floating: bool,
+    /// Position and size related properties of the Window.
+    pub location: WindowLocation,
```
YaLTeR, line 1042: Yeah as much as I have concerns about Geometry, I think it works better than Location for this. Let's go with Geometry

src/ipc/server.rs
```diff
@@ -622,6 +630,36 @@ impl State {
         let mut events = Vec::new();
         let layout = &self.niri.layout;
 
+        // Gather position/size information ahead of time, since we have to traverse
+        // the entire layout to get tile cell positions.
+        let mut window_locations = HashMap::new();
+        for (_, _, ws) in layout.workspaces() {
```
YaLTeR, line 636: This shouldn't be a separate loop from the one below. This function is called every single time niri does anything, so two loops instead of one is too costly. Perhaps convert the loop below to go over workspaces() instead?

YaLTeR (commented):
Yeah, this overall structure looks good. Let's add the other stuff and see how it looks.

src/ipc/server.rs
yrkv: Maybe it would be best to make a variant of `layout.with_windows` that provides the window/tile geometry information needed? Alternatively, it could just be added into `with_windows`...

I mostly implemented it that way to minimize changes to anything which may affect anything else. I don't immediately see that two loops over the windows is definitely too costly without seeing some benchmark showing it, but I agree it should be merged into one regardless for code quality reasons.

niri-ipc/src/lib.rs
yrkv: Sgtm.

In case you want to see the other options I considered:
I spent a while perusing synonyms, so here's all the ones I thought were viable: WindowLocation, WindowGeometry, WindowArea, WindowLayoutInfo, WindowArrangement. Could also add "Info" to location or geometry to indicate its purpose clearer.

YaLTeR: Hmm, out of these WindowLayout also sounds good (I'd like to avoid a third word). Not sure which one I like more been Geometry and Layout

src/ipc/server.rs
yegor-hyron: I'd like to get some more discussion on this, I'm seeing two reasonable ways to do this and I don't quite like either, so I'd like to hear what you prefer.
1. Something similar to `Layout::with_windows`, but it provides the `WindowLayout` to the provided closure too.
2. Replace the `layout.with_windows` call with a loop over workspaces as you mentioned. The catch is that we also have to consider the ongoing interactive move. The nicest way to do so that I can see so far is to make a closure same as before, but just use it locally and not pass it to `layout.with_windows`. I'm not the biggest fan of this pattern, but it should work fine.

I'm currently leaning towards (2), and to handle the ongoing interactive move I'm restructuring it like so:

```rust
let mut f = |mapped: &Mapped,
             window_layout: WindowLayout,
             ws_id: Option<WorkspaceId>| {
    // OMITTED: essentially the same closure that was previously given to `layout.with_windows`
};

if let Some(tile) = layout.current_interactive_move_tile() {
    f(tile.window(), WindowLayout {/* omitted */}, None);
}
for (_, _, ws) in layout.workspaces() {
    for (tile, cell) in ws.tiles_with_workspace_positions() {
        f(tile.window(), WindowLayout {/* omitted */}, Some(ws.id()));
    }
}
```

Basically, my question is if you want that bit of structure to be in `State::ipc_refresh_windows` or somewhere in `Layout`. (i.e. make a new fn `Layout::with_windows_and_window_layouts` or something like that)

yrkv: I'd like to get some more discussion on this, I'm seeing two reasonable ways to do this and I don't quite like either, so I'd like to hear what you prefer.
1. Something similar to `Layout::with_windows`, but it provides the `WindowLayout` to the provided closure too.
2. Replace the `layout.with_windows` call with a loop over workspaces as you mentioned. The catch is that we also have to consider the ongoing interactive move. The nicest way to do so that I can see so far is to make a closure same as before, but just use it locally and not pass it to `layout.with_windows`. I'm not the biggest fan of this pattern, but it should work fine.

I'm currently leaning towards (2), and to handle the ongoing interactive move I'm restructuring it like so:

```rust
let mut f = |mapped: &Mapped,
             window_layout: WindowLayout,
             ws_id: Option<WorkspaceId>| {
    // OMITTED: essentially the same closure that was previously given to `layout.with_windows`
};

if let Some(tile) = layout.current_interactive_move_tile() {
    f(tile.window(), WindowLayout {/* omitted */}, None);
}
for (_, _, ws) in layout.workspaces() {
    for (tile, cell) in ws.tiles_with_workspace_positions() {
        f(tile.window(), WindowLayout {/* omitted */}, Some(ws.id()));
    }
}
```

Basically, my question is if you want that bit of structure to be in `State::ipc_refresh_windows` or somewhere in `Layout`. (i.e. make a new fn `Layout::with_windows_and_window_layouts` or something like that)

YaLTeR: Actually I think `Layout::for_each_window_with_window_layout` would make sense. The structure will need to be in niri-ipc either way, but maybe Layout will need its own copy (or maybe not).

If it doesn't add much overhead to fetch the layout, then perhaps just add it to `with_windows()` directly?

druskus20:
I was able to use this PR to make a workspace / column widget - thanks for the work!

yrkv:
Finally got around to continuing this, sorry it took a while.

A few small miscellaneous comments:
- Should I use `tile.window_size()` or `tile.window().size()` for `window_size` in `WindowLayout`? Note that window and tile size are equal if border is disabled, effectively just putting unnecessary stuff in it.
- Note that `window_size` has been made non-optional, same as `tile_size`.
- I'm not 100% sure I understood what you meant with `*_pos_in_workspace_view` -- should those be set for "interactive move" windows, or just floating ones?

YaLTeR:
> sorry it took a while

no worries, take your time

> Should I use tile.window_size() or tile.window().size() for window_size in WindowLayout? Note that window and tile size are equal if border is disabled, effectively just putting unnecessary stuff in it.

For window_size you should use `tile.window().size()`, the raw `i32` in Logical space. This is what comes directly from Wayland from the window, which is why it is useful in the IPC.

> I'm not 100% sure I understood what you meant with *_pos_in_workspace_view -- should those be set for "interactive move" windows, or just floating ones?

With the overview, interactively moved window won't necessarily have a well defined pos in workspace view, so only floating windows.

Nitepone:
Hey, been working on a niri/taskbar waybar module that consumes this! Have an ask/suggestion.

Resizing windows using the mouse spits out a large number of IPC events where the window size is changing. In consumers that only care about window ordering (like a taskbar), these need to be filtered out. Consumers could each implement tracking changes in the fields they care about, but it could be nice to have an indication of what type of change has occurred to the layout from the IPC call.

As a suggested implementation, a new enum of ChangeType with variants like "Resize", "Move". Or perhaps only populating fields of `WindowLayoutsChanged::changes::WindowLayout` that have been changed?

Weathercold:
With this feature, using a wallpaper daemon that supports background translation like swww, we can theoretically write an external script to enable [background parallax](https://github.com/YaLTeR/niri/discussions/988)

cptpiepmatz:
How is it going?

Nitepone:
Would really love to get this merged. I have a bunch of free time right now, would be happy to pick up this MR to resolve any outstanding issues preventing merge!

cptpiepmatz:
> Would really love to get this merged. I have a bunch of free time right now, would be happy to pick up this MR to resolve any outstanding issues preventing merge!

Maybe you should just do it, @yrkv seems very inactive on Github

yrkv:
@YaLTeR What's the current state of this look like for you? I think I didn't really clarify clearly that the last commit applied all the changes (afaik) that were discussed

though it's a bit outdated relative to current commit now

YaLTeR:
Oh you did? I gotta take another look then

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
     ///
     /// If the window isn't floating then it is in the tiling layout.
     pub is_floating: bool,
+    /// Position and size related properties of the Window.
```
```suggestion
    /// Position- and size-related properties of the window.
```

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
...
+}
+
+/// Position and size related properties of a Window
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct WindowLayout {
+    /// Location of the window within a workspace in terms of (column index, tile index in column).
+    ///
+    /// Unset for floating windows.
+    pub tile_pos_in_scrolling_layout: Option<(usize, usize)>,
```
```suggestion
    pub pos_in_scrolling_layout: Option<(usize, usize)>,
```

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
     ///
     /// If the window isn't floating then it is in the tiling layout.
     pub is_floating: bool,
+    /// Position and size related properties of the Window.
+    pub location: WindowLayout,
+}
+
+/// Position and size related properties of a Window
```
```suggestion
/// Position- and size-related properties of a [`Window`].
///
/// Optional properties will be unset for some windows, do not rely on them always being present.
```

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct WindowLayout {
+    /// Location of the window within a workspace in terms of (column index, tile index in column).
```
```suggestion
    /// Location of a tiled window within a workspace in terms of (column index, tile index in column).
```

niri-ipc/src/lib.rs
YaLTeR: Might want to extract this into a struct for future extensibility.

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
...
+    ///
+    /// Unset for floating windows.
+    pub tile_pos_in_scrolling_layout: Option<(usize, usize)>,
+    /// Size of the tile this window is in.
```
```suggestion
    /// Size of the tile this window is in, including decorations like borders.
```

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
...
+    pub tile_size: (f64, f64),
+    /// Size of the window itself.
+    /// 
+    /// Note that Wayland windows can only be integer sized.
+    pub window_size: (i32, i32),
+    /// set for floating windows ("workspace view" is also used for gradients relative-to in the config)
```
```suggestion
    /// Tile position within the current view of the workspace.
    ///
    /// This is the same "workspace view" as in gradients' `relative-to` in the niri config.
```

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
...
+    pub tile_pos_in_workspace_view: Option<(f64, f64)>,
+    /// same but windows themselves
```
```suggestion
    /// Window position within the current view of the workspace.
    ///
    /// Unlike `tile_pos_in_workspace_view`, this is the position of the window contents, excluding borders and other decorations. You probably want to use `tile_pos_in_workspace_view` instead, since borders are a part of the window, as far as the user is concerned.
```

niri-ipc/src/lib.rs
```diff
@@ -1232,6 +1254,13 @@ pub enum Event {
         /// Id of the newly focused window, or `None` if no window is now focused.
         id: Option<u64>,
     },
+    /// Apply changes to the tile location and/or size of one or more tiles/windows.
```
```suggestion
    /// The layout of one or more windows has changed.
```

niri-ipc/src/lib.rs
```diff
@@ -1232,6 +1254,13 @@ pub enum Event {
...
+    ///
+    /// Note that this does not trigger for a window's physical location changing.
```
```suggestion
```
Not sure what this meant

niri-ipc/src/lib.rs
```diff
@@ -1232,6 +1254,13 @@ pub enum Event {
...
+    WindowLayoutsChanged {
+        /// Pairs consisting of a window id and new position/size information for the window.
```
```suggestion
        /// Pairs consisting of a window id and new layout information for the window.
```

niri-ipc/src/state.rs
```diff
@@ -132,6 +132,15 @@ impl EventStreamStatePart for WindowsState {
             Event::WindowsChanged { windows } => {
                 self.windows = windows.into_iter().map(|win| (win.id, win)).collect();
             }
+            Event::WindowLayoutsChanged { changes } => {
+                for (id, update) in changes {
+                    let w = self
+                        .windows
+                        .get_mut(&id)
+                        .expect("altered window missing from map");
```
YaLTeR:
```suggestion
                        .expect("changed window was missing from the map");
```

src/ipc/client.rs
```diff
@@ -429,6 +429,9 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
                     Event::WindowFocusChanged { id } => {
                         println!("Window focus changed: {id:?}");
                     }
+                    Event::WindowLayoutsChanged { changes } => {
+                        println!("Windows locations changed: {changes:?}");
```
YaLTeR:
```suggestion
                        println!("Window layouts changed: {changes:?}");
```

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,28 @@ pub struct Window {
...
+    pub location: WindowLayout,
+}
+
+/// Position and size related properties of a Window
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct WindowLayout {
+    /// Location of the window within a workspace in terms of (column index, tile index in column).
+    ///
+    /// Unset for floating windows.
```
```suggestion
```

src/layout/workspace.rs
```diff
@@ -1367,6 +1367,30 @@ impl<W: LayoutElement> Workspace<W> {
...
+            window_pos_in_workspace_view: None,
+        }));
+
+        let floating = self.floating.tiles_with_render_positions();
+        let floating = floating.map(move |(tile, pos)| (tile, WindowLayout {
+            tile_pos_in_scrolling_layout: None,
+            tile_size: tile.tile_size().into(),
+            window_size: tile.window().size().into(),
+            tile_pos_in_workspace_view: Some(pos.into()),
+            window_pos_in_workspace_view: Some((pos + tile.window_loc()).into()),
```
YaLTeR: This makes me think. Maybe it's a better idea to expose `window_loc`? It doesn't change much, and it can be exposed once instead of duplicating `tile_pos_in_...` and `window_pos_in_...` (we may get more of those positions later on).

YaLTeR (changes requested):
So, after reviewing the quickshell protocol proposals, I realized that something is missing here: window stacking order/z index. Not entirely sure how to best expose this. Ideally it should also avoid spam.

In the scrolling layout, there technically is stacking order, but since it's mostly invisible, it changes somewhat erratically. Perhaps however we expose the stacking order needn't include the scrolling layout windows?

src/layout/workspace.rs
yrkv: I was thinking somewhat similar -- it feels like information is being reconstructed when it may be better to expose it

yrkv: How exactly do you think this should be exposed?

YaLTeR: Hmm, maybe `window_geometry_offset` or `window_offset` or `window_offset_from_tile`. Value-wise, an (f64, f64) equal to `tile.window_loc()`.

YaLTeR:
For window stacking order, I don't have a very clear idea, and it should probably be done in a separate PR. One idea I had is to have the stacking order on a workspace as a `Vec<WindowId>`, then have some kind of `WorkspaceWindowsRestacked { ws_id: WorkspaceId, new_order: Vec<WindowId> }`. Any windows missing from that should be considered below, or something.

niri-ipc/src/lib.rs
```diff
@@ -1072,6 +1072,32 @@ pub struct Window {
...
+/// Position- and size-related properties of a [`Window`].
+///
+/// Optional properties will be unset for some windows, do not rely on them always being present.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct WindowLayout {
+    /// Location of a tiled window within a workspace in terms of (column index, tile index in column).
+    pub pos_in_scrolling_layout: Option<(usize, usize)>,
+    /// Size of the tile this window is in, including decorations like borders.
+    pub tile_size: (f64, f64),
```
atagen, line 1188: is there any scope here for an enum over the preset window widths ? ie. PresetWidth(usize) for indexing into user provided widths and Free(f64) for manually set widths

agausmann: Reasonable, but I'd be concerned that making an enum would make the effective value of `tile_size` more difficult to get. It may be more useful for the recipient if `tile_size` stays the way it is. In that case the preset index  that generated that size (if any) could be a separate field - `Option<usize>`. Or, we could still use an enum to be able to support other "sources" of tile size in the future - just in a separate field:

```rust
#[non_exhaustive]
enum DimensionSource {
    Free,
    Preset(usize),
}
```

Also note that heights can have preset sizes (the default binding is mod+shift+R)

YaLTeR: Can be added later as a separate field if needed

alicealysia:
What's outstanding to take this from a draft into a full PR?

I'm quite tempted to tackle it myself.

logantgt:
Commenting to show support. I wrote a small workspace/window indicator widget in Quickshell using @yrkv's fork to retrieve the window layout and it worked great. Having some ability to calculate what windows have their bounds completely within the display using some kind of geometry information would also be very useful.

YaLTeR:
> What's outstanding to take this from a draft into a full PR?

More or less addressing my open comments. You can do that and open a separate PR I suppose

YaLTeR:
Alright let me take a look at this

YaLTeR:
Cleaned things up, should be good to go. More stuff can be added later.

Anyone wants to help test?

YaLTeR:
Did a quick test with quickshell for floating windows, seems to work

https://github.com/user-attachments/assets/c88fdae1-23a5-46eb-966c-a29b8919bcad

YaLTeR:
Alright let's yolo it

YaLTeR:
Thanks @calops and @yrkv!

niri-ipc/src/lib.rs
yangyingchao: > I don't think this should be included; the clients can compute on their own. We can provide helpers and docs. (Bars will want to connect to the socket directly rather than niri msg event-stream anyway)

Hey @YaLTeR @yrkv , where can I find the related docs? I need to get geometry of focused window in my script. Thanks.

PR #1266, opened 2025-03-14
wiki update

chinatsu:
this change updates the wiki with

- more links leading to relevant sections in other files
  - unfortunately, this can't be done with regular `.md` ending links in GitHub-wikis :(
- development-section pages renamed to `Development:-${title}.md` to match configuration pages somewhat
- ~~restructured sidebar to have configuration pages higher up~~
  - plenty of other changes were made in the sidebar, and i'm not sure if i'm happy with those change yet. the order of elements don't feel super "correct"
  - maybe you don't agree w/ this change at all lol; can probably revert this :)

a generated version can be perused here https://github.com/chinatsu/niri/wiki


wiki/_Sidebar.md
```diff
@@ -1,35 +1,32 @@
...
-* [Developing niri](./Developing-niri.md)
-* [Fractional Layout](./Fractional-Layout.md)
-* [Redraw Loop](./Redraw-Loop.md)
-* [Animation Timing](./Animation-Timing.md)
+* Development
+    * [Design Principles](./Development:-Design-Principles.md)
+    * [Developing niri](./Development:-Developing-niri.md)
+    * [Fractional Layout](./Development:-Fractional-Layout.md)
+    * [Redraw Loop](./Development:-Redraw-Loop.md)
+    * [Animation Timing](./Development:-Animation-Timing.md)
```
YaLTeR, line 32: I kinda prefer the layout with the headers more?

chinatsu: i prefer headers too, but my desire to bring configuration up near the top breaks up the content a little too much. if you don't care about the configuration-part being high up, i can revert this change.

YaLTeR: I don't think it's that important yes

chinatsu:
> The reason I link .md without anchors is so that the links work both when you browse the wiki in the repository, and on the wiki page. Our GitHub action to publish the wiki strips .md extension automatically. Unfortunately, it cannot deal with anchors, which is why I never use anchors. Maybe it's been updated to support that, or there's a different action that does it?

ugh, you're right. these changes broke the "browsability" when clicking links in the repository. it appears the action we use hasn't been updated in a fairly long time, and we're on the latest commit.

i'll close the PR, nothing of particular value will be added once we restore the sidebar and removing anchor links, haha

chinatsu:
i'm reopening this, turned it into a draft. awaiting this PR to be approved before i call it ready: https://github.com/Andrew-Chen-Wang/github-wiki-action/pull/79

the action author seems to be fairly active on github, so i can't imagine it taking too long to get through. :)

YaLTeR:
I wouldn't mind too much depending on an exact commit in a fork fwiw

wiki/Configuration:-Layer-Rules.md
```diff
@@ -79,7 +79,7 @@ They will be replaced with solid black rectangles.
 This can be useful for notifications.
 
 The same caveats and instructions apply as for the `block-out-from` window rule.
-Please read the `block-out-from` section in the [window rules](./Configuration:-Window-Rules.md) wiki page for more details.
+Please read the `block-out-from` section in the [window rules](./Configuration:-Window-Rules.md#block-out-from) wiki page for more details.
```
YaLTeR, line 82: With working anchors, you can move the link to "`block-out-from` section". The only reason it was on "window rules" is the missing anchor.

Similarly in other places.

chinatsu: I'm out and about right now but I'll go through and fix some stuff like this later!

YaLTeR:
Thanks!

You know what else would be cool, to add a test that checks that all links + anchors are valid. We already have a test that checks all config snippets, could be something similar.

PR #1269, opened 2025-03-14
Support nested environment variable interpolation

deurzen:
Currently, the configured `environment` doesn't expand variables that are nested inside of the values of other variables. For example, the following configuration:

```kdl
environment {
    XDG_CONFIG_HOME "$HOME/.config"
    NIRI_CONFIG "$XDG_CONFIG_HOME/niri"
    PATH "$HOME/bin:$PATH"
}
```

would assign the strings as is to the variables, leaving any nested variables intact.

This PR adds support for variable interpolation, resolving previously defined variables to their values (and undefined variables to the empty string), similar to how a shell would do it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2243,6 +2243,7 @@ dependencies = [
  "log",
  "niri-config",
  "niri-ipc",
+ "once_cell",
  "ordered-float",
  "pango",
  "pangocairo",
@@ -2253,6 +2254,7 @@ dependencies = [
  "proptest",
  "proptest-derive",
  "rayon",
+ "regex",
  "sd-notify",
  "serde",
  "serde_json",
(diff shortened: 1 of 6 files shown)


YaLTeR:
Uhh, I'm not sure I want this. I feel that this is overcomplicating it. Definitely don't want anything involving the words "how a shell would do it" because that is a *massive* can of worms.

deurzen:
> Uhh, I'm not sure I want this. I feel that this is overcomplicating it. Definitely don't want anything involving the words "how a shell would do it" because that is a _massive_ can of worms.

Fair enough. Motivation is that configuring session-wide environment variables differs between distributions and init systems, and is generally a mess. Without variable expansion, some of the portability benefits to having compositor-level environment configuration are lost (no `$USER`, no `$HOME`).

Just intended to illustrate that it's _similar_ to how a shell expands variables for clarity; definitely not implying there should be any overflow from the shell world here. The same syntax is used all over the place (e.g., for defining environments in systemd, Docker, Kubernetes, CI/CD configurations, to name a few).

YaLTeR:
FWIW the niri environment block is not a good place for *session-wide* environment configuration because it's not imported into the session manager (systemd or dinit), so e.g. D-Bus-activated services won't get those variables. Instead, you can put them in something like .bashrc (which niri-session takes care to source and import environment from).

PR #1271, opened 2025-03-15
Dynamic screencast target

YaLTeR:
Implements #1081. Shows a special "window" in the portal dialog:

image

It starts out as a clear (transparent) stream. Then, you can use the new actions to change the cast target:

- `set-dynamic-cast-window` to use the focused window, or by id with `--id`.
- `set-dynamic-cast-monitor` to use the focused monitor, or by name.
- `clear-dynamic-cast-target` to go back to transparent.

You also use niri msg for things like:
```sh
niri msg action set-dynamic-cast-window --id $(niri msg --json pick-window | jq .id)
```

https://github.com/user-attachments/assets/12ad94fb-06e4-4dfc-9d47-2d52a5c4f3db

If the target disappears (e.g. window closes), it also goes back to transparent.

All dynamic casts share the same target, but new ones start out clear until the next time you change it (to avoid surprises and sharing sensitive info).

cc @elkowar who suggested, also everyone interested please test.

Some feedback I'm looking for:
- Are the action names good or maybe we can phrase them better?
- Is the fake window name "niri Dynamic Cast Target" good or can you suggest something better?
- Is it sufficient to show it in the window selector? We could also show it as a fake monitor in the monitor selector, but I think it shouldn't be necessary. It only increases clutter. Also, with windows there's at least some expectation that the stream size can change at runtime; with monitors I'm afraid there's some bug potential in client apps. Though, so far I don't think I've seen an app that would handle it wrong. Monitor casts also resize at runtime when changing the resolution so it's not like this is completely new behavior.


phisch:
Couldn't every screencast be from a dynamic cast target, but just pre-select the actual window/monitor selected in the screencast dialog? That way you could start any screencast and change what is being cast dynamically at all times.

YaLTeR:
It could, but I'm not sure this is desirable.
1. Static casts have their place. You may want to start up OBS and get your primary monitor every time.
2. If every cast is dynamic, what's the UI for picking which one to edit? In this PR, there's only one global dynamic cast target, so there's a simple set of actions. With multiple dynamic casts, you'd have to introduce IDs, and then mess with IPC to choose which one to edit, etc.

phisch:
Fair, I didn't think of multiple screencasts like that :thinking: 

The only good way would be to require a screencast identifier for the commands, unless there is only one dynamic screencast going. I think I would actually like that.

Re-opening OBS would then start screen-casting the initially selected target again, right?

So whenever you start your OBS, your setup is as expected, but you gain the ability to change the target as you want.

YaLTeR:
This could be an extension of this PR. So you have the explicit dynamic target modified by global commands, but you can also pass any "screencast ID" to modify that screencast.

That's more work though to expose all of this, can be done at a later point.

phisch:
I agree. So the interesting parts for a separate PR would be:
- make all screen-casts dynamic
- make dynamic screen-cast commands take a screen-cast-id argument
- make screen-cast-id argument optional if there is only **one single** screen-cast

YaLTeR:
Regarding the name: should it be dynamic cast *target* or *source*? I just realized that both kind of make sense.

- The Portal API calls it "source".
- The xdp-gnome dialog avoids this terminology altogether.
- The OBS plus button's tooltip is "Add Source".
- Firefox, Chromium, Telegram, Discord selectors also avoid the terminology.
- In the niri code I happened to call these *targets*, which is probably why I went for dynamic cast *target* instinctively.
- I also called it *target* in our `is-window-cast-target` window rule, and that's been released, so no going back on it.

I guess considering the last point, we should be consistent and keep calling it *target*?

PR #1279, opened 2025-03-16
screenshot: make selection area modifiable via keybinds

### LINKED ISSUE #980: Make floating move & resize work on the screenshot UI area
Now that we have floating window movement, the same movement could naturally apply to the screenshot UI area. 

Though, need to figure out how to best do it input-wise, since the screenshot UI intercepts the keys early. Should probably first check if it's a screenshot UI bind, then if not, run the usual search against configured binds, and if it matches one of the move/resize actions, apply it to the screenshot UI area.

`niri msg action` ones won't apply to the screenshot UI because they apply to the underlying layout even when the screenshot UI is up.

Idea by @FreeFull

nnyyxxxx:
resolves #980

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -7,7 +7,7 @@ use std::time::Duration;
 use calloop::timer::{TimeoutAction, Timer};
 use input::event::gesture::GestureEventCoordinates as _;
 use niri_config::{Action, Bind, Binds, Key, ModKey, Modifiers, SwitchBinds, Trigger};
-use niri_ipc::{LayoutSwitchTarget, SizeChange};
+use niri_ipc::LayoutSwitchTarget;
 use smithay::backend::input::{
     AbsolutePositionEvent, Axis, AxisSource, ButtonState, Device, DeviceCapability, Event,
     GestureBeginEvent, GestureEndEvent, GesturePinchUpdateEvent as _, GestureSwipeUpdateEvent as _,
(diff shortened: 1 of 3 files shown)


niri-config/src/lib.rs
```diff
@@ -1458,6 +1458,18 @@ pub enum Action {
...
     #[knuffel(skip)]
     ScreenshotTogglePointer,
+    MoveScreenshotLeft,
+    MoveScreenshotRight,
+    MoveScreenshotUp,
+    MoveScreenshotDown,
+    ResizeScreenshotLeft,
+    ResizeScreenshotRight,
+    ResizeScreenshotUp,
+    ResizeScreenshotDown,
```
YaLTeR, line 1468: So my idea for this was to *not* add new binds, but instead make the existing `MoveColumn/WindowLeft/Right/Up/Down` and Resize binds move the screenshot area when the screenshot UI is open, exactly as they would move a floating window.

nnyyxxxx: rebased with this idea

YaLTeR:
Thanks.

Your original implementation had some problems:
- moving outside the boundaries would reset the selection to the default rather than clamping
- resizing ignored the change and instead used a hardcoded value
- moving and resizing didn't take output scale into account

So I reimplemented it to do all of this.

Now, having tried it, the next logical steps are:
- support switch-preset-window-width/height
  - basically do the same as layout/floating.rs. Will be a bit annoying to keep track of preset widths/heights so that they're the same as in the layout Options
- support move-window-to-monitor-right/left/up/down to move the selection across monitors

PR #1282, opened 2025-03-16
ci/wiki: check that (local) links are well formed

chinatsu:
this change introduces a job that `publish-wiki` also depends upon (besides `build`), using [lychee](https://lychee.cli.rs/) v0.16.1 (specifically before the introduction of https://github.com/lycheeverse/lychee/issues/1574) to check that the links in our wiki are well formed and lead to places that exist.

this `check-links` step has thus far helped me find a few more trouble spots, which are also included in this PR.
the args provided to lychee ensures that only local/relative links are checked, as well as fragments/anchors.

one could consider removing `--offline`, but i ended up running into timeouts (though `--timeout n` (with a default of 20) is a valid flag that could be used to maybe mitigate some of those).

---

i've done some rudimentary tests in my fork,

- links to non-existing pages = failure
- links to non-existing anchors = failure

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,7 +229,6 @@ jobs:
         continue-on-error: true
   check-links:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-    needs: build
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,7 +229,6 @@ jobs:
         continue-on-error: true
 
   check-links:
-    needs: build
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
@@ -241,7 +240,9 @@ jobs:
 
   publish-wiki:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
-    needs: check-links
+    needs:
+      - build
+      - check-links
     permissions:
       contents: write
     runs-on: ubuntu-24.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,8 +227,8 @@ jobs:
 
       - run: nix flake check
         continue-on-error: true
+
   check-links:
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     needs: build
     runs-on: ubuntu-24.04
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,7 +234,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           show-progress: false
-      - uses: lycheeverse/lychee-action@v2.0.2 # don't bump until this is fixed: https://github.com/lycheeverse/lychee/issues/1574
+      - uses: lycheeverse/lychee-action@v2.0.2 # later versions break fragment checks. don't bump until this is fixed: https://github.com/lycheeverse/lychee/issues/1574
         with:
           args: --offline --include-fragments 'wiki/*.md'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,6 +229,7 @@ jobs:
         continue-on-error: true
   check-links:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    needs: build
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -228,9 +228,21 @@ jobs:
       - run: nix flake check
         continue-on-error: true
 
+  check-links:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@f87f0a62993c2647717456af92593666acb3a500 # lychee v0.16.1 seems to work well for our case at the moment
```
YaLTeR, line 237: Could you comment with what exactly is broken and link to issue? Also is there maybe a working tag rather than a commit hash?

.github/workflows/ci.yml
```diff
@@ -228,9 +228,21 @@ jobs:
       - run: nix flake check
         continue-on-error: true
 
+  check-links:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Link Checker
+        id: lychee
```
YaLTeR, line 236: Is this id needed? I'd remove both name and id

.github/workflows/ci.yml
```diff
@@ -228,9 +228,21 @@ jobs:
       - run: nix flake check
         continue-on-error: true
 
+  check-links:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
```
YaLTeR, line 234: With show progress false

wiki/_Sidebar.md
```diff
@@ -5,7 +5,7 @@
 * [Floating Windows](./Floating-Windows.md)
 * [Tabs](./Tabs.md)
 * [Screencasting](./Screencasting.md)
-* [Layer‐Shell Components](./Layer%E2%80%90Shell-Components.md)
+* [Layer‐Shell Components](./Layer‐Shell-Components.md)
```
YaLTeR, line 8: Does this work fine? Pretty sure I copied that URL from GitHub itself or something

chinatsu: yeah, it works fine! tbh, i think lychee might not care about this one, https://github.com/becheran/mlc did though.

chinatsu: browsers do prefer turning special characters into %55%52%4C%20%65%6E%63%6F%64%69%6E%67 (tl note: url encoding), but special characters work in links anyway.

chinatsu: i gave it a try with lychee locally, and the special character can remain special or url encoded. i figure you prefer it to be "safe", so i reverted it to the url encoded version

YaLTeR:
Thanks

chinatsu:
also, i want to clear up the feasibility to don't use `--offline`. running lychee locally returns these errors at the moment, which should clear up once freedesktop finishes their migration.

```
cn@squarewave ~/c/niri (main)> lychee --version
lychee 0.16.1
cn@squarewave ~/c/niri (main)> lychee --include-fragments 'wiki/*.md'
  175/175 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                            Issues found in 3 inputs. Find details below.

[wiki/Configuration:-Input.md]:
     [503] https://wayland.freedesktop.org/libinput/doc/latest/tapping.html#tap-and-drag
     [503] https://wayland.freedesktop.org/libinput/doc/latest/clickpad-softbuttons.html
     [503] https://wayland.freedesktop.org/libinput/doc/latest/device-configuration-via-udev.html

[wiki/Configuration:-Layout.md]:
     [503] https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/155

[wiki/Xwayland.md]:
     [503] https://gitlab.freedesktop.org/ofourdan/xwayland-run

🔍 175 Total (in 2s) ✅ 170 OK 🚫 5 Errors
```

it doesn't seem like we can allow failure for external urls, while failing on local ones, so perhaps it's best to stay offline and do a manual check for all links (external ones too) every now and then?

YaLTeR:
> it doesn't seem like we can allow failure for external urls, while failing on local ones, so perhaps it's best to stay offline and do a manual check for all links (external ones too) every now and then?

Yes, but in general I wouldn't want to ping all online links on every CI. Doing it manually is perfectly fine

PR #1302, opened 2025-03-19
wiki: Documented flags for Electron based applications

vanadium23:
Configured VsCode and other Electron apps (Obsidian) without FlatPak to run natively on Wayland, and niri as well.

### DIFF
--- a/wiki/Application-Issues.md
+++ b/wiki/Application-Issues.md
@@ -1,8 +1,27 @@
+### Electron based applications
+
+Electron based application can work directly on Wayland, but don't accept it without configuration. Configuration differs between versions. Infromation was grabbed from [ArchWiki](https://wiki.archlinux.org/title/Wayland#Electron).
+
+For Electron (>28) you need to set environment variable:
+```kdl
+environment {
+    ELECTRON_OZONE_PLATFORM_HINT "auto"
+}
+```
+
+For previous version you need to set flags in configuration file `~/.config/electron-flags.conf`:
+```
+--enable-features=UseOzonePlatform
+--ozone-platform-hint=auto
+```
+
 ### VSCode
 
-If you're having issues with some VSCode hotkeys, try starting `Xwayland` and setting the `DISPLAY=:0` environment variable for VSCode.
-That is, still running VSCode with the Wayland backend, but with `DISPLAY` set to a running Xwayland instance.
-Apparently, VSCode currently unconditionally queries the X server for a keymap.
+To run VSCode natively via Wayland you need to put flags in `~/.config/code-flags.conf`:
+```
+--enable-features=UseOzonePlatform
+--ozone-platform-hint=auto
+```
 
 ### WezTerm
 


wiki/Application-Issues.md
```diff
@@ -1,8 +1,27 @@
+### Electron based applications
+
+Electron based application can work directly on Wayland, but don't accept it without configuration. Configuration differs between versions. Infromation was grabbed from [ArchWiki](https://wiki.archlinux.org/title/Wayland#Electron).
```
chinatsu, line 3:
```suggestion
Electron based applications can work directly on Wayland, but they don't accept it without configuration. Configuration differs between versions. Information was grabbed from the [ArchWiki](https://wiki.archlinux.org/title/Wayland#Electron).
```

wiki/Application-Issues.md
```diff
@@ -1,8 +1,27 @@
...
+```
+--enable-features=UseOzonePlatform
+--ozone-platform-hint=auto
+```
+
 ### VSCode
 
-If you're having issues with some VSCode hotkeys, try starting `Xwayland` and setting the `DISPLAY=:0` environment variable for VSCode.
-That is, still running VSCode with the Wayland backend, but with `DISPLAY` set to a running Xwayland instance.
-Apparently, VSCode currently unconditionally queries the X server for a keymap.
```
YaLTeR, line 5: Why did you remove these lines? They apply to VSCode running natively on Wayland.

wiki/Application-Issues.md
```diff
@@ -1,8 +1,27 @@
+### Electron based applications
+
+Electron based applications can work directly on Wayland, but they don't accept it without configuration. Configuration differs between versions. Information was grabbed from the [ArchWiki](https://wiki.archlinux.org/title/Wayland#Electron).
+
+For Electron (>28) you need to set environment variable:
+```kdl
+environment {
+    ELECTRON_OZONE_PLATFORM_HINT "auto"
```
YaLTeR, line 8: Should this say "wayland" instead?

wiki/Application-Issues.md
```diff
@@ -1,8 +1,27 @@
...
+}
+```
+
+For previous version you need to set flags in configuration file `~/.config/electron-flags.conf`:
```
YaLTeR, line 12: This file is Arch-specific, so we can't document it here

wiki/Application-Issues.md
```diff
@@ -1,8 +1,27 @@
...
+--enable-features=UseOzonePlatform
+--ozone-platform-hint=auto
+```
+
 ### VSCode
 
-If you're having issues with some VSCode hotkeys, try starting `Xwayland` and setting the `DISPLAY=:0` environment variable for VSCode.
-That is, still running VSCode with the Wayland backend, but with `DISPLAY` set to a running Xwayland instance.
-Apparently, VSCode currently unconditionally queries the X server for a keymap.
+To run VSCode natively via Wayland you need to put flags in `~/.config/code-flags.conf`:
```
YaLTeR, line 20: I suspect this is also Arch-specific

YaLTeR (changes requested):

wiki/Application-Issues.md
vanadium23: because if you run VSCode via platform "wayland", then you don't need to start Xwayland and also no issues with hotkeys.

YaLTeR: I added this in response to someone literally having this exact issue when running VSCode on Wayland..

wiki/Application-Issues.md
vanadium23: Oh, I thought it is electron based, but you are right. I will remove those and just paste about command line flags.

wiki/Application-Issues.md
YaLTeR: I mean, maybe they fixed the problem since? But I wasn't affected by this so idk

wiki/Application-Issues.md
vanadium23: Auto first check wayland, then fallback to x11. I think it is good, if XWayland still present or niri used in combination with other WMs.
https://www.electronjs.org/docs/latest/api/environment-variables#electron_ozone_platform_hint-linux

YaLTeR: I see. Makes you wonder why it's still not the default

wiki/Application-Issues.md
vanadium23: Okay, I'll revert

wiki/Application-Issues.md
chinatsu: i believe so as well https://gitlab.archlinux.org/archlinux/packaging/packages/code/-/blob/main/code.sh?ref_type=heads#L5

wiki/Application-Issues.md
vanadium23: Reverted.

wiki/Application-Issues.md
vanadium23: Removed.

wiki/Application-Issues.md
vanadium23: Removed.

wiki/Application-Issues.md
phisch: Oh, I didn't know that electron apps respect this variable. I copied system `.desktop` files tom my users home to overwrite the `exec` commands, and append `--ozone-platform-hint=auto` instead. This is much better :heart:

wiki/Application-Issues.md
```diff
@@ -1,3 +1,22 @@
+### Electron based applications
```
YaLTeR, line 1:
```suggestion
### Electron applications
```

wiki/Application-Issues.md
```diff
@@ -1,3 +1,22 @@
...
+
+Electron based applications can work directly on Wayland, but they don't accept it without configuration. Configuration differs between versions. Information was grabbed from the [ArchWiki](https://wiki.archlinux.org/title/Wayland#Electron).
```
YaLTeR, line 3:
```suggestion
Electron-based applications can run directly on Wayland, but it's not the default.
```

wiki/Application-Issues.md
```diff
@@ -1,3 +1,22 @@
...
+
+For Electron (>28) you need to set environment variable:
```
YaLTeR, line 5:
```suggestion
For Electron > 28, you can set an environment variable:
```

wiki/Application-Issues.md
```diff
@@ -1,3 +1,22 @@
...
+```kdl
+environment {
+    ELECTRON_OZONE_PLATFORM_HINT "auto"
+}
+```
+
+For previous version you need to set flags via CLI:
```
YaLTeR, line 12:
```suggestion
For previous versions, you need to pass command-line flags to the target application:
```

wiki/Application-Issues.md
```diff
@@ -1,3 +1,22 @@
...
+```
+--enable-features=UseOzonePlatform
+--ozone-platform-hint=auto
```
YaLTeR, line 15:
```suggestion
--enable-features=UseOzonePlatform --ozone-platform-hint=auto
```

wiki/Application-Issues.md
```diff
@@ -1,3 +1,22 @@
...
+```
+
+If application has a [desktop entry](https://specifications.freedesktop.org/menu-spec/latest/menu-add-example.html), you can modify an `Exec` section.
```
YaLTeR, line 18:
```suggestion
If the application has a [desktop entry](https://specifications.freedesktop.org/menu-spec/latest/menu-add-example.html), you can put the command-line arguments into the `Exec` section.
```

PR #1319, opened 2025-03-21
window urgency

Duncaen:
quick draft of urgency support through serial-less xdg-activation. This PR adds an urgency boolean to windows and adds methods to set and get the urgency through the `LayoutElement` trait.
The next commit adds IPC events that indicate workspace urgency based on whether a window on the workspace has urgency, not sure if the IPC should also emit window events for that.

yambar patch:
```patch
From 90871daaac49fd87c29445b072ab7fc85d924eec Mon Sep 17 00:00:00 2001
From: Duncan Overbruck <mail@duncano.de>
Date: Fri, 21 Mar 2025 18:12:18 +0100
Subject: [PATCH] add urgency tag to niri-workspaces

---
 modules/niri-common.c     | 23 +++++++++++++++++++++++
 modules/niri-common.h     |  2 ++
 modules/niri-workspaces.c |  5 +++--
 3 files changed, 28 insertions(+), 2 deletions(-)

diff --git a/modules/niri-common.c b/modules/niri-common.c
index ac53921..f380d85 100644
--- a/modules/niri-common.c
+++ b/modules/niri-common.c
@@ -12,6 +12,7 @@
 #include <unistd.h>
 
 #include "../log.h"
+#include "json_object.h"
 #include "niri-common.h"
 
 #define LOG_MODULE "niri:common"
@@ -82,6 +83,7 @@ parser(char *response)
                 ws->active = json_object_get_boolean(json_object_object_get(ws_obj, "is_active"));
                 ws->focused = json_object_get_boolean(json_object_object_get(ws_obj, "is_focused"));
                 ws->empty = json_object_get_int(json_object_object_get(ws_obj, "active_window_id")) == 0;
+                ws->urgent = json_object_get_boolean(json_object_object_get(ws_obj, "is_urgent"));
 
                 char const *name = json_object_get_string(json_object_object_get(ws_obj, "name"));
                 if (name)
@@ -185,6 +187,27 @@ parser(char *response)
             events |= keyboard_layouts_switched;
         }
 
+        // "WorkspaceUrgencyChanged": {
+        //   "id": 1,
+        //   "urgent": true
+        // }
+        else if (strcmp(key, "WorkspaceUrgencyChanged") == 0) {
+            json_object *obj = json_object_iter_peek_value(&it);
+            int id = json_object_get_int(json_object_object_get(obj, "id"));
+            bool urgent = json_object_get_boolean(json_object_object_get(obj, "urgent"));
+
+            tll_foreach(instance.workspaces, it)
+            {
+                if (it->item->id == id) {
+                    it->item->urgent = urgent;
+                    break;
+                }
+            }
+            mtx_unlock(&instance.mtx);
+
+            events |= workspace_urgency_changed;
+        }
+
         json_object_iter_next(&it);
     }
 
diff --git a/modules/niri-common.h b/modules/niri-common.h
index 18afe38..8aa2f39 100644
--- a/modules/niri-common.h
+++ b/modules/niri-common.h
@@ -10,6 +10,7 @@ enum niri_event {
     workspace_active_window_changed = (1 << 2),
     keyboard_layouts_changed = (1 << 3),
     keyboard_layouts_switched = (1 << 4),
+    workspace_urgency_changed = (1 << 5),
 };
 
 struct niri_subscriber {
@@ -24,6 +25,7 @@ struct niri_workspace {
     bool active;
     bool focused;
     bool empty;
+    bool urgent;
 };
 
 struct niri_socket {
diff --git a/modules/niri-workspaces.c b/modules/niri-workspaces.c
index bca0150..99a8c83 100644
--- a/modules/niri-workspaces.c
+++ b/modules/niri-workspaces.c
@@ -56,8 +56,9 @@ content(struct module *module)
                 tag_new_bool(module, "active", it->item->active),
                 tag_new_bool(module, "focused", it->item->focused),
                 tag_new_bool(module, "empty", it->item->empty),
+                tag_new_bool(module, "urgent", it->item->urgent),
             },
-            .count = 5,
+            .count = 6,
         };
 
         exposable[i++] = private->label->instantiate(private->label, &tags);
@@ -86,7 +87,7 @@ run(struct module *module)
     if (private->niri == NULL)
         return 1;
 
-    int fd = niri_socket_subscribe(workspaces_changed | workspace_activated | workspace_active_window_changed);
+    int fd = niri_socket_subscribe(workspaces_changed | workspace_activated | workspace_active_window_changed | workspace_urgency_changed);
     if (fd == -1) {
         niri_socket_close();
         return 1;
-- 
2.49.0
```

And a yambar configuration example:
```yaml
  left:
    - niri-workspaces:
        content:
            map:
              default: {string: {text: "{id}", margin: 5}}
              conditions:
                urgent: {string: {text: "{id}", deco: {background: {color: af3029ff } }, margin: 5}}
                active: {string: {text: "{id}", deco: {background: {color: 383838ff } }, margin: 5}}
```

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -803,6 +803,24 @@ pub enum Action {
     OpenOverview {},
     /// Close the Overview.
     CloseOverview {},
+    /// Toggle urgent status of a window.
+    ToggleUrgent {
+        /// Id of the window to toggle urgent.
+        #[cfg_attr(feature = "clap", arg(long))]
+        id: u64,
+    },
+    /// Set urgent status of a window.
+    SetUrgent {
+        /// Id of the window to set urgent.
+        #[cfg_attr(feature = "clap", arg(long))]
+        id: u64,
+    },
+    /// Unset urgent status of a window.
+    UnsetUrgent {
+        /// Id of the window to unset urgent.
+        #[cfg_attr(feature = "clap", arg(long))]
+        id: u64,
+    },
 }
 
 /// Change in window or column size.
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1129,6 +1129,8 @@ pub struct Window {
     ///
     /// If the window isn't floating then it is in the tiling layout.
     pub is_floating: bool,
+    /// Whether this window requests your attention.
+    pub is_urgent: bool,
 }
 
 /// Output configuration change result.
@@ -1297,6 +1299,13 @@ pub enum Event {
     WindowFocusChanged {
         /// Id of the newly focused window, or `None` if no window is now focused.
         id: Option<u64>,
+    },
+    /// Window urgency changed.
+    WindowUrgencyChanged {
+        /// Id of the window.
+        id: u64,
+        /// The new urgency state of the window.
+        urgent: bool,
     },
     /// The configured keyboard layouts have changed.
     KeyboardLayoutsChanged {
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1150,6 +1150,8 @@ pub struct Workspace {
     ///
     /// Can be `None` if no outputs are currently connected.
     pub output: Option<String>,
+    /// Whether the workspace currently has an urgent window in its output.
+    pub is_urgent: bool,
     /// Whether the workspace is currently active on its output.
     ///
     /// Every output has one active workspace, the one that is currently visible on that output.
@@ -1223,6 +1225,13 @@ pub enum Event {
         /// This configuration completely replaces the previous configuration. I.e. if any
         /// workspaces are missing from here, then they were deleted.
         workspaces: Vec<Workspace>,
+    },
+    /// The workspace urgency changed.
+    WorkspaceUrgencyChanged {
+        /// Id of the workspace.
+        id: u64,
+        /// Whether this workspace has an urgent window.
+        urgent: bool,
     },
     /// A workspace was activated on an output.
     ///
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -192,6 +192,9 @@ layout {
         active-color "#ffc87f"
         inactive-color "#505050"
 
+        // Color of the border around windows that request your attention.
+        urgent-color "#9b0000"
+
         // active-gradient from="#ffbb66" to="#ffc880" angle=45 relative-to="workspace-view"
         // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view"
     }
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -707,6 +707,8 @@ impl GammaControlHandler for State {
 }
 delegate_gamma_control!(State);
 
+struct UrgentOnlyMarker;
+
 impl XdgActivationHandler for State {
     fn activation_state(&mut self) -> &mut XdgActivationState {
         &mut self.niri.activation_state
@@ -716,11 +718,10 @@ impl XdgActivationHandler for State {
         // Tokens without a serial are urgency-only. This is not specified, but it seems to be the
         // common client behavior.
         //
-        // We don't have urgency yet, so just ignore such tokens.
-        //
         // See also: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/150
         let Some((serial, seat)) = data.serial else {
-            return false;
+            data.user_data.insert_if_missing(|| UrgentOnlyMarker);
+            return true;
         };
         let Some(seat) = Seat::<State>::from_resource(&seat) else {
             return false;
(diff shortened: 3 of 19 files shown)


FreeFull:
I think it would also be nice to have a niri action to set a specific window as urgent, similar to `niri msg action focus-window --id <ID>`

src/window/mapped.rs
```diff
@@ -830,6 +839,16 @@ impl LayoutElement for Mapped {
         }
     }
 
+    fn set_urgent(&mut self, urgent: bool) {
+        let changed = self.is_urgent != urgent;
```
YaLTeR, line 843: If the window is focused, probably don't want to set urgent?

src/layout/mod.rs
```diff
@@ -1583,6 +1586,14 @@ impl<W: LayoutElement> Layout<W> {
         ws_idx == mon.active_workspace_idx
     }
 
+    pub fn urgent_window(&mut self, id: &W::Id) {
```
YaLTeR, line 1589: I don't think this is needed here, along with `LayoutElement::set_urgent()`. Since the layout code doesn't need to set urgency, it's always done externally where it can just use a `Mapped` method.

src/layout/tests.rs
```diff
@@ -31,6 +31,7 @@ struct TestWindowInner {
     is_fullscreen: Cell<bool>,
     is_windowed_fullscreen: Cell<bool>,
     is_pending_windowed_fullscreen: Cell<bool>,
+    is_urgent: Cell<bool>,
```
YaLTeR, line 34: If removing `set_urgent()`, then this is not necessary, can just return constant false from `is_urgent()`.

src/handlers/mod.rs
```diff
@@ -707,6 +707,11 @@ impl GammaControlHandler for State {
 }
 delegate_gamma_control!(State);
 
+pub enum ActivationContext {
+    UrgentOnly,
+    FocusIfPossible,
+}
```
YaLTeR, line 713: I'm thinking, maybe leave just a marker `struct UrgentOnlyMarker;`? Since we already filter tokens as valid/invalid by returning true/false from `token_created()`. Then only insert `UrgentOnlyMarker` in the urgent-only branch, and check if it's present in `request_activation()`.

src/handlers/mod.rs
```diff
@@ -761,10 +772,21 @@ impl XdgActivationHandler for State {
...
-                self.niri.layer_shell_on_demand_focus = None;
-                self.niri.queue_redraw_all();
+                match context {
+                    ActivationContext::FocusIfPossible => {
+                        self.niri.layout.activate_window(&window);
+                        self.niri.layer_shell_on_demand_focus = None;
+                        self.niri.queue_redraw_all();
+                    }
+                    ActivationContext::UrgentOnly => {
+                        self.niri.layout.urgent_window(&window);
```
YaLTeR, line 787: Needs `queue_redraw_all()` if it changes the focus ring/border color

src/layout/workspace.rs
```diff
@@ -1608,6 +1608,10 @@ impl<W: LayoutElement> Workspace<W> {
         self.scrolling.scroll_amount_to_activate(window)
     }
 
+    pub fn is_urgent(&self) -> bool {
```
YaLTeR, line 1611: I think this belongs in the next commit

niri-ipc/src/lib.rs
YaLTeR, line 1: What about Window is_urgent and WindowUrgencyChanged?

niri-ipc/src/lib.rs
```diff
@@ -1224,6 +1226,13 @@ pub enum Event {
         /// workspaces are missing from here, then they were deleted.
         workspaces: Vec<Workspace>,
     },
+    /// A window on a workspace became urgent.
```
YaLTeR, line 1247: I think this is better as just "The workspace urgency changed." because it could change due to window closing or moving away, etc.

src/ipc/server.rs
```diff
@@ -591,6 +591,12 @@ impl State {
             if is_active && !ipc_ws.is_active {
                 events.push(Event::WorkspaceActivated { id, focused: false });
             }
+
+            // Check if this workspace urgent state changed.
```
YaLTeR, line 595: Shouldn't this go higher, above the `continue;` in the focused branch?

src/layout/focus_ring.rs
```diff
@@ -69,6 +70,8 @@ impl FocusRing {
 
         let color = if is_active {
             self.config.active_color
+        } else if is_urgent {
```
YaLTeR, line 73: Should this maybe take precedence over is_active? So that in the overview you can see urgent active windows on other workspaces with urgent color.

resources/default-config.kdl
```diff
@@ -191,6 +191,7 @@ layout {
         width 4
         active-color "#ffc87f"
         inactive-color "#505050"
+        urgent-color "#9b0000"
```
YaLTeR, line 194: // Color of the border around windows that request your attention.

src/input/mod.rs
```diff
@@ -1971,6 +1971,37 @@ impl State {
                     self.niri.queue_redraw_all();
                 }
             }
+            Action::ToggleUrgent(id) => {
+                let window = self
+                    .niri
+                    .layout
+                    .workspaces_mut()
+                    .find_map(|ws| ws.windows_mut().find(|w| w.id().get() == id));
+                if let Some(window) = window {
```
YaLTeR, line 1980:
```
                let window = self.niri.layout.windows().find(|(_, m)| m.id().get() == id);
                if let Some((_, mapped)) = window {
```

YaLTeR (changes requested):

src/input/mod.rs
YaLTeR: Ah right. We have with_windows_mut() but not windows_mut()

YaLTeR:
All looks good but I think it would be good to include Window-level is_urgent into the IPC and event stream, rather than just Workspace-level?

YaLTeR:
Renamed is_urgent to urgent in the WindowUrgencyChanged event for consistency. Gave it a test, seems to work well.

YaLTeR:
Thanks! And thanks for separating the commits out cleanly.

Duncaen:
Awesome thank you. I'll look into up-streaming support for this into yambar and maybe some other tool bars if its not too complicated soon.

YaLTeR:
Cool! I just pushed some follow-ups for the missing settings.

I guess waybar would be nice and shouldn't be too hard if i recall correctly from when doing the initial impl

YaLTeR:
So, apparently something I've missed is that the actions should ideally be named un/set/toggle-*window*-urgent rather than just un/set/toggle-urgent. I'm thinking how bad of an idea it would be to correct this in a hotfix release.

Do you think that's a good idea? They are not bindable in the config so it's not a config-breaking change, and I don't think any bars use those actions yet?

Duncaen:
Seems reasonable and early enough that nothing really depends on it.

PR #1332, opened 2025-03-24
layout: add focus flag to move-window-to-workspace

nnyyxxxx:


YaLTeR:
I don't really like the name "silent", it's unclear in the window management context. Maybe something like `--focus false` or `--no-focus`? This is similar to how there's a window rule for `open-focused false`.

nnyyxxxx:
> I don't really like the name "silent", it's unclear in the window management context. Maybe something like `--focus false` or `--no-focus`? This is similar to how there's a window rule for `open-focused false`.

changed to focus

nnyyxxxx:
should be good now, did some minor touchup

niri-ipc/src/lib.rs
```diff
@@ -426,6 +426,11 @@ pub enum Action {
         /// Reference (index or name) of the workspace to move the window to.
         #[cfg_attr(feature = "clap", arg())]
         reference: WorkspaceReferenceArg,
+
+        /// When set to false, move the window to the workspace without switching
+        /// focus to that workspace. The focus remains on the current workspace.
+        #[cfg_attr(feature = "clap", arg(long))]
+        focus: Option<bool>,
```
YaLTeR, line 433: Should be just a `bool` that defaults to `true`.

src/input/mod.rs
```diff
@@ -1091,8 +1091,10 @@ impl State {
                             self.move_cursor_to_output(&output);
                         }
                     } else {
-                        self.niri.layout.move_to_workspace(None, index);
-                        self.maybe_warp_cursor_to_focus();
+                        self.niri.layout.move_to_workspace(None, index, focus);
+                        if focus {
+                            self.maybe_warp_cursor_to_focus();
```
YaLTeR, line 1096: This should remain unconditional because the cursor should still warp to the newly focused window after this one moves away.

src/input/mod.rs
```diff
@@ -1130,12 +1133,16 @@ impl State {
...
-                            self.niri.layout.move_to_workspace(Some(&window), index);
+                            self.niri
+                                .layout
+                                .move_to_workspace(Some(&window), index, focus);
 
                             // If we focused the target window.
-                            let new_focus = self.niri.layout.focus();
-                            if new_focus.is_some_and(|win| win.window == window) {
-                                self.maybe_warp_cursor_to_focus();
+                            if focus {
```
YaLTeR, line 1141: Probably unnecessary extra condition

src/layout/monitor.rs
```diff
@@ -490,9 +490,16 @@ impl<W: LayoutElement> Monitor<W> {
         }
         let new_id = self.workspaces[new_idx].id();
 
-        let activate = window.map_or(true, |win| {
-            self.active_window().map(|win| win.id()) == Some(win)
-        });
+        // Whether to activate the window based on focus parameter and whether the window
+        // is currently focused.
+        let activate = if focus {
```
YaLTeR, line 495: Could you replace `focus: bool` with `activate: ActivateWindow` then use `map_smart()` for this check? Pass `Smart` for the current behavior, and `No` for new behavior

YaLTeR (changes requested):

niri-ipc/src/lib.rs
nnyyxxxx: what?

YaLTeR:
```rs
#[cfg_attr(feature = "clap", arg(long, action = clap::ArgAction::Set, default_value_t = true))]
focus: bool,
```

src/layout/monitor.rs
```diff
@@ -578,7 +584,7 @@ impl<W: LayoutElement> Monitor<W> {
 
         let workspace = &mut self.workspaces[source_workspace_idx];
         if workspace.floating_is_active() {
-            self.move_to_workspace(None, idx);
+            self.move_to_workspace(None, idx, ActivateWindow::Yes);
```
YaLTeR, line 587:
```suggestion
            self.move_to_workspace(None, idx, ActivateWindow::Smart);
```

YaLTeR:
Thanks

YaLTeR:
So I somehow missed this but moving to a workspace on a different output does not honor the focus flag

YaLTeR:
Fixed in d91499486e505bba1e1764e1e055ce1ec5eda7fb

Issue #1284, opened 2025-03-16
Certain popup windows cannot open


realSaltyFish:
Under niri, GNOME Nautilus cannot open the file rename popup window (on the first attempt the popup will open then immediately disappear). It works fine under Hyprland, or xwayland-satellite.

### System Information

* niri version: `niri 25.02 (unknown commit)`

* Distro: `Arch Linux`

* GPU: `Intel Alder Lake-P GT2`

* CPU: `Intel i9-12900H`

YaLTeR:
Can't reproduce, works fine for me

sakarie9:
Same issue here.

Close fcitx5 makes rename popup works again, so maybe this is an ime related bug.

Some logs after press F2 rename key:

```
[ 506373.482] {Default Queue} wl_keyboard#38.key(1180286, 7957400, 60, 1)
[ 506374.168] {Default Queue}  -> xdg_wm_base#37.create_positioner(new id xdg_positioner#85)
[ 506374.183] {Default Queue}  -> xdg_positioner#85.set_size(372, 183)
[ 506374.188] {Default Queue}  -> xdg_positioner#85.set_anchor_rect(219, 150, 551, 44)
[ 506374.193] {Default Queue}  -> xdg_positioner#85.set_offset(0, 0)
[ 506374.199] {Default Queue}  -> xdg_positioner#85.set_anchor(2)
[ 506374.204] {Default Queue}  -> xdg_positioner#85.set_gravity(2)
[ 506374.208] {Default Queue}  -> xdg_positioner#85.set_constraint_adjustment(57)
[ 506374.213] {Default Queue}  -> xdg_positioner#85.set_reactive()
[ 506374.218] {Default Queue}  -> xdg_wm_base#37.get_xdg_surface(new id xdg_surface#89, wl_surface#51)
[ 506374.224]  -> xdg_surface#89.get_popup(new id xdg_popup#105, xdg_surface#59, xdg_positioner#85)
[ 506374.228] {Default Queue}  -> xdg_positioner#85.destroy()
[ 506374.234]  -> xdg_popup#105.grab(wl_seat#31, 1180286)
[ 506374.239]  -> wl_surface#51.commit()

(org.gnome.Nautilus:106945): Gdk-WARNING **: 21:53:50.258: ../gtk/gdk/wayland/gdkcursor-wayland.c:210 cursor image size (32) not an integer multiple of theme size (24)
[...]

Issue #1206, opened 2025-03-04
can't open 'Choose File' Dialog with `xdg-desktop-portal-gnome>=47.1-1`


AClon314:
See: https://bbs.archlinux.org/viewtopic.php?id=299593
Solution: `sudo dnf remove xdg-desktop-portal-gnome`
Suggest to remove the dependency at here: https://github.com/YaLTeR/niri/wiki/Important-Software#portals

In fedora 42 pre-release, the `xdg-desktop-portal-gnome` version bump to `v48`, which can't open 'Choose File' Dialog on any Xwayland Applications. Only firefox can popup the dialog.
```
Package                         Arch     Version                         Repository                Size
Installing:
 xdg-desktop-portal-gnome       x86_64   48~beta.2-1.fc42                updates-testing      878.8 KiB
```

Related threads: https://discussion.fedoraproject.org/t/blink-based-browsers-dont-open-file-dialog-in-sway-fedora-38/90403/2

YaLTeR:
xdp-gnome 48 requires nautilus for the file chooser dialog. It is a packaging error for xdp-gnome that a dependency on nautilus is missing for 48.

bugaevc:
@AClon314 why are you reporting this as an issue to Niri?

YaLTeR:
(also removing `xdg-desktop-portal-gnome` will prevent screencasting from working in niri, because niri works through the gnome portal for screencasting)

AClon314:
> [@AClon314](https://github.com/AClon314) why are you reporting this as an issue to Niri?
My bad, it's a packaging error and serachable from Internet. And I make a confussion between `discusson` and `issue`.

PR #1344, opened 2025-03-26
wiki: Mention file chooser dependency and settings for portals

GZGavinZhao:
Added to hopefully prevent confusion as seen in #1206 and #1284.

### DIFF
--- a/wiki/Important-Software.md
+++ b/wiki/Important-Software.md
@@ -22,6 +22,8 @@ Since we're using `xdg-desktop-portal-gnome`, Flatpak apps will read the GNOME U
 dconf write /org/gnome/desktop/interface/color-scheme '"prefer-dark"'
 ```
 
+Note that if you're using the provided `resources/niri-portals.conf`, you also need to install the `nautilus` file manager in order for file chooser dialogues to work properly. If you do not want to install `natuilus` (say you use `nemo` instead), you need to set `org.freedesktop.impl.portal.FileChooser=gtk;` in `niri-portals.conf` to use the GTK portal for file chooser dialogues.
+
 ### Authentication Agent
 
 Required when apps need to ask for root permissions. Something like `plasma-polkit-agent` works fine. Start it [with systemd](./Example-systemd-Setup.md) or with [`spawn-at-startup`](./Configuration:-Miscellaneous.md#spawn-at-startup).


YaLTeR:
Looks ok, but I'd rather emphasize that this is a packaging problem for the distro's xdg-desktop-portal-gnome package (it does not have the required nautilus runtime dependency). An arguably better hotfix is removing FileChooser from `/usr/share/xdg-desktop-portal/portals/gnome.portal` (I haven't tested it myself).

wiki/Important-Software.md
```diff
@@ -22,6 +22,8 @@ Since we're using `xdg-desktop-portal-gnome`, Flatpak apps will read the GNOME U
 dconf write /org/gnome/desktop/interface/color-scheme '"prefer-dark"'
 ```
 
+Note that if you're using the provided `resources/niri-portals.conf`, you also need to install the `nautilus` file manager in order for file chooser dialogues to work properly. If you do not want to install `natuilus` (say you use `nemo` instead), you need to set `org.freedesktop.impl.portal.FileChooser=gtk;` in `niri-portals.conf` to use the GTK portal for file chooser dialogues.
```
YaLTeR, line 25: What do you think about this?

```suggestion
Note that if you're using the provided `resources/niri-portals.conf`, you also need to install the `nautilus` file manager in order for file chooser dialogues to work properly. This is necessary because xdg-desktop-portal-gnome uses nautilus as the file chooser by default starting from version 47.0.

If you do not want to install `natuilus` (say you use `nemo` instead), you can set `org.freedesktop.impl.portal.FileChooser=gtk;` in `niri-portals.conf` to use the GTK portal for file chooser dialogues.
```

GZGavinZhao: Looks good. Thank you!

YaLTeR:
Thanks

PandorasFox:
> An arguably better hotfix is removing FileChooser from /usr/share/xdg-desktop-portal/portals/gnome.portal (I haven't tested it myself).

chiming in just to say i have tested approximately this and it works great :)

```
➜ cat ~/.config/xdg-desktop-portal/portals.conf
[preferred]
default=gtk
org.freedesktop.impl.portal.FileChooser=gtk
org.freedesktop.impl.portal.ScreenCast=gnome
```

`xdg-desktop-portal-gnome` functions perfectly for screencasting without nautilus and its ~20 extra dependencies installed, and the file-picker modals go through the gtk portal and default to dolphin or thunar or whatever is actually installed.

as far as I can tell, [on arch linux] xdg-desktop-portal-gnome only needs gnome-desktop-common & gnome-desktop-4 as supporting libraries for the screencasting interface to function.

PR #1416, opened 2025-04-10
Allow a user to disable all Input

dmarkham:
This is not expected to get merged or used, Just wanted to share something I found useful.  I'm building a niri2vnc client  to work like my old x2vnc and I needed to disable all Input via the IPC. I'm no Rust Dev,  It's all vibe coded. Works perfectly for me.


YaLTeR:
Okay then

PR #1430, opened 2025-04-15
chore: update smithay

nagisa:


YaLTeR:
Thanks!

YaLTeR:
Rolled this back for now bc there are problems https://github.com/Smithay/smithay/issues/1715

PR #1437, opened 2025-04-17
Addded a FAQ question for adding custom iconsets in niri

Livesi5e:
Title says it all

### DIFF
--- a/wiki/FAQ.md
+++ b/wiki/FAQ.md
@@ -46,3 +46,10 @@ To run X11 apps, you can use [xwayland-satellite](https://github.com/Supreeeme/x
 Check [the Xwayland wiki page](./Xwayland.md) for instructions.
 
 Keep in mind that you can run many Electron apps such as VSCode natively on Wayland by passing the right flags, e.g. `code --ozone-platform-hint=auto`
+
+### How to change cursor and system icons?
+
+You'll need a GTK theme editor like [nwg-look](https://github.com/nwg-piotr/nwg-look) to modify the cursor and system icons.
+
+After installing the progam, put your preferred icon theme into
+`~/.icons` or `~/local/share/.icons` for user wide and `/usr/share/icons` for system wide installation. Then select the iconset in nwg-look or a similar program.
--- a/wiki/FAQ.md
+++ b/wiki/FAQ.md
@@ -49,7 +49,9 @@ Keep in mind that you can run many Electron apps such as VSCode natively on Wayl
 
 ### How to change cursor and system icons?
 
-You'll need a GTK theme editor like [nwg-look](https://github.com/nwg-piotr/nwg-look) to modify the cursor and system icons.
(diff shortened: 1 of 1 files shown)


wiki/FAQ.md
```diff
@@ -46,3 +46,10 @@ To run X11 apps, you can use [xwayland-satellite](https://github.com/Supreeeme/x
 Check [the Xwayland wiki page](./Xwayland.md) for instructions.
 
 Keep in mind that you can run many Electron apps such as VSCode natively on Wayland by passing the right flags, e.g. `code --ozone-platform-hint=auto`
+
+### How to change cursor and system icons?
```
YaLTeR, line 50: Huh? But you change the cursor in the niri config: https://github.com/YaLTeR/niri/wiki/Configuration:-Miscellaneous#cursor

wiki/FAQ.md
```diff
@@ -46,3 +46,10 @@ To run X11 apps, you can use [xwayland-satellite](https://github.com/Supreeeme/x
...
+
+You'll need a GTK theme editor like [nwg-look](https://github.com/nwg-piotr/nwg-look) to modify the cursor and system icons.
```
YaLTeR, line 52: Why specifically GTK?

wiki/FAQ.md
```diff
@@ -46,3 +46,10 @@ To run X11 apps, you can use [xwayland-satellite](https://github.com/Supreeeme/x
...
+
+After installing the progam, put your preferred icon theme into
+`~/.icons` or `~/local/share/.icons` for user wide and `/usr/share/icons` for system wide installation. Then select the iconset in nwg-look or a similar program.
```
YaLTeR, line 55: Okay, two things here. First, `/usr/share/icons` should only be modified by system packages, so this should say something like "install the icon theme using your system's package manager", or download and put into `~/.local/share/icons/`.

Second, all this seems very unrelated to niri, and just general Linux knowledge? Not sure why it belongs to the niri wiki?

YaLTeR: Maybe instead should link to something like a relevant arch wiki page?

Livesi5e: That's true, it's mostly what's also in the arch wiki. I can also just link to that. But as a user I would expect to not be thrown around multiple wikis to find what I need. Maybe giving a quick rundown and then linking to the arch page might be more convenient?

wiki/FAQ.md
Livesi5e: Misunderstood GTK, my bad

wiki/FAQ.md
YaLTeR: I guess yeah

wiki/FAQ.md
Livesi5e: Yup, will remove that. Just saw that nwg-looks also changed the cursor

wiki/FAQ.md
Livesi5e: Updated the writing

Issue #850, opened 2024-12-08
Overview


YaLTeR:
Something that zooms out and lets you see multiple workspaces and windows on each monitor. The design concepts here are very close to what I have in mind: https://github.com/YaLTeR/niri/discussions/352#discussioncomment-10094739

Actual interactions in the overview (click on window to focus it scrolling it into view? moving windows around workspaces like in GNOME Shell?) will need some trying and evaluating hands-on.

There are some questions on what to do with layer-shell surfaces. It probably makes sense to draw a separate background layer on each workspace in the overview, and the top layers should stay on top. Once again will have to try and see what works.

mastoca:
Perhaps just hiding the layer-shell is ok as I doubt there's much need for an 'overview' of that layer. I could be wrong but as of now I don't even use it beyond the waybar currently.

What are other uses for layer-shell?

salman-farooq-sh:
@mastoca Launchers e.g. wofi, fuzzel, bemenu, etc. also use layer-shell surfaces. Still I also prefer that at least such launchers and bars should be hidden in the overview, will make implementation simpler too?. The bar is going to look tiny and unusable at that size anyway.

Uhh, notifications? Yeah, it's a mess. Best show only the normal app windows as that design concept shows I think.

mastoca:
@salman-farooq-sh I agree focusing on just the normal app windows would satisfy 95-99% of use cases. 

If we must get fancy, I would suggest a filter to omit certain windows from participating like the screen recording does (perhaps would be a nice phase II).

YaLTeR:
My idea is to show Background layer on every workspace, but show all other layers just on top, so you'll still see your bar as normal.

salman-farooq-sh:
@YaLTeR Yeah that is a good way to do it, it feels consistent with the normal (current non-overview) mode specifically that launchers, notifications, and bars remain at their own place when navigating columns/workspaces.

Regarding background, how about keeping it fixed too? The vertical column in the concept design designating the viewport also looks great by simply making it a bit darker.

![image](https://github.com/user-attachments/assets/8d130bdd-cf33-4bd6-b897-b9b76002feae)

Crdr0122:
Wondering how would floating windows show up in the overview? Stickied in the middle viewport?

YaLTeR:
> Regarding background, how about keeping it fixed too? The vertical column in the concept design designating the viewport also looks great by simply making it a bit darker.

[...]

Issue #849, opened 2024-12-08
Hot corners


YaLTeR:
Top-left hot corner to open the overview is my favorite gesture in GNOME Shell when using the mouse. We don't have an overview yet, but once we do, I will certainly want the hot corner.

Config wise, they can probably go in `binds {}` since they are discrete actions. Something like `TopLeftHotCorner { open-overview; }`. Though, one question is how to configure the outputs, and if that's even needed.

Implementation wise, can probably copy whatever GNOME Shell does, it works well there.

There's apparently a client-side solution for this: https://github.com/AndreasBackx/waycorner. I haven't tried it (tbh I'm not sure what I'd use a hot corner for other than opening the overview), but once the overview happens, I'll probably want the top-left hot corner by default, and this means it needs to be built into niri. It should be fine because I expect hot corners to be relatively light code-wise.

mastoca:
I've only ever used hot corners for entering overviews, start screensaver/lock-screen, and on places I have touch gestures to perhaps open a quick note (on ipad).

salman-farooq-sh:
I used waycorner for a while. It was janky, and thus very unreliable, to trigger it even after I tried with increased detection area config thingy it has.

YaLTeR:
I added the top-left corner for the Overview, and made sure that the config syntax allows future extensibility. However, I don't have concrete ideas for anything else, so as other feature requests, it will need to be discussed in Discussions first, taking into account that `waycorner` exists and covers some use cases.

scientiac:
It would be better if the hot corner wasn't so sensitive, I am using way-edges as my bar replacement and when I hover to the corner where my volume control is, the overview gets triggered. A forced push, sth like gnome/KDE has, would be great.

YaLTeR:
Made an issue for that: https://github.com/YaLTeR/niri/issues/1473

PR #1440, opened 2025-04-18
Overview

YaLTeR:
Implements #850. Related: #849.

This is an overview. It zooms out your workspaces and lets you move your windows around.

https://github.com/user-attachments/assets/c3a3fd5d-0a2c-490b-a47f-3fff7826e347

Open it with a new `toggle-overview` bind, or via the top-left hot corner, or via a touchpad four-finger swipe up. While in the overview, all keyboard navigation keeps working, while pointing devices get easier:
- Mouse: left click for interactive-move, right click to scroll a workspace left/right (no Mod required).
- Touchpad: two-finger gestures TBD, for now use the normal three-finger gestures.
- Touchscreen: TBD.

The overview hot corner has some differences from a layer-shell impl like waycorner:
- It is automatically disabled when a fullscreen window is on the monitor.
- It does not prevent direct scanout because it does not render anything.
- It works during interactive move and drag-and-drop.

Drag-and-drop will also scroll the workspaces up/down in the overview, and will activate the workspace in the overview if you hold it above for a moment. Combined with the hot corner, this lets you do a mouse-only DnD across workspaces.

https://github.com/user-attachments/assets/5f09c5b7-ff40-462b-8b9c-f1b8073a2cbb

You can also drag-and-drop to a new workspace above, below, or in-between existing workspaces.

https://github.com/user-attachments/assets/b76d5349-aa20-4889-ab90-0a51554c789d

New options:
- `overview-open-close` in `animations {}`
- `dnd-edge-workspace-switch` in `gestures {}`

I'm looking for testing and feedback:
- bugs and crashes
- wrong behavior with good justification on why it is wrong
- features with good justification; if it's small enough I may consider it for this PR, otherwise maybe for later
- config options **only** with good justification why they are needed and example how they should look in the config, which should be consistent with the rest of the config

Some issues that I know about:
- [ ] Mouse scroll to switch workspaces during interactive move in the overview doesn't work properly.
- [ ] Mouse scroll in the overview affects focused output instead of output under cursor.
- [ ] Drag-and-drop hold over a window outside the viewport doesn't scroll to that window after focusing the workspace.
- [ ] There's some animation jank when interactively moving a window to a new workspace above first.
- [ ] Zoom in + scroll window into view animation is not linear; I'm not yet sure if it's fixable or not for this PR, will have to look into it.

Some TBD things for this PR:
- [ ] Touchpad two-finger scrolling in the overview that will match three-finger scrolling.
- [ ] Touchscreen gestures within the overview.

Some things definitely out of scope for this PR:
- Window icons. This requires searching and keeping track of .desktop files to match Wayland app-id to icons, which is a whole endeavor all by itself.
- 1 pixel jank with thin borders, gaps, and in general in the overview. This requires a big refactor of the niri rendering. I gave it a honest attempt, and it is way too big and error prone to do for this PR.
- Scaling down windows, preserving gap sizes, other overview window layout ideas. This requires the same refactor as above, plus the design still needs experimentation.
- Animations for workspace movement, removal, addition: a lot of work.
- DnD entire workspaces: a lot of work, though would be cool, and is a natural overview feature.
- Put blurred background on the backdrop or w/e: we don't have blur yet, also we can't just take the bottom layer-shell layer because if you have desktop icons there then it will look weird. So idk this might be blocked on adding desktop backgrounds support to niri itself. Also unclear how it should work with per-workspace bg that will be a thing when that happens.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -23,7 +23,8 @@ use smithay::input::keyboard::xkb::{keysym_from_name, KEYSYM_CASE_INSENSITIVE};
 use smithay::input::keyboard::{Keysym, XkbConfig};
 use smithay::reexports::input;
 
-pub const DEFAULT_BACKGROUND_COLOR: Color = Color::from_array_unpremul([0.2, 0.2, 0.2, 1.]);
+pub const DEFAULT_BACKGROUND_COLOR: Color = Color::from_array_unpremul([0.25, 0.25, 0.25, 1.]);
+pub const DEFAULT_BACKDROP_COLOR: Color = Color::from_array_unpremul([0.15, 0.15, 0.15, 1.]);
 
 pub mod layer_rule;
 
@@ -984,6 +985,8 @@ pub struct Animations {
     pub config_notification_open_close: ConfigNotificationOpenCloseAnim,
     #[knuffel(child, default)]
     pub screenshot_ui_open: ScreenshotUiOpenAnim,
+    #[knuffel(child, default)]
+    pub overview_open_close: OverviewOpenCloseAnim,
 }
 
 impl Default for Animations {
@@ -999,6 +1002,7 @@ impl Default for Animations {
             window_resize: Default::default(),
             config_notification_open_close: Default::default(),
             screenshot_ui_open: Default::default(),
+            overview_open_close: Default::default(),
         }
     }
 }
@@ -1147,6 +1151,22 @@ impl Default for ScreenshotUiOpenAnim {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq)]
+pub struct OverviewOpenCloseAnim(pub Animation);
+
+impl Default for OverviewOpenCloseAnim {
+    fn default() -> Self {
+        Self(Animation {
+            off: false,
+            kind: AnimationKind::Spring(SpringParams {
+                damping_ratio: 1.,
+                stiffness: 800,
+                epsilon: 0.0001,
+            }),
+        })
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Animation {
     pub off: bool,
     pub kind: AnimationKind,
@@ -1183,6 +1203,8 @@ pub struct SpringParams {
 pub struct Gestures {
     #[knuffel(child, default)]
     pub dnd_edge_view_scroll: DndEdgeViewScroll,
+    #[knuffel(child, default)]
+    pub dnd_edge_workspace_switch: DndEdgeWorkspaceSwitch,
 }
 
 #[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
@@ -1199,6 +1221,26 @@ impl Default for DndEdgeViewScroll {
     fn default() -> Self {
         Self {
             trigger_width: FloatOrInt(30.), // Taken from GTK 4.
+            delay_ms: 100,
+            max_speed: FloatOrInt(1500.),
+        }
+    }
+}
+
+#[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
+pub struct DndEdgeWorkspaceSwitch {
+    #[knuffel(child, unwrap(argument), default = Self::default().trigger_height)]
+    pub trigger_height: FloatOrInt<0, 65535>,
+    #[knuffel(child, unwrap(argument), default = Self::default().delay_ms)]
+    pub delay_ms: u16,
+    #[knuffel(child, unwrap(argument), default = Self::default().max_speed)]
+    pub max_speed: FloatOrInt<0, 1_000_000>,
+}
+
+impl Default for DndEdgeWorkspaceSwitch {
+    fn default() -> Self {
+        Self {
+            trigger_height: FloatOrInt(50.),
             delay_ms: 100,
             max_speed: FloatOrInt(1500.),
         }
@@ -1716,6 +1758,7 @@ pub enum Action {
     SetDynamicCastWindowById(u64),
     SetDynamicCastMonitor(#[knuffel(argument)] Option<String>),
     ClearDynamicCastTarget,
+    ToggleOverview,
 }
 
 impl From<niri_ipc::Action> for Action {
@@ -1980,6 +2023,7 @@ impl From<niri_ipc::Action> for Action {
                 Self::SetDynamicCastMonitor(output)
             }
             niri_ipc::Action::ClearDynamicCastTarget {} => Self::ClearDynamicCastTarget,
+            niri_ipc::Action::ToggleOverview {} => Self::ToggleOverview,
         }
     }
 }
@@ -2950,6 +2994,21 @@ where
 }
 
 impl<S> knuffel::Decode<S> for ScreenshotUiOpenAnim
+where
+    S: knuffel::traits::ErrorSpan,
+{
+    fn decode_node(
+        node: &knuffel::ast::SpannedNode<S>,
+        ctx: &mut knuffel::decode::Context<S>,
+    ) -> Result<Self, DecodeError<S>> {
+        let default = Self::default().0;
+        Ok(Self(Animation::decode_node(node, ctx, default, |_, _| {
+            Ok(false)
+        })?))
+    }
+}
+
+impl<S> knuffel::Decode<S> for OverviewOpenCloseAnim
 where
     S: knuffel::traits::ErrorSpan,
 {
@@ -4459,6 +4518,18 @@ mod tests {
                         ),
                     },
                 ),
+                overview_open_close: OverviewOpenCloseAnim(
+                    Animation {
+                        off: false,
+                        kind: Spring(
+                            SpringParams {
+                                damping_ratio: 1.0,
+                                stiffness: 800,
+                                epsilon: 0.0001,
+                            },
+                        ),
+                    },
+                ),
             },
             gestures: Gestures {
                 dnd_edge_view_scroll: DndEdgeViewScroll {
@@ -4468,6 +4539,15 @@ mod tests {
                     delay_ms: 100,
                     max_speed: FloatOrInt(
                         50.0,
+                    ),
+                },
+                dnd_edge_workspace_switch: DndEdgeWorkspaceSwitch {
+                    trigger_height: FloatOrInt(
+                        50.0,
+                    ),
+                    delay_ms: 100,
+                    max_speed: FloatOrInt(
+                        1500.0,
                     ),
                 },
             },
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -764,6 +764,8 @@ pub enum Action {
     },
     /// Clear the dynamic cast target, making it show nothing.
     ClearDynamicCastTarget {},
+    /// Toggle the Overview.
+    ToggleOverview {},
 }
 
 /// Change in window or column size.
--- a/niri-visual-tests/src/cases/layout.rs
+++ b/niri-visual-tests/src/cases/layout.rs
@@ -266,6 +266,7 @@ impl TestCase for Layout {
             .monitor_for_output(&self.output)
             .unwrap()
             .render_elements(renderer, RenderTarget::Output, true)
+            .flat_map(|(_, iter)| iter)
             .map(|elem| Box::new(elem) as _)
             .collect()
     }
--- a/src/handlers/xdg_shell.rs
+++ b/src/handlers/xdg_shell.rs
@@ -153,7 +153,7 @@ impl XdgShellHandler for State {
 
         match start_data {
             PointerOrTouchStartData::Pointer(start_data) => {
-                let grab = MoveGrab::new(start_data, window);
+                let grab = MoveGrab::new(start_data, window, false);
                 pointer.set_grab(self, grab, serial, Focus::Clear);
             }
             PointerOrTouchStartData::Touch(start_data) => {
@@ -315,6 +315,9 @@ impl XdgShellHandler for State {
             return;
         } else if let Some(output) = self.niri.layout.active_output() {
             let layers = layer_map_for_output(output);
+
+            // FIXME: somewhere here we probably need to check is_overview_open to match the logic
+            // in update_keyboard_focus().
 
             if layers
                 .layer_for_surface(&root, WindowSurfaceType::TOPLEVEL)
--- a/src/handlers/xdg_shell.rs
+++ b/src/handlers/xdg_shell.rs
@@ -1062,6 +1062,19 @@ impl State {
         // The target geometry for the positioner should be relative to its parent's geometry, so
         // we will compute that here.
         let mut target = Rectangle::from_size(output_geo.size);
+
+        // Background and bottom layer popups render below the top and the overlay layer, so let's
+        // put them into the non-exclusive zone.
+        //
+        // FIXME: ideally this should use the "top and overlay layer" non-exclusive zone, but
+        // Smithay only computes the "all layers" non-exclusive zone atm.
+        //
+        // FIXME: related to the above, top layer popups should use the "overlay layer"
+        // non-exclusive zone.
+        if matches!(layer_surface.layer(), Layer::Background | Layer::Bottom) {
+            target = map.non_exclusive_zone();
+        }
+
         target.loc -= layer_geo.loc;
         target.loc -= get_popup_toplevel_coords(popup);
 
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -29,7 +29,7 @@ use smithay::input::touch::{
 };
 use smithay::input::SeatHandler;
 use smithay::output::Output;
-use smithay::utils::{Logical, Point, Rectangle, Transform, SERIAL_COUNTER};
+use smithay::utils::{Logical, Point, Rectangle, Size, Transform, SERIAL_COUNTER};
 use smithay::wayland::keyboard_shortcuts_inhibit::KeyboardShortcutsInhibitor;
 use smithay::wayland::pointer_constraints::{with_pointer_constraint, PointerConstraint};
 use smithay::wayland::selection::data_device::DnDGrab;
@@ -389,6 +389,15 @@ impl State {
                         .get_pointer()
                         .unwrap()
                         .unset_grab(this, serial, time);
+                    this.niri.suppressed_keys.insert(key_code);
+                    return FilterResult::Intercept(None);
+                }
+
+                if this.niri.keyboard_focus.is_overview()
+                    && pressed
+                    && matches!(raw, Some(Keysym::Escape | Keysym::Return))
+                {
+                    this.niri.layout.toggle_overview();
                     this.niri.suppressed_keys.insert(key_code);
                     return FilterResult::Intercept(None);
                 }
@@ -1915,10 +1924,18 @@ impl State {
             Action::ClearDynamicCastTarget => {
                 self.set_dynamic_cast_target(CastTarget::Nothing);
             }
+            Action::ToggleOverview => {
+                self.niri.layout.toggle_overview();
+                self.niri.queue_redraw_all();
+            }
         }
     }
 
     fn on_pointer_motion<I: InputBackend>(&mut self, event: I::PointerMotionEvent) {
+        let was_inside_hot_corner = self.niri.pointer_inside_hot_corner;
+        // Any of the early returns here mean that the pointer is not inside the hot corner.
+        self.niri.pointer_inside_hot_corner = false;
+
         // We need an output to be able to move the pointer.
         if self.niri.global_space.outputs().next().is_none() {
             return;
@@ -2095,6 +2112,18 @@ impl State {
 
         pointer.frame(self);
 
+        // contents_under() will return no surface when the hot corner should trigger.
+        if pointer.current_focus().is_none() {
+            let hot_corner = Rectangle::from_size(Size::from((1., 1.)));
+            if let Some((_, pos_within_output)) = self.niri.output_under(pos) {
+                let inside_hot_corner = hot_corner.contains(pos_within_output);
+                if inside_hot_corner && !was_inside_hot_corner {
+                    self.niri.layout.toggle_overview();
+                }
+                self.niri.pointer_inside_hot_corner = inside_hot_corner;
+            }
+        }
+
         // Activate a new confinement if necessary.
         self.niri.maybe_activate_pointer_constraint();
 
@@ -2119,6 +2148,10 @@ impl State {
         &mut self,
         event: I::PointerMotionAbsoluteEvent,
     ) {
+        let was_inside_hot_corner = self.niri.pointer_inside_hot_corner;
+        // Any of the early returns here mean that the pointer is not inside the hot corner.
+        self.niri.pointer_inside_hot_corner = false;
+
         let Some(pos) = self.compute_absolute_location(&event, None).or_else(|| {
             self.global_bounding_rectangle().map(|output_geo| {
                 event.position_transformed(output_geo.size) + output_geo.loc.to_f64()
@@ -2163,6 +2196,18 @@ impl State {
         );
 
         pointer.frame(self);
+
+        // contents_under() will return no surface when the hot corner should trigger.
+        if pointer.current_focus().is_none() {
+            let hot_corner = Rectangle::from_size(Size::from((1., 1.)));
+            if let Some((_, pos_within_output)) = self.niri.output_under(pos) {
+                let inside_hot_corner = hot_corner.contains(pos_within_output);
+                if inside_hot_corner && !was_inside_hot_corner {
+                    self.niri.layout.toggle_overview();
+                }
+                self.niri.pointer_inside_hot_corner = inside_hot_corner;
+            }
+        }
 
         self.niri.maybe_activate_pointer_constraint();
 
@@ -2235,10 +2280,54 @@ impl State {
             self.niri.pointer_hidden = false;
             self.niri.tablet_cursor_location = None;
 
+            let is_overview_open = self.niri.layout.is_overview_open();
+
+            if is_overview_open && !pointer.is_grabbed() && button == Some(MouseButton::Right) {
+                if let Some((output, ws)) = self.niri.workspace_under_cursor(true) {
+                    let ws_id = ws.id();
+                    let ws_idx = self.niri.layout.find_workspace_by_id(ws_id).unwrap().0;
+
+                    self.niri.layout.focus_output(&output);
+
+                    let location = pointer.current_location();
+                    let start_data = PointerGrabStartData {
+                        focus: None,
+                        button: button_code,
+                        location,
+                    };
+                    self.niri
+                        .layout
+                        .view_offset_gesture_begin(&output, Some(ws_idx), false);
+                    let grab = SpatialMovementGrab::new(start_data, output, ws_id, true);
+                    pointer.set_grab(self, grab, serial, Focus::Clear);
+                    self.niri
+                        .cursor_manager
+                        .set_cursor_image(CursorImageStatus::Named(CursorIcon::AllScroll));
+
+                    // FIXME: granular.
+                    self.niri.queue_redraw_all();
+
+                    // Don't activate the window under the cursor to avoid unnecessary
+                    // scrolling when e.g. Mod+MMB clicking on a partially off-screen window.
+                    return;
+                }
+            }
+
             if button == Some(MouseButton::Middle) && !pointer.is_grabbed() {
                 let mod_down = modifiers_from_state(mods).contains(mod_key.to_modifiers());
                 if mod_down {
-                    if let Some(output) = self.niri.output_under_cursor() {
+                    let output_ws = if is_overview_open {
+                        self.niri.workspace_under_cursor(true)
+                    } else {
+                        self.niri.output_under_cursor().and_then(|output| {
+                            let mon = self.niri.layout.monitor_for_output(&output)?;
+                            Some((output, mon.active_workspace_ref()))
+                        })
+                    };
+
+                    if let Some((output, ws)) = output_ws {
+                        let ws_id = ws.id();
+
                         self.niri.layout.focus_output(&output);
 
                         let location = pointer.current_location();
@@ -2247,7 +2336,7 @@ impl State {
                             button: button_code,
                             location,
                         };
-                        let grab = SpatialMovementGrab::new(start_data, output);
+                        let grab = SpatialMovementGrab::new(start_data, output, ws_id, false);
                         pointer.set_grab(self, grab, serial, Focus::Clear);
                         self.niri
                             .cursor_manager
@@ -2269,12 +2358,14 @@ impl State {
                 // Check if we need to start an interactive move.
                 if button == Some(MouseButton::Left) && !pointer.is_grabbed() {
                     let mod_down = modifiers_from_state(mods).contains(mod_key.to_modifiers());
-                    if mod_down {
+                    if is_overview_open || mod_down {
                         let location = pointer.current_location();
                         let (output, pos_within_output) = self.niri.output_under(location).unwrap();
                         let output = output.clone();
 
-                        self.niri.layout.activate_window(&window);
+                        if !is_overview_open {
+                            self.niri.layout.activate_window(&window);
+                        }
 
                         if self.niri.layout.interactive_move_begin(
                             window.clone(),
@@ -2286,11 +2377,14 @@ impl State {
                                 button: button_code,
                                 location,
                             };
-                            let grab = MoveGrab::new(start_data, window.clone());
+                            let grab = MoveGrab::new(start_data, window.clone(), is_overview_open);
                             pointer.set_grab(self, grab, serial, Focus::Clear);
-                            self.niri
-                                .cursor_manager
-                                .set_cursor_image(CursorImageStatus::Named(CursorIcon::Move));
+
+                            if !is_overview_open {
+                                self.niri
+                                    .cursor_manager
+                                    .set_cursor_image(CursorImageStatus::Named(CursorIcon::Move));
+                            }
                         }
                     }
                 }
@@ -2365,7 +2459,20 @@ impl State {
                     }
                 }
 
-                self.niri.layout.activate_window(&window);
+                if !is_overview_open {
+                    self.niri.layout.activate_window(&window);
+                }
+
+                // FIXME: granular.
+                self.niri.queue_redraw_all();
+            } else if let Some((output, ws)) = is_overview_open
+                .then(|| self.niri.workspace_under_cursor(false))
+                .flatten()
+            {
+                let ws_idx = self.niri.layout.find_workspace_by_id(ws.id()).unwrap().0;
+
+                self.niri.layout.focus_output(&output);
+                self.niri.layout.toggle_overview_to_workspace(ws_idx);
 
                 // FIXME: granular.
                 self.niri.queue_redraw_all();
@@ -2437,23 +2544,63 @@ impl State {
         let horizontal_amount_v120 = event.amount_v120(Axis::Horizontal);
         let vertical_amount_v120 = event.amount_v120(Axis::Vertical);
 
+        let is_overview_open = self.niri.layout.is_overview_open();
+
         // Handle wheel scroll bindings.
         if source == AxisSource::Wheel {
             // If we have a scroll bind with current modifiers, then accumulate and don't pass to
             // Wayland. If there's no bind, reset the accumulator.
             let mods = self.niri.seat.get_keyboard().unwrap().modifier_state();
             let modifiers = modifiers_from_state(mods);
-            if self.niri.mods_with_wheel_binds.contains(&modifiers) {
+            let should_handle =
+                is_overview_open || self.niri.mods_with_wheel_binds.contains(&modifiers);
+            if should_handle {
                 let horizontal = horizontal_amount_v120.unwrap_or(0.);
                 let ticks = self.niri.horizontal_wheel_tracker.accumulate(horizontal);
                 if ticks != 0 {
-                    let config = self.niri.config.borrow();
-                    let bindings = &config.binds;
-                    let bind_left =
-                        find_configured_bind(bindings, mod_key, Trigger::WheelScrollLeft, mods);
-                    let bind_right =
-                        find_configured_bind(bindings, mod_key, Trigger::WheelScrollRight, mods);
-                    drop(config);
+                    let (bind_left, bind_right) = if is_overview_open {
+                        if modifiers.is_empty() {
+                            let bind_left = Some(Bind {
+                                key: Key {
+                                    trigger: Trigger::WheelScrollLeft,
+                                    modifiers: Modifiers::empty(),
+                                },
+                                action: Action::FocusColumnLeft,
+                                repeat: true,
+                                cooldown: None,
+                                allow_when_locked: false,
+                                allow_inhibiting: false,
+                                hotkey_overlay_title: None,
+                            });
+                            let bind_right = Some(Bind {
+                                key: Key {
+                                    trigger: Trigger::WheelScrollRight,
+                                    modifiers: Modifiers::empty(),
+                                },
+                                action: Action::FocusColumnRight,
+                                repeat: true,
+                                cooldown: None,
+                                allow_when_locked: false,
+                                allow_inhibiting: false,
+                                hotkey_overlay_title: None,
+                            });
+                            (bind_left, bind_right)
+                        } else {
+                            (None, None)
+                        }
+                    } else {
+                        let config = self.niri.config.borrow();
+                        let bindings = &config.binds;
+                        let bind_left =
+                            find_configured_bind(bindings, mod_key, Trigger::WheelScrollLeft, mods);
+                        let bind_right = find_configured_bind(
+                            bindings,
+                            mod_key,
+                            Trigger::WheelScrollRight,
+                            mods,
+                        );
+                        (bind_left, bind_right)
+                    };
 
                     if let Some(right) = bind_right {
                         for _ in 0..ticks {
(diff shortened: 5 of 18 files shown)


Vortriz:
Since now the wallpapers as well as `background-color` property of `outputs` is limited to the workspaces, is there a way to configure this default grey background that is now present, through a config option perhaps.

Grey looks weird while switching workspaces or in the overview itself too, when the wallpaper is very brightly colored.

YaLTeR:
In the code I called that the "backdrop", so I guess we could have a `backdrop-color`.

Quackdoc:
A few things after using it for a bit, Very nice work so far, the hot corner as it is feels a bit... overzealous? It would be nice to have a higher default delay.  Also it would be really nice if keyboard control had an easier mode too. I found myself often trying to use the arrow keys themselves to navigate up and down. 

as for the questions

bugs and crashes : None so far
performance issues: Far from it, I have a a380, 4k screen, 1080p screen and a 960p screen and it is all very fluid even with a large amount of dummy windows open, don't have anything to stress the gpu at the moment however.

YaLTeR:
> the hot corner as it is feels a bit... overzealous? It would be nice to have a higher default delay.

When do you find hitting it by accident?

> I found myself often trying to use the arrow keys themselves to navigate up and down.

So you suggest arrows for focus-workspace-up/down? (As opposed to something like focus-window-up/down.) And left/right to focus-column-left/right?

Vortriz:
Also, I have rendering issues with window borders. There is a sort of flickering in the borders when
1. The windows move, not DnD, just regular horizontal movement.
2. During the transition between the normal view and the overview view.

Is this what you mean by
> 1 pixel jank with thin borders, gaps, and in general in the overview

I was trying to screenshot it and realized that the inbuilt tool takes screenshot of the focused workspace only. Is it possible to implement taking screenshot of the overview itself?

YaLTeR:
> Is this what you mean by

Yes

> I was trying to screenshot it and realized that the inbuilt tool takes screenshot of the focused workspace only. Is it possible to implement taking screenshot of the overview itself?

Huh? The built-in screenshot UI works fine here.

debugloop:
Awesome work 🚀 Everything is working as a expect it to so far. Mouse interaction and drag and drop is unexpected but awesome.

One thing: I use very thin 1px borders with no gaps and no ring, and it doesn't come out great. You can see the two blue dots at the bottom of the right hand terminal window here, which is all that remains for the focus indicator. Maybe you can ensure that the border remains at least 1px? Not sure how to tackle this best otherwise. On the upper workspace browser window, you can also see one of the focus pixels.

image

Quackdoc:
> When do you find hitting it by accident?

I have a high DPI mouse I use on the screen, so when I find myself going to focus tabs in cosmic-edit I often find that I overshoot the mark.

> So you suggest arrows for focus-workspace-up/down? (As opposed to something like focus-window-up/down.) And left/right to focus-column-left/right?

It's hard to say, intuitively, I think regardless of how it's done it's going to conflict with expectations depending on the user. Personally I use single window work spaces most of the time on my laptop, so that would by far be way more intuitive. On the other hand, someone using less workspaces more would for sure find it more intuitive to leave it default.

I think if there was a way to merge the behavior, so it would focus workspace if there is no window next would be ideal. Otherwise maybe this would be a good candidate for a config? `overview-arrows-workspace  false` or something along those lines? I am fine simply rebinding myself as I don't often find myself actually using both behaviors a lot at the same time.

I think it would also be nice to not need to hold `mod` for the for the `focus-{window,workspace}` to be triggered.

Just brainstorming here, maybe this could be punted to a config too. I suppose maybe overview could even go so far to "toggle" mod? It doesn't seem like key presses are passed to the focused window anyways in overview mode, so maybe a `overview-toggles-mod false` could work? not sure how well that would work, might be a bit clunky, but I could see myself getting used to it at least personally. 

This could be really beneficial for "mini" keyboards.

debugloop:
Oops, I missed that one, sorry. Alright, I got some more hopefully non-known weird things with 2px borders. From top to bottom workspace in the screenshot:

1. With a 50% window and several smaller ones, focusing left to right consistently misses drawing the left border on the newly focused window. Just the second window on the workspace, the one right of the 50% window. Might be something about my specific screen dimensions and rounding I guess.
2. The border with my vertical bar never shows.
3. Windows "overflowing" by bar hide the bar. I feel it'd be more natural to have the bar drawn on top of the window? Maybe that's just me.

image

YaLTeR:
> I think if there was a way to merge the behavior, so it would focus workspace if there is no window next would be ideal.

Already exists: focus-window-or-workspace-down. Question is if that's what should be bound to arrows or focus-workspace-down.

> I think it would also be nice to not need to hold mod for the for the focus-{window,workspace} to be triggered.

I'm reluctant to do this in general because maybe in the future we'll have some kind of typing, then it will conflict.

> This could be really beneficial for "mini" keyboards.

I'm on a 40% keyboard, don't really feel the need for this. If anything, I didn't even think about arrows before you pointed them out, because they are more annoying to press because I have them on a layer. But I think it's a good idea.

YaLTeR:
> I got some more hopefully non-known weird things with 2px borders.

All stem from the same problem.

> Windows "overflowing" by bar hide the bar. I feel it'd be more natural to have the bar drawn on top of the window?

Your bar is on the bottom layer. Bottom layer draws below windows.

mcs0042:
Thanks for writing this!

So far, the only issue I've run into not addressed in the original post is as follows: Pressing Return in overview mode turns off the overview (as expected), but it also looks like some (maybe all?) keybinds involving Return will turn off the overview instead of doing what they're supposed to when pressed in overview mode. For example, if I set Mod+Return to open fuzzel, then pressing Mod+Return in normal mode opens fuzzel as expected, but pressing Mod+Return in overview mode deactivates the overview without opening fuzzel.

Quackdoc:
> Already exists: focus-window-or-workspace-down. Question is if that's what should be bound to arrows or focus-workspace-down.

ah, I somehow completely missed that config entirely, If this config exists I think that's good enough.

> I'm reluctant to do this in general because maybe in the future we'll have some kind of typing, then it will conflict.

It would be a nice feature for some input devices, but if it could conflict, then it would for sure at most be punted by a off by default config. device expanded below. 

> I'm on a 40% keyboard, don't really feel the need for this. If anything, I didn't even think about arrows before you pointed them out, because they are more annoying to press because I have them on a layer. But I think it's a good idea.

for context, I use an air mouse a lot, (wechip one to be specific) with the standard "remote" on top and keyboard on bottom, which actually requires the user to flip the remote over to input a super key command. It's an odd device, but for a TV config (which I do actually have mostly done) this with the workspaces pretty much finishes off my setup (barring a missing launcher, but it is what it is).

YaLTeR:
> but it also looks like some (maybe all?) keybinds involving Return will turn off the overview instead of doing what they're supposed to when pressed in overview mode

Good catch.

> I use an air mouse a lot, (wechip one to be specific) with the standard "remote" on top and keyboard on bottom, which actually requires the user to flip the remote over to input a super key command. It's an odd device, but for a TV config (which I do actually have mostly done) this with the workspaces pretty much finishes off my setup (barring a missing launcher, but it is what it is).

Huh, that's interesting. You should record some sort of video on how it works, I'm curious.

Didn't you also run niri on a tablet? Really driving it on all the weird setups :p

YaLTeR:
> Personally I look forward and see if my regular workspace/column navigation/move keybindings still work for familiarity and some continuity. Especially because I am trying to have a keyboard centric workflow.
>
> For example, changing focus and moving columns around. I am fine having these binds configurable which I would rebind to the keys I already have.

They work.

RenanEuzebio:
It would be great to have an overview-padding config option. The current overview windows fills the full monitor, which works well by default—but adding padding (e.g., overview-padding-left 200) would make it easier to place layer shell components like fuzzel without overlapping windows.

YaLTeR:
> It would be great to have an overview-padding config option. The current overview windows fill the full monitor, which works well by default—but adding padding (e.g., overview-padding-left 200) would make it easier to place layer shell components without having them overlap the overview windows.

I'm not sure I understand how this helps? The windows will always go below layer-shell, just like they already do without the overview. I suggest using a solid background for layer-shell. For example, you can match the niri backdrop background. That will make it look more or less how you image it I think.

RenanEuzebio:
> > It would be great to have an overview-padding config option. The current overview windows fill the full monitor, which works well by default—but adding padding (e.g., overview-padding-left 200) would make it easier to place layer shell components without having them overlap the overview windows.
> 
> I'm not sure I understand how this helps? The windows will always go below layer-shell, just like they already do without the overview. I suggest using a solid background for layer-shell. For example, you can match the niri backdrop background. That will make it look more or less how you image it I think.

It's a small UI/UX detail, thanks for taking the time to discuss it, but yeah, in my opinion, for the overview mode specifically, windows drawing on front of components causes a visual overlap that conflict their UX (if they have gaps for example).

Adding a config like:

overview-padding-left 200 (right, top, and bottom too depending on where u want the layer shell component)

Would shift the window layout in overview. Which helps mainly because it would look better and also have better visibility (both windows and layer-shell components remain fully visible)

An idea of what I mean is like how the Vertical Workspaces gnome extension reserves space for the dock ans search.

In this very poorly drawn idea I try to show this too (dashed lines are what the padding is blocking)

red line is there to visualize where padding would be. 

image

Without this config, the windows would just draw behind the component, which for when there's gaps is much less clean.

YaLTeR:
If it's fuzzel then you can pass it CLI options to set a background color matching the niri backdrop, effectively doing what you see on the screen? Apart from uncentered workspaces, but that is a separate issue (not sure how much a fix is needed here).

YaLTeR:
> In addition to aforementioned backdrop-color, I'd appreciate a way to tweak the zoom level. Possibly as a factor on top of the current logic, maybe even expressed as n workspaces or something (i.e. the current implementation is 3 full workspaces visible, I feel like 1 full workspace and most of the two adjacent once would be enough).

Hm. I mean, it shouldn't be hard to add, but I picked the current zoom level so that:
1. You don't need to scroll or dnd too much to move windows around adjacent workspaces.
2. Part of the next workspace is visible to indicate that there are more workspaces that way.

debugloop:
Yeah, using the mouse is not really a use case on my end. I enjoy using the overview when reorganizing my windows, it's why I think your solution that keeps the original keybinds for focusing and moving intact is so nice. Possibly, I'd also use it to see whether there are more windows to the right (and then to move them somewhere).

Hence, one workspace with parts of the adjacent ones would be enough. Possibly it could also be an optional parameter to the action itself, and not a config option at all.

RenanEuzebio:
> If it's fuzzel then you can pass it CLI options to set a background color matching the niri backdrop, effectively doing what you see on the screen? Apart from uncentered workspaces, but that is a separate issue (not sure how much a fix is needed here).

Good point, that works well enough unless the user wants a different color for it. Another possible problem with this approach is it would not work if in a future PR backdrop wallpaper support gets added.

Vortriz:
Also, instead of `backdrop-color`, maybe keep `backdrop`. future proofing, incase you decide to add support for backdrop images.

YaLTeR:
> Good point, that works well enough unless the user wants a different color for it. Another possible problem with this approach is it would not work if in a future PR backdrop wallpaper support gets added.

Maybe can think about it sometime in the future

> Also, instead of backdrop-color, maybe keep backdrop. future proofing, incase you decide to add support for backdrop images.

We already have `background-color` which will have the exact same problem, better to be consistent.

arkboix:
This is great 🤩
everything seems to work smoothly, same for hot corners, i don't feel like its too sensitive.

currently it's using a solid color for the background,
im not sure if its possible to maybe use an image as a background 🤔 with less brightness?

Changing just the background color would be great!
we could also do just a transparent background, thought that'd be weird i guess...

other than that, awesome 👍

Quackdoc:
> Huh, that's interesting. You should record some sort of video on how it works, I'm curious.

Here is a video of a fairly vanilla config, using fuzzel but one day I hope to find a launcher that works a bit better. IMO Niri will be the defacto TV experience. Sorry for the poor video quality, my camera is remarkably bad and LOS refuses to encode the video right, and half the time, actually record with the wide angle lens. 

In the video I am manually pressing `mod` to trigger the menu and navigation. I have overview configured to the dedicated button which is `XF86Home` With a proper launcher, controller mapping software, and an OSK niri would be a great console config too https://files.catbox.moe/qurlwg.webm 

> Didn't you also run niri on a tablet? Really driving it on all the weird setups :p

yup. I have an update for that comming soonish but I am just at the tail end of a large move.

RenanEuzebio:
> > Huh, that's interesting. You should record some sort of video on how it works, I'm curious.
> 
> Here is a video of a fairly vanilla config, using fuzzel but one day I hope to find a launcher that works a bit better. IMO Niri will be the defacto TV experience. Sorry for the poor video quality, my camera is remarkably bad and LOS refuses to encode the video right, and half the time, actually record with the wide angle lens.
> 
> In the video I am manually pressing `mod` to trigger the menu and navigation. I have overview configured to the dedicated button which is `XF86Home` With a proper launcher, controller mapping software, and an OSK niri would be a great console config too https://files.catbox.moe/qurlwg.webm
> 
> > Didn't you also run niri on a tablet? Really driving it on all the weird setups :p
> 
> yup. I have an update for that comming soonish but I am just at the tail end of a large move.

Omg that's amazing lol

erdii:
The is really really awesome!

Some feedback after several hours of toying around with this:

Hot Corner works fine for me, although I think some users would like to be able turn it off or place it into another corner of the screen.

Config proposal:
```
overview {
  // Uncomment this line to disable the hot-corner.
  // disable-hot-corner

  // Where to place the hot-corner on each output, options are:
  // "upper-left", default
  // "upper-right",
  // "lower-left",
  // "lower-right"
  hot-corner-placement "upper-left"

  // Other overview specific options could be added to this section as well.
}
```

---

Toggling the overview with `Mod+Tab` bound to `toggle-overview` works well especially in combination with selecting windows with `Mod+H/J/K/L/Arrows` (with `k/j/up/down` bound to `focus-window-or-workspace-up/down`).

---

I personally don't need the special treatment of `Escape` or `Enter` and find it distracting because I have bound `Mod+Enter` to spawn a terminal. Maybe this could be dropped or made configurable as well?
The same applies for scrolling. Its weird that I have to stop holding Mod when the overview is open to be able to scroll through windows/workspaces, even though I Mod+Scroll when the overview is closed.

---

Drag and drop of windows in the overview is really neat!
It disables moving the overview around through all means though (Scrolling, `focus-...-up/down/left/right` but when the window is dropped the moves get "unbuffered" and the overview jumps to the place that was navigated to. 

Mod+Dragging a window in and out of the overview (either through the hot corner or `toggle-overview` works very well.

---

The zoom level works fine for me.
So does the gray background.
No crashes.
No glitches noticed.

YaLTeR:
> https://files.catbox.moe/qurlwg.webm

That is so fascinating.

> Hot Corner works fine for me, although I think some users would like to be able turn it off or place it into another corner of the screen.
>
> Config proposal:

I have a feeling that it should rather be a `hot-corner {}` kind of section so that in the future it's possible to bind other commands and/or multiple hot corners. But I'm not sure how exactly the syntax should look like; ideally it should be something that allows to keep just top-left for now, but extensible to other corners sometime later.

> I personally don't need the special treatment of Escape or Enter and find it distracting because I have bound Mod+Enter to spawn a terminal. Maybe this could be dropped or made configurable as well?

It is a bug that it overwrites the binds. It won't.

>  The same applies for scrolling. Its weird that I have to stop holding Mod when the overview is open to be able to scroll through windows/workspaces, even though I Mod+Scroll when the overview is closed.

Good catch. This shouldn't prevent Mod+scrolling from working.

> It disables moving the overview around through all means though (Scrolling, focus-...-up/down/left/right but when the window is dropped the moves get "unbuffered" and the overview jumps to the place that was navigated to.

Yes, known issue.

armerpunkt:
In general, it's working fairly well for me. Just a couple comments/suggestions:

1. I'm used to doing the four finger swipe up for Mission Control in MacOS. So I keep swiping up, which makes the windows start to zoom in but then they stop and go back to normal. Perhaps that shouldn't do anything unless you are in the overview already? Additionally, it would be nice if there was an option to reverse the swipe directions (even if it's just a debug option like `reverse-overview-swipe-direction` or something).

2. This has already commented on but I would really really like to be able to disable the hot corner. First, because I don't like them in general and always disable it in any desktop environment that I'm using. And second because I have a multiple monitor setup with 3 across and one on top. So I keep inadvertently triggering it when trying to move from the bottom right monitor to the top center monitor or vice versa.  I know you mentioned that the monitor positions could be modified but it would really throw me off to have to intentionally travel through the bottom center monitor to go between the two when I'm so used to just going through the corners. 

So I like erdii's suggestion of an overview configuration section. As that would also allow adding a swipe direction option.

Your suggestion for a hot-corner section is intriguing as well. Though likely outside the scope of this PR. You could do something like:
```
hot-corners {
    top-left { disable; }
    top-right { toggle-overview; }
    bottom-right { spawn "swaylock"; }
}
```
So if someone doesn't specify the top-left option, it can still do the normal toggle-overview, but if they don't want the hot corner there, they can disable it or move it to another corner. 

Another thought I had, though not sure if it's practical, is to put it in the output config, then users could configure different hot corners for each monitor. Although this would be annoying to have to disable them for each monitor. So it may be better to just have a hot-corners option of `limit-to-output` to specify the hot-corners are only active on that one monitor. Or if per monitor hot corners sounds nice, there could be something like a `disable-default-hot-corners` debug option.

YaLTeR:
> I'm used to doing the four finger swipe up for Mission Control in MacOS. So I keep swiping up, which makes the windows start to zoom in but then they stop and go back to normal. Perhaps that shouldn't do anything unless you are in the overview already? Additionally, it would be nice if there was an option to reverse the swipe directions (even if it's just a debug option like reverse-overview-swipe-direction or something).

The overview opens on four finger swipe up for me. Could it be the natural-scrolling setting? Should this gesture be unaffected by natural scrolling?

> Your suggestion for a hot-corner section is intriguing as well. Though likely outside the scope of this PR. You could do something like:

Yeah, that is closer to what I'm thinking. Or maybe this could be in the `gestures {}` section. Not sure yet.

> Another thought I had, though not sure if it's practical, is to put it in the output config, then users could configure different hot corners for each monitor.

Hmm... Yeah, not very clear what's the best way to structure the config here. I mean, could have it in several places (or rather do one thing while keeping in mind the possible extensibility).

YaLTeR:
> (1a) touchpad gestures and keyboard bindings will operate over two different workspaces often leading me to close the wrong window on a different workspace. As in my head i was interacting with the one i was scrolling. demo in gif

Yeah, touchpad operates on the workspace under cursor, whereas keyboard operates on the focused workspace. I think this makes much more sense, though I can see how it can get confusing if you use both keyboard and touchpad at the same time.

FWIW this is already how it works e.g. with outputs, it's just not that prominent there.

> (1b) so workspace borders, or some form of better indication of 'bind-active' workspace/pane during overview. if you main thin (1px-2px) focus rings/borders i find them very hard to read in the overview, not legible enough to know where i am on the grid. a slight zoom on active workspace even maybe?

I was thinking about adding a focus ring or something, not sure yet.

> (2) dragging a fullscreen window from one workspace to another loses the fullscreen state

Hmm, so this is a normal interactive move behavior. I guess it makes less sense in the overview. Not sure how painful it would be to special case.

> losing that direct-scanout on old gear makes for a big hiccups with some apps (browsers primarily)

Note that you won't get direct scanout in the overview regardless. Direct scanout only works when the surface covers the entire monitor. (Well unless it scans out to an overlay plane, which are disabled by default in niri.)

> (3) visual indicator of topmost workspace. I did attempt to toss a window above workspace one thinking i could do that. it felt like the natural thing to do.

What do you mean? Right now it should put the window on a new workspace above first. I think that makes sense?

armerpunkt:
> The overview opens on four finger swipe up for me. Could it be the natural-scrolling setting? Should this gesture be unaffected by natural scrolling?

Ah!  I had natural-scroll disabled, and enabling it does make it trigger the overview when swiping up.  I'd say the gesture should not be affected by the natural scrolling setting (though I haven't checked how it's handled other than in MacOS, so others might disagree). But, if not too much work, I do think having the direction be configurable would still be handy.

stefonarch:
Have to admit that I didn't read all the wall here so maybe it's a duplicate: 
What I'm missing is a "close" function for  mouse/touchpad (middle click?). Mod+Q works. 
Really nice work anyway as usual!

Also closing the overview when launching an app?

For some reason it's 4fingerscroll down here.

martinetd:
Awesome, thank you!

I also haven't found any obvious problem, just a few random thoughts:
- in overview mode, it changes of focus are properly sent to each window but it looks like regular input (e.g. anything that's not mapped) is blocked -- I assume that's on purpose, but I don't see the benefit, and would actually think it'd be nice to allow input (e.g. spawn a new shell and run a command while in overview mode, or cleanly exit a shell without relying on Mod+Q (^D)
- focus-follows-mouse might make sense to keep as well for non-touchpad users? That might resolve the above concern about closing window in the wrong workspace as well, if focus follows I'd assume it would close the correct window.
- regarding bindings, this is not something I've seen in niri, but having binding "modes" like sway would allow configuring binds for overlay e.g.
```
binds {
    // normal binds
    Mod+Return...

    mode "overview" {
        Down { focus-window-or-workspace-down; }
        // +up/left/right/etc
        Escape { exit-mode; }
        MouseMiddle { close-window; }
        // also need to repeat any other binding from default mode if not inherited?
        // Probably simpler to inherit and allow overriding to noop key
    }
}
```
That'd allow more flexibility, but if there isn't anything like that right now it's probably out of scope and best to keep thing focused here with hardcoded keys as a first step

Anyway, I think it's good enough as is and probably ok to adjust behaviour a bit as we figure it out; I'll certainly be using this branch going forward until it's merged :) Thank you!

YaLTeR:
> What I'm missing is a "close window" function for mouse/touchpad (middle click?).

I agree, though I'm not sure that something as easy as a middle click is enough considering some windows may not save their state etc. GNOME Shell shows X buttons in the overview that you can click, but this sounds like some effort to implement. Maybe for later, or when we do more spacious window layout.

> Also closing the overview when launching an app?

We can't do this reliably when it's some launcher running the app instead of niri, because we don't know if it successfully launched something.

> in overview mode, it changes of focus are properly sent to each window but it looks like regular input (e.g. anything that's not mapped) is blocked -- I assume that's on purpose, but I don't see the benefit, and would actually think it'd be nice to allow input (e.g. spawn a new shell and run a command while in overview mode, or cleanly exit a shell without relying on Mod+Q (^D)

Yes, that's on purpose. It's hard to see where you're inputting, so to avoid wrong input. Besides, it allows for doing something more interesting in the overview down the line.

> focus-follows-mouse might make sense to keep as well for non-touchpad users? That might resolve the above concern about closing window in the wrong workspace as well, if focus follows I'd assume it would close the correct window.

See my comment above; focus-follows-mouse can't work without scrolling to the target window's workspace (would be hard to change), and that would be very disorienting.

> regarding bindings, this is not something I've seen in niri, but having binding "modes" like sway would allow configuring binds for overlay e.g.

Not sure this is necessary. Considering all normal binds already work. And we can have a few extra hardcoded ones.

YaLTeR:
Updates:
- Added backdrop-color output option
- Fixed Mod+Return binds getting eaten in the overview
- Fixed Mod+Scroll binds getting eaten in the overview
- Made four-finger swipe touchpad gesture to always go up regardless of natural-scroll 
- Fixed hot corner working inside the screenshot UI
- Added arrow binds to the overview (see how it feels if we should change from window-or-workspace to workspace, especially with tabbed columns)

stefonarch:
> Made four-finger swipe touchpad gesture to always go up regardless of natural-scroll

Ok, that's the reason it was down here. But actually I like it more down for it ;)

martinetd:
> focus-follows-mouse can't work without scrolling to the target window's workspace (would be hard to change), and that would be very disorienting.

Hmm, as I see with the current code the focus scrolling target workspace works (through keybinds) so it's probably not too bad but it'd make it scroll within the output's view of it, not the overview, which I guess can be a bit disorienting..
Ah, if opening from the hot corner it'd definitely be annoying as you "pass through" the top workspaces to go below, yeah.
I've also just noticed regular click (not drag) focuses window and exits overview mode, so this also address the "open overview, find whatever I'm looking for and focus it" use; alright.

>> regular input (e.g. anything that's not mapped) is blocked
>
> Yes, that's on purpose. It's hard to see where you're inputting, so to avoid wrong input. Besides, it allows for doing something more interesting in the overview down the line.

Ok, I still think it could be interesting to be able to enter input in a window (perhaps a toggle to allow/disallow input pass-through), but I'll wait for your more interesting something before playing with it :)
(Something like click on a widow that's not focused = focus and allow input without exiting overview, click on the focused window = exit overview? But also just spawning a new shell now that works with mod+enter..)

>> binding "modes" like sway would allow configuring binds for overlay
>
> Not sure this is necessary. Considering all normal binds already work. And we can have a few extra hardcoded ones.

Someone will definitely ask for configuring these hardcoded bindings at some point, but I agree it can/should come later so we can focus on more interesting behavioural questions first. Thanks!

erdii:
I have a keybind (alt+tab) that opens a floating fuzzel instance with a list of all open windows to pick one that will be focused. (basically [this script](https://github.com/YaLTeR/niri/discussions/352#discussioncomment-12489627)

Two observations:
- Focusing a window via `niri msg action focus-window ...` works fine but it doesn't close the overview if it is open.
- Scrolling is not passed to focused floating fuzzel window, but scrolls the overview instead

kchibisov:
I'd say that animation when focusing the window that is would have required extra scroll to get into is a bit weird. Since it's basically a speed up version of transition from overview + scroll in workspace. I'd prefer if it would just directly zoomed into a newly focused window instead of _zoom + fast scrolling_.

Though, I guess for now it's just what was present just got reused.

Quackdoc:
> Added arrow binds to the overview (see how it feels if we should change from window-or-workspace to workspace, especially with tabbed columns)

There arrow binding is really nice, It's very intuitive. Testing it on a larger range of devices, I would for sure like a method for controlling zoom level as well, Using it on a large TV I find I would need to zoom in quite a bit for it to work out.

YaLTeR:
> Scrolling is not passed to focused floating fuzzel window, but scrolls the overview instead

Good catch; I'll see if I can fix this.

> I'd say that animation when focusing the window that is would have required extra scroll to get into is a bit weird. Since it's basically a speed up version of transition from overview + scroll in workspace. I'd prefer if it would just directly zoomed into a newly focused window instead of zoom + fast scrolling.

Yes, this is the:

>  Zoom in + scroll window into view animation is not linear; I'm not yet sure if it's fixable or not for this PR, will have to look into it.

> There arrow binding is really nice, It's very intuitive. Testing it on a larger range of devices, I would for sure like a method for controlling zoom level as well, Using it on a large TV I find I would need to zoom in quite a bit for it to work out.

Would a setting for the overview zoom level suffice?

tsdh:
I just gave it a try and really like it.

IMHO, the zoom level is a bit too high.  The two constraints you mentioned are more than fulfilled.  I would expect that as much space is taken so that everything fits on the screen, i.e., either the widest workspace or the number of workspaces + one empty one defines the maximum horizontal or vertical space.  (Well, at least in my one-output scenario.)

image

Quackdoc:
> Would a setting for the overview zoom level suffice?

absolutely.

YaLTeR:
Updates:
- Added a zoom option. It's at the top level of the config. It's clamped to 0.75 at the top because above that you can barely see that there are more workspaces, which would make it confusing.
    ```
    overview {
        zoom 0.25 // The default
    }
    ```
    Please use this setting responsibly.
- Added a way to disable the hot corner:
    ```
    gestures {
        hot-corners {
            off
        }
    }
    ```
- Added touchpad two-finger gestures in the overview that work the same as three-finger gestures.
- Not directly related to overview, but removed cancellation from the three finger workspace switch gesture. Before, if you e.g. lifted one finger in the middle, it would revert to before the gesture, now it will behave as if you lifted all fingers.

boomskats:
I have nothing else to add apart from that I am so irrationally happy right now. I don't think this is normal.

tsdh:
> > Would a setting for the overview zoom level suffice?
> 
> absolutely.

Yeah. But what would the value be? I'd argue that 1.0 should mean "fits exactly on screen without free vertical or horizontal space" which would suggest the current value is about 0.6.

tsdh:
@YaLTeR Wouldn't it make sense that `overview.zoom` set to 1.0 would be the largest possible value which would ensure that it's zoomed as much that everything (including an empty workspace at the bottom) fits on screen but not more?

YaLTeR:
That doesn't sound like a good idea because then your workspaces will constantly zoom back and forth as you add or remove workspaces. 

The current zoom setting is literally the workspace zoom, i.e. 0.5 means that everything is 2x smaller than normal.

tsdh:
Oh, I see. That makes sense.

debugloop:
It turns out that the Arrow keys are nice. What are your thoughts on adding hjkl in addition?

YaLTeR:
Nah, I want to leave the characters as is.

YaLTeR:
I'll try running with zoom 0.5 for a bit and see how it feels. Maybe that should be the default? (Instead of 0.25)

HyperSuperMetaCtrl:
I think I prefer 0.5 for default overview zoom

erdii:
focus-window-previous works fine cross-workspace without the overview
but the overview breaks it in two ways:
- when the overview is open, focus-window-previous only does something once and then it gets stuck
- after i focus a new window by clicking it in the overview, focus-window-previous sometimes jumps to a window which is definitely not the previous one

https://github.com/user-attachments/assets/5d4d01de-a071-4b6f-b84b-2f449bd516d8

in the Video
- Alt+Tab is bound to focus-window-previous
- Mod+Tab is bound to toggle-overview

YaLTeR:
Yeah so this is actually due to how focus-window-previous tracking works using the keyboard focus. In the overview no window has keyboard focus so it breaks. I'm not sure there's an easy way to fix that in this PR.

tsdh:
> I'll try running with zoom 0.5 for a bit and see how it feels. Maybe that should be the default? (Instead of 0.25)

I like that it's big enough so that I can still read text in apps. But it's harder to re-arrange windows, especially when there are many workspaces.

Is there a way to change the zoom level dynamically using key/mouse bindings in the overview?

martinetd:
zoom level: played with it a bit yesterday and I like 0.33, although 0.5 is nice too depending on what I'm looking for.. Perhaps something like `switch-preset-column-width` that'd make the toggle-overview binding cycle (no zoom, zoom#1, zoom#2...n, back to no zoom) would work as far as keyboard shortcuts go? four finger scroll could likewise go back one level at a time, but I'm not sure about corner -- entering corner again?
I do think it's overkill though, probably just as good as is for now :)
(BTW, I've disabled the hot corner, I hit it accidentally a few times, and conversely it's also made me scroll left accidentally when trying to reach the corner... Thank you for the config item.)

Also on keybinds, I'm sure arrows are nice when using a full size keyboards but I'd rather keep using mod+hjkl in overlay; perhaps it's make sense to make focus-window-up/down behave as focus-window-or-workspace-up/down when in overlay mode? I've tried to change my default briefly but I'm not ready to make the change of the non-overlay binding...
It's pretty cool to navigate though, with a keybind for overlay hitting overlay, move around, hit enter or overlay bind again to get out of it. I can see this becoming a habit when moving further than 2 windows away.

YaLTeR:
> Also on keybinds, I'm sure arrows are nice when using a full size keyboards but I'd rather keep using mod+hjkl in overlay; perhaps it's make sense to make focus-window-up/down behave as focus-window-or-workspace-up/down when in overlay mode?

I don't think this is really warranted. And I mean, you have all your binds working, including workspace switching, just use those if you need.

tsdh:
> > Is there a way to change the zoom level dynamically using key/mouse bindings in the overview?
> 
> There isn't. I am not sure that this is the best approach. Just more manual work before you get to the actual thing you wanted to do (rearrange windows).

I don't think there's a *best* approach at all.  It depends on how you organize your windows, and if and how many outputs you have attached etc.

One thing I noticed is that a lot of display space isn't used due to the currently focused thing being centered both horizontally and vertically.  While it feels natural to do so, you have a lot of unused space on top/bottom when the first/last workspace is focused, same for unused space at the left/right when the first/last column is focused.  One could imagine a different indication of what's focused in which utilizes the screen space better.  Maybe a compromise would be that centering still happens but is restricted such that it cannot pull in free space in which case the difference would only be visible when focusing things at the edges...

YaLTeR:
I think the centering is staying regardless because doing it another way is gonna be way too annoying and edge casy. And I'm not sure it will feel any better either.

YaLTeR:
Updates:
- Switching workspaces (via keyboard or mouse scroll) now works during DnD in the overview.
- DnD hold in the overview to focus window now correctly scrolls to the target window.
- Due to refactors that enabled the above fixes, interactively dragging out a window from the tiling layout will now bring the newly active window into focus, as it used to do before the DnD scrolling update.

HigherOrderLogic:
> also we can't just take the bottom layer-shell layer because if you have desktop icons there then it will look weird. 

In case it hasnt been mentioned yet, can we have this as an option? I dont know about other but for me, I have no desktop icon on the home screen so it'll be fine by me.

YaLTeR:
Some more updates:
- `open-overview` and `close-overview` actions for scripts
- tablet pen tap now closes the overview focusing the target window/workspace
- some minor input fixes with an interactively moved window

YaLTeR:
Updates: added touch support. The entire thing. Somehow it completely worked first try.

https://github.com/user-attachments/assets/77c7fd18-cea4-4771-8070-328f0fe7b343

Quackdoc:
Touch support works phenomenally, for me on my "traditional like" tablet. if I had to state a request, it would be that you could "fling" across workspace and let it scroll, but honestly, I don't even actually care that much, it's just a back of head thought. Very smooth and responsive. 

Perhaps a way to trigger a "close window" would be nice but that only saves a button press so it too is not something I actually care that much about. It would be nice, but it's not like I found myself needing it, the bar I use works fine for that anyways.

YaLTeR:
Good to hear!

> if I had to state a request, it would be that you could "fling" across workspace and let it scroll

It already kinda works this way, just the deceleration is fairly strong. If it's not a big deal I'll leave it as is, at least for now.

> Perhaps a way to trigger a "close window" would be nice

I agree, but I think this will be for later (and need to figure out the design). Could try some kind of X button at the corner of every window. Or @kchibisov suggested maybe some trash can area where you can drop a window to close it. Will need to experiment. This also somewhat depends on where the overall overview design will go, like if we end up making windows smaller with large gaps, that will allow for more space for the X buttons.

Vortriz:
> Or @kchibisov suggested maybe some trash can area where you can drop a window to close it.

this could be another area where customizable hot-corners can be used. simply bind a corner to `close-window` action.

YaLTeR:
Well I want something discoverable yet hard to misfire for closing a window

hackr-sh:
Thoughts on when you "grab" a window that the size of the window animates to signal to the user it's currently grabbed?

YaLTeR:
I think I don't want to pile extra anims for now. As for opacity, windows already become transparent during interactive move, but only once you drag them out of the layout

YaLTeR:
Another small update:
- Changed zoom default to 0.5. After using it for a bit I feel that it's a decent default; it also avoids the "unclear which workspace is focused" problem and the "heavy aliasing artifacts on lower scales due to lack of mipmaps" problem (this is not about the 1 px jank with thin borders etc., just window contents themselves).
- Fixed some animation jank when moving a window to a new workspace above first (especially with empty-workspace-above-first enabled).

YaLTeR:
Updates:
- DnD edge scroll up/down in the overview now takes the working area into account, meaning if you have some layer-shell docks or bars, it will start scrolling against their edge, rather than the output edge.
- Bottom and background layer-shell keyboard grabs will now break when opening the Overview (since in the Overview, keyboard input doesn't reach bottom and background layers).
- warp-mouse-to-focus while "holding" a workspace switch gesture will no longer try to compensate for the position you're holding (this edge case doesn't really make sense, and it would complicate the logic unnecessarily after this PR).

At this point I implemented everything I wanted for this PR and fixed all bugs that I wanted to fix. What's left is:
- a (somewhat long and painful) process of splitting out some changes from the "wip" commit
- refactoring and cleaning up some logic that was left unrefactored
- writing a wiki page

What should we bind toggle-overview to by default? I've been using Mod+O so far. Ideally it would be a release bind on Mod, but we don't have release binds yet and they are somewhat tricky to do. And perhaps hardcoding Mod itself wouldn't be too bad compared to general release binds, but I'm not sure it's good to do it without a way to unbind.

TyberiusPrime:
Beautiful.

This needs an undo button though :).

ravish0007:
Tried this out, this feature is sick. I'm gonna tell all my friends to jump to niri.

debugloop:
I'm on Mod+Space and I like it. However a hardcoded release bind for Mod itself sounds good. You're right though that it should have a way to unbind it too.

bphenriques:
> Mod+Space is conventionally bound to keyboard layout switch

`Mod+Space` for fuzzel launcher (similar to how MacOS works) and `Mod+O` for the overview and and `Mod+k` for keyboard. If we had "release" bindings, I would opt for `Mod` for fuzzel but that is off topic. In any case, as long as it is configurable I am fine with providing sensible defaults for new comers.

So far I am really enjoying. I would like to have a way to select and close windows exclusively using the keyboard but for now it is not a big issue for me and tbh, this iteration is already great 🎉  I would prefer keyboard approaches first before looking into a GUI (`X` button on top or a trash as mentioned).

YaLTeR:
> If we had "release" bindings, I would opt for Mod for fuzzel but that is off topic. In any case, as long as it is configurable I am fine with providing sensible defaults for new comers.

Yeah, when we have proper release binds I'll default just Mod to overview.

> I would like to have a way to select and close windows exclusively using the keyboard

But... Just your normal keyboard binds that still work in the overview?

algernon:
I switched my Mom over to this branch, and she's enjoying it very much. However, we ran into an issue with hot corners, which may or may not be overview related.

What happens is that if she moves the mouse to the top left corner, the screen freezes for seconds, and *sometimes* the overview then appears, sometimes I have to move the mouse for something to happen. It is kinda weird. At the same time, opening the overview with a keybind works without problems. And to make things stranger still: if I try to record the screen, *the recording* is fine, and shows the overview appear near instantenously when moving the mouse to the hot corner.

Attached are two videos: one made of the screen with my phone (potato quality, sorry!), and another of the same session, recorded with `wf-recorder`. Also attached are DRM debug logs (a different session, but includes reproducing the freeze).

[dmesg-drm.log](https://github.com/user-attachments/files/19890890/dmesg-drm.log)

https://github.com/user-attachments/assets/37579f3c-16a2-40c9-a005-a64ddda5faf4

https://github.com/user-attachments/assets/f8e48d2e-4611-41a6-8b70-9e4cf1dfc9a3

niri is running on a [ZBox BI322](https://www.zotac.com/my/product/mini_pcs/zbox-bi322) mini PC.

YaLTeR:
This issue is really strange. The DRM commits are failing for some seconds after opening the overview:
```
WARN niri::backend::tty: error queueing frame: The underlying drm surface encountered an error: DRM access error: Page flip commit failed on device `Some("/dev/dri/card0")` (Invalid argument (os error 22))
```
But *only* when triggering it through the hot corner. Triggering via the keybind works fine. There shouldn't be any difference between the two paths, apart from the cursor position simultaneously changing I guess.

@cmeissl maybe you can notice something weird in the DRM debug log above, when you've some free time?

martinetd:
Looking for EINVAL (-22) has some hits in the drm log, then going back up this is what fails:
```
drivers/gpu/drm/i915/display/intel_cursor.c
        /*
         * There's something wrong with the cursor on CHV pipe C.
         * If it straddles the left edge of the screen then
         * moving it away from the edge or disabling it often
         * results in a pipe underrun, and often that can lead to
         * dead pipe (constant underrun reported, and it scans
         * out just a solid color). To recover from that, the
         * display power well must be turned off and on again.
         * Refuse the put the cursor into that compromised position.
         */              
        if (IS_CHERRYVIEW(dev_priv) && pipe == PIPE_C &&
            plane_state->uapi.visible && plane_state->uapi.dst.x1 < 0) {
                drm_dbg_kms(&dev_priv->drm,
                            "CHV cursor C not allowed to straddle the left screen edge\n");
                return -EINVAL;
        }
```
```
...
[ 4851.870404] i915 0000:00:02.0: [drm:i9xx_check_cursor [i915]] CHV cursor C not allowed to straddle the left screen edge
[ 4851.871165] i915 0000:00:02.0: [drm:intel_atomic_check [i915]] [PLANE:92:cursor C] atomic driver check failed
[ 4851.871606] i915 0000:00:02.0: [drm:intel_crtc_state_dump [i915]] [CRTC:97:pipe C] enable: yes [failed]
...
[ 4851.883540] i915 0000:00:02.0: [drm:drm_atomic_check_only] atomic driver check for 00000000cbb09788 failed: -22
```

So basically cursor being slightly over the edge causes atomic flips to fail...

cmeissl:
We are quite optimistic when deciding if a state has to be retested to avoid atomic tests. In case of the cursor plane we assume that if it worked before and we only update the position that it will work again. Which seems to lead to this error during the page flip. I think it should be relatively easy to extend the check to retest when we enter or exit offscreen.

YaLTeR:
Updates:
- fixed misaligned input to layer-shell during a workspace switch
- did the remaining refactors and clean-ups
- split into commits
- wrote the wiki pages

YaLTeR:
Alright let's merge this and get some wider testing. Thanks everyone for feedback!

dragonnn:
Just installed it, looks amazing! I am wondering if it would be possible to have a per monitor variant. I think I would prefer that (by it I mean when I toggle it only the current active monitor gets into overview).

YaLTeR:
Nah, I think it'll stay global

dragonnn:
I didn't mean to change it, just have `toggle-overview` and `toggle-monitor-overview`. But if you are not interested in it sure I understand that. Love you work anyway!

YaLTeR:
Yes but that's, like, a lot of work to make it per-monitor..

Issue #291, opened 2024-04-15
Add a config to enable the NumLock key on login


YaLTeR:
### Discussed in https://github.com/YaLTeR/niri/discussions/290

<div type='discussions-op-text'>

<sup>Originally posted by **stepanzak** April 14, 2024</sup>
Hi, I wasn't able to find a simple way to enable the NumLock key on login to Niri session, so I don't have to press NumLock to be able to type on numerical keyboard. In fact, I wasn't able to find any way to do that. Running the `numlockl on` command using [Numlockl](https://gitlab.com/phoepsilonix/numlockl) turns the NumLock on, but running it with spawn-at-startup or as a user systemd service doesn't. Both Hyprland and Sway have a config to enable it, so I think it would be awesome if Niri had it too.</div>

cole-h:
TL;DR: I think Smithay needs to grow support for this for us to take advantage of.

---

Last weekend, I poked around at this (with my lacking knowledge of Wayland itself) and only got far enough to realize this likely needs some additional support in Smithay. Specifically, I couldn't find a way to set the locked modifiers (so as to enable numlock).

Here's the very, very barebones and not working diff I was using to play around (I also patched Smithay to make the state fields public so I could actually print them, but otherwise this is as far as I got):

```diff
diff --git a/Cargo.lock b/Cargo.lock
index eed1cf8..f3579bf 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3116,7 +3116,6 @@ checksum = "3c5e1a9a646d36c3599cd173a41282daf47c44583ad367b8e6837255952e5c67"
 [[package]]
 name = "smithay"
 version = "0.3.0"
-source = "git+https://github.com/Smithay/smithay.git#c5e9a697e41f50dc56b918d9ef1e3d2e52c84ac0"
 dependencies = [
  "appendlist",
  "bitflags 2.5.0",
diff --git a/Cargo.toml b/Cargo.toml
index 51c2ae0..a3d0d44 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,8 @@ tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 tracy-client = { version = "0.17.0", default-features = false }
 
 [workspace.dependencies.smithay]
-git = "https://github.com/Smithay/smithay.git"
-# path = "../smithay"
+# git = "https://github.com/Smithay/smithay.git"
+path = "../smithay"
 default-features = false
 
[...]

PR #1447, opened 2025-04-20
feat: implement support to enable numlock at startup

erdii:
This PR implements a config setting to enable numlock at startup as requested in https://github.com/YaLTeR/niri/issues/291.

The code works on my machine (aside from the bug mentioned by @YaLTeR in https://github.com/YaLTeR/niri/pull/127#issuecomment-1913650587 where the numlock led stays off until the first input event is received.)

I still have to find out what the idiomatic way to coming up with the new serialized modifier_state is.

Questions I have:
- [x] Do the expected values change between different keymappings?
  - Answer: possibly - implemented a lookup inside set_modifier_state in smithay
- [x] should the serialization step happen here in niri or inside the set_modifier_state method in smithay?
    - Answer: implemented ad hoc serialization inside set_modifier_state in smithay
- [x] Does this implementation enable numlock whenever a new keyboard is added? Is this a bug?
- [x] should I add some tests?

Depends on: https://github.com/Smithay/smithay/pull/1721.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -120,6 +120,8 @@ pub struct Keyboard {
     pub repeat_rate: u8,
     #[knuffel(child, unwrap(argument), default)]
     pub track_layout: TrackLayout,
+    #[knuffel(child)]
+    pub numlock: bool,
 }
 
 impl Default for Keyboard {
@@ -129,6 +131,7 @@ impl Default for Keyboard {
             repeat_delay: 600,
             repeat_rate: 25,
             track_layout: Default::default(),
+            numlock: Default::default(),
         }
     }
 }
@@ -4059,6 +4062,7 @@ mod tests {
                     repeat_delay: 600,
                     repeat_rate: 25,
                     track_layout: Window,
+                    numlock: false,
                 },
                 touchpad: Touchpad {
                     off: false,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2721,7 +2721,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c80231409c20246a13fddb31776fb942c38553c51e871f8cbd687a4cfb5843d"
 dependencies = [
  "phf_shared",
- "rand",
+ "rand 0.8.5",
 ]
 
 [[package]]
@@ -2965,8 +2965,8 @@ dependencies = [
  "bitflags 2.9.0",
  "lazy_static",
  "num-traits",
- "rand",
- "rand_chacha",
+ "rand 0.8.5",
+ "rand_chacha 0.3.1",
  "rand_xorshift",
  "regex-syntax 0.8.5",
  "rusty-fork",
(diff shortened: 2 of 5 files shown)


niri-config/src/lib.rs
```diff
@@ -144,6 +144,8 @@ pub struct Xkb {
     pub options: Option<String>,
     #[knuffel(child, unwrap(argument))]
     pub file: Option<String>,
+    #[knuffel(child)]
+    pub numlock: bool,
```
YaLTeR, line 148: Does this need to be configurable? I'd just enable it always by default

src/input/mod.rs
```diff
@@ -184,13 +184,14 @@ impl State {
...
-                    {
-                        device.led_update(led_state.into());
+                    if let Some(keyboard) = self.niri.seat.get_keyboard() {
+                        device.led_update(keyboard.led_state().into());
+
+                        if self.niri.config.borrow().input.keyboard.xkb.numlock {
+                            let mut modifier_state = keyboard.modifier_state();
+                            modifier_state.num_lock = true;
+                            keyboard.set_modifier_state(modifier_state);
+                        }
```
YaLTeR, line 194: This seems like the wrong place to do it? I would expect numlock to be enabled in Niri::new() when a keyboard is created.

YaLTeR (commented):
> should I add some tests?

Nothing easily testable here so no

niri-config/src/lib.rs
erdii: I'd argue yes.
I have a notebook laying around that overlays the numlock on top of regular keys.
Numlock has to be disabled by default on that machine for the keyboard to be usable.

YaLTeR: Oh god. Okay then, let's default it to enabled though

src/input/mod.rs
erdii: I've moved it over.

niri-config/src/lib.rs
FreeFull: I think defaulting to disabled would be better, since that matches the previous behaviour, and the way other things default to having numlock off too. I also used to have a laptop that had the numpad overlayed on top of the regular keys like that, too, so I imagine it's not too rare.

YaLTeR: I suggested just having it on in the default config. Similar to touchpad tap, etc

YaLTeR (approved):
Looks good. Let's merge after a Smithay update. Note to self: don't forget to document on the wiki.

erdii:
What do you think of adding the wiki update to this PR?
Wording up for discussion:

````diff
diff --git i/wiki/Configuration:-Input.md w/wiki/Configuration:-Input.md
index 4a4c62ec..6dbdfc80 100644
--- i/wiki/Configuration:-Input.md
+++ w/wiki/Configuration:-Input.md
@@ -23,6 +23,7 @@ input {
         // repeat-delay 600
         // repeat-rate 25
         // track-layout "global"
+        // numlock
     }
 
     touchpad {
@@ -166,6 +167,21 @@ input {
 }
 ```
 
+#### NumLock
+
+NumLock can be automatically turned on at startup by adding the `numlock` flag.
+This setting will not have an effect on keymaps without a numlock modifier.
+If you're using a laptop with a keyboard that overlays numlock keys on top of regular keys, you might have to disable the setting.
+To disable, omit the flag or comment it out.
+
+```kdl
+input {
+    keyboard {
+        numlock
+    }
+}
+```
+
 ### Pointing Devices
 
 Most settings for the pointing devices are passed directly to libinput.
````

**Comment by YaLTeR** [comment]
Sure.

1. "Num Lock" seems to be commonly written with a space
2. Uncomment it in the topmost config section since it's on by default

```diff
+Set the `numlock` flag to turn on Num Lock automatically at startup.
+
+You might want to disable (comment out) `numlock` if you're using a laptop with a keyboard that overlays Num Lock keys on top of regular keys.
```

erdii:
Added to the PR.

YaLTeR:
Do you mind rebasing the PR and updating Smithay?

YaLTeR:
Fix pls

erdii:
Done

YaLTeR:
Thanks, seems to work.

PR #1449, opened 2025-04-20
delayed blocker + syncobj + fifo + commit timing

cmeissl:
At this point just a poc, which obviously needs a major cleanup, showcasing and tracking https://github.com/Smithay/smithay/pull/1720

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -19,6 +19,7 @@ use smithay::backend::allocator::format::FormatSet;
 use smithay::backend::allocator::gbm::{GbmAllocator, GbmBufferFlags, GbmDevice};
 use smithay::backend::allocator::Fourcc;
 use smithay::backend::drm::compositor::{DrmCompositor, FrameFlags, PrimaryPlaneElement};
+use smithay::backend::drm::exporter::gbm::GbmFramebufferExporter;
 use smithay::backend::drm::{
     DrmDevice, DrmDeviceFd, DrmEvent, DrmEventMetadata, DrmEventTime, DrmNode, NodeType, VrrSupport,
 };
@@ -114,7 +115,7 @@ pub type TtyRendererError<'render> = <TtyRenderer<'render> as RendererSuper>::Er
 
 type GbmDrmCompositor = DrmCompositor<
     GbmAllocator<DrmDeviceFd>,
-    GbmDevice<DrmDeviceFd>,
+    GbmFramebufferExporter<DrmDeviceFd>,
     (OutputPresentationFeedback, Duration),
     DrmDeviceFd,
 >;
@@ -971,7 +972,7 @@ impl Tty {
             surface,
             None,
             allocator.clone(),
-            device.gbm.clone(),
+            GbmFramebufferExporter::new(device.gbm.clone()),
             SUPPORTED_COLOR_FORMATS,
             // This is only used to pick a good internal format, so it can use the surface's render
             // formats, even though we only ever render on the primary GPU.
@@ -1001,7 +1002,7 @@ impl Tty {
                     surface,
                     None,
                     allocator,
-                    device.gbm.clone(),
+                    GbmFramebufferExporter::new(device.gbm.clone()),
                     SUPPORTED_COLOR_FORMATS,
                     render_formats,
                     device.drm.cursor_size(),
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -51,6 +51,7 @@ use smithay::wayland::dmabuf::{DmabufFeedback, DmabufFeedbackBuilder, DmabufGlob
 use smithay::wayland::drm_lease::{
     DrmLease, DrmLeaseBuilder, DrmLeaseRequest, DrmLeaseState, LeaseRejected,
 };
+use smithay::wayland::drm_syncobj::{supports_syncobj_eventfd, DrmSyncobjState};
 use smithay::wayland::presentation::Refresh;
 use smithay_drm_extras::drm_scanner::{DrmScanEvent, DrmScanner};
 use wayland_protocols::wp::linux_dmabuf::zv1::server::zwp_linux_dmabuf_feedback_v1::TrancheFlags;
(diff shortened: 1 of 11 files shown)


src/backend/tty.rs
```diff
@@ -583,6 +587,12 @@ impl Tty {
                     }
                 }
             }
+
+            // Expose syncobj protocol if supported by primary GPU
+            if supports_syncobj_eventfd(&device_fd) {
+                let syncobj_state = DrmSyncobjState::new::<State>(&niri.display_handle, device_fd);
+                assert!(self.syncobj_state.replace(syncobj_state).is_none());
```
YaLTeR, line 594: Guess this should be removed once the primary node is removed?

src/niri.rs
```diff
@@ -647,7 +662,60 @@ impl State {
...
         self.niri.advance_animations();
 
-        self.niri.redraw_queued_outputs(&mut self.backend);
+        let mut queued_outputs = self.niri.queued_outputs();
+        queued_outputs.sort_by_key(|(_, target_presentation_time, _)| *target_presentation_time);
+
+        let mut min_next_schedule: Option<Duration> = None;
+
+        for (output, target_presentation_time, refresh_interval) in queued_outputs {
+            let next_deadline = self.signal_commit_timing(&output, target_presentation_time);
```
YaLTeR, line 671: Should this happen at the start of `refresh()` together with the call to `notify_blocker_cleared()`? As I understand, this signalling clears blockers for the surfaces that now became ready, which means it should happen earlier, so that the commits are considered by `layout.refresh()`.

src/niri.rs
```diff
@@ -3216,19 +3572,17 @@ impl Niri {
...
-            matches!(
-                state.redraw_state,
-                RedrawState::Queued | RedrawState::WaitingForEstimatedVBlankAndQueued(_)
-            )
-        }) {
-            trace!("redrawing output");
-            let output = output.clone();
-            self.redraw(backend, &output);
-        }
+    pub fn queued_outputs(&self) -> Vec<(Output, Duration, Option<Duration>)> {
```
YaLTeR, line 3575: I guess this is not exactly `queued_outputs()` since it doesn't check if the output is queued for redraw?

src/niri.rs
```diff
@@ -647,7 +662,60 @@ impl State {
...
+
+                min_next_schedule = min_next_schedule
+                    .filter(|min_next_schedule| *min_next_schedule <= next_schedule)
+                    .or(Some(next_schedule));
+            }
+
+            if self.niri.is_queued(&output) {
+                self.niri
+                    .redraw(&mut self.backend, &output, target_presentation_time);
+                self.signal_fifo(&output);
```
YaLTeR, line 702: I'm not sure I entirely understand how fifo is supposed/intended to work. In the protocol:

> This means that when the content update containing "set_barrier" was made active at a latching deadline, it will be active for at least one refresh cycle.

What happens if we redraw and signal fifo here, but the redraw doesn't actually submit a buffer (e.g. the damage was outside the output rect), then something else commits right away and we redraw and signal again, before one refresh cycle had passed? Doesn't this go against this line in the protocol?

YaLTeR, line 702: Same question here about clearing commit blockers at the end here, after `layout.refresh()`. Though this one is trickier because I understand it has to happen after a redraw which itself has to happen at the end.

sabisho:
When is it going to be implemented.
Only major blocker for nvidia users like me.
Also will there be support for wp-color-mangement-v1 (HDR) in future?

YaLTeR:
How exactly is this a blocker? Many people use niri on NVIDIA just fine.

sabisho:
It just flickers, or so I've heard. NVIDIA doesn’t support implicit sync.

YaLTeR:
You've heard wrong. The screencast flickers are unrelated to this PR. Nvidia has a workaround for lack of explicit sync since long ago and it works fine.

yasink21:
sorry for reviving the thread, but is there a roadmap for these features, especially fifo

ewgdg:
I think we should avoid making this PR depend on Smithay `SurfaceBarrier` / delayed blockers.

COSMIC has a similar cross-surface tiling/configure blocker model, but it only synchronizes the configure/commit side; dmabuf/syncobj readiness is still handled by normal per-surface blockers. That seems to be good enough in practice.

So the useful invariant appears to be “all participants committed the configured state”, not “all participants reached the same post-dmabuf-readiness point”. The latter mostly prevents a possible one-frame mismatch if one surface’s buffer becomes ready slightly later than another’s.

That feels like a small and speculative benefit. Meanwhile it increases code complexity quite a bit: special mapped-surface pre-commit handling, transaction lifetime tied to dmabuf readiness, and a new Smithay transaction primitive. I do not see an obvious benefit that justifies that complexity, especially if it blocks explicit syncobj support, which fixes real NVIDIA driver issues.

I’d rather simplify niri’s transaction model:

- keep one default dmabuf/syncobj pre-commit hook for per-surface readiness blockers;
- remove the mapped-surface dmabuf readiness workaround / double-hook special case;
- complete niri layout transactions when the matching configure commit arrives;
- let Smithay apply each surface when its own readiness blockers clear.

If a real visual bug shows up from that weaker invariant, we can revisit a Smithay pre-apply/coordinating blocker later. For now, the new Smithay primitive feels like too much complexity for no clear practical gain.

ewgdg:
After more thinking, `SurfaceBarrier` does have real value, but maybe it is substantial enough to split into its own focused PR.

dfseifert:
Just wanted to share an observation that might (or not) be useful context/information here.
When developing and running a godot game on niri, the engine emits the following warning at game startup:

```
W 0:00:00:130   init: FIFO protocol not found! Frame pacing will be degraded.
  <C++ Source>  platform/linuxbsd/wayland/wayland_thread.cpp @ init()
```

My guess is this happens because godots wayland backend looks for the `wp-fifo-v1` protocol to properly synchronize frame delivery with the compositor. But when it's absent, godot falls back to a manual frame-pacing workaround (https://github.com/godotengine/godot/pull/101454) which has a few known side effects, like shader `TIME` advancing incorrectly (https://github.com/godotengine/godot/issues/87963) and GPU particles dropping framerate (https://github.com/godotengine/godot/issues/88248).

I am unsure about the impact on games developed with other engines though.
Not raising this as a blocker, just thought it might be a helpful data point if `fifo-v1` support comes up in the future when working on gaming compatibility and performance optimizations.

Happy to test anything if that's useful.

Kind Regards and thanks for all the hard work done!

PR #1451, opened 2025-04-21
Add instructions to the ui on how to copy a screenshot without saving it

NeYurii:


NeYurii:
Maybe so. But this information is only presented in the wiki, which makes it a bit difficult to find, so it might be good to have it somewhere else

YaLTeR:
Well it's also not that terribly important arguably. Wiki is a good place. Maybe some comment in the default config, idk

erdii:
I'd argue that it would be nice to have the description in the screenshot UI since it is
a) only active there
b) other important keybinds for the screenshot UI are already described there

Atemo-C:
Most screenshots I take are just quick area screenshots that I simply want to copy+paste to friends. I do not want them all to be saved *(especially since I have a script that automatically optimizes the taken screenshots before converting them either to `.webp` images for web usage or `.jxl` for local usage, so it takes some time)*. I would then forget to clean them up from my "screenshots" directory, & would have to deal with dozens or hundreds of images all at once.

For the longest time, I thought Niri's built-in screenshot tool did not have a copy function, & this issue is the way I discovered it has one (& I seem to not be the only one here), I have read Niri's Wiki & was a bit confused at first to see the `Ctrl+C` hint being [here](https://github.com/YaLTeR/niri/wiki/Configuration:-Key-Bindings#screenshot-screenshot-screen-screenshot-window), but I suppose it could be a case of me not knowing how to look up information in documentation.

Additionally, it is not said that `Enter` can also be used to validate taking the screenshot, and `Ctrl+D` does not cancel the screenshot like `Escape`, which is a keybind that is still pretty commonly used in other places, but this is another issue entirely.

Adding that one line of text would be very neat, & if not, at least offer a command-line option to add it if we want to, or even let us create our own messages; I really do not know.

YaLTeR:
> Most screenshots I take are just quick area screenshots that I simply want to copy+paste to friends. I do not want them all to be saved (especially since I have a script that automatically optimizes the taken screenshots before converting them either to .webp images for web usage or .jxl for local usage, so it takes some time).

Sounds like you might want to just disable screenshot saving entirely? You can do that by setting screenshot-path to null.

> For the longest time, I thought Niri's built-in screenshot tool did not have a copy function, & this issue is the way I discovered it has one (& I seem to not be the only one here), I have read Niri's Wiki & was a bit confused at first to see the `Ctrl+C` hint being [here](https://github.com/YaLTeR/niri/wiki/Configuration:-Key-Bindings#screenshot-screenshot-screen-screenshot-window), but I suppose it could be a case of me not knowing how to look up information in documentation.

But... Niri *always* copies the screenshot. What Ctrl+C is doing is *disabling writing to disk* rather than enabling copying.

> Additionally, it is not said that `Enter` can also be used to validate taking the screenshot

Yeah, but how is this a problem? The most common key is documented. There are other obvious ones that do the same thing but they don't all have to be documented.

> and `Ctrl+D` does not cancel the screenshot like `Escape`, which is a keybind that is still pretty commonly used in other places, but this is another issue entirely.

I've never seen Ctrl+D used to cancel?

> Adding that one line of text would be very neat

I want to keep that panel minimal. No clutter. Ctrl+C to avoid saving to disk is not essential knowledge.

Atemo-C:
Yes, I suppose I explained myself rather poorly, but you still seem to have understood most of my points, thanks for that.

---

> Sounds like you might want to just disable screenshot saving entirely? You can do that by setting screenshot-path to null.

Yes, I did consider doing that, but there is no command-line option to do so _(`--write-to-disk=false` does not work with `niri msg action screenshot`)_, and I would rather avoid having to edit the configuration file every time I want to simply automatically copy a screenshot without saving it. Adding `--write-to-disk=false` to `niri msg action screenshot` would solve this issue.

Now, this is not too much of a problem either way, since I have my own script that works with Niri's built-in screenshot utility to achieve more of the features I want, and adding an automatic enablement/disablement of the screenshot path depending on the type of screenshot would be trivial. Though, it is possible I am blind and simply missing an obvious option — I would actually love for it to be the case.

> But... Niri always copies the screenshot. What Ctrl+C is doing is disabling writing to disk rather than enabling copying.

Yes, but I know it does now. My phrasing was poor, but I said that I *thought* it did not, so, sorry for the confusion. The reason for it being that I had not yet set-up Dunst for my new system yet, so I did not see this useful notification pop up:
image
Issue on my part, no doubt.

> Yeah, but how is this a problem? The most common key is documented. There are other obvious ones that do the same thing but they don't all have to be documented.

This is more of a nitpick than anything. I simply like to have everything documented, it is simply a symptom of how I write my own stuff. Even things like different wordings for the same commands, such as for my Screenshot script _(adapted from Hyprland to Niri)_, as seen here, are documented:
image
_(Edit: Oh, I noticed some minor inconsistency issues with this screenshot (wrong color for one of Niri and missing space after [ Examples], welp, fixed now.)_

Again, simply the way I do things, but I totally understand why it is not like that in Niri, it could be argued that it is over-complication for no real reason.

> I've never seen Ctrl+D used to cancel?

I have seen it used to cancel things just as it closes/quits things. When taking a simple screenshot, cancelling/exiting are basically the same thing, at least, to me.

> I want to keep that panel minimal. No clutter. Ctrl+C to avoid saving to disk is not essential knowledge.

Very well, I suppose we cannot change your opinion on that matter, so there is no need to continue arguing about it. Thank you for listening, still, and for correcting my brain at 3AM in lack of sleep, too.

NeYurii:
> Yeah, but how is this a problem? The most common key is documented. There are other obvious ones that do the same thing but they don't all have to be documented.

Ctrl+C isnt obvious key, not in the linux world at least. Many users can expect ctrl+c to interupt screenshot and y to copy, for example.

> I want to keep that panel minimal. No clutter. Ctrl+C to avoid saving to disk is not essential knowledge.

Its a widely used feature, that many people may want to use. If this isnt essential knowledge, is key binding to show/hide cursor one?

If it should be minimal, maybe it worth to add an option to disable it for people already learned keys and add all needed information? Or add this information to startup tip near other keys?

Issue #619, opened 2024-08-22
Implement alpha-modifier and single-pixel-buffer protocols


YaLTeR:
Important: it's not enough to just delegate them; niri's rendering helpers will also need to learn about these two. For example, snapshotting in https://github.com/YaLTeR/niri/blob/7bfdf87bf0138d602888fc3167921bc1d029b0ab/src/render_helpers/surface.rs#L56

PR #1463, opened 2025-04-24
Implement `wp_single_pixel_buffer_manager_v1` in render helpers

Eliasin:
I wanted to take a crack at #619 but don't have much prior experience with GPU programming or wayland dev, so I'm putting up my first bits of progress to make sure I'm doing the right thing.

I'm not sure exactly how to test whether the results of `render_snapshot_from_surface_tree` are correct, at first glance it's used in parts of the logic for when surfaces are "animated" by resizing, becoming floating, closing, etc. So far I've just made sure that the code path I'm writing is being taken during float and that the texture buffer is created in a way that seems to make sense.

Thanks in advance for any review :smile:

### DIFF
--- a/src/layout/tile.rs
+++ b/src/layout/tile.rs
@@ -989,23 +989,27 @@ impl<W: LayoutElement> Tile<W> {
                                         // program shader. To work around this, we can use a
                                         // `BorderRenderElement` when we detect that we need to clip
                                         // a solid color texture.
-                                        return BorderRenderElement::new(
-                                            geo.size,
-                                            Rectangle::from_size(geo.size),
-                                            GradientInterpolation::default(),
-                                            // color32_f uses premultiplied alpha which matches
-                                            // what `BorderRenderElement` expects
-                                            Color::from_color32f(*color32_f),
-                                            Color::from_color32f(*color32_f),
-                                            0.,
-                                            Rectangle::from_size(geo.size),
-                                            0.,
-                                            radius,
-                                            scale.x as f32,
-                                            1.,
-                                        )
-                                        .with_location(geo.loc)
-                                        .into();
+                                        let elem_geo =
+                                            elem.geometry(scale).to_f64().to_logical(scale);
+                                        if let Some(shader_geo) = geo.intersection(elem_geo) {
+                                            return BorderRenderElement::new(
+                                                shader_geo.size,
+                                                Rectangle::from_size(shader_geo.size),
+                                                GradientInterpolation::default(),
+                                                // color32_f uses premultiplied alpha which matches
+                                                // what `BorderRenderElement` expects
+                                                Color::from_color32f(*color32_f),
+                                                Color::from_color32f(*color32_f),
+                                                0.,
+                                                Rectangle::new(geo.loc - shader_geo.loc, geo.size),
+                                                0.,
+                                                radius,
+                                                scale.x as f32,
+                                                1.,
+                                            )
+                                            .with_location(shader_geo.loc)
+                                            .into();
+                                        }
                                     }
                                 }
                             }
--- a/src/layout/tile.rs
+++ b/src/layout/tile.rs
@@ -993,6 +993,8 @@ impl<W: LayoutElement> Tile<W> {
                                             geo.size,
                                             Rectangle::from_size(geo.size),
                                             GradientInterpolation::default(),
+                                            // color32_f uses premultiplied alpha which matches
+                                            // what `BorderRenderElement` expects
                                             Color::from_color32f(*color32_f),
                                             Color::from_color32f(*color32_f),
                                             0.,
--- a/src/layout/tile.rs
+++ b/src/layout/tile.rs
@@ -2,6 +2,7 @@ use core::f64;
 use std::rc::Rc;
 
 use niri_config::{Color, CornerRadius, GradientInterpolation};
+use smithay::backend::renderer::element::surface::WaylandSurfaceTexture;
 use smithay::backend::renderer::element::{Element, Kind};
 use smithay::backend::renderer::gles::GlesRenderer;
 use smithay::utils::{Logical, Point, Rectangle, Scale, Size};
@@ -967,14 +968,44 @@ impl<W: LayoutElement> Tile<W> {
                     if clip_to_geometry {
                         if let Some(shader) = clip_shader.clone() {
                             if ClippedSurfaceRenderElement::will_clip(&elem, scale, geo, radius) {
-                                return ClippedSurfaceRenderElement::new(
-                                    elem,
-                                    scale,
-                                    geo,
-                                    shader.clone(),
-                                    radius,
-                                )
-                                .into();
+                                match elem.texture() {
+                                    WaylandSurfaceTexture::Texture(_) => {
+                                        return ClippedSurfaceRenderElement::new(
+                                            elem,
+                                            scale,
+                                            geo,
+                                            shader.clone(),
+                                            radius,
+                                        )
+                                        .into();
+                                    }
+                                    WaylandSurfaceTexture::SolidColor(color32_f) => {
+                                        // In this branch we're rendering a solid color texture
+                                        // which smithay will render using a solid program shader.
+                                        // `ClippedSurfaceElement` is implemented by overriding the
+                                        // default texture program shader that smithay uses. This
+                                        // strategy won't work for solid color textures because
+                                        // smithay doesn't provide a way to override the solid
+                                        // program shader. To work around this, we can use a
+                                        // `BorderRenderElement` when we detect that we need to clip
+                                        // a solid color texture.
+                                        return BorderRenderElement::new(
+                                            geo.size,
+                                            Rectangle::from_size(geo.size),
+                                            GradientInterpolation::default(),
+                                            Color::from_color32f(*color32_f),
+                                            Color::from_color32f(*color32_f),
+                                            0.,
+                                            Rectangle::from_size(geo.size),
+                                            0.,
+                                            radius,
+                                            scale.x as f32,
+                                            1.,
+                                        )
+                                        .with_location(geo.loc)
+                                        .into();
+                                    }
+                                }
                             }
                         }
                     }
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -100,7 +100,6 @@ use smithay::wayland::shell::wlr_layer::{self, Layer, WlrLayerShellState};
 use smithay::wayland::shell::xdg::decoration::XdgDecorationState;
 use smithay::wayland::shell::xdg::XdgShellState;
 use smithay::wayland::shm::ShmState;
-#[cfg(test)]
 use smithay::wayland::single_pixel_buffer::SinglePixelBufferState;
 use smithay::wayland::socket::ListeningSocketSource;
 use smithay::wayland::tablet_manager::TabletManagerState;
@@ -298,13 +297,6 @@ pub struct Niri {
     pub activation_state: XdgActivationState,
     pub mutter_x11_interop_state: MutterX11InteropManagerState,
 
-    // This will not work as is outside of tests, so it is gated with #[cfg(test)] for now. In
-    // particular, shaders will need to learn about the single pixel buffer. Also, it must be
-    // verified that a black single-pixel-buffer background lets the foreground surface to be
-    // unredirected.
-    //
-    // https://github.com/YaLTeR/niri/issues/619
-    #[cfg(test)]
     pub single_pixel_buffer_state: SinglePixelBufferState,
 
     pub seat: Seat<State>,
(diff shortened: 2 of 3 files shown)


rustfmt.toml
```diff
@@ -1,4 +1,5 @@
 imports_granularity = "Module"
 group_imports = "StdExternalCrate"
+edition = "2021"
```
HigherOrderLogic, line 3: I believe you dont need to define this here. `rustfmt` automatically picks up the version in `Cargo.toml`.

Eliasin: Oops, was going to ask something about that because my setup didn't like it being missing but I forgot

src/niri.rs
```diff
@@ -293,7 +292,6 @@ pub struct Niri {
     // unredirected.
     //
     // https://github.com/YaLTeR/niri/issues/619
```
bbb651, line 294:
```suggestion
```
You should remove the comments as well (idk if it's possible to do a multiline suggestion for unchanged lines)

YaLTeR:
It needs to be supported by the `ClippedSurfaceRenderElement` too somehow. Not sure what's the best way there.

Eliasin:
> It needs to be supported by the `ClippedSurfaceRenderElement` too somehow. Not sure what's the best way there.

I poked around in there and exercised the code path for creating a `ClippedSurfaceRenderElement` in `Tile::render_inner` and everything *looks* ok to me. Reading the implementation for `ClippedSurfaceRenderElement` I'm also not sure what would need to change for single pixel buffer support since to me it looks like the underlying surface render element is treated pretty generically. Would the necessary changes be related to the shader? (unless you're talking about the alpha-modifier support)

YaLTeR:
`ClippedSurfaceRenderElement` clips the windows to their geometry with [`clip-to-geometry true`](https://github.com/YaLTeR/niri/wiki/Configuration:-Window-Rules#clip-to-geometry) window rule, also rounds corners. The way it works is by [setting a custom texture shader](https://github.com/YaLTeR/niri/blob/78e3daf5f82b5870e998faf0e1efeaa048730976/src/render_helpers/clipped_surface.rs#L253).

Single-pixel buffers don't have a texture and [use `draw_solid()`](https://github.com/Smithay/smithay/blob/f85d06a4629795f5008b573f6b48606db9ff6d71/src/backend/renderer/element/surface.rs#L401) which [uses a solid_program shader](https://github.com/Smithay/smithay/blob/f85d06a4629795f5008b573f6b48606db9ff6d71/src/backend/renderer/gles/mod.rs#L2420) rather than (an overridable) texture_program.

The good news is that it's still a shader (I thought it wasn't). So maybe we have some options here.

One is to add an `override_default_solid_program()` to Smithay similar to [`override_default_tex_program()`](https://smithay.github.io/smithay/smithay/backend/renderer/gles/struct.GlesFrame.html#method.override_default_tex_program), then hook that up. However, that will result in more shader code duplication on niri side.

Another option is to detect when we need to clip a single pixel buffer (there's already a function for this in clipped surface), then replace the clipped surface element with a `BorderRenderElement` [similar to how block-out-from already works](https://github.com/YaLTeR/niri/blob/78e3daf5f82b5870e998faf0e1efeaa048730976/src/layout/tile.rs#L993).

I think option 2 might be more convenient? Just need to make sure that it only replaces when necessary, so that direct scanout with a single pixel buffer black bg keeps working when there's no clipping or rounding (this will only work if WaylandSurfaceRenderElement itself is used without any replacement with BorderRenderElement).

Eliasin:
I think that should work, I tested it by disabling the `clip-to-geometry` and `geometry-corner-radius` rule and verifying that the `BorderRenderElement` wasn't used and checking that the border element was used for when the rule was turned on. It visually also looks correct with no clipping or rounding.
Screenshot from 2025-05-09 01-17-37
Screenshot from 2025-05-09 01-18-12

src/render_helpers/surface.rs
```diff
@@ -53,17 +55,37 @@ pub fn render_snapshot_from_surface_tree(
...
+                            f64::from(data.buffer_scale()),
+                            data.buffer_transform(),
+                            Vec::new(),
+                        )
+                    } else if let Some(single_pixel_buffer_user_data) = data
+                        .buffer()
+                        .and_then(|buffer| get_single_pixel_buffer(buffer).ok())
+                    {
+                        let pixel: [u8; 4] = single_pixel_buffer_user_data.rgba8888();
+                        let argb_pixel = [pixel[3], pixel[1], pixel[2], pixel[0]];
```
YaLTeR, line 72: Is this premultiplied? Does TextureBuffer::from_memory expect premultiplied? Add a comment here pls

src/layout/tile.rs
```diff
@@ -967,14 +968,44 @@ impl<W: LayoutElement> Tile<W> {
...
+                                        // smithay doesn't provide a way to override the solid
+                                        // program shader. To work around this, we can use a
+                                        // `BorderRenderElement` when we detect that we need to clip
+                                        // a solid color texture.
+                                        return BorderRenderElement::new(
+                                            geo.size,
+                                            Rectangle::from_size(geo.size),
+                                            GradientInterpolation::default(),
+                                            Color::from_color32f(*color32_f),
+                                            Color::from_color32f(*color32_f),
```
YaLTeR, line 997: Is this premultiplied? Does BorderRenderElement::new expect premultiplied? Add a comment here pls

src/render_helpers/surface.rs
```diff
@@ -53,17 +55,37 @@ pub fn render_snapshot_from_surface_tree(
...
-                    renderer,
-                    texture.clone(),
-                    f64::from(data.buffer_scale()),
-                    data.buffer_transform(),
-                    Vec::new(),
-                );
+                        TextureBuffer::from_memory(
+                            renderer,
+                            &argb_pixel,
+                            Fourcc::Argb8888,
```
YaLTeR, line 77: Maybe we can use Abgr and skip the pixel swapping?

YaLTeR (commented):
Looks reasonable. Could you please share your test client too?

src/render_helpers/surface.rs
Eliasin: Do you mean that we should use `Rgba8888`? I tried doing that initially but it turns out that the supported formats for `GlesRenderer` are

```rust
const SUPPORTED_MEM_FORMATS_2: &[Fourcc] = &[
    Fourcc::Abgr8888,
    Fourcc::Xbgr8888,
    Fourcc::Argb8888,
    Fourcc::Xrgb8888,
];

const SUPPORTED_MEM_FORMATS_3: &[Fourcc] = &[
    Fourcc::Abgr8888,
    Fourcc::Xbgr8888,
    Fourcc::Argb8888,
    Fourcc::Xrgb8888,
    Fourcc::Abgr2101010,
    Fourcc::Xbgr2101010,
    Fourcc::Abgr16161616f,
    Fourcc::Xbgr16161616f,
];
```

If you mean use `Abgr8888` and just `.reverse` the pixel from the `SinglePixelBufferUserData` then that also sounds like a good idea.

Either way I'll definitely add a comment here

src/render_helpers/surface.rs
Eliasin: my cursory look through `gles/mod.rs` shows that the renderer implementation expects premultiplied because it uses ` self.gl.BlendFunc(ffi::ONE, ffi::ONE_MINUS_SRC_ALPHA);` and `TextureBuffer` when used for `TextureRenderElement` passes the imported texture to the renderer with no real modification after import.

src/layout/tile.rs
Eliasin: `BorderRenderElement` calls `Color::to_array_unpremul` on these, so it expects premultiplied, also verified that the border fragment shader expects unpremultiplied, then internally premultiplies them

I checked that the `color32_f` matches the single pixel buffer data which would mean its also premultiplied

Eliasin:
Tester's here, kinda janky but it works :)
https://github.com/Eliasin/wayland-single-pixel-tester

YaLTeR:
Okay, I found and fixed one issue: the replacement border shader was drawing over the full window size, rather than over its element size, which you can see in `gtk4-demo --run=video_player` when you let the controls bar disappear.

Screenshot from 2025-05-10 08-26-56

The headerbar should still be visible as it is with `clip-to-geometry false` (and as it is after my fix).

However, now I realized that there are some more issues.

1. The fix is incomplete, it needs to do something when the intersection is zero, e.g. draw a 1×1 fully transparent element or something.
2. Something reasonable needs to happen when the border shader does not exist (can happen when it fails to compile or on older devices). This is not a super important case, so it doesn't need to be perfect.

    My idea is to draw a regular solid color over shader_geo in this case. Actually, it would be nice to draw a regular solid color instead of the border shader even just when rounded corners are not needed because the border shader is a bit more taxing on the GPU.
3. Since this fallback border shader is recreated every frame, it will produce full damage every frame, so whenever something else on screen redraws, the full area of the fallback border will also redraw. Not ideal from energy use perspective.

And now for the trickiest problem. Since we're replacing the Wayland surface render element along with its element `Id`, it means that the surface won't get frame callbacks, presentation feedbacks, and so on, because `update_primary_scanout_output` will think that the surface is not on screen. I'm not sure how to best deal with this.
* Window offscreening has the same problem and we have a good way of dealing with it through `window.set_offscreen_data()`. However, in this case it's just one surface we're replacing, not the whole window, so this mechanism isn't applicable.
* (IIUC) We cannot simply put the surface element `Id` (which is derived directly from wl_surface) on the replacement border element because then we need to match the commit counter between the two, and while we can copy the commit counter from the wayland surface render element, we cannot put back an updated commit counter once we're done with the border element.

    It might more-or-less work to copy the commit counter from the wayland surface render element and just not update it. This may result in changing the corner radius at runtime not updating right away, but it's probably not a big deal? Worth a try at least.

    Clipped surface does re-use the Id and commit counter directly, and it has the ExtraDamage thing to deal with the corner radius update problem. Actually, thinking about it, ExtraDamage probably already deals with this issue on all 4 corners, so this way may just work?

So yeah, I guess try putting the Id and commit counter on the replacement border element (add some `.with_id_and_commit_counter()` method to it and to `ShaderRenderElement` that just replaces their own Id/commit counter) and see if it works ok. This will also fix the damage every frame problem.

wildwestrom:
I found this issue after finding out that [`chayang`](https://github.com/b0o/chayang) doesn't work on Niri.
I tried this PR on my local machine and it works perfectly! I hope to see a version of this get merged!

I had to go through several shenanigans to get it actually running on NixOS, so here's the clanker's summary of what got it to run. Something to do with how Niri gets linked to Mesa.

<details>
<summary>
Repro notes: getting this running on NixOS
</summary>
<br>

**1. `nix develop` failed outright with a `rust-src` unpack error**

```
do not know how to unpack source archive .../rust-src-...-unknown
error: Cannot build '.../rust-src-1.90.0-nightly-....drv'
```

The fetched file was valid `.xz` data, just named `unknown` with no extension for Nix's
generic `unpackPhase` to key off. Root cause: the flake's `rust-overlay` input was pinned
to a commit from mid-2025, a full year behind `nixpkgs`. Fix — bump `rust-overlay` alone,
leave `nixpkgs` untouched (bumping `nixpkgs` too pulls forward C library versions like
`libdisplay-info` past what `Cargo.lock` requires, and introduces problem #2 below):

```
nix flake update rust-overlay
```

**2. Nested niri crashed with `Egl(DisplayNotSupported)`**

`RUST_LOG=trace` showed `Supported EGL client extensions: [""]` — libglvnd found *zero*
EGL vendors, not just a missing Wayland platform extension. Tracing it down to a raw
`dlopen()` of `libEGL_mesa.so.0`:

```
version `GLIBC_ABI_GNU2_TLS' not found (required by .../mesa-26.1.1/lib/libgallium-26.1.1.so)
```

NixOS's `/run/opengl-driver` always points at the system's *live* Mesa, built against
whatever glibc the current system generation uses. The devShell's own glibc comes from
the flake's pinned (older) `nixpkgs`, and can't `dlopen()` a Mesa built for a newer glibc
— a hard ABI wall, no env var works around it once you're pointed at the wrong Mesa.

Fix: point glvnd at the devShell's *own*, version-matched Mesa instead of the system's:

```
export __EGL_VENDOR_LIBRARY_FILENAMES=$(nix eval --raw nixpkgs#mesa)/share/glvnd/egl_vendor.d/50_mesa.json
cargo run -- -c /path/to/some/test-config.kdl
```

(Use a scratch KDL config, not your real dotfile config — it'll happily spawn every
app in your real `spawn-at-startup` block otherwise.)

Once both were fixed, the nested niri instance came up cleanly and the single-pixel-buffer
support worked as expected — this is purely a NixOS devShell/toolchain wrinkle, unrelated
to the PR's actual changes.
</details>

HigherOrderLogic:
> 2. Nested niri crashed with `Egl(DisplayNotSupported)`

This should be fixed after rebasing this PR on main.

PR #1465, opened 2025-04-24
Add `focus_target_workspace` arg to `MoveColumnToWorkspace` actions (#457)

yzy-1:
## Description:

This PR introduces a new `focus-target-workspace` argument for `move-column-to-workspace*` actions to address the use case described in #457, while maintaining backward compatibility with existing configurations.

## Changes:

- Added a **new optional argument** `focus-target-workspace` (default: `true`) to control workspace focus behavior
  - When set to `true` (default): Maintains original behavior - moves column and switches focus to target workspace
  - When set to `false`: Moves column without changing focus
- Example configurations:
  ``` 
  # New behavior: Move column without switching workspace
  Mod+Ctrl+1 { move-column-to-workspace 1 focus-target-workspace=false; }
  
  # Equivalent to original behavior (explicit)
  Mod+Ctrl+2 { move-column-to-workspace 2 focus-target-workspace=true; }
  
  # Equivalent to original behaviour (implicit default)
  Mod+Ctrl+3 { move-column-to-workspace 3; }
  ```
- Benefits of this implementation:
  - Enables batch column organization across workspaces without disruptive context switching
  - Allows temporary workspace decluttering while maintaining current focus
  - Preserves compatibility with all existing keybindings through default argument
  - Reduces code duplication by extending existing actions
- Original workflow remains unchanged for users who prefer immediate workspace switching

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -1888,7 +1888,7 @@ impl<W: LayoutElement> Layout<W> {
             }
         }
 
-        self.move_column_to_output(output);
+        self.move_column_to_output(output, None, true);
         true
     }
 
@@ -1899,7 +1899,7 @@ impl<W: LayoutElement> Layout<W> {
             }
         }
 
-        self.move_column_to_output(output);
+        self.move_column_to_output(output, None, true);
         true
     }
 
@@ -2213,31 +2213,25 @@ impl<W: LayoutElement> Layout<W> {
         monitor.move_to_workspace(window, idx, activate);
     }
 
-    pub fn move_column_to_workspace_up(&mut self) {
+    pub fn move_column_to_workspace_up(&mut self, activate: bool) {
         let Some(monitor) = self.active_monitor() else {
             return;
         };
-        monitor.move_column_to_workspace_up();
+        monitor.move_column_to_workspace_up(activate);
     }
 
-    pub fn move_column_to_workspace_down(&mut self) {
+    pub fn move_column_to_workspace_down(&mut self, activate: bool) {
         let Some(monitor) = self.active_monitor() else {
             return;
         };
-        monitor.move_column_to_workspace_down();
+        monitor.move_column_to_workspace_down(activate);
     }
 
-    pub fn move_column_to_workspace(&mut self, idx: usize) {
+    pub fn move_column_to_workspace(&mut self, idx: usize, activate: bool) {
         let Some(monitor) = self.active_monitor() else {
             return;
         };
-        monitor.move_column_to_workspace(idx);
-    }
-
-    pub fn move_column_to_workspace_on_output(&mut self, output: &Output, idx: usize) {
-        self.move_column_to_output(output);
-        self.focus_output(output);
-        self.move_column_to_workspace(idx);
+        monitor.move_column_to_workspace(idx, activate);
     }
 
     pub fn switch_workspace_up(&mut self) {
@@ -3559,7 +3553,12 @@ impl<W: LayoutElement> Layout<W> {
         }
     }
 
-    pub fn move_column_to_output(&mut self, output: &Output) {
+    pub fn move_column_to_output(
+        &mut self,
+        output: &Output,
+        target_ws_idx: Option<usize>,
+        activate: bool,
+    ) {
         if let MonitorSet::Normal {
             monitors,
             active_monitor_idx,
@@ -3583,8 +3582,10 @@ impl<W: LayoutElement> Layout<W> {
                 return;
             };
 
-            let workspace_idx = monitors[new_idx].active_workspace_idx;
-            self.add_column_by_idx(new_idx, workspace_idx, column, true);
+            let workspace_idx = target_ws_idx
+                .unwrap_or(monitors[new_idx].active_workspace_idx)
+                .min(monitors[new_idx].workspaces.len() - 1);
+            self.add_column_by_idx(new_idx, workspace_idx, column, activate);
         }
     }
 
--- a/src/layout/tests.rs
+++ b/src/layout/tests.rs
@@ -473,9 +473,9 @@ enum Op {
         #[proptest(strategy = "0..=4usize")]
         workspace_idx: usize,
     },
-    MoveColumnToWorkspaceDown,
-    MoveColumnToWorkspaceUp,
-    MoveColumnToWorkspace(#[proptest(strategy = "0..=4usize")] usize),
+    MoveColumnToWorkspaceDown(bool),
+    MoveColumnToWorkspaceUp(bool),
+    MoveColumnToWorkspace(#[proptest(strategy = "0..=4usize")] usize, bool),
     MoveWorkspaceDown,
     MoveWorkspaceUp,
     MoveWorkspaceToIndex {
@@ -508,7 +508,13 @@ enum Op {
         #[proptest(strategy = "proptest::option::of(0..=4usize)")]
         target_ws_idx: Option<usize>,
     },
-    MoveColumnToOutput(#[proptest(strategy = "1..=5usize")] usize),
+    MoveColumnToOutput {
+        #[proptest(strategy = "1..=5usize")]
+        output_id: usize,
+        #[proptest(strategy = "proptest::option::of(0..=4usize)")]
+        target_ws_idx: Option<usize>,
+        activate: bool,
+    },
     SwitchPresetColumnWidth,
     SwitchPresetWindowWidth {
         #[proptest(strategy = "proptest::option::of(1..=5usize)")]
@@ -1074,9 +1080,9 @@ impl Op {
                 let window_id = window_id.filter(|id| layout.has_window(id));
                 layout.move_to_workspace(window_id.as_ref(), workspace_idx, ActivateWindow::Smart);
             }
-            Op::MoveColumnToWorkspaceDown => layout.move_column_to_workspace_down(),
-            Op::MoveColumnToWorkspaceUp => layout.move_column_to_workspace_up(),
-            Op::MoveColumnToWorkspace(idx) => layout.move_column_to_workspace(idx),
+            Op::MoveColumnToWorkspaceDown(focus) => layout.move_column_to_workspace_down(focus),
+            Op::MoveColumnToWorkspaceUp(focus) => layout.move_column_to_workspace_up(focus),
+            Op::MoveColumnToWorkspace(idx, focus) => layout.move_column_to_workspace(idx, focus),
             Op::MoveWindowToOutput {
                 window_id,
                 output_id: id,
@@ -1097,13 +1103,17 @@ impl Op {
                     ActivateWindow::Smart,
                 );
             }
-            Op::MoveColumnToOutput(id) => {
+            Op::MoveColumnToOutput {
+                output_id: id,
+                target_ws_idx,
+                activate,
+            } => {
                 let name = format!("output{id}");
                 let Some(output) = layout.outputs().find(|o| o.name() == name).cloned() else {
                     return;
                 };
 
-                layout.move_column_to_output(&output);
+                layout.move_column_to_output(&output, target_ws_idx, activate);
             }
             Op::MoveWorkspaceDown => layout.move_workspace_down(),
             Op::MoveWorkspaceUp => layout.move_workspace_up(),
@@ -1560,10 +1570,10 @@ fn operations_dont_panic() {
             window_id: None,
             workspace_idx: 2,
         },
-        Op::MoveColumnToWorkspaceDown,
-        Op::MoveColumnToWorkspaceUp,
-        Op::MoveColumnToWorkspace(1),
-        Op::MoveColumnToWorkspace(2),
+        Op::MoveColumnToWorkspaceDown(true),
+        Op::MoveColumnToWorkspaceUp(true),
+        Op::MoveColumnToWorkspace(1, true),
+        Op::MoveColumnToWorkspace(2, true),
         Op::MoveWindowDown,
         Op::MoveWindowDownOrToWorkspaceDown,
         Op::MoveWindowUp,
@@ -1735,11 +1745,11 @@ fn operations_from_starting_state_dont_panic() {
             window_id: None,
             workspace_idx: 3,
         },
-        Op::MoveColumnToWorkspaceDown,
-        Op::MoveColumnToWorkspaceUp,
-        Op::MoveColumnToWorkspace(1),
-        Op::MoveColumnToWorkspace(2),
-        Op::MoveColumnToWorkspace(3),
+        Op::MoveColumnToWorkspaceDown(true),
+        Op::MoveColumnToWorkspaceUp(true),
+        Op::MoveColumnToWorkspace(1, true),
+        Op::MoveColumnToWorkspace(2, true),
+        Op::MoveColumnToWorkspace(3, true),
         Op::MoveWindowDown,
         Op::MoveWindowDownOrToWorkspaceDown,
         Op::MoveWindowUp,
@@ -2058,8 +2068,8 @@ fn workspace_transfer_during_switch_gets_cleaned_up() {
         },
         Op::RemoveOutput(1),
         Op::AddOutput(2),
-        Op::MoveColumnToWorkspaceDown,
-        Op::MoveColumnToWorkspaceDown,
+        Op::MoveColumnToWorkspaceDown(true),
+        Op::MoveColumnToWorkspaceDown(true),
         Op::AddOutput(1),
     ];
 
@@ -3352,6 +3362,69 @@ fn interactive_resize_on_pending_unfullscreen_column() {
     ];
 
     check_ops(&ops);
+}
+
+#[test]
+fn move_column_to_workspace_unfocused_with_multiple_monitors() {
+    let ops = [
+        Op::AddOutput(1),
+        Op::SetWorkspaceName {
+            new_ws_name: 101,
+            ws_name: None,
+        },
+        Op::AddWindow {
+            params: TestWindowParams::new(1),
+        },
+        Op::FocusWorkspaceDown,
+        Op::SetWorkspaceName {
+            new_ws_name: 102,
+            ws_name: None,
+        },
+        Op::AddWindow {
+            params: TestWindowParams::new(2),
+        },
+        Op::AddOutput(2),
+        Op::FocusOutput(2),
+        Op::SetWorkspaceName {
+            new_ws_name: 201,
+            ws_name: None,
+        },
+        Op::AddWindow {
+            params: TestWindowParams::new(3),
+        },
+        Op::AddWindow {
+            params: TestWindowParams::new(4),
+        },
+        Op::MoveColumnToOutput {
+            output_id: 1,
+            target_ws_idx: Some(0),
+            activate: false,
+        },
+        Op::FocusOutput(1),
+    ];
+
+    let layout = check_ops(&ops);
+
+    assert_eq!(layout.active_workspace().unwrap().name().unwrap(), "ws102");
+
+    for (mon, win) in layout.windows() {
+        let mon = mon.unwrap();
+        let ws = mon
+            .workspaces
+            .iter()
+            .find(|w| w.has_window(win.id()))
+            .unwrap();
+
+        assert_eq!(
+            ws.name().unwrap(),
+            match win.id() {
+                1 | 4 => "ws101",
+                2 => "ws102",
+                3 => "ws201",
+                _ => unreachable!(),
+            }
+        );
+    }
 }
 
 #[test]
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1734,9 +1734,12 @@ pub enum Action {
         reference: WorkspaceReference,
         focus: bool,
     },
-    MoveColumnToWorkspaceDown,
-    MoveColumnToWorkspaceUp,
-    MoveColumnToWorkspace(#[knuffel(argument)] WorkspaceReference),
+    MoveColumnToWorkspaceDown(#[knuffel(property(name = "focus"), default = true)] bool),
+    MoveColumnToWorkspaceUp(#[knuffel(property(name = "focus"), default = true)] bool),
+    MoveColumnToWorkspace(
+        #[knuffel(argument)] WorkspaceReference,
+        #[knuffel(property(name = "focus"), default = true)] bool,
+    ),
     MoveWorkspaceDown,
     MoveWorkspaceUp,
     MoveWorkspaceToIndex(#[knuffel(argument)] usize),
@@ -1970,10 +1973,14 @@ impl From<niri_ipc::Action> for Action {
                 reference: WorkspaceReference::from(reference),
                 focus,
             },
-            niri_ipc::Action::MoveColumnToWorkspaceDown {} => Self::MoveColumnToWorkspaceDown,
-            niri_ipc::Action::MoveColumnToWorkspaceUp {} => Self::MoveColumnToWorkspaceUp,
-            niri_ipc::Action::MoveColumnToWorkspace { reference } => {
-                Self::MoveColumnToWorkspace(WorkspaceReference::from(reference))
+            niri_ipc::Action::MoveColumnToWorkspaceDown { focus } => {
+                Self::MoveColumnToWorkspaceDown(focus)
+            }
+            niri_ipc::Action::MoveColumnToWorkspaceUp { focus } => {
+                Self::MoveColumnToWorkspaceUp(focus)
+            }
+            niri_ipc::Action::MoveColumnToWorkspace { reference, focus } => {
+                Self::MoveColumnToWorkspace(WorkspaceReference::from(reference), focus)
             }
             niri_ipc::Action::MoveWorkspaceDown {} => Self::MoveWorkspaceDown,
             niri_ipc::Action::MoveWorkspaceUp {} => Self::MoveWorkspaceUp,
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -438,14 +438,35 @@ pub enum Action {
         focus: bool,
     },
     /// Move the focused column to the workspace below.
-    MoveColumnToWorkspaceDown {},
+    MoveColumnToWorkspaceDown {
+        /// Whether the focus should follow the target workspace.
+        ///
+        /// If `true` (the default), the focus will follow the column to the new workspace. If
+        /// `false`, the focus will remain on the original workspace.
+        #[cfg_attr(feature = "clap", arg(long, action = clap::ArgAction::Set, default_value_t = true))]
+        focus: bool,
+    },
     /// Move the focused column to the workspace above.
-    MoveColumnToWorkspaceUp {},
+    MoveColumnToWorkspaceUp {
+        /// Whether the focus should follow the target workspace.
+        ///
+        /// If `true` (the default), the focus will follow the column to the new workspace. If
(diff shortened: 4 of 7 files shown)


the-fuckin-nobody:
Tried it and its working nicely. But, I think adding a `silent` flag would make sense(as yalter said while typing this response), as it would open more possibilities to silently move windows to workspaces with actions like, `move-column-to-workspace-up`, move-column-to-workspace-down`, etc.  Also similar thing for moving windows silently would be great too.

YaLTeR:
(a flag should not be called "silent", but rather something descriptive, like focus-target-workspace)

r0skar:
Sorry to chime in here but shouldnt this follow the same naming scheme as https://github.com/YaLTeR/niri/pull/1332?

YaLTeR:
Yes, thanks

yzy-1:
> Sorry to chime in here but shouldnt this follow the same naming scheme as #1332?

Done.

YaLTeR:
Did some minor fixes, and also decided to revert adding --focus to MoveColumnToWorkspace. It did not work correctly: it would still focus the target output. I looked into it and it would be a more complex refactor than I'd like to fix it properly (putting the column straight to the right workspace on an output rather than first to output and then moving). Are you fine with this?

yzy-1:
That is strange, the `--focus=false` option of `MoveColumnToWorkspace` works fine for me.

Could you please describe the steps that caused the bug?

My config file for testing (Almost the same as the default configuration. Only the name of the terminal emulator is changed and the focus=false parameter is added):

```kdl
// This config is in the KDL format: https://kdl.dev
// "/-" comments out the following node.
// Check the wiki for a full description of the configuration:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Overview

// Input device configuration.
// Find the full list of options on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Input
input {
    keyboard {
        xkb {
            // You can set rules, model, layout, variant and options.
            // For more information, see xkeyboard-config(7).

            // For example:
            // layout "us,ru"
        }
    }

    // Next sections include libinput settings.
    // Omitting settings disables them, or leaves them at their default values.
    touchpad {
        // off
        tap
        // dwt
        // dwtp
        // drag false
        // drag-lock
        natural-scroll
        // accel-speed 0.2
        // accel-profile "flat"
        // scroll-method "two-finger"
        // disabled-on-external-mouse
    }

    mouse {
        // off
        // natural-scroll
        // accel-speed 0.2
        // accel-profile "flat"
        // scroll-method "no-scroll"
    }

    trackpoint {
        // off
        // natural-scroll
        // accel-speed 0.2
        // accel-profile "flat"
        // scroll-method "on-button-down"
        // scroll-button 273
        // middle-emulation
    }

    // Uncomment this to make the mouse warp to the center of newly focused windows.
    // warp-mouse-to-focus

    // Focus windows and outputs automatically when moving the mouse into them.
    // Setting max-scroll-amount="0%" makes it work only on windows already fully on screen.
    // focus-follows-mouse max-scroll-amount="0%"
}

// You can configure outputs by their name, which you can find
// by running `niri msg outputs` while inside a niri instance.
// The built-in laptop monitor is usually called "eDP-1".
// Find more information on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Outputs
// Remember to uncomment the node by removing "/-"!
/-output "eDP-1" {
    // Uncomment this line to disable this output.
    // off

    // Resolution and, optionally, refresh rate of the output.
    // The format is "<width>x<height>" or "<width>x<height>@<refresh rate>".
    // If the refresh rate is omitted, niri will pick the highest refresh rate
    // for the resolution.
    // If the mode is omitted altogether or is invalid, niri will pick one automatically.
    // Run `niri msg outputs` while inside a niri instance to list all outputs and their modes.
    mode "1920x1080@120.030"

    // You can use integer or fractional scale, for example use 1.5 for 150% scale.
    scale 2

    // Transform allows to rotate the output counter-clockwise, valid values are:
    // normal, 90, 180, 270, flipped, flipped-90, flipped-180 and flipped-270.
    transform "normal"

    // Position of the output in the global coordinate space.
    // This affects directional monitor actions like "focus-monitor-left", and cursor movement.
    // The cursor can only move between directly adjacent outputs.
    // Output scale and rotation has to be taken into account for positioning:
    // outputs are sized in logical, or scaled, pixels.
    // For example, a 3840×2160 output with scale 2.0 will have a logical size of 1920×1080,
    // so to put another output directly adjacent to it on the right, set its x to 1920.
    // If the position is unset or results in an overlap, the output is instead placed
    // automatically.
    position x=1280 y=0
}

// Settings that influence how windows are positioned and sized.
// Find more information on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Layout
layout {
    // Set gaps around windows in logical pixels.
    gaps 16

    // - "never", default behavior, focusing an off-screen column will keep at the left
    //   or right edge of the screen.
    // - "always", the focused column will always be centered.
    // - "on-overflow", focusing a column will center it if it doesn't fit
    //   together with the previously focused column.
    center-focused-column "never"

    // You can customize the widths that "switch-preset-column-width" (Mod+R) toggles between.
    preset-column-widths {
        // Proportion sets the width as a fraction of the output width, taking gaps into account.
        // For example, you can perfectly fit four windows sized "proportion 0.25" on an output.
        // The default preset widths are 1/3, 1/2 and 2/3 of the output.
        proportion 0.33333
        proportion 0.5
        proportion 0.66667

        // Fixed sets the width in logical pixels exactly.
        // fixed 1920
    }

    // You can also customize the heights that "switch-preset-window-height" (Mod+Shift+R) toggles between.
    // preset-window-heights { }

    // You can change the default width of the new windows.
    // default-column-width {}

    // By default focus ring and border are rendered as a solid background rectangle
    // behind windows. That is, they will show up through semitransparent windows.
    // This is because windows using client-side decorations can have an arbitrary shape.
    //
    // If you don't like that, you should uncomment `prefer-no-csd` below.
    // Niri will draw focus ring and border *around* windows that agree to omit their
    // client-side decorations.
    //
    // Alternatively, you can override it with a window rule called
    // `draw-border-with-background`.

    // You can change how the focus ring looks.
    focus-ring {
        // Uncomment this line to disable the focus ring.
        // off

        // How many logical pixels the ring extends out from the windows.
        width 4

        // Colors can be set in a variety of ways:
        // - CSS named colors: "red"
        // - RGB hex: "#rgb", "#rgba", "#rrggbb", "#rrggbbaa"
        // - CSS-like notation: "rgb(255, 127, 0)", rgba(), hsl() and a few others.

        // Color of the ring on the active monitor.
        active-color "#7fc8ff"

        // Color of the ring on inactive monitors.
        inactive-color "#505050"

        // You can also use gradients. They take precedence over solid colors.
        // Gradients are rendered the same as CSS linear-gradient(angle, from, to).
        // The angle is the same as in linear-gradient, and is optional,
        // defaulting to 180 (top-to-bottom gradient).
        // You can use any CSS linear-gradient tool on the web to set these up.
        // Changing the color space is also supported, check the wiki for more info.
        //
        // active-gradient from="#80c8ff" to="#bbddff" angle=45

        // You can also color the gradient relative to the entire view
        // of the workspace, rather than relative to just the window itself.
        // To do that, set relative-to="workspace-view".
        //
        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view"
    }

    // You can also add a border. It's similar to the focus ring, but always visible.
    border {
        // The settings are the same as for the focus ring.
        // If you enable the border, you probably want to disable the focus ring.
        off

        width 4
        active-color "#ffc87f"
        inactive-color "#505050"

        // active-gradient from="#ffbb66" to="#ffc880" angle=45 relative-to="workspace-view"
        // inactive-gradient from="#505050" to="#808080" angle=45 relative-to="workspace-view"
    }

    // You can enable drop shadows for windows.
    shadow {
        // Uncomment the next line to enable shadows.
        // on

        // By default, the shadow draws only around its window, and not behind it.
        // Uncomment this setting to make the shadow draw behind its window.
        //
        // Note that niri has no way of knowing about the CSD window corner
        // radius. It has to assume that windows have square corners, leading to
        // shadow artifacts inside the CSD rounded corners. This setting fixes
        // those artifacts.
        // 
        // However, instead you may want to set prefer-no-csd and/or
        // geometry-corner-radius. Then, niri will know the corner radius and
        // draw the shadow correctly, without having to draw it behind the
        // window. These will also remove client-side shadows if the window
        // draws any.
        // 
        // draw-behind-window true

        // You can change how shadows look. The values below are in logical
        // pixels and match the CSS box-shadow properties.

        // Softness controls the shadow blur radius.
        softness 30

        // Spread expands the shadow.
        spread 5

        // Offset moves the shadow relative to the window.
        offset x=0 y=5

        // You can also change the shadow color and opacity.
        color "#0007"
    }

    // Struts shrink the area occupied by windows, similarly to layer-shell panels.
    // You can think of them as a kind of outer gaps. They are set in logical pixels.
    // Left and right struts will cause the next window to the side to always be visible.
    // Top and bottom struts will simply add outer gaps in addition to the area occupied by
    // layer-shell panels and regular gaps.
    struts {
        // left 64
        // right 64
        // top 64
        // bottom 64
    }
}

// Add lines like this to spawn processes at startup.
// Note that running niri as a session supports xdg-desktop-autostart,
// which may be more convenient to use.
// See the binds section below for more spawn examples.

spawn-at-startup "waybar"

// Uncomment this line to ask the clients to omit their client-side decorations if possible.
// If the client will specifically ask for CSD, the request will be honored.
// Additionally, clients will be informed that they are tiled, removing some client-side rounded corners.
// This option will also fix border/focus ring drawing behind some semitransparent windows.
// After enabling or disabling this, you need to restart the apps for this to take effect.
// prefer-no-csd

// You can change the path where screenshots are saved.
// A ~ at the front will be expanded to the home directory.
// The path is formatted with strftime(3) to give you the screenshot date and time.
screenshot-path "~/Pictures/Screenshots/Screenshot from %Y-%m-%d %H-%M-%S.png"

// You can also set this to null to disable saving screenshots to disk.
// screenshot-path null

// Animation settings.
// The wiki explains how to configure individual animations:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Animations
animations {
    // Uncomment to turn off all animations.
    // off

    // Slow down all animations by this factor. Values below 1 speed them up instead.
    // slowdown 3.0
}

// Window rules let you adjust behavior for individual windows.
// Find more information on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Window-Rules

// Work around WezTerm's initial configure bug
// by setting an empty default-column-width.
window-rule {
    // This regular expression is intentionally made as specific as possible,
    // since this is the default config, and we want no false positives.
    // You can get away with just app-id="wezterm" if you want.
    match app-id=r#"^org\.wezfurlong\.wezterm$"#
    default-column-width {}
}

// Open the Firefox picture-in-picture player as floating by default.
window-rule {
    // This app-id regular expression will work for both:
    // - host Firefox (app-id is "firefox")
    // - Flatpak Firefox (app-id is "org.mozilla.firefox")
    match app-id=r#"firefox$"# title="^Picture-in-Picture$"
    open-floating true
}

// Example: block out two password managers from screen capture.
// (This example rule is commented out with a "/-" in front.)
/-window-rule {
    match app-id=r#"^org\.keepassxc\.KeePassXC$"#
    match app-id=r#"^org\.gnome\.World\.Secrets$"#

    block-out-from "screen-capture"

    // Use this instead if you want them visible on third-party screenshot tools.
    // block-out-from "screencast"
}

// Example: enable rounded corners for all windows.
// (This example rule is commented out with a "/-" in front.)
/-window-rule {
    geometry-corner-radius 12
    clip-to-geometry true
}

binds {
    // Keys consist of modifiers separated by + signs, followed by an XKB key name
    // in the end. To find an XKB name for a particular key, you may use a program
    // like wev.
    //
    // "Mod" is a special modifier equal to Super when running on a TTY, and to Alt
    // when running as a winit window.
    //
    // Most actions that you can bind here can also be invoked programmatically with
    // `niri msg action do-something`.

    // Mod-Shift-/, which is usually the same as Mod-?,
    // shows a list of important hotkeys.
    Mod+Shift+Slash { show-hotkey-overlay; }

    // Suggested binds for running programs: terminal, app launcher, screen locker.
    Mod+T hotkey-overlay-title="Open a Terminal: alacritty" { spawn "foot"; }
    Mod+D hotkey-overlay-title="Run an Application: fuzzel" { spawn "fuzzel"; }
    Super+Alt+L hotkey-overlay-title="Lock the Screen: swaylock" { spawn "swaylock"; }

    // You can also use a shell. Do this if you need pipes, multiple commands, etc.
    // Note: the entire command goes as a single argument in the end.
    // Mod+T { spawn "bash" "-c" "notify-send hello && exec alacritty"; }

    // Example volume keys mappings for PipeWire & WirePlumber.
    // The allow-when-locked=true property makes them work even when the session is locked.
    XF86AudioRaiseVolume allow-when-locked=true { spawn "wpctl" "set-volume" "@DEFAULT_AUDIO_SINK@" "0.1+"; }
    XF86AudioLowerVolume allow-when-locked=true { spawn "wpctl" "set-volume" "@DEFAULT_AUDIO_SINK@" "0.1-"; }
    XF86AudioMute        allow-when-locked=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SINK@" "toggle"; }
    XF86AudioMicMute     allow-when-locked=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SOURCE@" "toggle"; }

    Mod+Q { close-window; }

    Mod+Left  { focus-column-left; }
    Mod+Down  { focus-window-down; }
    Mod+Up    { focus-window-up; }
    Mod+Right { focus-column-right; }
    Mod+H     { focus-column-left; }
    Mod+J     { focus-window-down; }
    Mod+K     { focus-window-up; }
    Mod+L     { focus-column-right; }

    Mod+Ctrl+Left  { move-column-left; }
    Mod+Ctrl+Down  { move-window-down; }
    Mod+Ctrl+Up    { move-window-up; }
    Mod+Ctrl+Right { move-column-right; }
    Mod+Ctrl+H     { move-column-left; }
    Mod+Ctrl+J     { move-window-down; }
    Mod+Ctrl+K     { move-window-up; }
    Mod+Ctrl+L     { move-column-right; }

    // Alternative commands that move across workspaces when reaching
    // the first or last window in a column.
    // Mod+J     { focus-window-or-workspace-down; }
    // Mod+K     { focus-window-or-workspace-up; }
    // Mod+Ctrl+J     { move-window-down-or-to-workspace-down; }
    // Mod+Ctrl+K     { move-window-up-or-to-workspace-up; }

    Mod+Home { focus-column-first; }
    Mod+End  { focus-column-last; }
    Mod+Ctrl+Home { move-column-to-first; }
    Mod+Ctrl+End  { move-column-to-last; }

    Mod+Shift+Left  { focus-monitor-left; }
    Mod+Shift+Down  { focus-monitor-down; }
    Mod+Shift+Up    { focus-monitor-up; }
    Mod+Shift+Right { focus-monitor-right; }
    Mod+Shift+H     { focus-monitor-left; }
    Mod+Shift+J     { focus-monitor-down; }
    Mod+Shift+K     { focus-monitor-up; }
    Mod+Shift+L     { focus-monitor-right; }

    Mod+Shift+Ctrl+Left  { move-column-to-monitor-left; }
    Mod+Shift+Ctrl+Down  { move-column-to-monitor-down; }
    Mod+Shift+Ctrl+Up    { move-column-to-monitor-up; }
    Mod+Shift+Ctrl+Right { move-column-to-monitor-right; }
    Mod+Shift+Ctrl+H     { move-column-to-monitor-left; }
    Mod+Shift+Ctrl+J     { move-column-to-monitor-down; }
    Mod+Shift+Ctrl+K     { move-column-to-monitor-up; }
    Mod+Shift+Ctrl+L     { move-column-to-monitor-right; }

    // Alternatively, there are commands to move just a single window:
    // Mod+Shift+Ctrl+Left  { move-window-to-monitor-left; }
    // ...

    // And you can also move a whole workspace to another monitor:
    // Mod+Shift+Ctrl+Left  { move-workspace-to-monitor-left; }
    // ...

    Mod+Page_Down      { focus-workspace-down; }
    Mod+Page_Up        { focus-workspace-up; }
    Mod+U              { focus-workspace-down; }
    Mod+I              { focus-workspace-up; }
    Mod+Ctrl+Page_Down { move-column-to-workspace-down; }
    Mod+Ctrl+Page_Up   { move-column-to-workspace-up; }
    Mod+Ctrl+U         { move-column-to-workspace-down; }
    Mod+Ctrl+I         { move-column-to-workspace-up; }

    // Alternatively, there are commands to move just a single window:
    // Mod+Ctrl+Page_Down { move-window-to-workspace-down; }
    // ...

    Mod+Shift+Page_Down { move-workspace-down; }
    Mod+Shift+Page_Up   { move-workspace-up; }
    Mod+Shift+U         { move-workspace-down; }
    Mod+Shift+I         { move-workspace-up; }

    // You can bind mouse wheel scroll ticks using the following syntax.
    // These binds will change direction based on the natural-scroll setting.
    //
    // To avoid scrolling through workspaces really fast, you can use
    // the cooldown-ms property. The bind will be rate-limited to this value.
    // You can set a cooldown on any bind, but it's most useful for the wheel.
    Mod+WheelScrollDown      cooldown-ms=150 { focus-workspace-down; }
    Mod+WheelScrollUp        cooldown-ms=150 { focus-workspace-up; }
    Mod+Ctrl+WheelScrollDown cooldown-ms=150 { move-column-to-workspace-down; }
    Mod+Ctrl+WheelScrollUp   cooldown-ms=150 { move-column-to-workspace-up; }

    Mod+WheelScrollRight      { focus-column-right; }
    Mod+WheelScrollLeft       { focus-column-left; }
    Mod+Ctrl+WheelScrollRight { move-column-right; }
    Mod+Ctrl+WheelScrollLeft  { move-column-left; }

    // Usually scrolling up and down with Shift in applications results in
    // horizontal scrolling; these binds replicate that.
    Mod+Shift+WheelScrollDown      { focus-column-right; }
    Mod+Shift+WheelScrollUp        { focus-column-left; }
    Mod+Ctrl+Shift+WheelScrollDown { move-column-right; }
    Mod+Ctrl+Shift+WheelScrollUp   { move-column-left; }

    // Similarly, you can bind touchpad scroll "ticks".
    // Touchpad scrolling is continuous, so for these binds it is split into
    // discrete intervals.
    // These binds are also affected by touchpad's natural-scroll, so these
    // example binds are "inverted", since we have natural-scroll enabled for
    // touchpads by default.
    // Mod+TouchpadScrollDown { spawn "wpctl" "set-volume" "@DEFAULT_AUDIO_SINK@" "0.02+"; }
    // Mod+TouchpadScrollUp   { spawn "wpctl" "set-volume" "@DEFAULT_AUDIO_SINK@" "0.02-"; }

    // You can refer to workspaces by index. However, keep in mind that
    // niri is a dynamic workspace system, so these commands are kind of
    // "best effort". Trying to refer to a workspace index bigger than
    // the current workspace count will instead refer to the bottommost
    // (empty) workspace.
    //
    // For example, with 2 workspaces + 1 empty, indices 3, 4, 5 and so on
    // will all refer to the 3rd workspace.
    Mod+1 { focus-workspace 1; }
    Mod+2 { focus-workspace 2; }
    Mod+3 { focus-workspace 3; }
    Mod+4 { focus-workspace 4; }
    Mod+5 { focus-workspace 5; }
    Mod+6 { focus-workspace 6; }
    Mod+7 { focus-workspace 7; }
    Mod+8 { focus-workspace 8; }
    Mod+9 { focus-workspace 9; }
    Mod+Ctrl+1 { move-column-to-workspace 1 focus=false; }
    Mod+Ctrl+2 { move-column-to-workspace 2 focus=false; }
    Mod+Ctrl+3 { move-column-to-workspace 3 focus=false; }
    Mod+Ctrl+4 { move-column-to-workspace 4 focus=false; }
    Mod+Ctrl+5 { move-column-to-workspace 5 focus=false; }
    Mod+Ctrl+6 { move-column-to-workspace 6 focus=false; }
    Mod+Ctrl+7 { move-column-to-workspace 7 focus=false; }
    Mod+Ctrl+8 { move-column-to-workspace 8 focus=false; }
    Mod+Ctrl+9 { move-column-to-workspace 9 focus=false; }

    // Alternatively, there are commands to move just a single window:
    // Mod+Ctrl+1 { move-window-to-workspace 1; }

    // Switches focus between the current and the previous workspace.
    // Mod+Tab { focus-workspace-previous; }

    // The following binds move the focused window in and out of a column.
    // If the window is alone, they will consume it into the nearby column to the side.
    // If the window is already in a column, they will expel it out.
    Mod+BracketLeft  { consume-or-expel-window-left; }
    Mod+BracketRight { consume-or-expel-window-right; }

    // Consume one window from the right to the bottom of the focused column.
    Mod+Comma  { consume-window-into-column; }
    // Expel the bottom window from the focused column to the right.
    Mod+Period { expel-window-from-column; }

    Mod+R { switch-preset-column-width; }
    Mod+Shift+R { switch-preset-window-height; }
    Mod+Ctrl+R { reset-window-height; }
    Mod+F { maximize-column; }
    Mod+Shift+F { fullscreen-window; }

    // Expand the focused column to space not taken up by other fully visible columns.
    // Makes the column "fill the rest of the space".
    Mod+Ctrl+F { expand-column-to-available-width; }

    Mod+C { center-column; }

    // Finer width adjustments.
    // This command can also:
    // * set width in pixels: "1000"
    // * adjust width in pixels: "-5" or "+5"
    // * set width as a percentage of screen width: "25%"
    // * adjust width as a percentage of screen width: "-10%" or "+10%"
    // Pixel sizes use logical, or scaled, pixels. I.e. on an output with scale 2.0,
    // set-column-width "100" will make the column occupy 200 physical screen pixels.
    Mod+Minus { set-column-width "-10%"; }
    Mod+Equal { set-column-width "+10%"; }

    // Finer height adjustments when in column with other windows.
    Mod+Shift+Minus { set-window-height "-10%"; }
    Mod+Shift+Equal { set-window-height "+10%"; }

    // Move the focused window between the floating and the tiling layout.
    Mod+V       { toggle-window-floating; }
    Mod+Shift+V { switch-focus-between-floating-and-tiling; }

    // Toggle tabbed column display mode.
    // Windows in this column will appear as vertical tabs,
    // rather than stacked on top of each other.
    Mod+W { toggle-column-tabbed-display; }

    // Actions to switch layouts.
    // Note: if you uncomment these, make sure you do NOT have
    // a matching layout switch hotkey configured in xkb options above.
    // Having both at once on the same hotkey will break the switching,
    // since it will switch twice upon pressing the hotkey (once by xkb, once by niri).
    // Mod+Space       { switch-layout "next"; }
    // Mod+Shift+Space { switch-layout "prev"; }

    Print { screenshot; }
    Ctrl+Print { screenshot-screen; }
    Alt+Print { screenshot-window; }

    // Applications such as remote-desktop clients and software KVM switches may
    // request that niri stops processing the keyboard shortcuts defined here
    // so they may, for example, forward the key presses as-is to a remote machine.
    // It's a good idea to bind an escape hatch to toggle the inhibitor,
    // so a buggy application can't hold your session hostage.
    //
    // The allow-inhibiting=false property can be applied to other binds as well,
    // which ensures niri always processes them, even when an inhibitor is active.
    Mod+Escape allow-inhibiting=false { toggle-keyboard-shortcuts-inhibit; }

    // The quit action will show a confirmation dialog to avoid accidental exits.
    Mod+Shift+E { quit; }
    Ctrl+Alt+Delete { quit; }

    // Powers off the monitors. To turn them back on, do any input like
    // moving the mouse or pressing any other key.
    Mod+Shift+P { power-off-monitors; }
}
```

YaLTeR:
It will work fine on the same monitor but if you move to a different monitor, then the target monitor will get focused even with focus=false

yzy-1:
Ok I'm trying to figure this out. Is there any way to simulate a test environment with multiple monitors? If not I might need some time to set up a complicated VM to test.

YaLTeR:
No easy way unfortunately. Though I guess you can get halfway there by writing a src/layout/tests unit test that does the operation and asserts that the focus remained where it should have

yzy-1:
I finished the code to work on multiple monitors and wrote simple tests. Does it work on your device? @YaLTeR

src/layout/mod.rs
```diff
@@ -2227,17 +2227,31 @@ impl<W: LayoutElement> Layout<W> {
...
+        if activate {
+            self.focus_output(output);
+            self.move_column_to_workspace(idx, activate);
+            return;
+        }
+
+        let Some(monitor) = self.monitor_for_output_mut(&output) else {
+            return;
+        };
+        monitor.move_column_to_workspace(idx, activate);
```
YaLTeR, line 2254: If I'm not mistaken, this will break if activate = false: you move a column to output without activation. So whatever column was previously focused on that output will remain focused. Then, `move_column_to_workspace()` will move that column, rather than the new column, to the target workspace.

yzy-1: Oops, that was my mistake! I'll try to fix it.

yzy-1: OK, is it working now?

src/layout/tests.rs
```diff
@@ -473,9 +473,14 @@ enum Op {
...
-    MoveColumnToWorkspaceDown,
-    MoveColumnToWorkspaceUp,
-    MoveColumnToWorkspace(#[proptest(strategy = "0..=4usize")] usize),
+    MoveColumnToWorkspaceDown(bool),
+    MoveColumnToWorkspaceUp(bool),
+    MoveColumnToWorkspace(#[proptest(strategy = "0..=4usize")] usize, bool),
+    MoveColumnToWorkspaceOnOutput(
+        #[proptest(strategy = "0..=5usize")] usize,
+        #[proptest(strategy = "0..=4usize")] usize,
+        bool,
```
YaLTeR, line 482: Make these named fields pls

src/layout/tests.rs
```diff
@@ -3354,6 +3367,61 @@ fn interactive_resize_on_pending_unfullscreen_column() {
...
+            ws_name: None,
+        },
+        Op::AddWindow {
+            params: TestWindowParams::new(1),
+        },
+        Op::FocusWorkspaceDown,
+        Op::SetWorkspaceName {
+            new_ws_name: 102,
+            ws_name: None,
+        },
```
YaLTeR, line 3385: Add a window here in order for this test to catch the last error that I pointed out

src/layout/tests.rs
```diff
@@ -3354,6 +3367,61 @@ fn interactive_resize_on_pending_unfullscreen_column() {
...
+            .iter()
+            .find(|w| w.has_window(win.id()))
+            .unwrap();
+
+        assert_eq!(
+            ws.name().unwrap(),
+            match win.id() {
+                1 | 3 => "ws101",
+                2 => "ws201",
+                _ => panic!("bad case"),
```
YaLTeR, line 3419:
```suggestion
                _ => unreachable!(),
```

src/layout/mod.rs
```diff
@@ -3595,7 +3595,9 @@ impl<W: LayoutElement> Layout<W> {
                 return;
             };
 
-            let workspace_idx = workspace_idx.unwrap_or(monitors[new_idx].active_workspace_idx);
+            let workspace_idx = workspace_idx.map_or(monitors[new_idx].active_workspace_idx, |v| {
```
YaLTeR, line 3598: I think it would look better if you first unwrap_or() then min() with len - 1 (clamp is unnecessary)

yzy-1:
Does it work now?

YaLTeR:
Seems to work now, thanks

YaLTeR:
what conflicts :skull:

PR #1479, opened 2025-04-26
Added an appendage to Steam Black Screen issue.

Quackdoc:
`-system-composer` can also sometimes fix the black screen issue without disabling gpu acceleration, greatly increases performance on high resolution displays.


YaLTeR:
Interesting, thanks

PR #1502, opened 2025-05-01
wiki: Add keyring info to electron in Application-Issues.md

marcinjahn:
I had issues running Mongo Compass, and the solution I applied is worth sharing I think.

### DIFF
--- a/wiki/Application-Issues.md
+++ b/wiki/Application-Issues.md
@@ -1,5 +1,7 @@
 ### Electron applications
 
+#### Wayland
+
 Electron-based applications can run directly on Wayland, but it's not the default.
 
 For Electron > 28, you can set an environment variable:
@@ -14,7 +16,23 @@ For previous versions, you need to pass command-line flags to the target applica
 --enable-features=UseOzonePlatform --ozone-platform-hint=auto
 ```
 
+
+
+#### Keyring
+
+Electron apps that store sensitive information use keyring providers. On [some desktop environments](https://www.electronjs.org/docs/latest/api/safe-storage), Electron will automatically 
+select the right keyring provider that the environment provides. When using Niri, you have to inform the Electron runtime about the keyring provider you want to use for a given app.
+As mentioned in the [Important Software section](./Important-Software.md), you will most likely want to install the `gnome-keyring` component on your system. To use it with Electron apps,
+you will have to provide the following flags to the app:
+
+```
+--password-store="gnome-libsecret" --ignore-additional-command-line-flags
+```
+
+---
+
 If the application has a [desktop entry](https://specifications.freedesktop.org/menu-spec/latest/menu-add-example.html), you can put the command-line arguments into the `Exec` section.
+
 
 ### VSCode
 


wiki/Application-Issues.md
```diff
@@ -14,8 +16,24 @@ For previous versions, you need to pass command-line flags to the target applica
 --enable-features=UseOzonePlatform --ozone-platform-hint=auto
 ```
 
+
+
+#### Keyring
+
+Electron apps that store sensitive information use keyring providers. On [some desktop environments](https://www.electronjs.org/docs/latest/api/safe-storage), Electron will automatically 
+select the right keyring provider that the environment provides. When using Niri, you have to inform the Electron runtime about the keyring provider you want to use for a given app.
```
YaLTeR, line 24: But I never had to do this and it works fine?

marcinjahn: Have you tried running https://flathub.org/apps/com.mongodb.Compass?

marcinjahn: My setup is Fedora 42 Workstation with niri session.

YaLTeR: I tried running this app, indeed without the CLI flag it shows the unable to access the credential storage toast. However, I'm also running VSCode (in a toolbox), and it can access the secret storage fine provided that libsecret is installed. Which makes me think this might not be niri related?

marcinjahn: Possibly VS Code has some internal logic that is able to find out which keyring to use and mongo compass does not have that, I don’t know.
Anyway, I thought that tip might be useful, feel free to close the PR if you think otherwise.

YaLTeR: Wonder if this happens with any other apps or on other compositors?

marcinjahn: That's the only app I saw having that issue until now

mohammedbilalns:
I had the same issues on Arch Linux with MongoDB Compass (from AUR) and Chromium ., they only retain secrets when started with the flag as @marcinjahn  mentioned.

wiki/Application-Issues.md
mohammedbilalns: I had the same issues on Arch Linux with MongoDB Compass (from AUR) and Chromium , they only retain secrets when started with the flag as @marcinjahn mentioned.

erdii: Signal Desktop (and lately also Matrix Desktop) have the same problem.

YaLTeR: Guess electron broke something?.. ugh

wiki/Application-Issues.md
```diff
@@ -14,8 +16,24 @@ For previous versions, you need to pass command-line flags to the target applica
...
+As mentioned in the [Important Software section](./Important-Software.md), you will most likely want to install the `gnome-keyring` component on your system. To use it with Electron apps,
+you will have to provide the following flags to the app:
+
+```
+--password-store="gnome-libsecret" --ignore-additional-command-line-flags
+```
+
+---
+
```
YaLTeR, line 33:
```suggestion
```

wiki/Application-Issues.md
```diff
@@ -14,8 +16,24 @@ For previous versions, you need to pass command-line flags to the target applica
...
 If the application has a [desktop entry](https://specifications.freedesktop.org/menu-spec/latest/menu-add-example.html), you can put the command-line arguments into the `Exec` section.
 
+
```
YaLTeR, line 36:
```suggestion
```

wiki/Application-Issues.md
YaLTeR: How do they pick by default? As in, why does it no longer work on niri by default. Do they look at XDG_CURRENT_DESKTOP or something else

YaLTeR, line 24: 1. On the wiki we wrap by sentence, so one newline after every `.` and nowhere else
2. It's lowercase niri when in the middle of a sentence

niklaskorz: Yes they do, see https://github.com/chromium/chromium/blob/95283ef55eed3e99dd63436f9072d788196f0c21/components/os_crypt/sync/key_storage_util_linux.cc#L25-L70 (Electron just calls the Chromium function directly, the only custom logic on the Electron side is checking the command line parameter to override it, but no equivalent environment variable unfortunately)

They also explicitly detect Cosmic to use gnome-libsecret already

niklaskorz: Setting `XDG_CURRENT_DESKTOP=niri:GNOME` also makes all Electron apps use gnome-libsecret by default

ghost: > Setting `XDG_CURRENT_DESKTOP=niri:GNOME` also makes all Electron apps use gnome-libsecret by default

Thanks. That works fine. Ente and Signal are now accessing the keyring :)

PR #1503, opened 2025-05-01
`center-focused-column "enclosed"`

ET-arded:
`center-focused-column "enclosed"`
Focused column will be centered if it is not the outermost column

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -174,6 +174,7 @@ pub enum CenterFocusedColumn {
     /// Focusing a column will center it if it doesn't fit on the screen together with the
     /// previously focused column.
     OnOverflow,
+    Enclosed,
 }
 
 #[derive(knuffel::DecodeScalar, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


ET-arded:
> Why?

it's like “always,” but without the waste of space for the outermost columns

YaLTeR:
Why would outermost columns waste space, but not inside columns? Also, this will still "waste space" in a setup like this:

image

I understand the idea, just not sure if it's useful enough over the existing options to justify adding another.

Also, there are questions like: should this behave more like always-center when centering middle columns, i.e. keep them centered even during/after resizing? Probably not because we don't have a robust way of doing it currently, but then the feature ends up a bit janky.

ET-arded:
> I tried to fix this by aligning the view to the leftmost or rightmost column, but that still left unused space when focusing on the middle columns (regarding the question), which would probably require to realign all columns? honestly I don't think this feature can be done correctly anymore.

It is also impossible to do correctly in the case where all columns are small enough to fit completely in the view

Issue #530, opened 2024-07-08
Multi-monitor window management checklist:


SalmanFarooqShiekh:
- [x] `focus-column-or-monitor-left/right` (merged to main)
- [x] `focus-window-or-monitor-up/down` (merged to main)
- [ ] `focus-window-or-workspace-or-monitor-up/down skip-empty=true/false` (nobody working on it) (default value/existence of `skip-empty` up for discussion)

- [ ] `move-column-left/right-or-to-monitor-left/right` (WIP in https://github.com/YaLTeR/niri/pull/528)
- [ ] `move-window-up/down-or-to-monitor-up/down` (nobody working on it)
- [ ] `move-window-up/down-or-to-workspace-up/down-or-to-monitor-up/down` (nobody working on it)

dyfrgi:
I think that right now, the implemented movement operations are pretty unintuitive. move-column-left-or-to-monitor-left doesn't move a column into the right-most column of the moved-to workspace, but rather it puts it to the right of the currently active column in the target workspace. Same for move-column-right-or-to-monitor-right. The result is that if you move a column back and forth it doesn't wind up in the same spot.
```
[A, B] [C, D*, E]
move-column-left-or-monitor-left
[A, B] [D*, C, E]
move-column-left-or-monitor-left
[A, B, D*] [C, E]
move-column-right-or-monitor-right
[A, B] [C, D*, E]  !!!
```
I'm not sure if this falls under the purview of this tracking ticket.

YaLTeR:
Want to make a PR fixing that? Shouldn't be too hard

SergioRibera:
Hi, is anyone working on these implementations? or can I take them? if taking them all in one PR or do you prefer separate PR?

YaLTeR:
Not sure

furcom:
This still isn't working, right? I couldn't find an option for "move-window-up/down-or-monitor-up/down".  
Currently, I am using a script I built with ChatGPT.  You can execute the script with the following options:

-mu = move window up
-md = move window down

Edit: Replaced the old Bash script with a faster Python version.
~Currently, the script does not work in Overview mode. It seems that move commands sent via IPC are blocked while Overview is active, so the script cannot control windows in that mode.~

Edit 2: Nevermind, now window movement is working again in overview. 😅

```
#!/usr/bin/env python3
"""
Usage:
    ./niri_move_updown.py [-mu|-md]
"""

import os
import sys
import json
import socket
import subprocess

[...]

PR #1509, opened 2025-05-02
Multi-monitor window management

SergioRibera:
Based on #530 

- [x] `move-window-up/down-or-to-monitor-up/down`
- [x] `focus-window-or-workspace-or-monitor-up/down skip-empty=true/false`
- [x] `move-window-up/down-or-to-workspace-up/down-or-to-monitor-up/down skip-empty=true/false`

At the moment I did not manage to replicate correctly the expected behavior of some functions but I open the PR as a draft in case someone wants to contribute or guide me in some things.

TODO:
- [ ] implement skip_empty


niri-config/src/lib.rs
```diff
@@ -1635,6 +1635,12 @@ pub enum Action {
     FocusWindowBottom,
     FocusWindowDownOrTop,
     FocusWindowUpOrBottom,
+    FocusWindowOrWorkspaceOrMonitorUp(
+        #[knuffel(property(name = "skip-empty"), default = true)] bool,
```
YaLTeR, line 1639: Can we do without skip_empty for now? Especially since it's unimplemented?..

src/layout/monitor.rs
```diff
@@ -598,16 +598,20 @@ impl<W: LayoutElement> Monitor<W> {
         }
     }
 
-    pub fn focus_window_or_workspace_down(&mut self) {
+    pub fn focus_window_or_workspace_down(&mut self) -> bool {
         if !self.active_workspace().focus_down() {
             self.switch_workspace_down();
+            return true;
```
YaLTeR, line 608: As far as I can tell, this will always return true regardless of whether the workspace is the last one, so it will never focus a different output. Did you test these actions?

src/layout/monitor.rs
```diff
@@ -586,16 +586,20 @@ impl<W: LayoutElement> Monitor<W> {
         true
     }
 
-    pub fn move_down_or_to_workspace_down(&mut self) {
+    pub fn move_down_or_to_workspace_down(&mut self) -> bool {
         if !self.active_workspace().move_down() {
             self.move_to_workspace_down();
+            return true;
```
YaLTeR, line 592: Same question

niri-config/src/lib.rs
SergioRibera: Hey, this is precisely the reason why the PR is in draft, I still don't know how to do the skip so I'm leaving it for later before having the PR ready

src/layout/monitor.rs
SergioRibera: I have tested it and I just had bugs with the behavior, maybe this is it, I haven't had time to test it again yet 😅

SergioRibera:
Well, this weekend I've been trying to advance in this PR but I can't, I'm going to close it for basically “skill issue” so if someone can do it, I would really appreciate it because every day I miss being able to move between all the screens with just keyboard

PR #1521, opened 2025-05-04
feat: import config 

HigherOrderLogic:
This PR allows to import/merge with other config, which would make it easier to integrate with other tools such as `matugen`.

I havent implement the actual merging functionality yet, as this is only a draft PR to gather some feedback on the merging strategy and the implementation direction.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -627,9 +627,9 @@ dependencies = [
 
 [[package]]
 name = "clap_complete"
-version = "4.5.50"
+version = "4.5.49"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c91d3baa3bcd889d60e6ef28874126a0b384fd225ab83aa6d8a801c519194ce1"
+checksum = "07ae023020f3bbb76bfd6c7b9dd3f903b40f60e4dc60696c303457c5c01e6cbe"
 dependencies = [
  "clap",
 ]
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -35,6 +35,8 @@ pub use utils::RegexEq;
 
 #[derive(knuffel::Decode, Debug, PartialEq)]
 pub struct Config {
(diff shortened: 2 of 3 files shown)


YaLTeR:
Well, you conveniently skipped basically all of the hard parts here.

- Obviously, merging is gonna be a huge refactor. My idea was to start with a kind of importing that doesn't require merging (so you can only have sections once, but this is already useful).
- Positionality. Window rules are positional (they are processed in order). If you include some window rules in the middle of other window rules, their positionality should work as expected.
- Watching. Included configs must also be live reloaded, including transitive includes, and breaking any cycles.

HigherOrderLogic:
> Positionality. Window rules are positional (they are processed in order). If you include some window rules in the middle of other window rules, their positionality should work as expected.

My idea for the merging strategy is whatever appear last will take precedence, or be appended to the array. The problem here is what if user import a file in the middle of another file? The current `Config` struct isnt really designed to handle this scenario though.

HigherOrderLogic:
How about instead of importing the config, we use `source PATH` and just simply replace the `source PATH` clause with the file content when parsing?

YaLTeR:
That would be the

> kind of importing that doesn't require merging

You still need to solve the

> Watching. Included configs must also be live reloaded, including transitive includes, and breaking any cycles.

And also you need to ensure that error messages point at the correct file which is harder with sourcing.

HigherOrderLogic:
@YaLTeR 

Would only watching and live realoading the main config and already imported files, and not live reloading when import new config files be okay?

YaLTeR:
No, it should live reload properly

HigherOrderLogic:
I mean, if you add new import in the main `config.kdl` file, the config will still reloaded, but if you edit the newly imported file (which isn't tracked) and save, Niri will not reload. Does that still count?

YaLTeR:
This case should also work correctly

PR #1526, opened 2025-05-04
Implement IPC for the overview state

CharlieQLe:
This adds the ability to query the overview's visible state and adds an event tracking the overview state. This seems to work fine, but I haven't tested it for long.

### DIFF
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -40,6 +40,9 @@ pub struct EventStreamState {
 
     /// State of the keyboard layouts.
     pub keyboard_layouts: KeyboardLayoutsState,
+
+    /// State of the overview.
+    pub overview: OverviewState,
 }
 
 /// The workspaces state communicated over the event stream.
@@ -61,6 +64,13 @@ pub struct WindowsState {
 pub struct KeyboardLayoutsState {
     /// Configured keyboard layouts.
     pub keyboard_layouts: Option<KeyboardLayouts>,
+}
+
+/// The overview state communicated over the event stream.
+#[derive(Debug, Default)]
+pub struct OverviewState {
+    /// State of the overview
+    pub opened: bool,
 }
 
 impl EventStreamStatePart for EventStreamState {
@@ -186,6 +196,22 @@ impl EventStreamStatePart for KeyboardLayoutsState {
                 let kb = self.keyboard_layouts.as_mut();
                 let kb = kb.expect("keyboard layouts must be set before a layout can be switched");
                 kb.current_idx = idx;
+            }
+            event => return Some(event),
+        }
+        None
+    }
+}
+
+impl EventStreamStatePart for OverviewState {
+    fn replicate(&self) -> Vec<Event> {
+        vec![Event::OverviewToggled { opened: self.opened }]
+    }
+
+    fn apply(&mut self, event: Event) -> Option<Event> {
+        match event {
+            Event::OverviewToggled { opened } => {
+                self.opened = opened;
             }
             event => return Some(event),
         }
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -207,7 +207,9 @@ impl EventStreamStatePart for KeyboardLayoutsState {
 
 impl EventStreamStatePart for OverviewState {
     fn replicate(&self) -> Vec<Event> {
-        vec![Event::OverviewOpenedOrClosed { is_open: self.is_open }]
+        vec![Event::OverviewOpenedOrClosed {
+            is_open: self.is_open,
+        }]
     }
 
     fn apply(&mut self, event: Event) -> Option<Event> {
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -70,7 +70,7 @@ pub struct KeyboardLayoutsState {
 #[derive(Debug, Default)]
 pub struct OverviewState {
     /// State of the overview
-    pub opened: bool,
+    pub is_open: bool,
 }
 
 impl EventStreamStatePart for EventStreamState {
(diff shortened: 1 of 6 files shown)


YaLTeR:
What's the use case?

CharlieQLe:
> What's the use case?

Auto-starting a launcher when the overview opens similar to how GNOME's overview works, or having a toggle button in a bar reflect the overview state.

src/ipc/client.rs
```diff
@@ -435,9 +436,27 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+
+            if json {
+                let response =
+                    serde_json::to_string(&response).context("error formatting response")?;
+                println!("{response}");
+                return Ok(());
+            }
+
+            let Overview { opened } = response;
+            println!("Overview: {opened}");
```
YaLTeR, line 458: Should print "Overview is open" or "Overview is closed"

src/ipc/server.rs
```diff
@@ -428,6 +428,16 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
             Response::FocusedOutput(output)
         }
         Request::EventStream => Response::Handled,
+        Request::Overview => {
+            let (tx, rx) = async_channel::bounded(1);
+            ctx.event_loop.insert_idle(move |state| {
+                let overview_state = state.niri.layout.is_overview_open();
+                let _ = tx.send_blocking(overview_state);
+            });
```
YaLTeR, line 436: Not needed, it's in `ctx.event_stream_state`

src/niri.rs
```diff
@@ -728,6 +728,7 @@ impl State {
         self.refresh_ipc_outputs();
         self.ipc_refresh_layout();
         self.ipc_refresh_keyboard_layout_index();
+        self.ipc_overview_toggled();
```
YaLTeR, line 731: Move call into `ipc_refresh_layout()`

src/niri.rs
```diff
@@ -663,7 +663,7 @@ impl State {
         state.ipc_keyboard_layouts_changed();
         // Focus the default monitor if set by the user.
         state.focus_default_monitor();
-
+
```
YaLTeR, line 666: Unrelated diff

src/ipc/server.rs
```diff
@@ -690,4 +700,22 @@ impl State {
             server.send_event(event);
         }
     }
+
+    pub fn ipc_overview_toggled(&mut self) {
```
YaLTeR, line 704: `ipc_refresh_overview()`

CharlieQLe:
Having used this for a bit now, I believe this is good to be if no more changes need to be made.

@YaLTeR Please let me know if there are any more changes I need to make.

niri-ipc/src/state.rs
```diff
@@ -63,6 +66,13 @@ pub struct KeyboardLayoutsState {
...
+/// The overview state communicated over the event stream.
+#[derive(Debug, Default)]
+pub struct OverviewState {
+    /// State of the overview
+    pub is_open: bool,
+}
+
 impl EventStreamStatePart for EventStreamState {
     fn replicate(&self) -> Vec<Event> {
         let mut events = Vec::new();
```
YaLTeR, line 78: Forgot to add it here and below. Which means that if you start event-stream when the overview is open, it won't be notified as such.

YaLTeR (changes requested):

niri-ipc/src/state.rs
CharlieQLe: I think I resolved this? Let me know if it looks good.

CharlieQLe:
> Pls fix formatting

:+1: Formatting should be fixed now.

YaLTeR:
Works well, just one thing that came to mind. Do we really want to take up a whole `niri msg overview` for this? Big subcommand name with very little utility as of now

CharlieQLe:
> Works well, just one thing that came to mind. Do we really want to take up a whole `niri msg overview` for this? Big subcommand name with very little utility as of now

Good point, renaming it to something like `is-overview-opened` might work out better.

YaLTeR:
Is it even needed much?

CharlieQLe:
I did just have a thought actually, on GNOME, the apps grid "hides" the overview when open (It really just shrinks the workspaces and moves them upwards to below the search bar). My own use-case for my own launcher to present a search bar when the overview is open, then I start typing, display a list of applications on the system. It works pretty well, but with the workspaces visible below the list, it isn't exactly the behavior I want, especially since I eventually want an app grid for my tablet in the future.

Should we fold additional overview properties such as `zoom` into the overview state, and have the ability to change them on the fly? There is precedence for temporarily adjusting the configuration, since that is what `niri msg output` does, but that would definitely require rethinking `niri msg overview` since it simply returns the current state of the overview. Additionally, it would probably be better to have a dedicated configuration IPC for temporarily adjusting the config.

YaLTeR:
This is way out of scope for this PR at least. Also reporting rapidly changing properties (zoom changes every frame during anims and gestures) needs design discussion, and transient configuration changes via IPC need their own whole separate design discussion.

For this PR, I suggest that we remove at least `niri msg overview`. Now if you need the one-time IPC request, we can keep it (perhaps renaming to OverviewState or something).

CharlieQLe:
> This is way out of scope for this PR at least. Also reporting rapidly changing properties (zoom changes every frame during anims and gestures) needs design discussion, and transient configuration changes via IPC need their own whole separate design discussion.

Yeah, I had a feeling that suggestion would be completely out of scope

> Though, maybe indeed in this case leave some kind of `niri msg overview-state`... Idk, hard bikeshedding :p

Sounds good to me.

YaLTeR:
Could you please rename them then?

CharlieQLe:
> Could you please rename them then?

:+1: Think I got it all.

YaLTeR:
Thanks!

PR #1548, opened 2025-05-09
fix: don't try to create a default config at path that exists

sullyj3:
Currently this bug has no actual consequences, we just continue silently on AlreadyExists in main()
(this line: https://github.com/YaLTeR/niri/blob/e9c6f08906143c3fec1ad1301d538bef4cbc1978/src/main.rs#L151).

This commit just eliminates the redundant attempt.


YaLTeR:
Makes sense

PR #1549, opened 2025-05-09
refactor(main): eliminate a `mut` from config load code in main

sullyj3:
I think this makes for marginally better readability, since you don't have to wonder whether config_errored is set anywhere else. It's also slightly terser.

Not fussed at all if you want to reject this since it's such a bikeshed/matter of opinion.


YaLTeR:
Eh

PR #1552, opened 2025-05-10
Stop including broken LFS files in source tarball

FedericoCeratto:
The release tarball at https://github.com/YaLTeR/niri/tags contains broken files due to github using `git archive` without LFS support. This change prevents the LFS entities  and the .gitattributes file from being included (source tarballs should be git-agnostic).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,12 @@
+# LFS configuration for images from the wiki
 *.png filter=lfs diff=lfs merge=lfs -text
+
+# Exclude LFS-tracked files from the tarball
+/wiki/img/ export-ignore
+
+# exclude .gitattributes itself from the tarball
+.gitattributes export-ignore
+
+# tip: can be tested using
+# git archive --format=tar.gz --output=source.tar.gz HEAD && \
+# tar tfvz source.tar.gz | grep -e '.png' -e '.gitattributes'


.gitattributes
```diff
@@ -1 +1,12 @@
+# LFS configuration for images from the wiki
 *.png filter=lfs diff=lfs merge=lfs -text
+
+# Exclude LFS-tracked files from the tarball
+/wiki/img/ export-ignore
```
YaLTeR, line 5: There are some .svg diagram sources there, is it ok to exclude them too?

YaLTeR (commented):
>  and the .gitattributes file from being included (source tarballs should be git-agnostic)

Why is that? Is this a problem?

.gitattributes
FedericoCeratto: Source tarballs are meant to contain only files needed for installation. It's not uncommon for Linux distributions and others to track source tarball contents in git where spurious "stray" file from git like .gitattributes .gitignore or the .git directory can create confusion.
I haven't seen broken .svg files in the tarball so I don't think there's a need to exclude them, but right now it's not that important. Ideally in future it would be nice to have the full content of the wiki (with the "real" png data) as part of the release.

YaLTeR: What about .gitignore? I see it's there when I download ZIP. Should it be excluded too?

FedericoCeratto: Ideally yes.

FedericoCeratto: Ideally yes (but the focus of this PR was to remove the broken files). Want me to add the filter? Or you can edit the change.

YaLTeR: Please add yes

dimitry-ishenko:
> Source tarballs are meant to contain only files needed for installation.

There are literally thousands of documentation packages in Debian/Ubuntu:
```
dimitry@medtm7cY0d:~$ apt list *-doc | wc -l
4351
```
And, many of them include some sort of image files, eg:
```
dimitry@medtm7cY0d:~$ dpkg -L libasio-doc | grep png | wc -l
37
```
These files qualify as "needed for installation" and to exclude them from the release archive is wrong.

<strike>Could you please explain what you mean by "broken files". What's broken about them?</strike>

Additionally, to guarantee reproducible builds once a release is tagged, the official released archive should include all files in the release. It makes it easier to verify the archive has not been compromised.

dimitry-ishenko:
@YaLTeR is there a reason the PNG files have to live in LFS? Are they too big for the main repo? I am asking because if you move them into the main repo, it would solve the problem for everyone IMHO.

EDIT:

From what I can tell, they only take up 3.4MB of space:
```
dimitry@laptop:~/arch/desktop/niri.d/niri$ du -hs docs/wiki/img/
3.4M	docs/wiki/img/
```

YaLTeR:
Dunno I did it mostly as a habit. In any case they're in LFS now and it works fine

YaLTeR:
One consideration is that some months ago GitHub really cut down on the ability to use it as an image CDN. I used to upload some screenshot to a GitHub issue, get the image link, and put on our wiki. But you can no longer do that, the images won't load from another domain.

So any new images that we might want to put on our wiki will have to go in the repo. With LFS it's fine, but without LFS it'll keep growing the repo size. I'm not sure I want to be constrained by this when thinking whether to add some new image to the wiki.

dimitry-ishenko:
> Dunno I did it mostly as a habit. In any case they're in LFS now and it works fine

> One consideration is that some months ago GitHub really cut down on the ability to use it as an image CDN. I used to upload some screenshot to a GitHub issue, get the image link, and put on our wiki. But you can no longer do that, the images won't load from another domain.
> 
> So any new images that we might want to put on our wiki will have to go in the repo. With LFS it's fine, but without LFS it'll keep growing the repo size. I'm not sure I want to be constrained by this when thinking whether to add some new image to the wiki.

LOL, these are pretty lame excuses 😜 but OK I am cool with that. If you want to stay with LFS, that's totally fine.

All I am trying to say is that you've accepted the above PR because someone's tooling didn't work with LFS and now your release tarballs are missing several images. It would make sense to either completely remove the wiki from the release tarball or package the images with it. 

I've been building Debian packages for niri (nice work by the way!) and having mismatching releases and tree is just a bit of a pain in the ass.

YaLTeR:
I mean I'm not holding my hopes up that anyone ever checks the wiki bundled in the package, let alone view the images

YaLTeR:
Also I must say that calling these lame excuses doesn't exactly motivate me to help the singular distro where this seems to be causing any sort of problem

YaLTeR:
But, thanks for the packaging work. I hope we can figure something out

dimitry-ishenko:
> Also I must say that calling these lame excuses doesn't exactly motivate me to help the singular distro where this seems to be causing any sort of problem

You are right, emotions got the better of me. It's your project and it is up to you to decide how to package it. I thought I was bringing up a valid point, but if you don't see that way, that's OK.

Yours is actually the only project I've come across that is kind enough to bundle vendored Rust dependencies, which I appreciate a lot.

If you ever change your mind, let me know.

dimitry-ishenko:
BTW I found this, which will probably make everyone happy:

[Managing Git LFS objects in archives of your repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-git-lfs-objects-in-archives-of-your-repository)

You can keep all the images in LFS, but they will be included as actual images in the release archives. This will handle the OP's and my problem -- two stones with one bird.

YaLTeR:
Okay, I think I enabled the checkbox now

dimitry-ishenko:
@YaLTeR thank you. Should I send a PR reverting the filter?

YaLTeR:
Sure, if it works now

YaLTeR:
Well this is awkward; I had to disable LFS in archives again because I got an email that niri-wm used 9 out of 10 GB of free LFS traffic, which I guess means that the niri tarball is downloaded enough to where it added up rather quickly. So since that didn't work, in https://github.com/niri-wm/niri/commit/38191826cb1e5fb9051a7e141fefe4941a2b4bed I moved LFS files in-repo as you suggested (after also optimizing them).

dimitry-ishenko:
Oh right, that was the other reason I stopped using it a while ago... :)

PR #1592, opened 2025-05-18
layer-shell: properly handle re-map

alex-huff:
According to the zwlr_layer_surface_v1 documentation: Unmapping a layer_surface means that the surface cannot be shown by the compositor until it is explicitly mapped again. The layer_surface returns to the state it had right after layer_shell.get_layer_surface. The client can re-map the surface by performing a commit without any buffer attached, waiting for a configure event and handling it as usual.

Before this commit, no configure event was sent when a client performed a commit without any buffer attached.

### DIFF
--- a/src/handlers/layer_shell.rs
+++ b/src/handlers/layer_shell.rs
@@ -176,6 +176,20 @@ impl State {
                 } else {
                     self.niri.mapped_layer_surfaces.remove(layer);
                     self.niri.unmapped_layer_surfaces.insert(surface.clone());
+
+                    // After layer surface unmaps it has to perform the initial commit-configure
+                    // sequence again. This is a workaround until Smithay properly resets
+                    // initial_configure_sent upon the surface unmapping itself as it does for
+                    // toplevels.
+                    with_states(surface, |states| {
+                        let mut data = states
+                            .data_map
+                            .get::<LayerSurfaceData>()
+                            .unwrap()
+                            .lock()
+                            .unwrap();
+                        data.initial_configure_sent = false;
+                    });
                 }
             } else {
                 let scale = output.current_scale();


YaLTeR:
Shouldn't the surface do an initial commit first? At least for xdg-toplevel, when a surface is unmapped we just add it back to the list of unmapped surfaces, then the initial configure logic will trigger next time a surface commits.

I suspect the correct fix is for Smithay to update initial its `initial_configure_sent` upon unmap. A similar fix was done recently for toplevels: https://github.com/Smithay/smithay/pull/1542, https://github.com/Smithay/smithay/pull/1571

alex-huff:
> I suspect the correct fix is for Smithay to update initial its `initial_configure_sent` upon unmap. A similar fix was done recently for toplevels: [Smithay/smithay#1542](https://github.com/Smithay/smithay/pull/1542), [Smithay/smithay#1571](https://github.com/Smithay/smithay/pull/1571)

I agree this sounds like a better fix.

alex-huff:
I raised an issue with smithay: https://github.com/Smithay/smithay/issues/1735.

YaLTeR:
This makes kitty work fine right? If so then this looks good for now

alex-huff:
> This makes kitty work fine right? If so then this looks good for now

Correct.

YaLTeR:
Thanks

YaLTeR:
After this commit, fuzzel starts with a 1-2 s delay when on the second monitor. Any idea?

fidelix:
> After this commit, fuzzel starts with a 1-2 s delay when on the second monitor. Any idea?

Just confirming that I can also reproduce this.
I think this other bug was also introduced:
#1640

alex-huff:
Looking in to it.

alex-huff:
@fidelix I believe popups and fuzzel should work now.

PR #1597, opened 2025-05-18
feat: close fuzzel if keybind is triggered again

HigherOrderLogic:
This PR allows the user to close Fuzzel by simply triggering `Mod+D` again, without reaching for other key like Esc.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -342,7 +342,7 @@ binds {
 
     // Suggested binds for running programs: terminal, app launcher, screen locker.
     Mod+T hotkey-overlay-title="Open a Terminal: alacritty" { spawn "alacritty"; }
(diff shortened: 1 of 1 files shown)


YaLTeR:
Is it worth it the extra bash and pkill though?.. I don't think

HigherOrderLogic:
> Is it worth it the extra bash and pkill though?.. I don't think

I think it's worth it, for 2 reasons:
- It allows user to toggle on and off without moving their hand. Given how the repo is built around "productivity", this is the better way
- This is an UI element, and the eyes arent sensitive enough to recognize the speed difference

This is just my 2 cents. Feel free to close the PR if you see fit.

YaLTeR:
Well it's just the default config, I think it's ok for ppl to change it if they want

HigherOrderLogic:
So should I just close this then?

YaLTeR:
Sure

Issue #1214, opened 2025-03-06
Add a way to skip cursor-only updates during VRR


Gwenodai:
I recently decided to give Niri another shot considering there's been quite a bit of development since I last checked in. Everything is really great so far, but I did notice something that makes me hesitant to use Niri as my daily driver, or at least not when I want to play certain games.

Normally when VRR is enabled the games refresh rate dictactes the the monitors refresh rate. However, with games where there is a visible on screen cursor this doesn't seem to be the case. Of the few games I've tested that make use of a visible cursor (BG3 and Mini Airways) they both seem to ignore the game refresh rate when the cursor is in motion. In that situation the cursor gets updated at the speed of the monitor, which for me is 144hz. This causes output to the monitor to be delivered at a full 144hz despite the game itself outputing a much lower framerate.

I'm not sure if this is expected behaviour in Niri at this point in its development or not, so I figured I should make a bug report. Especially since I'd previously made a [similar bug report](https://github.com/YaLTeR/niri/issues/855) which turned out to be a different issue due to my particular hardware situation at the time (which is no longer the case). 

I will note that I've tried the `disable-cursor-plane` debug flag in hopes that it might solve the issue, but it didn't work (although I doubted it would help here).

### System Information

* niri version: niri 25.02 (0.0.git.1904.26618f8d)

* Distro: Fedora 41

* GPU: AMD RX 6700 XT

* CPU: AMD R9 3900X

Gwenodai:
Ok there's one other strange thing I'd like to mention that I just realised after opening the issue. I just tested the issue in the Pixmark-Piano benchmark and noticed that while the issue also exists there (which makes sense as it's not doing anything special with the cursor, just rendering a GPU heavy scene), it doesn't when the benchmark is taken out of fullscreen. When only maximized, cursor movements don't incur extra VRR redraws at all and the actual refresh rate matches the benchmarks refresh rate. (See clip below)

This struck me as odd as I had expected that if anything, not being fullscreened (and assumedly not directly scanning out) would cause the cursor to redraw as often as it could. Having noticed this, I tried doing the same in BG3 and Mini Airways to no success. I also tried running them in a gamescope session to see if that would have any effect on the issue, but that also didn't change anything.

I don't know if this behaviour is specific to the benchmark, or if the VRR cursor issue is specific to just these games while not being present in other games with visible cursors of which I've not tested.

[...]

PR #1616, opened 2025-05-19
Add debug option to skip cursor-only updates while VRR is active

Gwenodai:
This PR adds a debug option that allows disabling cursor-only screen redraws when VRR is active.

This is a toggleable implementation of the patch supplied by @YaLTeR in https://github.com/YaLTeR/niri/issues/1214.

I've been using this for a few weeks myself and see no issues at all when used along side VRR on-demand in games. This change makes such a huge difference in certain games that I feel like it should be included as a debug option for now until a proper bug-free implementation is ready.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -2329,6 +2329,8 @@ pub struct DebugConfig {
     pub strict_new_window_focus_policy: bool,
     #[knuffel(child)]
     pub honor_xdg_activation_with_invalid_serial: bool,
+    #[knuffel(child)]
+    pub skip_cursor_only_updates_while_vrr: bool,
 }
 
 #[derive(knuffel::DecodeScalar, Debug, Clone, Copy, PartialEq, Eq)]
@@ -5286,6 +5288,7 @@ mod tests {
                 disable_monitor_names: false,
                 strict_new_window_focus_policy: false,
                 honor_xdg_activation_with_invalid_serial: false,
+                skip_cursor_only_updates_while_vrr: false,
             },
             workspaces: [
                 Workspace {
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -1422,6 +1422,12 @@ impl Tty {
             if debug.disable_cursor_plane {
                 flags.remove(FrameFlags::ALLOW_CURSOR_PLANE_SCANOUT);
             }
+            if debug.skip_cursor_only_updates_while_vrr {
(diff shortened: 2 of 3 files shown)


src/backend/tty.rs
```diff
@@ -1422,6 +1422,12 @@ impl Tty {
             if debug.disable_cursor_plane {
                 flags.remove(FrameFlags::ALLOW_CURSOR_PLANE_SCANOUT);
             }
+            if debug.skip_cursor_only_updates_while_vrr {
+                let output_state = niri.output_state.get_mut(output).unwrap();
```
YaLTeR, line 1426:
```suggestion
                let output_state = niri.output_state.get(output).unwrap();
```

wiki/Configuration:-Debug-Options.md
```diff
@@ -275,6 +276,20 @@ debug {
 }
 ```
 
+### `skip-cursor-only-updates-while-vrr`
+
+
+Forces the screen not to redraw at all from cursor input while vrr is active.
+
+Useful for games where the cursor isn't drawn internally to prevent erratic vrr shifts in response to cursor movement.
```
YaLTeR, line 285:
```suggestion
Skips redrawing the screen from cursor input while variable refresh rate is active.

Useful for games where the cursor isn't drawn internally to prevent erratic VRR shifts in response to cursor movement.

Note that the current implementation has some issues, for example when there's nothing redrawing the screen (like a game), the rendering will appear to completely freeze (since cursor movements won't cause redraws).
```

niri-config/src/lib.rs
```diff
@@ -2329,6 +2329,8 @@ pub struct DebugConfig {
     pub strict_new_window_focus_policy: bool,
     #[knuffel(child)]
     pub honor_xdg_activation_with_invalid_serial: bool,
+    #[knuffel(child)]
+    pub skip_cursor_only_updates_while_vrr: bool,
```
YaLTeR, line 2333:
```suggestion
    pub skip_cursor_only_updates_during_vrr: bool,
```

YaLTeR (approved):

Gwenodai:
Sorry, that request was a misclick. I was checking the PR on my phone and hit the little icon while scrolling.

YaLTeR (approved):

YaLTeR:
I want to do a hotfix release before adding new features

YaLTeR:
Thanks

PR #1618, opened 2025-05-20
add option to hide unbound actions in hotkey overlay

chinatsu:
while it is possible to hide actions with [custom hotkey overlay titles](https://github.com/YaLTeR/niri/wiki/Configuration:-Key-Bindings#custom-hotkey-overlay-titles), it would be nice to be able to hide actions without needing to bind them. this PR introduces a configuration flag, `hotkey-overlay { hide-unbound }`.

in my particular case, i prefer to always use 1 window per column layouts, so i have `consume-or-expel-window-(left|right)` unbound (plus a plethora of other actions). in order to remove them from the hotkey overlay, i currently need to assign a bogus hotkey and set a none custom title.

<details><summary>Without `hide-unbound`</summary>
<p>

image

</p>
</details> 

<details><summary>With `hide-unbound`</summary>
<p>

image

</p>
</details> 

since `hide-unbound` is an opt-in flag, it should not break/change hotkey overlays for anyone after an update.

also, i'm not sure if i've selected a good/descriptive name for this option, so i'm open for better name suggestions :)

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1032,7 +1032,7 @@ pub struct HotkeyOverlay {
     #[knuffel(child)]
     pub skip_at_startup: bool,
     #[knuffel(child)]
-    pub hide_unbound: bool,
+    pub hide_not_bound: bool,
 }
 
 #[derive(knuffel::Decode, Debug, Default, Clone, Copy, PartialEq, Eq)]
@@ -4610,7 +4610,7 @@ mod tests {
             },
             hotkey_overlay: HotkeyOverlay {
                 skip_at_startup: true,
-                hide_unbound: false,
+                hide_not_bound: false,
             },
             animations: Animations {
                 off: false,
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -4610,6 +4610,7 @@ mod tests {
             },
             hotkey_overlay: HotkeyOverlay {
                 skip_at_startup: true,
+                hide_unbound: false,
             },
             animations: Animations {
                 off: false,
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1031,6 +1031,8 @@ pub struct Struts {
 pub struct HotkeyOverlay {
     #[knuffel(child)]
     pub skip_at_startup: bool,
+    #[knuffel(child)]
+    pub hide_unbound: bool,
 }
 
 #[derive(knuffel::Decode, Debug, Default, Clone, Copy, PartialEq, Eq)]
--- a/src/ui/hotkey_overlay.rs
+++ b/src/ui/hotkey_overlay.rs
@@ -125,12 +125,7 @@ impl HotkeyOverlay {
     }
 }
 
-fn format_bind(
-    binds: &[Bind],
-    mod_key: ModKey,
-    action: &Action,
-    hide_not_bound: bool,
-) -> Option<(String, String)> {
+fn format_bind(binds: &[Bind], mod_key: ModKey, action: &Action) -> Option<(String, String)> {
     let mut bind_with_non_null = None;
     let mut bind_with_custom_title = None;
     let mut found_null_title = false;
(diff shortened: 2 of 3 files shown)


wiki/Configuration:-Miscellaneous.md
```diff
@@ -233,4 +235,13 @@ hotkey-overlay {
 }
 ```
 
+#### `hide-unbound`
+
+Set the `hide-unbound` flag if you don't want to see actions that haven't been bound.
```
YaLTeR, line 240:
```suggestion
By default, niri will show the most important actions even if they aren't bound to any key, to prevent confusion.
Set the `hide-unbound` flag if you want to hide all actions not bound to any key.
```

src/ui/hotkey_overlay.rs
```diff
@@ -125,7 +125,12 @@ impl HotkeyOverlay {
     }
 }
 
-fn format_bind(binds: &[Bind], mod_key: ModKey, action: &Action) -> Option<(String, String)> {
+fn format_bind(
+    binds: &[Bind],
+    mod_key: ModKey,
+    action: &Action,
+    hide_unbound: bool,
```
YaLTeR, line 132: I feel like it'll be better to filter out unbound keys in render() before calling this function, instead of introducing this extra argument.

niri-config/src/lib.rs
```diff
@@ -1031,6 +1031,8 @@ pub struct Struts {
 pub struct HotkeyOverlay {
     #[knuffel(child)]
     pub skip_at_startup: bool,
+    #[knuffel(child)]
+    pub hide_unbound: bool,
```
YaLTeR, line 1035: Maybe `hide_not_bound` so it matches the "(not bound)" string? Or is that too weird

YaLTeR (changes requested):

niri-config/src/lib.rs
chinatsu: i don't mind that!

src/ui/hotkey_overlay.rs
chinatsu: i agree. i'll have a look into this

chinatsu:
Had to leave for an appointment so I rushed through the suggestion to filter actions before the format function. Might not be the best way to do this, haha.

src/ui/hotkey_overlay.rs
```diff
@@ -285,6 +285,21 @@ fn render(
...
+        // If hiding unbound hotkeys is desired
+        // filter out actions that aren't present in the binds
+        actions = actions
+            .into_iter()
+            .filter(|action| {
+                binds
+                    .iter()
+                    .map(|bind| bind.action.clone())
+                    .collect::<Vec<Action>>()
+                    .contains(action)
```
YaLTeR, line 298: This is collecting all binds into a freshly allocated Vector for every action that you check. What you need is something along `.iter().any(|bind| bind.action == action)`

chinatsu: ah, yes of course.

i was thinking perhaps it'd be nice to do this filtering on 
```rust
let strings = actions
        .into_iter()
        // .. something clever here
        .filter_map(|action| format_bind(binds, mod_key, action))
        .collect::<Vec<_>>();
```

but perhaps that's not necessary. :)

src/ui/hotkey_overlay.rs
```diff
@@ -285,6 +285,15 @@ fn render(
         }
     }
 
+    if config.hotkey_overlay.hide_not_bound {
+        // If hiding unbound hotkeys is desired
+        // filter out actions that aren't present in the binds
+        actions = actions
+            .into_iter()
+            .filter(|action| binds.iter().any(|bind| bind.action == **action))
+            .collect();
```
YaLTeR, line 294: Actually, even better: you can use https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain

chinatsu: :bulb:

YaLTeR (approved):

YaLTeR:
Thanks

PR #1643, opened 2025-05-22
layer-shell: don't dismiss popups because of unmapped layer surfaces

### LINKED ISSUE #1640: Context menus/dropdowns don't work when kitty quick terminal running
### System Information

* niri version: 
niri 25.05 (v25.05-3-g71251a70)

* Distro: 
* Arch

* GPU: 
* RTX 3080 TI

* CPU:
12900k

No context menus or dropdowns etc work when the kitty quick-access-terminal process is running (it doesn't even have to be open).
Tried both qt and gtk apps.

Nothing related to the context menus and dropdowns even shows up in niri `msg event-stream`.
Everything starts working again immediately after I kill kitty.

alex-huff:
Fixes #1640


YaLTeR:
Good catch

alex-huff:
`update_keyboard_focus` resets `layer_shell_on_demand_focus` if it was unmapped. Is this not sufficient?

alex-huff:
> It may run later. Well doesn't hurt to just put the && in the right place

Ok.

alex-huff:
> Why write it in two places? Just write once and put the other thing in ()?

Good point.

PR #1650, opened 2025-05-22
input: do not force redraw to hide an already hidden cursor

Duncaen:
This fixes a visual issue I've noticed where it would "activate" "surfaces" which are not yet activated for a really short time. Maybe there is some deeper issue and this fix is just hiding it, not exactly sure.

I've tried to record a video of the issue, in this example I move the cursor above an element in the "Gnome Font Viewer", switch windows to deactivate the focus and then when i switch the font viewer back into the video but focus a terminal, when I type, the font viewers entry (second row first column 13s in) will flash gray shortly.

https://github.com/user-attachments/assets/17a99574-db5d-4278-9ebb-ad1d1f552fc5

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -472,6 +472,8 @@ impl State {
     }
 
     fn hide_cursor_if_needed(&mut self) {
+        // If the pointer is already invisible, don't reset it back to Hidden causing one frame
+        // of hover.
         if !self.niri.pointer_visibility.is_visible() {
             return;
         }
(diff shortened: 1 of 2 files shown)


YaLTeR:
Good catch! This indeed fixes the issue, but not because it skips a redraw, but rather because it no longer reverts the pointer visibility from fully invisible to the intermediate state, causing a one frame of hover. I added a comment and a similar fix to the other hiding place (hide on timeout).

YaLTeR:
Pushed another fix, now I believe it should work properly, with no unintended hovers during cursor warping (warp-mouse-to-focus or when focusing a different output)

PR #1670, opened 2025-05-25
Add cycle back feature for presets of column/window width/height.

two-horned:
Hello,

this is _kind of_ a **DUPLICATE** since there's already draft #1367 that attacks issue / discussion #1000, but
it hasn't seen any updates since more than a month! So I hereby request to use this solution.

I simply made the existing functions that switch to the next preset `const generics`.
The `FORWARDS` constant can be set to either true or false, and depending on the value
we cycle for- or backwards through the presets.

Code compiles and passes tests. Wiki has been (slightly) updated.

Best Regards
S.K.

### DIFF
--- a/src/layout/floating.rs
+++ b/src/layout/floating.rs
@@ -790,7 +790,7 @@ impl<W: LayoutElement> FloatingSpace<W> {
         let idx = self.idx_of(id).unwrap();
 
         let tile = &mut self.tiles[idx];
-        tile.floating_preset_width_idx = None;
+        tile.floating_preset_height_idx = None;
 
         let available_size = self.working_area.size.h;
         let win = tile.window();
--- a/src/layout/floating.rs
+++ b/src/layout/floating.rs
@@ -624,7 +624,7 @@ impl<W: LayoutElement> FloatingSpace<W> {
         }
     }
 
-    pub fn toggle_window_width(&mut self, id: Option<&W::Id>) {
+    pub fn toggle_window_width(&mut self, id: Option<&W::Id>, forwards: bool) {
         let Some(id) = id.or(self.active_window_id.as_ref()).cloned() else {
             return;
         };
@@ -632,18 +632,22 @@ impl<W: LayoutElement> FloatingSpace<W> {
 
         let available_size = self.working_area.size.w;
 
+        let len = self.options.preset_column_widths.len();
         let tile = &mut self.tiles[idx];
         let preset_idx = if let Some(idx) = tile.floating_preset_width_idx {
-            (idx + 1) % self.options.preset_column_widths.len()
+            (idx + if forwards { 1 } else { len - 1 }) % len
         } else {
             let current_window = tile.window_expected_or_current_size().w;
             let current_tile = tile.tile_expected_or_current_size().w;
 
-            self.options
+            let mut it = self
+                .options
                 .preset_column_widths
                 .iter()
-                .position(|preset| {
-                    let resolved = resolve_preset_size(*preset, available_size);
+                .map(|preset| resolve_preset_size(*preset, available_size));
+
+            if forwards {
+                it.position(|resolved| {
                     match resolved {
                         // Some allowance for fractional scaling purposes.
                         ResolvedSize::Tile(resolved) => current_tile + 1. < resolved,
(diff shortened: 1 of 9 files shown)


two-horned:
Last merge-commit simply synced my fork to avoid future conflicts.

niri-config/src/lib.rs
```diff
@@ -1870,12 +1870,19 @@ pub enum Action {
     #[knuffel(skip)]
     ResetWindowHeightById(u64),
     SwitchPresetColumnWidth,
+    SwitchPresetColumnWidthBack,
     SwitchPresetWindowWidth,
+    SwitchPresetWindowWidthBack,
     #[knuffel(skip)]
     SwitchPresetWindowWidthById(u64),
+    #[knuffel(skip)]
+    SwitchPresetWindowWidthByIdBack(u64),
```
YaLTeR: `ById` should go in the end, so `SwitchPresetWindowWidthBackById`, same below

src/layout/floating.rs
```diff
@@ -608,7 +608,23 @@ impl<W: LayoutElement> FloatingSpace<W> {
         }
     }
 
-    pub fn toggle_window_width(&mut self, id: Option<&W::Id>) {
+    pub fn toggle_window_width<const FORWARDS: bool>(&mut self, id: Option<&W::Id>) {
```
YaLTeR: Could you please replace all const generics with normal function parameters? We don't need to codegen all of this twice.

src/layout/floating.rs
```diff
@@ -608,7 +608,23 @@ impl<W: LayoutElement> FloatingSpace<W> {
...
+    }
+
+    pub fn toggle_window_height<const FORWARDS: bool>(&mut self, id: Option<&W::Id>) {
+        self.toggle_window::<FORWARDS, false>(id);
+    }
+
+    // We use const generic IS_WIDTH, because we cannot specifically pass the option as parameter
+    // without cloning it. This is because we'd need to pass immutable and mutable references,
+    // which is not possible.
+    fn toggle_window<const FORWARDS: bool, const IS_WIDTH: bool>(&mut self, id: Option<&W::Id>) {
```
YaLTeR: This does not need to be deduplicated with the function below, at least not in this PR. Not to mention you forgot to update a bunch of the code... So please revert this change, which also gets rid of the IS_WIDTH generic along with the problem in your comment.

wiki/Configuration:-Layout.md
```diff
@@ -169,7 +169,7 @@ layout {
 
 ### `preset-column-widths`
 
-Set the widths that the `switch-preset-column-width` action (Mod+R) toggles between.
+Set the widths that the `switch-preset-column-width` (and `switch-preset-column-width-back`) action (Mod+R) toggles between.
```
YaLTeR:
```suggestion
Set the widths that the `switch-preset-column-width` action (Mod+R) toggles between.
```
I thought about it and I don't think this is needed here tbh (same below)

YaLTeR (changes requested):
Thanks, some small changes.

niri-config/src/lib.rs
two-horned: Ok, fixed in my latest commits.

src/layout/floating.rs
two-horned: Ok, changed in my latest commits. Tbf, I only used `const generics` because the draft wanted to define distinct functions that deal with cycling back (which was obviously introducing redundancy), so I tried adhering as closely as possible.

wiki/Configuration:-Layout.md
two-horned: Ok, fixed in the latest commits.

src/layout/floating.rs
two-horned: This piece of code was essential to understand cycling forwards/backwards of presets. 
Code generation and logic is identical to the previous version, just more readable.

If you want to remove the `const generic` and use a flag instead I am happy to change that, but it's kinda weird to reintroduce duplicate code again.

wiki/Configuration:-Layout.md
two-horned: All in all: happy to see the first review after a little more than 3 months

src/layout/floating.rs
YaLTeR: Do you mind reintroducing the duplicate functions? It'll result in a smaller PR diff overall too. It's not a big deal, and we can do refactoring separately down the line.

YaLTeR:
> All in all: happy to see the first review after a little more than 3 months

Sorry, but there's a lot of open PRs, and I'm also fairly busy outside niri (and have been traveling).

Anyway, I think we can get this in before the release this Saturday.

YaLTeR:
Alright, I fixed it and finished it up. (Sidenote, the de-duplicated code did in fact still contain mistakes like using width in both width and height cases, even after the latest push.)

Also while testing, managed to find a small bug in preset height handling for floating windows (in a separate commit).

YaLTeR:
Thanks

two-horned:
Hey, just recently saw your messages.
Thanks for merging this feature.

PR #1671, opened 2025-05-25
feat: not focus follow mouse when column is fullwidth

The-Lost-Light:
This PR add an property `not-scroll-on-maximized` for `focus-follows-mouse`
When set `focus-follows-mouse not-scroll-on-maximized=true`, If current Column is at full width, the focus will not follow the mouse.

For 2 reasons:
- not sure a bug(since triggering instability), even if I maximize the window, the edge of the screen may still trigger the focus to shift to the out-of-view window.
- In the following settings, you can display adjacent objects on the edge without triggering focus when the mouse moves over them.
```kdl
layout{
  struts {
    left 16
    right 16
  }
}
```

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -373,6 +373,8 @@ pub struct Touch {
 pub struct FocusFollowsMouse {
     #[knuffel(property, str)]
     pub max_scroll_amount: Option<Percent>,
+    #[knuffel(property)]
+    pub not_scroll_on_maximized: Option<bool>,
 }
 
 #[derive(knuffel::Decode, Debug, PartialEq, Eq, Clone, Copy)]
@@ -4302,6 +4304,7 @@ mod tests {
                 focus_follows_mouse: Some(
                     FocusFollowsMouse {
                         max_scroll_amount: None,
+                        not_scroll_on_maximized: None,
                     },
                 ),
                 workspace_auto_back_and_forth: true,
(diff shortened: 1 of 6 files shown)


YaLTeR:
? just set `focus-follows-mouse max-scroll-amount="0%"` (like in the default config)

The-Lost-Light:
I don't think this will meet my needs.
If it were two 75%+50% window, I would still want to follow the focus.
However, if use like `max-scroll-amount="25%"`, there will be a problem that the current window is 100% width, the adjacent window is 25% width, but the mouse moves to the edge of the screen and switches to it

The-Lost-Light:
In fact, I always think that the width of the current window is small enough to make you willing to switch focus, which is more intuitive than how far you need to scroll to focus on the new window.

The-Lost-Light:
Assume a property min-scroll-current-width
- For max-scroll-amount="0%", can set min-scroll-current-width="99%"
- For tiny size mismatch, min-scroll-current-width can't represent(But in most settings, this will also shift the focus)
- But in some complex situations, like 50%(25% in screen)+50%(current screen)+100%(25% in screen)`min-scroll-current-width="50%"`might make some sense(especially when you always keep the focus center), in this case, It is difficult for most people to remember how wide the window is to be moved (and then deduct the current display to get the scroll length), but it is obvious when you simply see that the current window is only 50% wide.

Because these two methods cannot completely cover each other, which one is more suitable depends on your needs, but I think both methods can exist at the same time and be left to the user to choose, and can even be used at the same time (as long as one of the conditions is met)

PR #1677, opened 2025-05-26
Portrait mode

spacedentist:


YaLTeR:
> One thing I haven't tackled yet at all is the key bindings. The binds for moving left/right now move you up/down, which is confusing obviously. I was wondering whether introducing new bindings called smart-{left,right,up,down} would be a good solution - the idea being that they do "the right thing" to move you in the given direction. But then there are probably different ideas what "the right thing" should be.

Imo in portrait mode the focus-column-right/left should instead focus windows inside "row" right/left, and focus-window-up/down should instead focus "rows" up/down, and so on. This way your binds keep working "as expected". Same way as for floating windows, focus-column-right/left will actually focus "window" right/left.

PR #1679, opened 2025-05-26
Assume Mod key as held in Overview to replace hardcoded bindings

iostapyshyn:
The simplified navigation in Overview mode relies on [hardcoded keybindings](https://github.com/YaLTeR/niri/blob/8ba57fcf25d2fc9565131684a839d58703f1dae7/src/input/mod.rs#L4113), which currently include the arrow keys and the [scroll wheel](https://github.com/YaLTeR/niri/blob/8ba57fcf25d2fc9565131684a839d58703f1dae7/src/input/mod.rs#L2685). This is somewhat limiting for users who prefer custom navigation methods (like HJKL) or have defined further navigation keybindings. These custom bindings still work, but require holding down Mod, which feels inconsistent since arrow keys don’t.

This PR makes Overview mode treat the Mod key as held while open, both for keyboard and mouse.  All user-defined navigation keybindings (including arrow keys and scroll wheel in the default config) continue working as expected. Additionally, actions such as closing windows, moving/focusing windows, jumping to first/last windows, switching monitors, navigating workspaces with PageUp/PageDown get easier out of the box while in Overview.

For now, I removed `FocusColumn{Left,Right}UnderMouse` as they are not accessible from the config. I am not completely sure what is the use case for changing a focus on an inactive workspace and whether its worth complicating regular `FocusColumn{Left,Right}` to keep this functionality.

Also previously suggested by @Quackdoc in #1440

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1696,11 +1696,7 @@ pub enum Action {
     FocusWindowInColumn(#[knuffel(argument)] u8),
     FocusWindowPrevious,
     FocusColumnLeft,
-    #[knuffel(skip)]
-    FocusColumnLeftUnderMouse,
     FocusColumnRight,
-    #[knuffel(skip)]
-    FocusColumnRightUnderMouse,
     FocusColumnFirst,
     FocusColumnLast,
     FocusColumnRightOrFirst,
@@ -1751,11 +1747,7 @@ pub enum Action {
     CenterWindowById(u64),
     CenterVisibleColumns,
     FocusWorkspaceDown,
-    #[knuffel(skip)]
-    FocusWorkspaceDownUnderMouse,
     FocusWorkspaceUp,
-    #[knuffel(skip)]
-    FocusWorkspaceUpUnderMouse,
     FocusWorkspace(#[knuffel(argument)] WorkspaceReference),
     FocusWorkspacePrevious,
     MoveWindowToWorkspaceDown,
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -872,38 +872,12 @@ impl State {
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
-            Action::FocusColumnLeftUnderMouse => {
-                if let Some((output, ws)) = self.niri.workspace_under_cursor(true) {
-                    let ws_id = ws.id();
-                    let ws = {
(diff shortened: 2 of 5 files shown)


YaLTeR:
Well, my thoughts still stand.

> I'm reluctant to do this in general because maybe in the future we'll have some kind of typing, then it will conflict.

Besides, we have different binds in the overview than Mod + something. Notably, right mouse drag is scrolling in the overview and Mod+RMB is resizing outside. This is intentional, I think this way works better for the overview.

iostapyshyn:
What exactly do you mean by typing?

> Besides, we have different binds in the overview than Mod + something. Notably, right mouse drag is scrolling in the overview and Mod+RMB is resizing outside. This is intentional, I think this way works better for the overview.

I agree and these bindings are kept as is. In particular because resizing in Overview with mouse may feel really awkward. Other the other side, I have rebound my scroll wheel and now it does something completely different in the overview than outside.

YaLTeR:
> What exactly do you mean by typing?

Like, type to launch app or type to search.

iostapyshyn:
I see your point, having an input field with automatic focus would be indeed nice. Worth noting, this is already possible with existing launchers or window switchers. Having set `Mod+Slash { spawn "rofi" "-show" "window"; }`, I am one keystroke away from (`/`) searching a window (works fabulously also in Overview), or (`d`) starting an application, or both depending on the launcher configuration. With #977 or by subscribing `niri msg event-stream` the list can be sorted MRU, something I wanted to implement for my workflow. I am curious about your ideas what features a niri-native search/launch would bring to top that.

But the issue remains that until (and if) such input field gets implemented, Overview is restricted to a limited set of hardcoded keybindings, which may end up getting inconsistent with navigation outside of Overview after customizing. And even with typing, hardcoding a list of nonconflicting bindings is hardly a solution.

YaLTeR:
There are other solutions like some special bind flag that makes it work in the overview so you could bind individual keys to spawn a launcher or something. (I'm not saying I'm committed to that one, or other ideas yet.) Just, allowing all keys without Mod both locks out future native search, and isn't entirely consistent due to hardcoded binds

ArijanJ:
I would also love to be able to use hjkl to focus windows in the overview, but I would also like to be able to configure these to esdf or anything else custom that don't match my regular binds.

Ideally, it would be a separate `overview-binds {}` block.

PR #1685, opened 2025-05-27
feat: add dynamic overview zoom functionality

trbjo:
Regarding https://github.com/YaLTeR/niri/discussions/1673, here is what i meant. 

It uses a very simple algorithm to decide zoom. It maintains backwards compatibility so that users who prefer the current static zoom level will not be affected.

I did not change any logic pertaining to the centered workspace.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1331,10 +1331,82 @@ pub struct HotCorners {
     pub off: bool,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct ZoomRange {
+    pub min: FloatOrInt<0, 1>,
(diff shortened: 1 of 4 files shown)


cvanelteren:
> I see. Well, I'm still not sure if I want this. As I mentioned in the discussion, I think the current static zoom works fine. Besides complexity, dynamic zoom will make things harder to animate for adding/removing workspaces.
> 
> So I guess let's leave the PR here for now, maybe there will be more feedback

Why would it be harder to animate?  The animation isn't tied to the zoom level is it? Personally I would love to get a bigger overview on where my windows are. I frequently run into the issue that there is a window hidden horizontally that I am looking for. Naturally there would be other solutions for this but an overview would make it nice too.

PR #1690, opened 2025-05-27
Introduce WorkspaceDimensions struct

spacedentist:
This commit introduces WorkspaceDimensions to replace the triplet of view_size/workspace_area/scale that is passed from Workspace to ScrollingSpace and FloatingSpace.

This will also make it easier to support horizontal columns for portrait monitors, as we can add the notion of column axis to WorkspaceDimensions and also add member functions to access the various flavours of dimensions (like workspace size along the primary and secondary axis).

### DIFF
--- a/src/layout/dimensions.rs
+++ b/src/layout/dimensions.rs
@@ -1,0 +1,58 @@
+use smithay::utils::{Logical, Rectangle, Size};
+
+#[derive(Debug, Clone)]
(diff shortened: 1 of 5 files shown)


YaLTeR:
Can you check what other places use similar data? I know `Tile` and `layer::Mapped` use view_size + scale (I don't think they need working area, at least not yet). Floating uses view_size + scale + working_area.

PR #1699, opened 2025-05-30
Fix wayland event loop overflow after only a few hundred millisecond

KosmX:
This is needed on higher-end systems, because if I use a high polling rate mice (1kHz), and the app stops receiving events for 200 ms, that is enough to overflow the event buffer, and crashes the app.

With this patch, the buffer can be increased, and these apps can have more time handling inputs

the default buffer size is 4096 bytes, which is pretty small.  
The buffer can be set with the global `max-buffer-size` config.  
If the config is not specified, the value won't be overwritten.  

I'll set ready-to-merge on this PR when https://github.com/Smithay/wayland-rs/pull/815 is merged and released.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -174,7 +174,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Create the compositor.
     let mut event_loop = EventLoop::try_new().unwrap();
-    let display = Display::new().unwrap();
+    let mut display = Display::new().unwrap();
+    display.backend().set_max_buffer_size(1024 * 1024);
     let mut state = State::new(
         config,
         event_loop.handle(),


KosmX:
It will definitely fail to build without the updated wayland-rs :D

YaLTeR:
Good idea, but please no config option for this. Why would anyone want to configure their system to be broken? Just make it work by default. Copy whatever Mutter does

KosmX:
Okay, it is 1 MB (1024*1024)
https://gitlab.gnome.org/GNOME/mutter/-/blob/6b9aa7c8893787289e334a4487f1d1e1129877bc/src/wayland/meta-wayland.c#L936

KosmX:
Force pushed the buffer size, this time without config. And I did a quick rebase.

YaLTeR:
Looks good, don't forget to run rustfmt

KosmX:
rustfmt? what is that? :D
I'm not familiar with rust tooling, and i kinda did not know about rust formatter.

ids1024:
Wayland-rs is released, so this should now be able to use https://docs.rs/wayland-server/0.31.12/wayland_server/struct.DisplayHandle.html#method.set_default_max_buffer_size.

YaLTeR:
Added in b82d52705e1424cf47b26dd7b096832901c31f56

PR #1704, opened 2025-05-31
Window MRU, now with a UI

rustn00b:
This PR is an updated take on #977 which this PR supercedes. It 
provides a UI to navigate currently open windows sorted in most-recently-used
order (MRU). At present it is not yet ready for prime time, most notably
because features like animations are missing. However (imho) it has proven to
be quite usable!

Because it includes a separate UI to navigate thumbnails of the open windows, it
should not suffer from the same shortcoming as #977 where the latter would mess
up the current disposition of tabs and workspaces because it relied on actual
focus changes to perform the MRU traversal.

Here's the list of what is supported:
- activate the MRU UI using either `Alt-Tab` or `Alt-Grave`, the former traverses
  among all open windows ("None" filter), while the latter only traverses
  windows that share the same `app-id` as the currently focused window ("AppId"
  filter).

  Each window is represented by a thumbnail version of the original.

  Adding `Shift` to the key binding makes the traversal happen in the opposite
  order.
- when the UI is opened:
  - regular key bindings are filtered.
  - using `Alt-Tab` or `Alt-Grave` allow switching.
    traversal modes on the fly. Toggling the `AppId` filter uses the `app-id`
    of the current MRU selection (as opposed to the window that had the focus
    when the UI was opened).
  - releasing the `Alt` key transfers the focus to whatever thumbnail was
    selected in the MRU UI.
    
    Note: This implies that using key bindings that don't use the `Alt` modifier
    won't behave quite as expected because the UI will not close when they
    are released. So for now at least it's `Alt-*` only.
  - `Esc` key closes the MRU UI and returns focus to the
    window that originally held it when the UI was invoked.
  - `q` close the window/applicatoin for the currently selected
    thumbnail.
  - `a`, `o` or `w` changes the "scope" of the windows shown in the MRU UI:
    - `a`: consider all open windows
    - `o`: consider only windows on the current _Output_
    - `w`: consider only windows in the current _Workspace_
  - `j`/`k` and `Left`/`Right` can also be used to navigate thumbnails.
  - `Home`/`End` jump to the first (respectively last) thumbnail.

Key bindings are all **hard coded**. Those that open the MRU UI (`Alt-Tab`,
`Alt-Grave` and their `Shift`-ed variants) use a lower binding priority so
they shouldn't interfere with bindings already set in the configuration file.

Most glaring remaining todos:
- [ ] Animations
- [ ] Support for pointer actions, eg. clicking on thumbnail should close
      the UI and focus the corresponding window.
- [ ] Maybe some others, ymmv.

Despite all that, hope you all enjoy!


jbms:
It would be great for all the key bindings to be fully customizable. 

Also it would be great if this could be fully controlled by ipc (e.g. list of window IDs to include in preview), so that you could have layer shell application that handles its own key bindings and then modifies the preview.

rustn00b:
>  It would be great for all the key bindings to be fully customizable.

The tricky bit here is that the UI needs to be dismissed with the mod key (`Alt` for now) is released. Hard-coding the bindings side-steps that problem by just wiring the behavior directly in Niri. Something more flexible would involve figuring out what mod (mods?) were used to open the UI to then detect when they are released. Doable, but might be clunky, especially if there is a more general way to deal with key release bindings in the wings.

Wrt bindings used within the UI itself, either the config would need to support some form of "mode" or maybe a separate binding section would be needed for this UI. Idk.

> Also it would be great if this could be fully controlled by ipc.

This shouldn't be too difficult, basically the UI first builds a list of window IDs when it first opens. To fit into what you describe, the only difference would be to allow that list to be fed from IPC.

Although these ideas are definitely interesting, I think there's more work to be done on just the base functionality before tackling more new stuff (at least on my side)!

YaLTeR:
Before looking into the code, would it be feasible to have 1 commit with the previous MRU PR (that I reviewed) and then 1 commit with all the new stuff? So it's easier for me to review just the new changes.

> basically the UI first builds a list of window IDs when it first opens

Does it work with windows opening/closing dynamically?

rustn00b:
> Does it work with windows opening/closing dynamically?

Yes, at least that is the intent. You can actually use the `Alt-q` from within the UI to close windows as you navigate.

> would it be feasible to have 1 commit with the previous MRU PR (that I reviewed) and then 1 commit with all the new stuff.

Aside from the focus timestamp tracking, most of the MRU code has changed. I can try to create a branch from current main, apply/squash the old PR onto there and do that again with the new PR. From there do you mean that I should close the current PR and open a new one containing just those two commits?

rustn00b:
I set up a new branch with just the two commits on top of 8ba57fc, the first matches the most recent state of #977, and the second has the rolled up changes for this PR. It is available [here](https://github.com/rustn00b/niri/tree/window-mru-ui-squashed).

@YaLTeR, should I just close this PR and reopen another based on the "squashed" branch? Tell me whatever makes things simplest for you.

YaLTeR:
Oh, if most of the code changed, then I guess that's not necessary. Thanks for the branch, I can use it as a reference, so just keep it up for now.

I see that you bumped MSRV from 1.80 to 1.83 and applied some clippy lints (is_none_or). Is that strictly necessary? I don't have a concrete MSRV policy for niri but if it's just something minor like is_none_or then I'd prefer keeping 1.80 MSRV for now. Regardless, those changes should be in a separate PR.

There's also one formatting change to `window_matches()` that I noticed, why did that happen?

rustn00b:
> I see that you bumped MSRV from 1.80 to 1.83 and applied some clippy lints (is_none_or). Is that strictly necessary? 

It was needed to support `ControlFlow::Break` / `ControlFlow::Continue`, which made things look clearer than the version I came up with before.

> Regardless, those changes should be in a separate PR.

So, I'll wait on your opinion regarding the appropriateness of using `ControlFlow` in this PR, and assuming you are in agreement, then I'll break out the MSRV part and do a separate PR.

> There's also one formatting change to window_matches() that I noticed, why did that happen?

That was `cargo-fmt`, definitely nothing manual.

rustn00b:
> Regardless, those changes should be in a separate PR.

Available in PR #1717

rustn00b:
ac4c5d3b adds some animations giving the whole thing a more polished feel. Ideally transitions out of the UI should make it more obvious how the focus transfers to the focused window. I'm leaning toward a resize + move animation to make things nice and obvious.

rustn00b:
This last update adds a thumbnail selection animation that improves the UX as it gives a visual indication of where the window for the selected thumbnail will appear. For now, the selected thumbnail doesn't properly handle cases where the window is resized while the thumbnail is expanding (the animation skips directly to the requested aspect ratio instead of scaling smoothly) and it also doesn't display the focus ring. Implementing these would involve duplicating stuff already better done in tile.rs. Unfortunately I haven't managed to grok how to use the OffscreenBuffer properly, which seems like the way to do it right.

madeleine313:
Hello.

May I ask whether an option to disable this feature would be available? My concern being that I cannot use Alt and Tab at the same time as it interferes with some software/clients I use, and being able to disable this feature would be a work-around for not being able to customise the modifier(s)/keybindings if the bindings end up being hardcoded.

I would personally like to be able to use Super + Tab (which also brings up the "alt-tab menu" in GNOME and if I remember correctly Plasma as well) instead of Alt + Tab.

Thank you for your work and time!

rustn00b:
Hi there!

> May I ask whether an option to disable this feature would be available?

Not a bad idea, one could easily imagine a user either not wanting interfering keybinds or just wanting to use some different external tool to navigate recently used windows. I'll push this onto the todo list.

 > I would personally like to be able to use Super + Tab

Amusingly the PR from which this one is derived _did_ use `Super + Tab`, but the consensus on the comment thread at the time was that `Alt + Tab` was preferable! Anyway, this seems like a pretty good indication that a little time should be spent on making this at least somewhat configurable. One possible step in that direction would be to allow configuration of the modifier that opens the UI, e.g. `<MOD> + Tab`. Also added to the todos.

graham00:
All this looks great!

> > Also it would be great if this could be fully controlled by ipc.
> 
> This shouldn't be too difficult, basically the UI first builds a list of window IDs when it first opens. To fit into what you describe, the only difference would be to allow that list to be fed from IPC.

Just to add to that - I heavily use a type-to-seek ipc<->fuzzel script to jump to specifically titled applications and/or named workspaces:

```
#!/usr/bin/env bash
windows_list=$(niri msg -j windows | jq -r '.[] | "\(.id)\t\(.title) (\(.app_id))"')
workspaces_list=$(niri msg -j workspaces | jq -r '.[] | select(.name != null) | "WS: \(.name)"')
selected_line=$(printf "%s\n%s" "$windows_list" "$workspaces_list" | fuzzel --dmenu --width=200 --lines=55)
if [ -n "$selected_line" ]; then
    if [[ "$selected_line" == "WS: "* ]]; then
        workspace_name=$(echo "$selected_line" | sed 's/^WS: //')
        niri msg action focus-workspace "$workspace_name"
    else
        window_id=$(echo "$selected_line" | cut -f1)
        niri msg action focus-window --id "$window_id"
    fi
fi
```

Having MRU data available via IPC would be a godsend.

rustn00b:
> Having MRU data available via IPC would be a godsend.

30c5cf38 includes focus timestamp data in the the JSON output of `niri msg windows` (and in the `event-stream`). Timestamps are serialized using two fields, e.g.:

```json
    "focus_timestamp": {
      "secs_since_epoch": 1755189524,
      "nanos_since_epoch": 71588761
    }
```

If for whatever reason a window never received focus, the `focus_timestamp` should be `null` (I haven't tested this), though Serde also offers the option to just skip the field in that case which may be a better solution. idk.

This isn't quite "full control over IPC", but it should still provide enough data to implement external MRU navigation.

elcste:
This is amazing!!! I tried sway a few years ago but missed a good MRU switcher and went back to Gnome. Alt+grave for same app is the icing on the cake and `q` to close is the cherry on top :-)

I know this is not complete, but my feedback is that the animation is a little much for me when quickly switching back and forth between two windows. It looks good when pressing tab multiple times to select a window. I think Gnome Shell has a timeout and doesn't show an animation for their switcher if it is invoked very quickly.

Thanks again!

rustn00b:
@elcste, that makes sense. Adding a threshold time to trigger animations shouldn't be too hard to do so I'll just add that to the Todos. Thanks for the comment!

rustn00b:
0a7df425 inhibits the thumbnail selection animation unless the MRU UI was open for longer than a threshold duration. The threshold is hard-coded to 250ms, if that settings turns out not to be comfortable for everyone then it could be migrated to the "recent-windows" configuration section. Until such a time, I'm partial to keeping config clutter under control.

elcste:
That's a great improvement. I might play with the timeout and try slightly longer. I'll report back if I really prefer it.

One other thing I've noticed: If I press Alt+Tab with only one window open – or Alt+Grave with only one window open of the same app – focus doesn't return correctly to the window. Clicking in the window doesn't help. I have have to do something outside that window, like switch workspaces, open fuzzel or invoke this switcher (like I can Alt+Tab if I got stuck on Alt+Grave). The apps I tried with were Alacritty, Ptyxis, Firefox and Sublime Text in case it matters.

rustn00b:
> If I press Alt+Tab with only one window open – or Alt+Grave with only one window open of the same app – focus doesn't return correctly to the window

@elcste, I haven't run into this personally although I do use alacritty as my "test subject". My usual (manual...) tests include the cases you describe (calling the UI with 0, 1, several windows). It's also intriguing that even clicking in the window doesn't trigger a focus update. The only similar situation I've encountered is with niri-in-niri and switching windows in the "outer-niri" while a modifier key was active in the "inner-niri".

Is this something you've encountered with just the latest version from yesterday or is this something you've noticed in older versions?

elcste:
I think I noticed it before you added the animation threshold. I can't remember now if you had added the IPC commit or not.

A little more I noticed. It seems like it could relate to the Alt key still being active somehow? I'm seeing things that also happen if I hold the Alt key.
- After using the switcher with one Ptyxis window, when I type I get the app error "visual bell" of invalid input.
- In an app with a menu, the mnemonic letters are underlined.
  - In Sublime Text, the main menu bar. And focus seems to be stuck on the menubar, which I can flick
  - In Ptyxis and Gnome Disks if I press Alt+Grave with the app menus open, I see the mnemonic letters in those menus. 

I use some xkb options that move around some modifier keys like Alt, but I get the same result with the stock config file.

rustn00b:
@elcste: Ah, thanks for those indications. With your description I was able to hit it too, and now I'm wondering how I could have missed noticing it up until then! The tldr is that while the Mod (e.g. Alt) key press event was sent to the focused app, the corresponding release event was getting intercepted. So 63de4d08 now forwards the release event and this should hopefully address the issue you were seeing.

elcste:
@rustn00b That must have been it – I'm not seeing it after your latest updates :-)

Sempyos:
I've been playing with this for a while and have had no issues from my fiddling, great works !

YaLTeR:
Haven't looked at the code yet, some feedback from 2 minutes of playing:
1. Wow this is really cool! With anims and all even
2. Wait you even got the window titles working, nice
3. It should be `H` and `L`, not `J` and `K`
4. Not sure about the `Q`. It's right next to Tab, and there's no confirmation, making it very easy to close a window by accident.
5. The window shouldn't animate from within the Alt-Tab to its position on the workspace. Just fade out or close the menu, no extra anim necessary.
6. The anim when switching between `A`-`O`-`W` modes is a bit jank, the windows move horizontally out of nowhere and other windows jump in.
7. Pressing Alt-Tab shouldn't close the Overview, it should just draw over it.
8. It should be sized relatively to the monitor logical height (like the Overview). Right now, switching from 2560x1440 to 1280x720, or equivalently changing scale from 1 to 2, results in a 2x taller Alt-Tab, when it should result in the same size Alt-Tab.
9. Windows don't redraw in Alt-Tab as can be seen with `weston-simple-egl`, they should.
    - Pay attention to windows that change size like `wleird-resize-loop`. This should work fine.
11. Mouse should be able to at least click to select the target window (and touch and tablet too).

    Right now, mouse seems to go "below" the Alt-Tab. I was actually planning to look into refactoring our Keyboard/PointerTarget to better support modal dialogs, I believe that would be useful for Alt-Tab too to solve this kind of issue.
12. Moving mouse to another monitor switches the Alt-Tab to that monitor. This is fairly abrupt; I think it should stay where it was. Also, all monitors should probably be dimmed, not just the monitor with the Alt-Tab.
12. Not sure how you draw the border in Alt-Tab, but it should definitely be thicker, and also it breaks on rounded corners:

    image

    I've also seen the bottom border to be missing, which leads me to believe that you draw the border on the original Tile then downscale it, kinda like the Overview. Instead, perhaps try drawing the border on the already-downscaled Tile so that the border itself is not scaled?
13. Design-wise will need to play around and think more about it, remember what GNOME Shell does and what others do. Not prepared to give any feedback rn.

madeleine313:
> It should be H and L, not J and K

I think J and K are fine and work just like they do in e.g. qutebrowser, and I would believe that is the intention.
Is there anything wrong with keeping the current J and K behaviour and adding H and L as well?

YaLTeR:
> I was actually planning to look into refactoring our Keyboard/PointerTarget to better support modal dialogs

Scratch that, after looking into it I'm not sure it's worth the trouble, at least for now

rustn00b:
@YaLTeR: thanks for the feedback!

I'm not sure how prompt I'll be to get around to all of it though.

> It should be H and L, not J and K

I guess I'll just add H and L and keep J and K for now. (It's fairly trivial to change).

> Not sure about the Q. It's right next to Tab, and there's no confirmation, making it very easy to close a window by accident.

There isn't much finger motion going on there so I have never accidentally closed a window. And I actually have found it quite convenient to just hit Q when I come across a window that has is no longer useful. Technically Mod+Q doesn't ask for a confirmation either. I also got to play with a Mac and their equivalent doesn't prompt either (assuming the app doesn't have it own close confirmation dialog).

> The window shouldn't animate from within the Alt-Tab to its position on the workspace. Just fade out or close the menu, no extra anim necessary.

 Bummer, I kind of really like that (it was also some effort to put together) :/ An older version of this PR was actually implemented without a transition but I found it less intuitive. You would select a thumbnail and the focus would go ... somewhere... With the thumbnail selection animation your eyes track the thumbnail and the direction it leaves towards, so you end up looking where you need to to find the newly focused window.

> The anim when switching between A-O-W modes is a bit jank, the windows move horizontally out of nowhere and other windows jump in.

I played around and indeed got into a situation where switching between A-O-W made the selected thumbnail move even though there were enough other thumbnails (for a scope filters) to fill the space to either side. Under those conditions I would expect the selected thumbnail not to move. Back to the drawing board...

> Pressing Alt-Tab shouldn't close the Overview, it should just draw over it.

That's actually a recent change. It's a bit of a cop out because the thumbnail selection (I know you've mentioned it should be removed) didn't have logic to resize the thumbnail to the overview size as opposed to the regular window size. It didn't look very good so I took the easy way out on that one.

> It should be sized relatively to the monitor logical height (like the Overview). Right now, switching from 2560x1440 to 1280x720, or equivalently changing scale from 1 to 2, results in a 2x taller Alt-Tab, when it should result in the same size Alt-Tab.

For now the thumbnails are just scaled to half the logical size of their window counterpart. I've been using it that way for a while with two screens with very different sizes and it's been fine. However I do see situations like using portrait mode with a very tall screen that could lead to things looking weird. I also haven't experimented with changes in scale. I guess I would need to think more about this one.

> Windows don't redraw in Alt-Tab as can be seen with weston-simple-egl, they should.

I only figured out how to use the OffscreenBuffer later in the game, also using a snapshot avoids having to deal with thumbnails resizing on their own. From a usability stand point I haven't been bothered by the fact that the thumbnails don't have the latest content (the use case isn't quite the same as the overview's).

> Mouse should be able to at least click to select the target window (and touch and tablet too).

Agreed, this was already in the todos. And I often click on thumbnails as a reflex and am reminded I have to get around to this one.

> Moving mouse to another monitor switches the Alt-Tab to that monitor. This is fairly abrupt; I think it should stay where it was. Also, all monitors should probably be dimmed, not just the monitor with the Alt-Tab.

Indeed. I wasn't sure what the best approach would be here so I left it the way it was when I started. Dimming all monitors and only displaying the UI on whichever one it was initially opened on shouldn't be too difficult.

> perhaps try drawing the border on the already-downscaled Tile so that the border itself is not scaled?

The focus ring is drawn around the already downscaled thumbnail render, however the border is indeed rendered to a texture which is later downscaled.
For now the radius is static so there is a possible discrepancy between what the thumbnail and the window might be doing.

YaLTeR:
> I'm not sure how prompt I'll be to get around to all of it though.

No worries :slightly_smiling_face: 

> There isn't much finger motion going on there so I have never accidentally closed a window. And I actually have found it quite convenient to just hit Q when I come across a window that has is no longer useful. Technically Mod+Q doesn't ask for a confirmation either. I also got to play with a Mac and their equivalent doesn't prompt either (assuming the app doesn't have it own close confirmation dialog).

What key do they have?

Actually we could probably bind Alt+F4 which is a common shortcut, and which seems like it should work nicely with Alt-Tab? Alternatively, maybe hardcode Backspace to close?

> Bummer, I kind of really like that (it was also some effort to put together) :/ An older version of this PR was actually implemented without a transition but I found it less intuitive. You would select a thumbnail and the focus would go ... somewhere... With the thumbnail selection animation your eyes track the thumbnail and the direction it leaves towards, so you end up looking where you need to to find the newly focused window.

Hm, I'll need to play around with it more to see how bad the problem is and how else it can be solved. I guess there should be a simple way to disable this anim for that purpose?

My main problem with the current anim is that it doesn't make spatial sense. Like, your Alt-Tab window previews are clearly overlaid over your window layout, then once you confirm, out of nowhere the original window disappears, and the window preview animates into that position.

> I only figured out how to use the OffscreenBuffer later in the game, also using a snapshot avoids having to deal with thumbnails resizing on their own. From a usability stand point I haven't been bothered by the fact that the thumbnails don't have the latest content (the use case isn't quite the same as the overview's).

Sure it's usable but it also feels incomplete this way, so I'd very much prefer it to live-update. Regarding resizing windows, I'm curious what other Alt-Tabs do with those, like GNOME or KDE.

elcste:
> What key do they have?
> 
> Actually we could probably bind Alt+F4 which is a common shortcut, and which seems like it should work nicely with Alt-Tab? Alternatively, maybe hardcode Backspace to close?

MacOS uses Q (to close an app. It's maybe a little more natural since the switcher is Cmd+Tab and Cmd+Q is the regular shortcut to close an app.) I haven't heard of anyone saying its too easy to press, although probably a majority of Mac users don't use Cmd+Tab much ;-)

GNOME also has Q during the Super+Tab "Switch applications" to close an app.

As a former Mac user, I really really like this feature :-D It's something I've missed on other Linux switchers. Maybe it could be disabled by default or user a configurable binding if you don't love it?

YaLTeR:
Hmm, okay. Didn't know about Q in GNOME but indeed. And I never hit it there by accident. Okay, we can leave it then.

Perhaps the problem is with the W in that case? It's also right next to Q, and easier to slip to Q from W by accident than from Tab.

jhasse:
> I haven't heard of anyone saying its too easy to press, [...]

It's too easy to press. Usability nightmare and I hate it that GNOME copies it from macOS. If you look around you can find lots of people complaining about the Q shortcut, e.g.: https://www.reddit.com/r/firefox/comments/iml4tk/brilliance_of_ctrlq/

Chromium also only activates the shortcut if you hold it for a few seconds.

I've also lost work because of Ctrl+Q in the past and nowadays disable it like this: https://askubuntu.com/questions/1919/how-can-i-disable-ctrlq-for-all-applications-system-wide

elcste:
@jhasse To each their own :-) – I bind Ctrl+Q to close-window so I can close all windows that way, even apps that don't set that shortcut themselves. (Off topic: You could bind Ctrl+Q in Niri to some no-op behavior so Niri will intercept the shortcut before it gets to any app. I actually bind it to a script that looks at the window title and skips one app I don't want to be able to close.)

But back to this MRU switcher, I love the Q to close feature, but I think it would be great for it to be configurable, either which key or just enable/disable.

I agree with @YaLTeR that the W shortcut seems tricky. I like the idea of the A, O and W filters but feel like maybe they should be all next to each other rather them mnemonics. Like 1, 2 and 3 or maybe even [, ] and \ (varying keyboard layouts could make this hard).

rustn00b:
Commit 12ecae6 addresses the following:
- Changed key binding to close a thumbnail to `<Mod>+Shift+q`. So this isn't
  really making it configurable as suggested by @elcste, but should reduce the
  likelihood of accidentally closing an application (because the most frequent
  form of navigation - moving from most recent to least recent - doesn't
  involve the `Shift` key).
- Keep the MRU UI on the display it was initially opened in. The mouse pointer
  is also boxed into that display.
- Thumbnails can be selected using the mouse pointer (and presumably using
  a touch action, but I couldn't test if it actually works).
- Added thumbnail fade-out animation when the MRU UI is closed. This was the initial intent
  but I forgot to include it at some point.
- Fix thumbnail open animation that only had alpha on the Window title but not
  on the thumbnail itself.
- Added `<Mod>+h` and `<Mod>+l` for MRU navigation (without removing `j`/`k`)
- Fix unwanted motion when switching between scopes (point raised by @YaLTeR)

Regarding `A`/`O`/`W`, my thinking is that having a single binding, say
`<Mod>+s` (for scope) that just cycles through the different scopes is
probably more intuitive. A little indication of the current scope would
also be nice I suppose.

As a first step on the road to acceptance that the selection animation is going
to get tossed, I added an `enable-selection-animation` toggle for it in the
`recent-windows` part of the configuration (defaults to `false`).

To address the "not knowing where to look" issue that drove my perceived need
for a thumbnail selection animation, I was thinking that  the `FocusWindow`
action could adjust the view so as to position the selected window in the same
on-screen location as when it was last focused. While this won't really tell you
where to look, it would leverage your memory of where the window _should_ be.
This would involve keeping some extra state in the `ScrollingSpace`.

elcste:
> presumably using a touch action, but I couldn't test

I can test touch in the next couple days :-)

elcste:
Touch does work to select thumbnails.

I like the reduced animations when switching.

TIAcode:
Needs to be updated on Arch Linux, can't build without the libdisplay changes in a1dcced

Been using it for few weeks, it's absolutely great, hard to live without. Only tiny problem is I have one monitor that is exactly 2x2 bigger than my other one so when I bring up the UI on the smaller one and scroll to a full screen window on the bigger one, it's scaled to be exactly full screen on the smaller monitor.

madeleine313:
> Changed key binding to close a thumbnail to <Mod>+Shift+q. So this isn't
really making it configurable as suggested by @ elcste, but should reduce the
likelihood of accidentally closing an application (because the most frequent
form of navigation - moving from most recent to least recent - doesn't
involve the Shift key).

I haven't been able to test the latest changes, but can Shift be used as the configurable modifier, and if so, do the recent changes introducing the \<Mod\>+Shift+q binding account for it? Not that I think anyone would realistically want to use Shift as the main modifier, but you never know.

rustn00b:
@elcste & @madeleine313, thanks for the feedback on touch!

> Only tiny problem is I have one monitor that is exactly 2x2 bigger than my other one

Yup, that one's going to be next on the todo list (@YaLTeR also pointed it out). I just need to wrap up the Scope indicator and Scope cycling I've been tinkering with before.

> can Shift be used as the configurable modifier, and if so, do the recent changes introducing the <Mod>+Shift+q binding account for it? Not that I think anyone would realistically want to use Shift as the main modifier, but you never know.

Yes, shift can be used as the general modifier. I actually tried it out that way with niri-in-niri, and you indeed probably don't want that as your baseline config :)
Switching from just `q` to `Shift+q` doesn't really account for it, i.e. the shortcut would work but there is no special handling. Trickier is `Tab` vs `Shift+Tab`, if Shift is the general modifier for the MRU UI, then the "plain" `tab` action is no longer be usable.

jbms:
The key bindings all really ought to be configurable --- I guess the issue is just the lack of a general mechanism for doing that in niri at the moment?

YaLTeR:
They don't have to be configurable for this initial MR imo. Alt-Tab and Alt-Shift-Tab are natural and follow other systems. `q` also follows other systems as we've established.

As an update, I haven't had a chance to take another look at this yet (I always have it in the back of my mind though). Some refactors I'm working on may cause conflicts, but should be nothing too complex hopefully.

rustn00b:
> Only tiny problem is I have one monitor that is exactly 2x2 bigger than my other one so when I bring up the UI on the smaller one and scroll to a full screen window on the bigger one, it's scaled to be exactly full screen on the smaller monitor.

@TIAcode, hopefully d074d77 should fix the issue.

Vortriz:
I was testing this out and everything has been working perfectly!

Just one thing that I noticed: the window title displayed under the thumbnail is quite blurry. I am using my display on 1.5 scale rn, not sure if that could be the issue.

PS – Would be nice to have a configuration option for thumbnail size.

Sempyos:
Just tested this again today, my notes from this:
1. I agree with the post above about the title being blurry, I'm on 1.2 scaling
2. Mod+Shift+Q is a bit annoying. I personally preferred the Mod+Q for being inline with the rest of niri
3. Touch works great, however, it may be nice to be able to toggle the MRU once and be able to scroll through thumbnails via touch

Everything else works lovely !

YaLTeR:
Gave this another test. Some things I've noticed:
1. Definitely way clearer with the scope overlay. Though, it is a bit of an unfamiliar concept for Alt-Tab. But maybe it's fine since if you don't know about it then you don't have to care about it, and you get the normal behavior.
2. The fade-out animation looks much more natural, I like it!
3. Moving the mouse over windows doesn't visually change the focus; I expect it to do so.
4. In GNOME Alt-Tab, all application titles are always shown. I think this would be good; helps to navigate the windows quicker. Now of course you're showing windows here, not applications like GNOME, and window titles can get long, but I see you already added a fade effect to deal with this (nice touch on that!), so it should work fine?
5. Clicking outside the windows should close the thing.
6. Remind me, why is the 1.83 MSRV bump necessary?
7. Some stuff I already mentioned: it should close the overview, thumbnails should animate. I can take a look at this myself if you'd like.
8. Regarding the mod key. It's kinda weird how it's configured separately from the binds. Maybe it's possible to, like, only have the binds, and then when an MRU bind is recognized, use whatever mod keys are pressed at that moment?

Also, to my surprise (and delight), rebasing on top of the true maximize branch has only 2 trivial conflicts. So that resolves my reservation about merging true maximize. Please squash everything together into one commit when you rebase, makes it easier to deal with. (I can do it for you if you want, jj makes it very easy.)

I'm taking a *very* brief look at the code, I see you added an Offscreen render target. That doesn't really make sense, those render targets are only for block-out-from functionality. Alt-Tab should just block out windows as normal rendering, i.e. pass through the normal render target. I guess this ties in to the troubles you had with the overview zoom—Alt-Tab shouldn't go through that rendering path at all I think, just render tiles more directly?

Now, the design. I will have to use it some more to get a better feel, but perhaps we should consider something more like Windows Alt-Tab?

| Normal | Many windows |
| --- | --- |
| image | image |

It also shows full windows like we do here (unlike GNOME which shows application icons), but it tries to fit them all on one screen without scrolling.

Some problems that I feel with the current design in this PR:
- It's kind of too imposing, fading out the entire screen and showing large windows. It also feels a bit disorienting when I just want to Alt-Tab to the last window, in a way that the Windows Alt-Tab doesn't.
- It shows like 3.5 windows, and for the rest you have to (painstakingly) scroll to find—kind of the bringing back one of the problems Alt-Tab is designed to solve. With fullscreen windows, you can only see 2 of them, making the problem much worse.
- If I have to scroll, then I immediately get the urge that this is similar to opening the Overview, so I expect all of my gestures and stuff to work, like right click to scroll. (Otherwise it's way too unwieldy to use with a mouse.) I really don't want to have all this trouble and complexity in an Alt-Tab.

Here's the GNOME Alt-Tab for reference, and also how it handles very many windows (things just get smaller). I think for our case, something more Windows-like makes more sense.

| Normal | Many windows |
| --- | --- |
| image_2025-10-14_17-16-46 | image_2025-10-14_17-15-39 |

moving-hawk:
I'm wondering if we can group the windows by their icons/app-ids. If multiple windows belong to the same app-id, Alt-Tab will enter their window preview. Then we can use Alt-Tab to choose another one. Maybe we can also  use another key to go back to different apps if we choose the wrong app. 
But I'm not sure if this is a good idea.

YaLTeR:
1. I don't want the whole app icon resolution and rendering logic in niri, at least not now. 
2. I have a suspicion that niri users may generally have many terminal windows open at once, making this application grouping less useful than on GNOME.

rustn00b:
> The window title displayed under the thumbnail is quite blurry

@Vortriz and @Sempyos, My guess is that I failed to align the location of the texture containing titles with actual physical pixels. I have an un-pushed patch that I hope will help with this. It still needs to be cleaned up before I feel comfortable pushing it though.

@Sempyos, I agree with your opinion on the Mod+Shift+Q. Assuming there's a consensus, it's trivial to revert back to Mod+Q.

> it may be nice to be able to toggle the MRU once and be able to scroll through
  thumbnails via touch

Agreed, it's been on my todo list for a while now. It turned out to not be as straightforward to implement as I had hoped, so I've been kicking it down the pile. But I haven't given up on it yet!

@Yalter, thanks for the review!

> 3. Moving the mouse over windows doesn't visually change the focus; I expect   it to do so.

Added to the todos.

> 4. In GNOME Alt-Tab, all application titles are always shown. I think this  would be good

Added to the todos, this should be trivial. Initially it was intended to
make the selected thumbnail even more obvious (than when there's only the focus ring).

> 5. Clicking outside the windows should close the thing.

Added to the todos, no biggie.

> 6. Remind me, why is the 1.83 MSRV bump necessary?

I don't quite remember what had required it at some point. Reverting to 1.80.1 would require replacing a handful of `is_none_or`s.

> 7. it should close the overview, thumbnails should animate.

I had misunderstood your previous comment, I thought that you wanted the overview not to be closed when the MRU was opened. My plan was to take advantage of the close animation disable toggle to make this possible.

> 8. It's kinda weird how it's configured separately from the binds.

Yeah, but this was a way to deal with it being a _release_ binding. I haven't checked if any recent changes could maybe help improve this.

> I see you added an Offscreen render target. That doesn't really make sense

I'm afraid you're making obvious my very limited grasp of how things work on the Smithay end. I'll take the comment to mean I should probably be doing things differently.

> It's kind of too imposing, fading out the entire screen and showing large  windows.

My typical workflow using the UI goes something like: I used this window in the last X minutes, so I just tab through the first few and read the content directly in the thumbnails (typically shells). I guess there is some margin for that content to still remain readable while making the thumbnails smaller, but not all that much...

If I find myself scrolling for too long in the MRU list, I usually switch to the overview because I usually have a broad sense of where the window I'm looking for should be in the general layout.

Making the scale configurable, as suggested by @Sempyos, could be an option. Or maybe using the scroll wheel? A more Windows-y layout (spread on multiple rows) would require a more significant rework (I'm not quite ready to take that on just yet!).

I unfortunately won't have much spare time in the coming weeks. I'll try to get through the simpler stuff quickly and back out the fixes I was trying to sort out for the rounded corners (which weren't working anyway).

YaLTeR:
Thanks for the response!

> I'm afraid you're making obvious my very limited grasp of how things work on the Smithay end. I'll take the comment to mean I should probably be doing things differently.

This RenderTarget is a niri-specific thing actually, I made it for block-out-from (it's only used for that). Probably not obvious from the generic-sounding name.

> Reverting to 1.80.1 would require replacing a handful of `is_none_or`s.

If that's not a big deal, I'd prefer to do that. Doesn't sound like an important enough thing on its own to warrant bumping the MSRV. (Even if realistically bumping to 1.83 is likely a non-issue).

> I just tab through the first few and read the content directly in the thumbnails (typically shells). I guess there is some margin for that content to still remain readable while making the thumbnails smaller, but not all that much...

Hmm, that is a good point. With a Windows-style menu you don't get to do much reading. Now that I think about it, on Windows, hovering Alt-Tab for a bit will reveal the window, letting you inspect and pick the right one, but I'm not sure this will work well on niri (where it will cause much unnecessary scrolling around). I will try the current MRU design some more then and see how it feels.

> Making the scale configurable, as suggested by @Sempyos, could be an option. Or maybe using the scroll wheel? A more Windows-y layout (spread on multiple rows) would require a more significant rework (I'm not quite ready to take that on just yet!).

Let's hold off on any changes here for now until I got a better feel.

> I unfortunately won't have much spare time in the coming weeks. I'll try to get through the simpler stuff quickly and back out the fixes I was trying to sort out for the rounded corners (which weren't working anyway).

Alright, I'll wait for you to deal with the easy fixes and notify me here, then I'll see if I can finish this PR myself. Good timing before I dive into some other large refactor :)

Sempyos:
I'll test this later today and provide my feedback

YaLTeR:
@rustn00b is it good to take this over?

Sempyos:
from my testing:

1. yay having the Q keybind back for closing windows is so much nicer

2. There is sometimes a noticeable delay when clicking outside of the MRU to close it which skips the fade out animation

3. The MRU does not close on a tap/touch outside of it

4. The MRU does not inhibit applications below from receiving inputs while open, which may or may not be an issue 

5. Window titles are a lot less fuzzy on fractional scaling to me now

6. Touching thumbnails does work perfectly fine. However, it may be ideal for touchscreen support to be able to drag to scroll thumbnails

7. Thumbnails have no corner radius. I don't personally mind but may bug others. (could pull from config.kdl..?)

8. Currently difficult to tell which is the focused window in the picker without a border/focus ring

 sorry this took a bit

YaLTeR:
> @rustn00b is it good to take this over?

Well I already got started with my changes to this PR so yeah

rustn00b:
@YaLTeR , I have a few minor things still queued up (no longer close overview if MRU is called up and some clean ups). I'll push them today.

There are still some outstanding issues:
- Corner radius handling is incorrect (I was aiming to copy the corner radius set on the window)
- The selection animation still uses an Offscreen render (bcs I haven't figured out how to scale without first going through a texture)
- I just recently noticed there's wonkiness when CSD is used in the selection animation: there's a gap between the title and window content.

@Sempyos also pointed out the following:

> 2. There is sometimes a noticeable delay when clicking outside of the MRU to close it which skips the fade out animation

This one I hadn't noticed yet but I must also admit that I don't use the feature much so I'm less likely to come across it.

>  3. The MRU does not close on a tap/touch outside of it

Strange, I thought that was handled. Intended behavior was that when a thumbnail is touched/tapped then that triggers the MRU close with that thumbnail selected. Conversely, if the tap/touch was outside any thumbnail then the MRU is just dismissed.

 > 4. The MRU does not inhibit applications below from receiving inputs while open, which may or may not be an issue
 
 Strange again, I thought inputs (keyboard, pointer and tap) were all intercepted. I'm assuming I missed some other source.

> 6. ..., However, it may be ideal for touchscreen support to be able to drag to scroll thumbnails

Agreed, that one was on the todo list but I haven't gotten around to it.

> 7. Thumbnails have no corner radius.

This was the one I've been trying to address recently, as yet somewhat unsuccessfully.

> 8.  Currently difficult to tell which is the focused window in the picker without a border/focus ring

That's what initially led me to only add the title under the focused thumbnail (though that was removed in 
[ae71349](https://github.com/YaLTeR/niri/pull/1704/commits/ae71349f8d7b63ecc3f79803a8eca575faaa62aa).

Maybe a thicker focus ring is needed, idk.


> is it good to take this over?

@YaLTeR, just to state the obvious: you're more than welcome to plug away at this PR! :) I can't exactly claim that I've been blazingly fast getting it done.

YaLTeR:
> (no longer close overview if MRU is called up and some clean ups). I'll push them today.
> 
> There are still some outstanding issues:
> 
>     * Corner radius handling is incorrect (I was aiming to copy the corner radius set on the window)
> 
>     * The selection animation still uses an Offscreen render (bcs I haven't figured out how to scale without first going through a texture)
> 
>     * I just recently noticed there's wonkiness when CSD is used in the selection animation: there's a gap between the title and window content.

FWIW I'm already in the middle of reworking rendering and either fixed or will fix these. I may need to override your recent push. If you did any other changes, please tell what you did so I can merge them in locally.

rustn00b:
Thanks for your time looking at this! The only significant recent change was the one that stopped closing the overview when MRU was opened. The other stuff I have locally are half-baked experiments to render corners properly using ClippedSurfaceRenderElement. What you are in the process of doing now will most likely make these redundant.

YaLTeR:
Got it, thanks!

YaLTeR:
Interestingly, the GNOME Shell Alt-Tab binds Q to close application, and W to close one window (when in the window selector): https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/9d904a804e73c97a1ecde406f395cd77a53f10e7/js/ui/altTab.js#L199

For niri I think it makes sense for Q to close a window given the presumably-common "several terminals" usecase.

YaLTeR:
I'm gonna be pushing to https://github.com/YaLTeR/niri/tree/wip/window-mru-ui for now bc I'm in the middle of refactoring and some stuff is half-broken. Later on I can force-push into this PR to merge it

Sempyos:
Very excited to see what comes !

YaLTeR:
Ok this is still in the middle of refactoring, just figured I'd push here since I unbroke most things that I broke. The first commit is this PR in squashed form, all my changes follow on top. The changes are roughly:
- Made window previews live. Everything should be handled correctly, including resizing windows. I also made block-out-from work in alt-tab. (Though it should probably also block out the window title now that I think about it?)
- Window titles are correctly scaled incl. for fractional scaling.
- Some movement animation fixes. That said, the closing window animation is still missing bc it's be more involved, we'll see.
- Removed open fade in favor of open delay. This follows GNOME Shell, and the idea is to make quick alt-tabs not change anything on screen (to reduce annoyance).
- Made mod key Ctrl in nested so Alt-Tab is easily testable.
- Changed the visuals a bit.

For the visuals, the idea was:
1. Make windows smaller to fit more of them on larger screens, while still keeping them big enough to support the

    > I just tab through the first few and read the content directly in the thumbnails (typically shells).

    case. @rustn00b does it work for you? Maybe we can reduce size even further? You can play with the `PREVIEW_MAX_HEIGHT` constant.
2. Make it so overly large windows don't appear overly large (this is easy to hit with differently sized monitors).
3. Make selection clearly visible and mouse-selectable even in cases like:
    - fully white window preview
    - fully white window in the background
    - very tiny windows
    - weird focus ring settings

    So here I went with a big gray background rect instead.

Vortriz:
I tested the new changes. excellent! scaling and delay are working nicely.

One thing though, the window title has a tendency to kinda blend with the background, despite the gray padding.

image

I am not sure what, if any, improvements can be made regarding this.

YaLTeR:
I noticed this too, mostly a problem when there's some text underneath. Could try removing/reducing transparency from the gray bg

YaLTeR:
Added some pointer/touch input fixes, plus made Alt-Tab have a separate keyboard focus, meaning it doesn't result in an Alt press/release if you select the same window.

YaLTeR:
I reworked the keyboard handling. Here's how it works now: there's a `binds` section inside `recent-windows` specifically for binds that open the MRU. These binds take precedence over the general binds. The defaults are:
```kdl
recent-windows {
    // off

    binds {
        Alt+Tab         { next-window; }
        Alt+Shift+Tab   { previous-window; }
        Alt+grave       { next-window filter="app-id"; }
        Alt+Shift+grave { previous-window filter="app-id"; }

        Mod+Tab         { next-window; }
        Mod+Shift+Tab   { previous-window; }
        Mod+grave       { next-window filter="app-id"; }
        Mod+Shift+grave { previous-window filter="app-id"; }
    }
}
```

If this section is not present anywhere in the config, then these default binds are used. If this section *is* present anywhere in the config, then the default binds are not used, and you get to fill them in manually as you wish. This way:
1. If you have an old config without `recent-windows` then you will get the new Alt-Tab binds automatically.
2. If you don't like that you can set your own binds to whatever you want.
3. These binds **must** have a modifier key. The new key handling will close the Alt-Tab menu whenever all modifiers are released, this way you can bind whatever modifier you want and it will work.
4. Only available actions are `next-window` and `previous-window`, with optional properties `filter="app-id"` and `scope="all"`, `scope="output"`, `scope="workspace"`.

Then, while in the MRU UI, there are still a few hardcoded binds (left and right arrows), but also it loads in some of your regular binds:
- focus-column-left/right + variations become previous/next-window. So if you have Mod+h and Mod+l bound to this they will work in the MRU.
- focus-column-first/last to first/last in the MRU.
- close-window to close window in the MRU.

This way we don't have to hardcode HJKL, and if the user has, say, Colemak-DH binds instead, they will just work.

These binds loaded from regular ones will work without modifiers in the Alt-Tab menu, so if you have Mod+Shift+c bound to close-window, in the Alt-Tab just pressing `c` will be enough.

@rustn00b question, is the `s` to cycle scope needed much? Or we can leave just a/w/o. Also: maybe we should rename O - Output to M - Monitor?

YaLTeR:
Also for everyone testing: please write about any problems, bugs or jank. I know there's a whole bunch of internal cleanups and fixes left, but just to make sure that I don't miss something.

Also, if anyone has ideas for visual improvements, feel free to write.

yuukibarns:
I wonder if it is intended that the mru ui is independent of the `focus-window-previous` action? Window switching via mru ui seems not to update the previous window in `focus-window-previous` action. For example, if the `focus-window-previous` is between window A and B, and we are focusing on window A.  Then when we switch to window C via mru ui and call `focus-window-previous` again, we will come back to window B instead of window A.

YaLTeR:
I'm planning to replace focus-window-previous logic with the same as what MRU uses as the latter is better

yuukibarns:
The prompt `S`cope: `A`ll `W`orkspace `O`utput doesn't scale with the `scale` option.

YaLTeR:
Updates:
- window title is now also hidden when window is blocked out
- improved visibility somewhat especially when there's text underneath the window title
- urgent windows are now indicated in alt-tab
- pointer selection no longer scrolls the ui
- added WindowFocusTimestampChanged event, now it should be good to make custom MRU switchers or logic

rustn00b:
@YaLTeR: Awesome work you've been doing on this, thanks!

#### Scope binding
> question, is the s to cycle scope needed much? Or we can leave just a/w/o. Also: maybe we should rename O - Output to M - Monitor?

Using "monitor" instead of "output" might indeed more appropriate since that's probably what users would expect it to be called (but since the term "output" is also used elsewhere - e.g. `niri msg outputs`, idk). The `s` binding was added in response to the following comment from @elcste:

>  I like the idea of the A, O and W filters but feel like maybe they should be all next to each other rather them mnemonics.

The  shortcut was intended to keep everything in a single location for those who like that better, while still having a somewhat intuitive key binding. I've actually used it occasionally too.

#### Thumbnail scale
> does it work for you? Maybe we can reduce size even further? You can play with the PREVIEW_MAX_HEIGHT constant.

With my monitor it's already on the too-small-to-read side of things, I can still figure out which window is which using the shape of the text but I can no longer just read it. Not exactly a blocking issue, but I found the larger thumbnails more comfortable (though I may be alone in that respect ¯\_(ツ)_/¯ ).

YaLTeR:
> With my monitor it's already on the too-small-to-read side of things, I can still figure out which window is which using the shape of the text but I can no longer just read it. Not exactly a blocking issue, but I found the larger thumbnails more comfortable (though I may be alone in that respect ¯_(ツ)_/¯ ).

What (logical) resolution do you have? Also, if you play with PREVIEW_MAX_HEIGHT and PREVIEW_MAX_SCALE, what would be comfortable values? (The logic is something like `min(PREVIEW_MAX_HEIGHT, output height * PREVIEW_MAX_SCALE)`)

YaLTeR:
More updates! More config:

```kdl
recent-windows {
    open-delay-ms 150

    highlight {
        active-color "#999999ff"
        urgent-color "#ff9999ff"
        padding 30
    }

    preview {
        max-height 480
        max-scale 0.5
    }
}
```

Also:
- moved scope panel down, fixed scaling
- window previews now draw with their geometry-corner-radius
- focus-follows-mouse and touchpad swipes no longer trigger while MRU is open
- screenshot bind now works in the MRU letting you screenshot it (note that on some keyboards, Alt+PrtSc is a SysRq which may interfere)
- fixed redrawing when a window updates on a different monitor than the one the MRU is displayed on
- some animation fixes
- more internal fixes/cleanups

YaLTeR:
Some more updates:
- renamed `preview {}` to `previews {}`
- moved the scope panel back up since it's annoying to screenshot otherwise
- increased contrast on the scope panel after feedback from @stefonarch
- fixed a border radius first frame rendering bug present for god knows how long how did this never come up before
- added `highlight { corner-radius 32; }` setting (defaults to 0)

image

stefonarch:
Everything I tried  for ` previews { corner-radius 32; }` doesn't validate here.

YaLTeR:
Oops sorry it's `highlight { corner-radius 32; }` not previews of course

stefonarch:
ok, the ";" is optional I see :)

default 32 looks too much to me.

YaLTeR:
Default is 0, this was just an example

YaLTeR:
Something I've noticed about the bindings. For the filter (Alt-Tab vs. Alt-grave) you want it to always apply on the binding, i.e. pressing Alt-Tab should always switch to no filter, and pressing Alt-grave should always switch to current app's windows filter. However, for the scope (All/Workspace/Output) you don't really want it to always apply on every press, because then you can't switch between scopes while Alt-Tab is open. So maybe the `scope=""` property in binds should rather mean *default* scope (and only apply when initally opening Alt-Tab), and not scope on every key press?

Sempyos:
I'm going to attempt to play around with this later and provide my feedback, but from the commits it's been looking very nices ! 
Lovely works

YaLTeR:
Updates:
- did most of the internal cleanups
- improved closing fade-out anim (made it offscreen the contents to avoid transparent compositing artifacts)
- changed bind `scope` to affect only initial opening of the switcher, as I mentioned in the comment above
- reimplemented `focus-window-previous` action in terms of the (better) MRU focus timestamp logic
- wrote all docs: https://github.com/YaLTeR/niri/pull/1704/commits/c65916e273d603d7ff15ac5d4f3d7d9bfa3aec11

By now this is very close to merging, I just need to do a final pass over the diff. So if you find any more issues, please write about them.

Vortriz:
The docs mention:

> The way this works is by finding all regular binds corresponding to these actions and taking just the trigger key without modifiers.
For example, if you have <kbd>Mod</kbd><kbd>Shift</kbd><kbd>C</kbd> bound to `close-window`, in the window switcher pressing <kbd>C</kbd> on its own will close the window.

What will happen if <kbd>Mod</kbd><kbd>Shift</kbd><kbd>C</kbd> and <kbd>Mod</kbd><kbd>C</kbd> both are binded to permissible actions? Which one will be triggered?

YaLTeR:
> What will happen if ModShiftC and ModC both are binded to permissible actions? Which one will be triggered?

The one with fewest modifiers, so Mod+C

YaLTeR:
And in case of Mod+C and Alt+C, the first one in the binds list

yuukibarns:
I seem to find something not quite intuitive: when we switch from window A to an empty workspace below, then mru ui will still put the window A at the first, which means we will switch to window B after the window A in the list instead of more intuitive one, i.e., window A itself. However `focus-window-previous` does switch back to window A and then cycle between it and window B.

YaLTeR:
> I seem to find something not quite intuitive: when we switch from window A to an empty workspace below, then mru ui will still put the window A at the first, which means we will switch to window B after the window A in the list instead of more intuitive one, i.e., window A itself. However `focus-window-previous` does switch back to window A and then cycle between it and window B.

I did notice this today actually when reimplementing `focus-window-previous`. I cross-checked with GNOME, and their Alt-Tab also puts the window second, like niri currently does. I agree this feels less intuitive. I wonder what others do here like KDE or labwc?

stefonarch:
Labwc doesn't change order if the switcher is opened on an empty workspace, kwin does a strange thing when not filtering per workspace: 3 windows, switching between 1 and 2 as most recent, when opening the switcher on the empty workspace 3 is selected...

YaLTeR:
I additionally checked Windows. There, if you minimize all windows, Alt-Tab keeps the first window selected (doesn't switch next). But Alt-Shift-Tab also keeps the first window selected.

I think it's the most intuitive if in this case Alt-Tab keeps the first window, while Alt-Shift-Tab still selects the very last window. So that's what I just pushed in niri.

Also made scroll wheel binds work in the MRU (they are also in the "automatically picked up from normal binds" category, so make sure they're bound).

YaLTeR:
Some more cleanups and fixes, incl. made tablet input work in the MRU and regular mouse binds get autoconverted too.

YaLTeR:
Alright, everything's squashed now, should be good to merge. Unless some last minute issues come up.

Vortriz:
Tested with default config. Working without any hitches :+1:

yuukibarns:
I have a somewhat nuance feature request for the case when there is no window focused. Currently, the selected window will popup abruptly without any sliding animation but in the normal case there is a sliding animation from the right to the center.

YaLTeR:
Hmm? Not sure I understand

yuukibarns:
Sorry for my bad English but I mean that when there is a focused window (the normal case), the switcher will arise as the second window slides in from the right to the center as if the switcher just appears in the middle of the sliding animation. But in the case when there is no focused window, the current behaviour is popping up directly without any animation. I hope there can also be some "slide from right to center" animation for the "no focused window" case.

YaLTeR:
In the last commits I made it so that opening Alt-Tab from unfocused picks the *first* window rather than the second. That's why you don't see an animation. In general it's still there if you try Alt-Shift-Tab for example.

yuukibarns:
Yes, I see but I mean that I wonder if we can add some slide animation even for the first window. I know this is not for everyone (that's why I call it "somewhat nuance"), but I think the current popup behavior of the first window is a little abrupt.

YaLTeR:
Eh, idk. On other systems it also appears without animation. In niri (and GNOME) specifically there's already a 150 ms delay, so any further animation is adding even more delay until things "settle", I'd rather not probably.

YaLTeR:
Alright let's continue on main. Thanks again @rustn00b for all of the work, and everyone for intermediate testing!

rustn00b:
Thanks to all of those who chimed in here and made this happen!

PR #1706, opened 2025-05-31
force xdg deactivation on invisable workspaces

sashomasho:
This debug option provides a workaround for many Chromium-based chat applications that fail to show notifications when they're active in a workspace that's not currently visible and don't have keyboard focus

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -2329,6 +2329,8 @@ pub struct DebugConfig {
     pub strict_new_window_focus_policy: bool,
     #[knuffel(child)]
     pub honor_xdg_activation_with_invalid_serial: bool,
+    #[knuffel(child)]
+    pub force_xdg_deactivation_on_invisible_workspaces: bool,
 }
 
 #[derive(knuffel::DecodeScalar, Debug, Clone, Copy, PartialEq, Eq)]
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -361,6 +361,7 @@ pub struct Options {
     // Debug flags.
     pub disable_resize_throttling: bool,
     pub disable_transactions: bool,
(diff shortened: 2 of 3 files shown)


YaLTeR:
1. This is not a proper way to do it, a proper way would need some more refactoring, I'll need to think about it
2. It's not related to xdg-activation and this config name isn't good. I guess since I need to look into this myself anyway I'll come up with a better name when I do

sashomasho:
I wasn't too happy with the name either, but it was the best I could come up with at the time.

As for the code, I actually struggled to find where I could enforce this behavior. This issue in Chromium was a dealbreaker for me and ultimately kept me from using Niri. I placed it there because the logic for determining the active output seemed to be in that section - but of course, you know your codebase better than I do. Any fix or suggestion would be greatly appreciated.

YaLTeR:
https://github.com/YaLTeR/niri/blob/09da884cd86d343c50c449f6ab8ba5d775bdd380/src/layout/scrolling.rs#L3486 it should be here. When this flag is on, it should also ignore `is_tabbed` in the check above. You should pass a new bool argument into Workspace::refresh() called something like `is_focused: bool` in addition to is_active, because is_active should remain unchanged for the purposes of focus ring and all the other stuff.

This way will still keep them active while on a lock screen, but fixing that is a bigger and more complex refactor.

For the name, maybe `deactivate-unfocused-windows`?

sashomasho:
I'm not fully understanding the logic around is_tabbed, and obviously this PR doesn't address windows in tabs which will not be deactivated, but to fix that it looks like more work is required,  which probably will be an overkill for a debug flag.

YaLTeR:
Alright, this is good to go now. Please test it again when you have a chance

YaLTeR:
Thanks!

sashomasho:
Thanks, this is working even for tabbed windows, which is even better.

PR #1716, opened 2025-06-02
Add a flag to allow no importing environment when running as --session

wrvsrx:
Fix https://github.com/YaLTeR/niri/discussions/1715

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -26,6 +26,10 @@ pub struct Cli {
     /// on a TTY as your non-main compositor instance, to avoid messing up the global environment.
     #[arg(long)]
     pub session: bool,
+
+    #[arg(long)]
+    pub no_import_environment: bool,
+
     /// Command to run upon compositor startup.
     #[arg(last = true)]
     pub command: Vec<OsString>,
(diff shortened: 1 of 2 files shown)


YaLTeR:
I don't really see a valid use case where you don't want to import these into systemd while running as a session

wrvsrx:
I run `niri` manually to start a wayland session. I dump niri env to an `EnvironmentFile` and load them into my gui service. I don't want all my user services to be poluted by gui related env.

Is there any other way to disable environment importing without recompiling niri? Currently I patch niri on my nixos.

YaLTeR:
How is it "polluting" when it is required for normal operation?.. Idk this kinda sounds like a weird setup so I suggest continuing to patch this on your end. Especially considering it's a trivial patch

wrvsrx:
`Pollute` was a miswording on my part—apologies. I simply want to minimize the environment variables accessible to non-GUI services in my user systemd setup. Therefore, I'm using `EnvironmentFile` to import Niri's environment variables for the necessary GUI applications. If you think this isn't needed, I can maintain the patch myself.

PR #1717, opened 2025-06-02
Bump MSRV to 1.83

rustn00b:
- Adds support for ControlFlow::Break/Continue (used in #1704)
- Requires applying a few new clippy lints

This PR was initially just blended into #1704, but is rather independent so submitted separately at @YaLTeR's request.


bluelinden:
Closing as niri's MSRV is now 1.85, thanks so much for the contribution!

PR #1728, opened 2025-06-04
Automatic xwayland-satellite integration

YaLTeR:
When a recent enough (with [this PR](https://github.com/Supreeeme/xwayland-satellite/pull/169)) xwayland-satellite is detected, niri will create the X11 sockets and set `DISPLAY`, then automatically spawn `xwayland-satellite` when an X11 client tries to connect. If Xwayland dies, niri will keep watching the X11 socket and restart `xwayland-satellite` as needed. This is very similar to how built-in Xwayland works in other compositors.

How to test:
1. Install niri with this PR and xwayland-satellite with [that PR](https://github.com/Supreeeme/xwayland-satellite/pull/169).
2. Remove any manual `spawn-at-startup "xwayland-satellite"` and `environment { "DISPLAY": ":0"; }`.
3. It should just work. Start niri, open some X11 app. Try killing xwayland-satellite or Xwayland, then open some X11 app again.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -65,6 +65,8 @@ pub struct Config {
     pub overview: Overview,
     #[knuffel(child, default)]
     pub environment: Environment,
+    #[knuffel(child, default)]
+    pub xwayland_satellite: XwaylandSatellite,
     #[knuffel(children(name = "window-rule"))]
     pub window_rules: Vec<WindowRule>,
     #[knuffel(children(name = "layer-rule"))]
@@ -1362,6 +1364,23 @@ pub struct EnvironmentVariable {
     pub name: String,
     #[knuffel(argument)]
     pub value: Option<String>,
+}
+
+#[derive(knuffel::Decode, Debug, Clone, PartialEq, Eq)]
+pub struct XwaylandSatellite {
+    #[knuffel(child)]
+    pub off: bool,
+    #[knuffel(child, unwrap(argument), default = Self::default().path)]
+    pub path: String,
+}
+
+impl Default for XwaylandSatellite {
+    fn default() -> Self {
+        Self {
+            off: false,
+            path: String::from("xwayland-satellite"),
+        }
+    }
 }
 
 #[derive(knuffel::Decode, Debug, Clone, PartialEq, Eq)]
@@ -4796,6 +4815,10 @@ mod tests {
                     },
                 ],
             ),
+            xwayland_satellite: XwaylandSatellite {
+                off: false,
+                path: "xwayland-satellite",
+            },
             window_rules: [
                 WindowRule {
                     matches: [
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,11 +17,11 @@ use niri::dbus;
 use niri::ipc::client::handle_msg;
 use niri::niri::State;
 use niri::utils::spawning::{
-    spawn, store_and_increase_nofile_rlimit, CHILD_ENV, REMOVE_ENV_RUST_BACKTRACE,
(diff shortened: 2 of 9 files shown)


two-horned:
It is just my personal opinion, but I really like how you can control whether or not you have a xorg-server running or not.

After this change, I suppose you won't have this anymore? Can we still have an option that prevents automatic startup of these applications? Please let me illustrate one use case:

I am discovering wayland compositors and open my favorite applications. One application does not start. I can assume it tries to run on an xorg server (for example, chromium based browsers). I go to my system files and create a shell script that starts chromium with my desired settings, make it visible in of my $PATH directories (maybe also making it the default).

I have also discovered, some application that have a choice between xorg and wayland always prefer xorg. If some xorg-server was running I would run into the issue that half my running programs don't utilize the wayland protocol :(

So it would be nice to still allow users to disable automatic startup of any xorg-server by possibly introducing a new setting.

two-horned:
> It is just my personal opinion, but I really like how you can control whether or not you have a xorg-server running or not.

From what I see in the commit, you already have a setting that control the desired behavior :+1:

YaLTeR:
Yes, you can disable it with `xwayland-satellite { off; }`

Stzx:
Will there be support for set additional parameters for `xwayland-satellite`? (e.g. `-rootless`), or is this left to `xwayland-satellite { off; }`?

YaLTeR:
There are no additional parameters. xwayland-satellite does not support any arguments.

Stzx:
> There are no additional parameters. xwayland-satellite does not support any arguments.

Aha, now that you mention it, I went back to look.

I had only been passing one parameter all along, whereas this position in the code should have been reserved for `display`.

I accidentally used it for parameter passing, relying on a "bug" to run.......

https://github.com/Supreeeme/xwayland-satellite/blob/da2ecb5be816de35e2efe23a408a1c49fe8b11ba/src/main.rs#L62

https://github.com/Supreeeme/xwayland-satellite/blob/da2ecb5be816de35e2efe23a408a1c49fe8b11ba/src/lib.rs#L71-L73

YaLTeR:
Pretty sure that shouldn't let you pass any extra parameters.

Either way, you're not supposed to pass any custom parameters to Xwayland on any WM that integrates with it. xwl-s already passes `-rootless` on its own

IVKLD:
Extremely odd decision to include xwayland in the best wayland composer. People won't start using wayland 100% unless you force them to use wayland. They will continue to use xorg software if they don't stop boycotting xorg. I would like to make it so that niri can be built without mentioning xorg.

YaLTeR:
Well I don't see Steam finally switching to Wayland because some obscure WM doesn't run Xwayland out of the box. So no, the only outcome of niri not integrating some Xwayland out of the box is that it's harder and more annoying for new users, and detracts people from trying niri.

If you don't like it you can turn it off in the config, then it will do nothing (exactly the same as before this PR). Or just not install xwayland-satellite. If you really don't like it, it's easy enough to patch out.

IVKLD:
No complaints then, if you can cut out xwayland. I fear for niri, as it is the only good product a product one uses and is happy with. No problems have been observed.

YaLTeR:
Note also that there's no trace of Xwayland in niri itself. All this code is doing is creating X11 sockets and spawning xwayland-satellite

abdeoliveira:
> Extremely odd decision to include xwayland in the best wayland composer. People won't start using wayland 100% unless you force them to use wayland. They will continue to use xorg software if they don't stop boycotting xorg. I would like to make it so that niri can be built without mentioning xorg. 

Could not disagree more. In the current state of affairs it is impossible to live without x11 applications in a normal production environment. For many people (me included), niri without the  xwayland-sattellite support would be impracticable.

PR #1757, opened 2025-06-09
Add link to Brodie Robertson's youtube video featuring `niri`.

voidstar-null:
`niri` is getting some recognition!! This is a massive W for the project.


voidstar-null:
Seems like, he's gonna upload a dedicated video on `niri`.
image

YaLTeR:
Yeah; also not sure what the policy should be on third-party showcase videos. Like, there are a few niri videos from less popular channels, we probably don't want to add them? Where do you draw the line?

I'd say if the dedicated niri video gives a good introduction and overview, it would be worthwhile to link as a video guide. What do you think?

voidstar-null:
> Yeah; also not sure what the policy should be on third-party showcase videos. Like, there are a few niri videos from less popular channels, we probably don't want to add them? Where do you draw the line?
>
> I'd say if the dedicated niri video gives a good introduction and overview, it would be worthwhile to link as a video guide. What do you think?

voidstar-null:
> I'd say if the dedicated niri video gives a good introduction and overview, it would be worthwhile to link as a video guide. What do you think?

Yeah, that would be better. 

> Yeah; also not sure what the policy should be on third-party showcase videos. Like, there are a few niri videos from less popular channels, we probably don't want to add them? Where do you draw the line?

I think big youtubers who drive the opinions of majority of the linux community should be worth mentioning as they help a project to grow from a niche one to something generic. Scrolling layout is amazing and I can't see why anyone who likes productivity won't use that, but sadly most people are just unaware of it and that is the main reason it's still a niche thing in the world of window managers/wayland compositors.

PR #1763, opened 2025-06-10
Set logind LockedHint on lock/unlock

notpeelz:
See #1745

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -2924,6 +2924,8 @@ impl Niri {
                     let lock = confirmation.ext_session_lock().clone();
                     confirmation.lock();
                     self.lock_state = LockState::Locked(lock);
+                    #[cfg(feature = "dbus")]
+                    self.set_locked_hint(true);
                 } else {
                     // Still waiting.
                     self.lock_state = LockState::Locking(confirmation);
@@ -4372,6 +4374,8 @@ impl Niri {
                         let lock = confirmation.ext_session_lock().clone();
                         confirmation.lock();
                         self.lock_state = LockState::Locked(lock);
+                        #[cfg(feature = "dbus")]
+                        self.set_locked_hint(true);
                     } else {
                         // Still waiting for other outputs.
                         self.lock_state = LockState::Locking(confirmation);
@@ -5603,6 +5607,9 @@ impl Niri {
             confirmation.lock();
             self.lock_state = LockState::Locked(lock);
 
+            #[cfg(feature = "dbus")]
+            self.set_locked_hint(true);
+
             return;
         }
 
@@ -5617,6 +5624,9 @@ impl Niri {
             let lock = confirmation.ext_session_lock().clone();
             confirmation.lock();
             self.lock_state = LockState::Locked(lock);
+
+            #[cfg(feature = "dbus")]
+            self.set_locked_hint(true);
         } else {
             // There are outputs which we need to redraw before locking. But before we do that,
             // let's wait for the lock surfaces.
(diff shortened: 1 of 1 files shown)


src/niri.rs
```diff
@@ -5700,12 +5713,77 @@ impl Niri {
             self.event_loop.remove(deadline_token);
         }
 
+        if let LockState::Locked(_) = prev {
+            #[cfg(feature = "dbus")]
+            self.set_locked_hint(false);
```
YaLTeR: Seeing how many places this is in, I think it's better if you move it into refresh_and_flush_clients(). Just check there if is locked != current locked hint (might need to store it on Niri) then call dbus.

src/niri.rs
```diff
@@ -5700,12 +5713,77 @@ impl Niri {
...
 
+    #[cfg(feature = "dbus")]
+    fn set_locked_hint(&mut self, locked: bool) {
+        let res = thread::Builder::new()
+            .name("LockedHint updater".to_owned())
+            .spawn(move || {
+                let session_id = match std::env::var("XDG_SESSION_ID") {
+                    Ok(session_id) => session_id,
+                    Err(_) => {
+                        error!("no session id?");
```
YaLTeR: error!() is reserved for definite niri bugs: https://github.com/YaLTeR/niri/wiki/Developing-niri#logging-levels

src/niri.rs
```diff
@@ -5700,12 +5713,77 @@ impl Niri {
...
+                        return;
+                    }
+                };
+
+                let conn = match zbus::blocking::Connection::system() {
+                    Ok(conn) => conn,
+                    Err(err) => {
+                        error!("failed to get session path: {err:?}");
```
YaLTeR: I suggest extracting all this into a function returning anyhow::Result, then using .context("msg")?

notpeelz:
I made the changes you suggested. I'm still not totally sure if LockedHint is the right solution, since e.g. a nested niri instance might mess with the state. Other programs can modify it and niri doesn't have any authority over it. It just doesn't feel like a good solution.

Personally, I like the Wayland protocol approach better (see [feat-hyprland-lock-notify](https://github.com/notpeelz/niri/tree/feat-hyprland-lock-notify)) because it removes the ambiguity. You get a clear, authoritative answer from the compositor about whether it's locked.

Is it the protocol approach itself you're not a fan of, or just the idea of relying on a custom protocol?

YaLTeR:
> I'm still not totally sure if LockedHint is the right solution, since e.g. a nested niri instance might mess with the state.

Good point; there's an obvious solution: only do it when it's the session instance (running as niri --session).

notpeelz:
Added the `--session` check. Should I should add a debug config entry to forcefully enable the feature?

YaLTeR:
Don't think it's needed

notpeelz:
Alright cool. Let me know if there's anything else to be done. I'll squash my changes and rebase once ready to merge (or you can do it yourself if GitHub lets you do that?)

notpeelz:
@YaLTeR want me to squash the fixups?

YaLTeR:
It's alright, I've got it

src/niri.rs
```diff
@@ -5741,6 +5755,81 @@ impl Niri {
...
+                Some("org.freedesktop.login1.Session"),
+                "SetLockedHint",
+                &(locked),
+            )
+            .context("failed to call SetLockedHint")?;
+
+            Ok(())
+        }
+
+        let locked = self.is_locked();
```
notpeelz, line 5811: @YaLTeR Wouldn't this will set the hint while it's in the `Locking` state? Could this result in flashing the desktop on system resume?

YaLTeR: Oh, so that's why it was `Locked`. Should've added a comment about that. I'll fix myself

YaLTeR: Done

PR #1772, opened 2025-06-11
Add vim navigation to the overview

paperbenni:


YaLTeR:
No; all letters are intentionally not bound to anything to allow for type to search in the future. You can keep using Super+HJKL.

paperbenni:
Ah, I didn't know that was planned :)

PR #1783, opened 2025-06-12
default-config.kdl: add repeat=false to close-window

schuelermine:
This is an unintuitive default, I believe


YaLTeR:
It's sometimes useful but I agree that this is a safer default

YaLTeR:
Thanks

PR #1795, opened 2025-06-13
Optional screenshot notification

ferrreo:
We don't always want notifications triggered when screenshoting things, this is an easy way to disable. 

Let me know if I need to change anything or reformat anything as I didn't see a contributors guide anywhere etc.


YaLTeR:
Is this really necessary in niri? Maybe it can be filtered out on the notification daemon side?

ferrreo:
> Is this really necessary in niri? Maybe it can be filtered out on the notification daemon side?

1. It's more surprising niri does this at all, most screenshot software does not force notifications upon the user. It's especially surprising when using the overlay gui, in that workflow it's not like the user is going to accidentally take a screenshot and not know it happened. It's also fairly expected from other WMs that a notifiy-send could be added to the keybind etc if wanting a notification.
2. Not all notification daemons have a filter system - especially now we have custom ones written in quickshell, astral, ewww etc. Forcing the user to then write something custom to filter it out is a bit unnecessary, when it's like what 4 lines of code the niri side?
3. If the user does not want the notification then currently extra work is being done on both the niri side and the notification daemon side for literally zero gain. Sure it isn't a lot of resource use but it also isn't zero.
4. This has been specifically requested by multiple users of niri on my distro and I'd rather not maintain a patch forever if possible but I'd also like to keep users happy.

YaLTeR:
Some considerations here:
1. If we're doing this, I'd probably want a new `screenshot {}` section (like `clipboard {}`) where this will live. Add `path` into that section and soft-deprecate the existing `screenshot-path` while maintaining backwards compatibility: if `screenshot { path` is explicitly set then use it, otherwise fall back to `screenshot-path`. This may be awkward with an `Option<Option<`. Update the default config to use the new syntax.
2. I had some ideas for the future to do for example an animation into a draggable screenshot preview on capture. It doesn't make sense to show notification + that animation at once. As is, I can just delete the notification code and replace it with something else. If this setting is added, then we'll end up with a setting that does nothing and is unused. Thereby, adding a setting kind of locks me into the existing design which is not necessarily something I want. What do we do about this?
3. "It's more surprising niri does this at all" no, it's not surprising, when I was adding the screenshot UI in GNOME that was an explicit design choice because otherwise as a new user you don't know what happened or if something happened at all.
4. "Not all notification daemons have a filter system - especially now we have custom ones written in quickshell, astral, ewww etc." I mean I get your point but also custom notification systems are exactly the ones where it is easy to add a filter so this not a good argument.

ferrreo:
For point 2 I guess there are two potential approaches:

First, just keep this patch/branch around till you are done that then either delete it or I can alter it to fit the new code. So anyone that wants this now can just patch it in. This should limit it pigeonholing you.

Second, make the setting more generically named so it can disable any post notification UI things, so works for both now and future. This still restricts you design wise a little bit though.

Quinntyx:
For point 3, I actually would like to second that the screenshot sending a notification was unexpected for me. 

Blocking it in a notification daemon config is indeed pretty easy, but it feels a little weird to me to have it enabled by default and not disableable inside of niri itself

On point 2 I do agree that it may not be good long-term to have a config item that might become useless, and I don't really know what to do about that

Quinntyx:
Maybe one way to consider it could be to have a setting for what type of screenshot feedback you want? You could set it to notification or none, and then also the potential future screenshot animation thing. Because maybe some users just liked the simple screenshot notification, and would still want it even after the animation is added

PR #1797, opened 2025-06-14
window: allow to set per-window view offset

ein-shved:
This allows to set view_offset for window, which makes workspace to shift it view while window is focused.

## Rationale

While vim still does not fully supports external windows, I'm working on workaround, which allows me to create windows with no limit. Niri implements ideal concept of no-limit tape of windows for this workaround. So, in theory I can create os-window of no limit width and place there no limit number of vim-windows of fixed width side-by-side.

For this Niri lucks of single one functionality: the dynamic focus on special offset of os-window which is wider then screen. 

Here is video which demonstrates work of my current setup with vim, my [integration utility](https://github.com/ein-shved/niri-integration) and Niri, patched with this PR.

https://github.com/user-attachments/assets/f57ac773-f4c7-4818-93b7-0e79dda080c5

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -385,6 +385,11 @@ pub enum Action {
     MruSetScope(MruScope),
     #[knuffel(skip)]
     MruCycleScope,
+    #[knuffel(skip)]
+    ViewOffset {
+        id: Option<u64>,
+        offset: f64,
+    },
 }
 
 impl From<niri_ipc::Action> for Action {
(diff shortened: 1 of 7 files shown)


YaLTeR:
What's the usecase?

ein-shved:
> What's the usecase?

Hi! I've updated the original comment

ein-shved:
@YaLTeR Hi! Is there is any chance to apply such functionality? Shall I continue with stabilization of code?

YaLTeR:
I haven't got to this PR yet so I haven't thought about it further yet

PR #1804, opened 2025-06-15
Add config option for first-column-width

Bahnschrift:
This would address something like #859, but with more options for configuration.

This option makes it so windows opened workspaces without any existing windows (not counting floating) may have a specific width applied. It exists as both a global layout rule and also as a window rule.

A potential issue with this could be the behaviour around leaving this option unset? If it is not set at all, then it falls back to `default-column-width`, but there is no way to explicitly set this option and keep this functionality. Happy to take a look at rewriting this if that's a problem, but it seems like it would require new config syntax in order to be able to have all the options of:
- fall back to `default-column-width`
- let the window decide
- explicitly set the window's width

There is also no way in this implementation to specify falling back to the global `default-column-width` in a window rule if `first-column-width` is already set globally.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -529,6 +529,8 @@ pub struct Layout {
     pub preset_column_widths: Vec<PresetSize>,
     #[knuffel(child)]
     pub default_column_width: Option<DefaultPresetSize>,
+    #[knuffel(child)]
+    pub first_column_width: Option<DefaultPresetSize>,
     #[knuffel(child, unwrap(children), default)]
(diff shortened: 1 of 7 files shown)


YaLTeR:
I'm kinda thinking with how specific this is it might be better left to scripts. Should be fairly simple to script this using event-stream

Bahnschrift:
> I'm kinda thinking with how specific this is it might be better left to scripts. Should be fairly simple to script this using event-stream

Yeah that's fair. I wonder if setting some sort of custom scripting for window widths (and possibly other rules) would be possible that way too? Might take a look into it.

PR #1824, opened 2025-06-17
Support backlight adjustment keys in default config

dev-nicolaos:
[Related Discussion](https://github.com/YaLTeR/niri/discussions/364)

This PR adds support for setting the backlight of a display using the brightness control keys found on many (most?) laptops in the default config. The motivation here is to have one less thing that a new user is forced to go configure themselves to get there hardware working as expected when first starting. I chose [`brightnessctl`](https://github.com/Hummer12007/brightnessctl) over [`light`](https://gitlab.com/dpeukert/light) because the latter appears to be archived.

I chose +/-10% somewhat randomly, if folks would prefer a different default value I think that's fine.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -364,7 +364,7 @@ binds {
     XF86AudioMute        allow-when-locked=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SINK@" "toggle"; }
     XF86AudioMicMute     allow-when-locked=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SOURCE@" "toggle"; }
 
-    // Example brightness key mappings for brightnessctl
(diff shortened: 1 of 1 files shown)


resources/default-config.kdl
```diff
@@ -364,6 +364,10 @@ binds {
     XF86AudioMute        allow-when-locked=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SINK@" "toggle"; }
     XF86AudioMicMute     allow-when-locked=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SOURCE@" "toggle"; }
 
+    // Example brightness key mappings for brightnessctl
```
YaLTeR:
```suggestion
    // Example brightness key mappings for brightnessctl.
```

YaLTeR:
Thanks

PR #1829, opened 2025-06-17
Emit config reloaded event through IPC

HigherOrderLogic:
Add the `ConfigReloaded { failed: bool }` which is emitted though IPC's `event-stream`.

Also add the option to disable the built-in "failed to reload config" notification.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -42,6 +42,8 @@ pub struct Config {
     #[knuffel(child, default)]
     pub layout: Layout,
     #[knuffel(child, default)]
+    pub no_failed_config_reloaded_notification: bool,
+    #[knuffel(child, default)]
     pub prefer_no_csd: bool,
     #[knuffel(child, default)]
     pub cursor: Cursor,
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1341,6 +1341,11 @@ pub enum Event {
         /// The new state of the overview.
         is_open: bool,
     },
+    /// The configuration is reloaded
+    ConfigReloaded {
+        /// Whether the operation failed or not
+        failed: bool
+    }
 }
 
 impl FromStr for WorkspaceReferenceArg {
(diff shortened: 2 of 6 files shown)


YaLTeR:
Question: what should happen at compositor startup if the config failed to parse? Should this event be emitted and the panel hidden if the setting is set? In that case maybe it needs to be called `ConfigLoaded`? In which case it should also be emitted at startup on success?

YaLTeR:
Although wait, there's nobody at compositor startup to listen to the event. Though, it could be buffered and sent to clients at connection to the event stream? Then it will always arrive in the initial batch with either success or fail. Not sure if that's a good idea?

HigherOrderLogic:
> Question: what should happen at compositor startup if the config failed to parse? Should this event be emitted and the panel hidden if the setting is set?

If Niri failed to parse the config, which means that `no_failed_config_reloaded_notification` will fallback to the default `false`, wouldnt the the built-in notification display as normal?

> Although wait, there's nobody at compositor startup to listen to the event.

Yes, and if the config failed to load then there will be nothing spawned to listen to the events either.

I haven't considered the idea of buffering and what to do then since events buffering doesnt exist in Niri yet(?).

YaLTeR:
> If Niri failed to parse the config, which means that no_failed_config_reloaded_notification will fallback to the default false, wouldnt the the built-in notification display as normal?

...good point

> I haven't considered the idea of buffering and what to do then since events buffering doesnt exist in Niri yet(?).

See any other event stream event. Like how at the start you get the current keyboard layout. With this you'd get the "last config load was successful"

YaLTeR:
For the config option, how about something more like:
```
config-notification {
    show-failed false
}
```
This way it can be extended in the future if needed. I don't like how this is an option with an argument (not very consistent with others), but I can't come up with a good negative flag name.

YaLTeR:
Went with:
```kdl
config-notification {
    disable-failed
}
```

YaLTeR:
Thanks, please re-test the new version

PR #1846, opened 2025-06-19
Update niri-session for better compatibility and output

Hellx2:
- Add a --help message
- Throw an error when called with arguments or options other than --help
- Replace `#!/bin/sh` with `#!/usr/bin/env sh` to support using on systems that have `sh` in other locations in the path, such as `/usr/local/bin` or `~/.local/bin`.

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -1,4 +1,23 @@
-#!/bin/sh
+#!/usr/bin/env sh
+
+if ! [ -z $1 ]; then
+    CODE=1
+    if [[ $1 = --* ]]; then
+        if [ "$1" = "--help" ]; then
+            CODE=0
+        else
+            echo -e "\x1b[31;1mError: Unrecognized option $1"
+        fi
+    else
+        echo -e "\x1b[31;1mError: niri-session should be called with no arguments."
+    fi
+
+    echo -e "\n\x1b[0;1mUsage:\x1b[0m niri-session [--help]\n"
+    echo -e "Runs niri as a session on systemd or dinit. On systems using OpenRC, runit, and other init systems,\
+the command \x1b[1mniri --session\x1b[0m should be used instead.\n"
+    echo -e "\x1b[1mOPTIONS\x1b[0m\n--help\t\tShows this help message\n"
+    exit $CODE
+fi
 
 if [ -n "$SHELL" ] &&
    grep -q "$SHELL" /etc/shells &&


resources/niri-session
```diff
@@ -1,4 +1,23 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
YaLTeR, line 1: Not sure about this one. I cargo culted from https://github.com/alebastr/sway-systemd/blob/main/src/session.sh#L1 and similar scripts

resources/niri-session
```diff
@@ -1,4 +1,23 @@
...
+        else
+            echo -e "\x1b[31;1mError: Unrecognized option $1"
+        fi
+    else
+        echo -e "\x1b[31;1mError: niri-session should be called with no arguments."
+    fi
+
+    echo -e "\n\x1b[0;1mUsage:\x1b[0m niri-session [--help]\n"
+    echo -e "Runs niri as a session on systemd or dinit. On systems using OpenRC, runit, and other init systems,\
+the command \x1b[1mniri --session\x1b[0m should be used instead.\n"
```
YaLTeR:
```suggestion
    echo -e "Runs niri as a session on systemd or dinit. If using OpenRC, runit, or other init systems, run \x1b[1mniri --session\x1b[0m instead.\n"
```

Hellx2:
@YaLTeR Using `/usr/bin/env sh` accounts for when the `sh` executable is superseded by another location, such as if the user has created a user-specific override in `~/.local/bin`, or if they have it overridden at a system level using `/usr/local/bin`. Although `sh` must always exist in `/usr/bin`, it's typically better to avoid using any executables directly from a shebang, especially anything that has a possibility to be in any location in a standard scenario.

Hellx2:
Also I have now altered the script to remove the ANSI escape codes if the `SHELL` env variable (which should be the same as what /usr/bin/sh is linked to) is anything other than zsh, bash, or fish.

YaLTeR:
Well this got really ugly now to the point where I don't think it pulls its weight. I'd be in favor of a simple message without escape codes over these shell checks.

And I'll need confirmation that it works on dash because last time we forgot about it and someone else had to fix it

resources/niri-session
johannesrld: `#!/usr/bin/env sh` will search for the `sh` variable on your `$PATH`, so its generally more compatible then just hoping `sh` will be located in `/bin`, especially for systems that have done a `/usr` merger (although those usually have `/bin` `/lib` etc symlinked to /usr/bin and /usr/lib for compatibility anyways, so its not a *massive* deal most of the time.)

YaLTeR: Do any other session scripts do this?

The reason why I'm so picky about this line here is that if the niri-session script breaks then the entire session doesn't run at all. So not breaking is a priority over supporting some weird edge case sh installation

nekopsykose: > although those usually have /bin /lib etc symlinked to /usr/bin and /usr/lib for compatibility anyways

is there a single distro that /doesn't/ symlink these when they're usrmerged? the usr-merge everyone talks about requires you to have these links.

> so its generally more compatible then just hoping sh will be located in /bin

/bin/sh is required to exist by posix, so even nixos has it there (funnily, it's the only thing it has in /bin)

this part of the change really doesn't make any sense, and /bin/sh is perfectly fine to have

resources/niri-session
```diff
@@ -1,4 +1,34 @@
...
+    case $(basename $SHELL) in
+        fish|bash|zsh)
+            echo -e $@
+            ;;
+        *)
+            # Remove all \x1B escapes from the text.
+            echo $@ | sed 's@\\x1b\[[0-9;]*m@@gi';
+            ;;
+    esac
+}
```
nekopsykose, line 13: you could also just use `printf`? all of the shells you match for `echo -e` have a working printf (and many you don't, including dash/ash etc), and it even has a posix spec, so you can call it directly

resources/niri-session
```diff
@@ -1,4 +1,34 @@
...
+if ! [ -z $1 ]; then
+    CODE=1
+    if case $1 in --*) true;; *) false;; esac; then
+        if [ "$1" = "--help" ]; then
+            CODE=0
+        else
+            print "\x1b[31;1mError: Unrecognized option $1"
+        fi
+    else
+        print "\x1b[31;1mError: niri-session should be called with no arguments."
```
nekopsykose, line 24: this looks like adding errors for the sake of having errors and doesn't seem to accomplish something

nekopsykose:
> Replace #!/bin/sh with #!/usr/bin/env sh to support using on systems that have sh in other locations in the path, such as /usr/local/bin or ~/.local/bin.

if you have a system that doesn't have a `#!/bin/sh` then `niri-session` is going to be far from the only thing that will be broken. that is a universally-held assumption by a huge amount of scripts in existence, and they're not wrong for making it.

that change is ultimately not very useful and doesn't actually improve compatibility, it only papers over a broken (hypothetical) system you've created where `/bin/sh` doesn't exist for some reason (which is none of the systems that niri runs on).

---

a slightly more controversial opinion from me would be that `#!/bin/sh` is more likely to exist than `#!/usr/bin/env`. specifically only for `sh`, of course; for any other binaries using `env` is the safer pick.

PR #1861, opened 2025-06-20
Add an option to disable moving the view when the user clicks a window

Olekoop:
This PR adds an option to disable moving the view when the user clicks a window. 
The option name is a bit weird but I couldn't find a better one. I wouldn't have any problem if the name would be changed to something better.

### DIFF
--- a/wiki/Configuration:-Input.md
+++ b/wiki/Configuration:-Input.md
@@ -99,6 +99,7 @@ input {
     // disable-power-key-handling
     // warp-mouse-to-focus
     // focus-follows-mouse max-scroll-amount="0%"
+    // disable-moving-view-when-focus-clicked
     // workspace-auto-back-and-forth
 
     // mod-key "Super"
@@ -315,6 +316,12 @@ input {
     focus-follows-mouse max-scroll-amount="0%"
 }
 ```
+
+#### `disable-moving-view-when-focus-clicked`
+
+By default, niri will move the view in order to fit the focused window under any circumstances.
+This option allows to disable moving the view only when the user clicks a window. The window will be focused but the view will not change.
+Any other way that focuses a window (e.g. bindings, moving or resizing) is not affected by this option.
 
 #### `workspace-auto-back-and-forth`
 
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -104,6 +104,8 @@ pub struct Input {
     #[knuffel(child)]
     pub focus_follows_mouse: Option<FocusFollowsMouse>,
     #[knuffel(child)]
+    pub disable_moving_view_when_focus_clicked: bool,
+    #[knuffel(child)]
     pub workspace_auto_back_and_forth: bool,
     #[knuffel(child, unwrap(argument, str))]
     pub mod_key: Option<ModKey>,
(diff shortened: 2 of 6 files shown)


wiki/Configuration:-Input.md
```diff
@@ -316,6 +317,12 @@ input {
 }
 ```
 
+#### `disable-moving-view-when-focus-clicked`
+
+By default, niri will move the view in order to fit the focused window under any circumstances.
+This option allows to disable moving the view only when the user clicks a window. The window is will be focused but the view will not change.
```
phisch: Little typo here: `The window is will be focused`

phisch:
Flipping it around into `pan_viewport_on_focus_click` and setting that to `true` by default could be better. But that's up to debate, just brain-dumping here.

Also to add my use-case for this:

I often have half a window on the side of my view-port, and I see some text I want to select, or something similar, and starting a click then moves the window, messing up my started drag gesture, and trying to select text with an offset is impossible to get used to. There are probably tons of use-cases I could come up with for that, that I have been working around so far.

wiki/Configuration:-Input.md
Olekoop: thanks for noticing

Olekoop:
@phisch I'm not sure if it's possible to do this option this way. From the looks of it, niri doesn't have a way to enable an option without having it declared in a config. You can only enable things in the config (in order to turn off a device, you enable the off option). Unless we want to break configs, `pan_viewport_on_focus_click` isn't going to work.

YaLTeR:
My design question here, is this the right thing to add? What is the list of the actual problematic cases intended to be addressed? Maybe something else is a better solution, like a view lock key? Should we make a discussion about this first?

phisch:
I think both would solve my use-cases. But there is no right or wrong answer here, they are both just slightly different user experiences.

Olekoop:
@YaLTeR I have a similar use-case to @phisch. It's basically where e.g. you have a terminal half-size of your monitor and a web browser full-size next to the terminal. It's impossible to just copy text from browser without focusing (and moving the view). With this new option, selecting text and middle-click pasting it is far more easier and convenient.

I wouldn't mind compromising for the view lock key, if it can be used for different type of a use-case than mine and @phisch.

phisch:
I was initially suggesting a scroll lock like half a year ago, but I think in most use-cases your implementation might be a slightly better UX since it doesn't require to lock/unlock the viewport before switching focus with intention (like with a shortcut).

The question is, can the behavior of this PR be scripted with a scroll lock?

Lets say you use `mod+left` and `mod+right` to switch your focus and viewport to the left/right. Now if you come up with binds like this:
```kdl
Mod+Right { 
    unlock-viewport-scroll;
    focus-column-right-or-first;
    lock-viewport-scroll;
}
Mod+Left {
    unlock-viewport-scroll;
    focus-column-left-or-last;
    lock-viewport-scroll;
}
```

This would essentially mirror the behavior of what this PR does, right? The viewport would be locked, but you can still move the focus around, which would pan the viewport.

Olekoop:
After thinking about the view look, it seems to provide more functionality than my solution.
I have opened a PR about the view look #1877 and I'm closing the old PR

PR #1877, opened 2025-06-22
Implement a view lock

Olekoop:
This PR implements a view lock. It allows the user to lock a view in place and nothing can move it unless the user tries to center a window, a column or visible columns. In that case the view moves but the lock is still active. It also disables automatic focus for a new window.

There are at least two scenarios that greatly benefit from this feature.
One of those is when debugging a GUI application. The user can have a full-size IDE window open and debug the program without the worry about niri moving the view to the application's window.
A different one is when trying to copy something from one window to another. The view lock allows to copy from every window that can be seen no matter how much visible it is. It is possible to create a script that auto-locks the view every time that the user tries to move the view using key bindings.

I have tested this feature with mouse and keyboard bindings but I didn't do it with touchpad and touch movements (my PC doesn't have any of those things). I would be grateful if someone tests this with touchpad or touch.


Olekoop:
Resolved the merge conflicts

bczhc:
I helped do the rebase.

[0001-Implement-a-view-lock.patch](https://github.com/user-attachments/files/28983163/0001-Implement-a-view-lock.patch)
[0002-Change-border-color-when-view-lock-is-locked.patch](https://github.com/user-attachments/files/28983164/0002-Change-border-color-when-view-lock-is-locked.patch)

Olekoop:
Rebased the PR based on @bczhc work

kpko:
Hi, really looking forward to this - what's the current state on the pr? Are there outstanding issues that need help? :)

Olekoop:
Now the only thing that stops it from being merged is the conflicts, you can try to resolve them if you want to

PR #1881, opened 2025-06-22
feat: add hint to disable "Important Hotkeys" in the default config file

sharunkumar:
Motivation: Avoids the need to google how to turn it off, for new users (I faced this as well)

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -266,6 +266,11 @@ layout {
 // This line starts waybar, a commonly used bar for Wayland compositors.
 spawn-at-startup "waybar"
 
+// Uncomment the following to disable "Important Hotkeys" pop-up at startup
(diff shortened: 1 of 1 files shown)


resources/default-config.kdl
```diff
@@ -266,6 +266,11 @@ layout {
 // This line starts waybar, a commonly used bar for Wayland compositors.
 spawn-at-startup "waybar"
 
+// Uncomment the following to disable "Important Hotkeys" pop-up at startup
+/-hotkey-overlay {
+    skip-at-startup
+}
+
```
YaLTeR:
```suggestion
hotkey-overlay {
    // Uncomment this line to disable the "Important Hotkeys" pop-up at startup.
    // skip-at-startup
}

```

YaLTeR:
Thanks

PR #1891, opened 2025-06-23
Change wiki to a static site

Hasnep:
This PR moves the wiki from a GitHub wiki to a static site built with mdbook and hosted on GitHub Pages. An example of the built wiki is available [here](https://hasnep.github.io/niri/wiki).

I had to make some decisions when setting up the site, so I just went with the simplest option I could think of to get something working. I'd like feedback on these decisions (and anything else of course)

- Small nitpicky question, I assume the wiki was called a wiki because it was hosted on the GitHub wiki feature, but it's not actually a wiki because you need to modify the repo to edit it. Now that it's not on the GitHub wiki, do we want to rename it to `docs` or something else?
- Currently, the repo's GitHub Pages site contains the rustdoc documentation. I've moved the rustdoc documentation to `/rustdoc` and put the wiki in `/wiki`. This will break any links to the current rustdoc pages, but it doesn't seem like there are many of those on the internet. Some alternatives I can think of would be:
  - Keep the rustdoc documentation in the root of the website and put the wiki in `/wiki` which will keep any existing links to the rustdoc site
  - If we assume we're going to break the links to the rustdoc site, we could put the wiki in the root of the site and put the rustdoc documentation in `/rustdoc`. This puts the more useful thing at the root and the more nice docs in a subdir.
  - Just remove the rustdoc docs if nobody uses them.
- I used mdbook to build the wiki which is written in Rust, which means it can be added to the `Cargo.toml` to pin the version of mdbook so that the site doesn't break when mdbook updates. I was hoping this would also avoid requiring developers to manually install mdbook to build the wiki locally, but Cargo doesn't automatically install the `mdbook` binary, only the library. Instead, I used a Cargo workspace in the `wiki` directory to install `mdbook` and use a script `wiki.rs` to import the mdbook library to build the wiki (`cd wiki && cargo run --bin wiki`). This means you just need Cargo installed to build the wiki.


YaLTeR:
How does it do on the requirements list from https://github.com/YaLTeR/niri/discussions/1830? Could you post there for discussion please?

Hasnep:
Oh, cool I didn't know there was an issue for it, I'll add to the discussion there :)

Hasnep:
Closing as this has been superseded by #2147 :)

PR #1907, opened 2025-06-25
fix hot reloading `/etc/niri/config.kdl`

sodiboo:
I've refactored how the config loading works to use a new enum for the two modes of operation, and modified the watcher logic to handle the "regular" case correctly.

now, changes to `/etc/niri/config.kdl` will actually hot reload.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -2372,14 +2372,19 @@ pub enum PreviewRender {
 
 #[derive(Debug, Clone)]
 pub enum ConfigPath {
+    /// Explicitly set config path.
+    ///
     /// Load the config only from this path, never create it.
     Explicit(PathBuf),
-    /// Normal operation: Prioritize the user path, fallback to the system path,
-    /// else maybe create the user path if we're starting (but not reloading).
+
+    /// Default config path.
+    ///
+    /// Prioritize the user path, fallback to the system path, fallback to creating the user path
+    /// at compositor startup.
     Regular {
-        /// User config path, usually `$XDG_CONFIG_HOME/niri/config.kdl`
+        /// User config path, usually `$XDG_CONFIG_HOME/niri/config.kdl`.
         user_path: PathBuf,
-        /// System config path, usually `/etc/niri/config.kdl`
+        /// System config path, usually `/etc/niri/config.kdl`.
         system_path: PathBuf,
     },
 }
@@ -2391,9 +2396,7 @@ impl ConfigPath {
 
         self.load_inner(|user_path, system_path| {
             Err(miette::miette!(
-                "No config file found. Create one at {user_path} or {system_path}",
-                user_path = user_path.display(),
-                system_path = system_path.display(),
+                "no config file found; create one at {user_path:?} or {system_path:?}",
             ))
         })
         .context("error loading config")
@@ -2406,7 +2409,7 @@ impl ConfigPath {
     /// If the config was created, but for some reason could not be read afterwards,
     /// this may return `(Some(_), Err(_))`.
     pub fn load_or_create(&self) -> (Option<&Path>, miette::Result<Config>) {
-        let _span = tracy_client::span!("ConfigPath::load_or_create_with");
+        let _span = tracy_client::span!("ConfigPath::load_or_create");
 
         let mut created_at = None;
 
@@ -2414,54 +2417,18 @@ impl ConfigPath {
             .load_inner(|user_path, _| {
                 Self::create(user_path, &mut created_at)
                     .map(|()| user_path)
-                    .map_err(|err| err.context(format!("error creating config at {user_path:?}")))
+                    .with_context(|| format!("error creating config at {user_path:?}"))
             })
             .context("error loading config");
 
         (created_at, result)
     }
 
-    fn create<'a>(path: &'a Path, created_at: &mut Option<&'a Path>) -> miette::Result<()> {
-        if let Some(default_parent) = path.parent() {
-            fs::create_dir_all(default_parent)
-                .into_diagnostic()
-                .map_err(|err| {
-                    err.context(format!(
-                        "error creating config directory {default_parent:?}"
-                    ))
-                })?;
-        }
-
-        // Create the config and fill it with the default config if it doesn't exist.
-        let mut new_file = match File::options()
-            .read(true)
-            .write(true)
-            .create_new(true)
-            .open(path)
-        {
-            Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => return Ok(()),
-            res => res,
-        }
-        .into_diagnostic()
-        .map_err(|err| err.context(format!("error opening config file at {path:?}")))?;
-
-        *created_at = Some(path);
-
-        let default = include_bytes!("../../resources/default-config.kdl");
-
-        new_file
-            .write_all(default)
-            .into_diagnostic()
-            .map_err(|err| err.context(format!("error writing default config to {path:?}")))?;
-
-        Ok(())
-    }
-
     fn load_inner<'a>(
         &'a self,
         maybe_create: impl FnOnce(&'a Path, &'a Path) -> miette::Result<&'a Path>,
     ) -> miette::Result<Config> {
-        Config::load_internal(match self {
+        let path = match self {
             ConfigPath::Explicit(path) => path.as_path(),
             ConfigPath::Regular {
                 user_path,
@@ -2475,17 +2442,45 @@ impl ConfigPath {
                     maybe_create(user_path.as_path(), system_path.as_path())?
                 }
             }
-        })
+        };
+        Config::load(path)
+    }
+
+    fn create<'a>(path: &'a Path, created_at: &mut Option<&'a Path>) -> miette::Result<()> {
+        if let Some(default_parent) = path.parent() {
+            fs::create_dir_all(default_parent)
+                .into_diagnostic()
+                .with_context(|| format!("error creating config directory {default_parent:?}"))?;
+        }
+
+        // Create the config and fill it with the default config if it doesn't exist.
+        let mut new_file = match File::options()
+            .read(true)
+            .write(true)
+            .create_new(true)
+            .open(path)
+        {
+            Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => return Ok(()),
+            res => res,
+        }
+        .into_diagnostic()
+        .with_context(|| format!("error opening config file at {path:?}"))?;
+
+        *created_at = Some(path);
+
+        let default = include_bytes!("../../resources/default-config.kdl");
+
+        new_file
+            .write_all(default)
+            .into_diagnostic()
+            .with_context(|| format!("error writing default config to {path:?}"))?;
+
+        Ok(())
     }
 }
 
 impl Config {
     pub fn load(path: &Path) -> miette::Result<Self> {
-        let _span = tracy_client::span!("Config::load");
-        Self::load_internal(path).context("error loading config")
-    }
-
-    fn load_internal(path: &Path) -> miette::Result<Self> {
         let contents = fs::read_to_string(path)
             .into_diagnostic()
             .with_context(|| format!("error reading {path:?}"))?;
--- a/src/utils/watcher.rs
+++ b/src/utils/watcher.rs
@@ -25,7 +25,8 @@ impl Watcher {
         process: impl FnMut(&ConfigPath) -> T + Send + 'static,
         changed: SyncSender<T>,
     ) -> Self {
-        Self::with_start_notification(path, process, changed, None)
+        let interval = Duration::from_millis(500);
+        Self::with_start_notification(path, process, changed, None, interval)
     }
 
     pub fn with_start_notification<T: Send + 'static>(
@@ -33,6 +34,7 @@ impl Watcher {
         mut process: impl FnMut(&ConfigPath) -> T + Send + 'static,
         changed: SyncSender<T>,
         started: Option<mpsc::SyncSender<()>>,
+        polling_interval: Duration,
     ) -> Self {
         let should_stop = Arc::new(AtomicBool::new(false));
 
@@ -78,7 +80,7 @@ impl Watcher {
                     }
 
                     loop {
-                        thread::sleep(Duration::from_millis(500));
+                        thread::sleep(polling_interval);
 
                         if should_stop.load(Ordering::SeqCst) {
                             break;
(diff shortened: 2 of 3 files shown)


src/utils/watcher.rs
```diff
@@ -130,8 +149,12 @@ mod tests {
 
         let (tx, rx) = sync_channel(1);
         let (started_tx, started_rx) = mpsc::sync_channel(1);
-        let _watcher =
-            Watcher::with_start_notification(config_path.clone(), |_| (), tx, Some(started_tx));
+        let _watcher = Watcher::with_start_notification(
```
YaLTeR: Could you add a test for `ConfigPath::Regular` for the case of watching system_path, then creating user_path and watching that? (And then perhaps removing user_path and back to watching system_path)

src/utils/watcher.rs
```diff
@@ -38,20 +39,41 @@ impl Watcher {
...
+                    // and /nix/store keeps no mtime (= 1970-01-01)
                     // so, symlink targets change frequently when mtime doesn't.
-                    let mut last_props = path
-                        .canonicalize()
-                        .and_then(|canon| Ok((canon.metadata()?.modified()?, canon)))
-                        .ok();
+                    //
+                    // therefore, we must also store the canonical path, along with its mtime
+
+                    fn see_path(path: &Path) -> std::io::Result<(std::time::SystemTime, PathBuf)> {
```
YaLTeR: Can impot `io`, `SystemTime`

src/utils/watcher.rs
```diff
@@ -64,14 +86,11 @@ impl Watcher {
...
                         }
 
-                        if let Ok(new_props) = path
-                            .canonicalize()
-                            .and_then(|canon| Ok((canon.metadata()?.modified()?, canon)))
-                        {
+                        if let Ok(new_props) = see(&config_path) {
                             if last_props.as_ref() != Some(&new_props) {
-                                trace!("file changed: {}", path.to_string_lossy());
+                                trace!("config file changed.");
```
YaLTeR: no good way to print which one here?

niri-config/src/lib.rs
```diff
@@ -2370,6 +2370,65 @@ pub enum PreviewRender {
     ScreenCapture,
 }
 
+#[derive(Debug, Clone)]
+pub enum ConfigPath {
+    /// Load the config only from this path, never create it.
+    Explicit(PathBuf),
+    /// Normal operation: Prioritize the user path, fallback to the system path,
+    /// else maybe create the user path if we're starting (but not reloading).
```
YaLTeR: First line of doc comment should be a brief title. So please adjust, or change to `//` (even tho we currently don't generate rustdocs for this)

niri-config/src/lib.rs
```diff
@@ -2370,6 +2370,65 @@ pub enum PreviewRender {
...
+    Regular {
+        user_path: PathBuf,
```
YaLTeR, line 2386:
```suggestion
        /// Config file in `~/.config/niri/`
        user_path: PathBuf,
```

niri-config/src/lib.rs
```diff
@@ -2370,6 +2370,65 @@ pub enum PreviewRender {
...
+        system_path: PathBuf,
```
YaLTeR, line 2388:
```suggestion
        /// Config file in `/etc/niri/`
        system_path: PathBuf,
```

niri-config/src/lib.rs
```diff
@@ -2370,6 +2370,65 @@ pub enum PreviewRender {
...
+
+    // Those lifetimes are really not flexible at all, but it's all we need for niri.
+    pub fn load_or_create_with<'a>(
+        &'a self,
+        maybe_create: impl FnOnce(&'a Path, &'a Path) -> miette::Result<&'a Path>,
+    ) -> miette::Result<Config> {
+        let _span = tracy_client::span!("ConfigPath::load_or_create_with");
+
+        self.load_inner(maybe_create)
+            .context("error loading config")
```
YaLTeR: Maybe this `.context()` should be inside `load_inner()`?

YaLTeR: or actually I think in the original we would only have this context in the watcher and not in the initial loading? I think the intention was that if we print "error creating config" then we shouldn't also print "error loading config". What would be the right place to put it?

niri-config/src/lib.rs
```diff
@@ -2370,6 +2370,65 @@ pub enum PreviewRender {
     ScreenCapture,
 }
 
+#[derive(Debug, Clone)]
+pub enum ConfigPath {
```
YaLTeR, line 2374: Hm, perhaps this whole enum should live in `main.rs`? Seeing how it mainly deals with creating the right files and such. Then maybe some creation logic can also be inlined into its functions, simplifying them.

Especially using `maybe_create` to output an error feels kinda weird idk

YaLTeR (changes requested):

niri-config/src/lib.rs
sodiboo: The main motivation i have for putting this in `niri-config` is because it is potentially useful for other consumers of `niri-config`.

Although `niri-config` is *not* a public crate, and we don't encourage external use of it, others writing Rust-based tooling that is very tightly coupled to niri for their own systems might want to load the niri config; in which case, the precise logic of when which specific config path is loaded or when it "should" be created, can be quite useful indeed.

I agree, the pattern of `maybe_create` is kinda weird. I did it mainly to write the "minimal changes necessary"; the logic to create the file needed no modifications, so why move it. But I also think the logic should be inlined into `ConfigPath`, but i'd prefer to put it entirely in `niri-config`, seeing as this is very much specific to how niri's config works. I don't see why it would be moved to `main.rs`.

YaLTeR: > Although `niri-config` is _not_ a public crate, and we don't encourage external use of it, others writing Rust-based tooling that is very tightly coupled to niri for their own systems might want to load the niri config; in which case, the precise logic of when which specific config path is loaded or when it "should" be created, can be quite useful indeed.

Well, this is kind of a theoretical concern for now; the practical concern though is that this would require adding various xdg directories deps to niri-config which it doesn't have right now, so it would be cleaner to leave all file path manipulation in main.rs. If this becomes a concern we can move things to niri-config as needed

niri-config/src/lib.rs
sodiboo: Yea, idk, i mostly just put it there to keep symmetry with the way Config::load` is implemented.

> I think the intention was that if we print "error creating config" then we shouldn't also print "error loading config"

In the current implementation, this already isn't the behaviour.

```sh
$ XDG_CONFIG_HOME=/nonexistent-nonwritable/ niri
INFO niri: starting version unstable 2025-06-24 (commit 37458d94b288945f6cfbd3c5c233f634d59f246c)
WARN niri: error creating config directories "/nonexistent-nonwritable/niri": Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
WARN niri:   × error loading config
  ├─▶ error reading "/nonexistent-nonwritable/niri/config.kdl"
  ╰─▶ No such file or directory (os error 2)
```

With my changes, this error message is *identical*. I think the duplicate error is fine, actually, as it hints to the way niri actually handles the config: It failed to write the config, and it failed to load the config that it wants to have already written.

However, i actually think that's a mistake. The implementation for `niri validate` looks way nicer.

Current:
```sh
$ XDG_CONFIG_HOME=/nonexistent-nonwritable/ niri validate
Error:   × error loading config
  ├─▶ error reading "/nonexistent-nonwritable/niri/config.kdl"
  ╰─▶ No such file or directory (os error 2)
```

Mine:
```sh
$ XDG_CONFIG_HOME=/nonexistent-nonwritable/ niri validate
Error:   × error loading config
  ╰─▶ No config file found. Create one at /nonexistent-nonwritable/niri/config.kdl or /etc/niri/config.kdl
  ```
  
The "error creating config" when starting as a compositor should be part of the "error loading config" message; not printed separately.

File: niri-config/src/lib.rs (hunk 8 above)
Comment (sodiboo) [reply niri-config/src/lib.rs:2374]: I don't think it requires adding new dependencies to `niri-config`. I only want it to include the logic for when to use/crate which path. The logic for *what paths should be considered* (e.g. loading xdg dirs) is still in `main.rs`, because this depends on the exact CLI interface of the code that is consuming `niri-config`.

File: niri-config/src/lib.rs (hunk 7 above)
Comment (sodiboo) [reply niri-config/src/lib.rs]: I refactored a bunch of the creation logic. The error for "failed to create config" now looks like this:

```sh
$ XDG_CONFIG_HOME=/nonexistent-nonwritable/ niri
INFO niri: starting version 25.05.1 (v25.05.1-47-g0f7e7791-modified)
ERROR niri:   × error loading config
  ├─▶ error creating config at "/nonexistent-nonwritable/niri/config.kdl"
  ├─▶ error creating config directory "/nonexistent-nonwritable/niri"
  ╰─▶ Permission denied (os error 13)
  ```
  
also notably, `WARN` -> `ERROR` because this is in fact an error.

File: niri-config/src/lib.rs (hunk 8 above)
Comment (sodiboo) [reply niri-config/src/lib.rs:2374]: I moved the creation logic into `niri-config`. The interface on `ConfigPath` is much cleaner now: just `ConfigPath::load(&self)` and `ConfigPath::load_or_create(&self)`.

This also results in nicer errors.

File: niri-config/src/lib.rs (hunk 7 above)
Comment (sodiboo) [reply niri-config/src/lib.rs]: > Maybe this `.context()` should be inside `load_inner()`?

It should not be. If it's placed at the end of that function (`fn load_inner(...) { Config::load_internal(...).context(...) }`), then we lose the `error loading config` context when there was an error in creating the config file. I think that context *should* stay as the top level, *even* if we never did actually try to "load" from any path.

File: src/utils/watcher.rs (hunk 3 above)
Comment (sodiboo) [reply src/utils/watcher.rs]: not really. or at least, it would be messy.

could debug print the `new_props` or the canonical path of it?

loading the config will already print the path it loads from; so give or take for a race condition between when the file changed and the config is loaded, the path will be printed anyways. (and that race condition is fine; because we would have reloaded again after that happened even if these operations were atomic)

```sh
2025-07-14T10:58:11.745653Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:11.906400Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/user-config.kdl"
2025-07-14T10:58:11.906894Z TRACE niri::niri: redrawing output
2025-07-14T10:58:12.406551Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:12.564523Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/user-config.kdl"
2025-07-14T10:58:12.565056Z TRACE niri::niri: redrawing output
2025-07-14T10:58:13.064685Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:13.226758Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/user-config.kdl"
2025-07-14T10:58:13.227155Z TRACE niri::niri: redrawing output
2025-07-14T10:58:13.726891Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:13.881937Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/user-config.kdl"
2025-07-14T10:58:13.882304Z TRACE niri::niri: redrawing output
2025-07-14T10:58:24.883929Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:25.048014Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/user-config.kdl"
2025-07-14T10:58:25.048523Z TRACE niri::niri: redrawing output
2025-07-14T10:58:28.048658Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:28.211372Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/system-config.kdl"
2025-07-14T10:58:28.211838Z TRACE niri::niri: redrawing output
2025-07-14T10:58:31.712065Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:31.878812Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/system-config.kdl"
2025-07-14T10:58:31.879253Z TRACE niri::niri: redrawing output
2025-07-14T10:58:32.378960Z TRACE niri::utils::watcher: config file changed.
2025-07-14T10:58:32.533339Z DEBUG niri_config: loaded config from "/home/sodiboo/niri/system-config.kdl"
2025-07-14T10:58:32.533738Z TRACE niri::niri: redrawing output
```

(here i've patched the config path to always point into my working directory, so i don't need to mess with system config)

YaLTeR: alright, then just remove the dot since we don't have it elsewhere

src/utils/watcher.rs
sodiboo: added some tests for `ConfigPath::Regular`. the existing test wrapper was hardcoded on a single path, so i basically rewrote it. the existing tests should behave basically the same, except they're not statically checking for creating a final change (there's no good way to do so generically, because with `Regular` there are two different paths to test there, so i just removed that part)

[i actually went way further with a fully monomorphic/generic typestate test builder thing](https://github.com/sodiboo/niri/commit/54ec612f5080f5029cd166e74e81de2e96e8106f) but uh that caused rustc to consume several gigabytes of memory and lock up my session, so i undid all of those changes.

niri-config/src/lib.rs
YaLTeR: > also notably, `WARN` -> `ERROR` because this is in fact an error.

Our `error!()` is for programming errors and after considering this case again, I don't think it warrants an exception, so I changed it back

YaLTeR:
Thanks. Please test again on main

YaLTeR:
I changed some time values before merging to make the tests not very slow. The CI just hit this spurious failure:
```
failures:

---- utils::watcher::tests::change_file_in_linked_dir stdout ----
$ ln -s niri2 niri

thread 'utils::watcher::tests::change_file_in_linked_dir' panicked at src/utils/watcher.rs:269:13:
assertion `left == right` failed: watcher should not have noticed any changes
  left: Some("/tmp/xshell-tmp-dir-1/niri/config.kdl")
 right: None
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/panicking.rs:75:14
   2: core::panicking::assert_failed_inner
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/panicking.rs:427:23
   3: core::panicking::assert_failed
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/panicking.rs:387:5
   4: assert_unchanged
             at ./src/utils/watcher.rs:269:13
   5: run<niri::utils::watcher::tests::change_file_in_linked_dir::{closure_env#1}>
             at ./src/utils/watcher.rs:249:18
   6: change_file_in_linked_dir
             at ./src/utils/watcher.rs:380:14
   7: {closure#0}
             at ./src/utils/watcher.rs:373:39
   8: call_once<niri::utils::watcher::tests::change_file_in_linked_dir::{closure_env#0}, ()>
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/ops/function.rs:250:5
```
I'm not sure tbh which time value exactly could trigger it this way. Could you take a look please?

PR #1920, opened 2025-06-27
Allow resolving key codes with a fixed layout.

sbroglie:
This is a draft for resolving a key press with respect to a fixed layout. See  https://github.com/YaLTeR/niri/discussions/1919 for the issue background.

The implementation is straight forward: the `raw_latin_sym_or_raw_current_sym()` call is replaced with a check if a fixed layout should be used.
In this case they layouts are not traversed like in [raw_latin_sym_or_raw_current_sym](https://smithay.github.io/smithay/smithay/input/keyboard/struct.KeysymHandle.html#method.raw_latin_sym_or_raw_current_sym) but the layout at the configured index is used directly, which becomes a three-liner. If the user does not configure such an index, then the previous resolution behaviour is used.

I briefly considered, whether it would be possible to specify a complete separate xkb-keyboard configuration for resolving keybindings of Niri (e.g. `us`) and another (e.g. `fr`) for everything else.  In particular keyboad layout switching would not be done as both are configs with a *single* layout.
However, this approach seems to be complicated as there is a global xkb keyboard state.  Hence the user has to configure multiple layouts in a single config (e.g. `us,fr`) and an index.

### Things to consider:

 * Sway allows to bind keycodes (roughly the "physical key" without symbolic meaning of a layout, see `bindcode` command in *sway(5)*).  This might be an alternative method to fixing the layout, which I have not investigated.
 * The config name `resolve_keycode_with_layout` is up to debate
 * The config option type is currently `Option<LayoutIndex>`.  There might be a more suitable type, e.g. `Layout`. Note that `LayoutIndex` is defined as `u32` (https://docs.rs/xkbcommon/latest/xkbcommon/xkb/type.LayoutIndex.html).
 * We currently do not check if the user-configured layout index is between 0 and n-1.  However, the function [key_get_syms_by_level](https://docs.rs/xkbcommon/latest/xkbcommon/xkb/struct.Keymap.html#method.key_get_syms_by_level) used under the hood, does the clipping for us.  In my tests it was a reduction module n.
   We could add an explicit (Niri side) range check for that, if required.
 * Instead of having the "fallback to latin keybindings" (which is a misnomer as it falls back to ASCII-bindings!) we could always resolve with respect to the current layout.  If ASCII fallback is wanted, setting the resolving layout index should provide that behaviour.
   
### What is missing:

If this seems like a sensible solution to the problem, I would
 * provide documentation about the new config knob
 * add an action to change the resolving layout index to Niri's IPC

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -19,7 +19,7 @@ use niri_ipc::{
 };
 use smithay::backend::renderer::Color32F;
 use smithay::input::keyboard::keysyms::KEY_NoSymbol;
-use smithay::input::keyboard::xkb::{keysym_from_name, KEYSYM_CASE_INSENSITIVE};
+use smithay::input::keyboard::xkb::{LayoutIndex, keysym_from_name, KEYSYM_CASE_INSENSITIVE};
 use smithay::input::keyboard::{Keysym, XkbConfig};
 use smithay::reexports::input;
 
@@ -152,6 +152,8 @@ pub struct Xkb {
     pub options: Option<String>,
     #[knuffel(child, unwrap(argument))]
     pub file: Option<String>,
+    #[knuffel(child, unwrap(argument))]
+    pub resolve_keycode_with_layout: Option<LayoutIndex>
 }
 
 impl Xkb {
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -17,7 +17,7 @@ use smithay::backend::input::{
     TabletToolTipState, TouchEvent,
 };
 use smithay::backend::libinput::LibinputInputBackend;
-use smithay::input::keyboard::{keysyms, FilterResult, Keysym, Layout, ModifiersState};
+use smithay::input::keyboard::{keysyms, FilterResult, Keysym, KeysymHandle, Layout, ModifiersState};
 use smithay::input::pointer::{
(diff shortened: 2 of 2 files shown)


YaLTeR:
Hm, I can see how this can be useful. But not sure off the bat if this is the best approach. It's probably fine considering it's quite small in code changes.

A textual string would be better than index, but I can immediately see a problem with using a textual string: you can have "us,us" with ",dvorak" variants, so you can't just match by the layout string.

Since this is a global option, wouldn't it be better for it to fall back to the current logic if the key isn't found in the specified layout? Otherwise it's too easy to break and get left with no bindings at all.

FYI the current logic works for cases like: "us,ru" or "us(dvorak),ru" where ru has no latin letters, and all binds must keep using the layout with latin letters (whichever one that is). That is the expected behavior across compositors and apps for cyrillic languages.

@kchibisov what do you think about this?

kchibisov:
I think you just want physical key bindings honestly. Like first of all such configurations are not that used on linux (I mean wanting qwerty bindings on dvorak), thus you'd have e.g. all the applications resolving to dvorak bindings and the problem is solved only for niri in particular.

I'd suggest to add a physical key bindings, since your issue doesn't solve a problem in general and tunes niri for _your use case_ and if speaking of that, I'd suggest to add a functionality that is more useful, like binding by scancode, since users may have setups with rather weird keys, etc, etc that don't really resolve great within layouts, so the functionality will be useful for general user.

PR #1922, opened 2025-06-28
Added IPC command to change layer opacity dynamically

Tordarus:
layer opacity can be changed using the following syntax:
`niri msg layer <namespace> opacity <opacity>`

The dynamic configuration gets overriden with a live-reload of the configuration file.

---

I don't know anything about the niri codebase or Rust in general, but I was desperate enough to implement this feature anyway. It might be of low quality or buggy because of my non-existent experience in Rust.

It will also be incomplete because I couldn't find a way to return `LayerConfigChanged::LayerNotFound` to the client without the compiler yelling at me.

But maybe this works as inspiration or a starting point for a new feature with far more options than just opacity.

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -97,6 +97,17 @@ pub enum Request {
         /// Configuration to apply.
         action: OutputAction,
     },
+    /// Change layer configuration temporarily.
+    ///
+    /// The configuration is changed temporarily and not saved into the config file. If the layer
+    /// configuration subsequently changes in the config file, these temporary changes will be
+    /// forgotten.
+    Layer {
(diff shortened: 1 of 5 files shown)


YaLTeR:
What's the use case?

Tordarus:
My use case might be a little bit specific.

I'm using mpvpaper to show music videos as my background wallpaper. But not all my tracks are videos. Some are just audio tracks. I want to be able to hide the mpvpaper layer when there is no video or it is currently paused. I could agree that this could be a feature in mpvpaper instead.

Another use case would be to hide the wallpaper when the overview opens which already has its own wallpaper different from the desktop wallpaper.

Issue #605, opened 2024-08-17
Modifier-only keybinds exist


miku4k:
You can set a modifier-only shortcut by using the xkb name for the modifier as the second key (e.g. Super+Super_L). Hitting just the modifier, or any keybind that includes the modifier, triggers the mod-only bind. 
I've only tested with Super+Super_L and the spawn action, but I believe other combinations behave similarly if not the same.

Checking that the non-mod key of a keybind isnt secretly a mod should be enough to fix this.

### System Information

* niri version: niri 0.1.7 (unknown commit)
(most recent version currently available in NixOS unstable)

* Distro: NixOS (unstable)

* GPU: AMD integrated

* CPU: AMD Ryzen 5 5500u

YaLTeR:
Hm, I guess until we have release binds, this is more or less the best we can do. Though, maybe someone does use it, by carefully avoiding binds using that modifier...

dev-nicolaos:
Is there an existing issue for release binds? I searched but couldn't find one. It would be nice to be able to bind an action to just Super without breaking all the other binds that use it as a modifier.

YaLTeR:
I'm not sure there's an issue, but implementing release binds would involve some non trivial logic I imagine.

jakesarjeant:
Is allowing mod-only binds necessarily a bug? I can think of at least a few use cases _for_ it, but I'm not clear on what problem it causes... Anyone who doesn't want mod-only binds can just not make any, but it doesn't seem like there's any reason to disallow them for people who want them (of course, release binds are kind of necessary for them to be useful). I'm also bouncing around the idea of showing a workspace overview (see #624) while the mod key is held instead of putting it in my bar, so it seems useful to bind opening the view to `Super+Super_L` with a timer to close it after a few seconds.

miku4k:
they can lead to unexpected behavior if they arent accounted for, which they arent
especially the fact that they get executed

the disallowing them was more a temporary suggestion since i too think release binds would be complex to implement

and yes mod-only keybinds definitively have their uses (super opens some kind of menu on almost all desktops)

jakesarjeant:
Maybe a better solution would be to introduce an `allow-mod-only;` option in the `binds` section of the config and disallow them otherwise?

dev-nicolaos:
[...]

PR #1923, opened 2025-06-28
Implement release keybinds and mod-only binds

Glitch752:
Hi!

I figured I would try my hand at implementing on-release binds.

This PR:
- Adds support for binds with release=true that are triggered on key release when no other binds were triggered between press and release
- Adds test cases and cleans up the existing tests to be more readable
- Adds the ability to assign a bind to only `Mod`. I bundled this in this PR since it feels like the primary use-case for release binds.
  - Adds another trigger, `Trigger::Compositor`, which depends on the current compositor key
- Changes `suppressed_keys` to a `HashMap<Keycode, bool>` where the value stores whether a binding is "still valid". This is set to false when another key is pressed so multiple release bindings (e.g. Mod and Mod+Q) don't fire on a single interaction.

In my testing, I found that some applications misbehave when suppressing mod (either alt or super); I'm not exactly sure why. Therefore, I updated the filter behavior to use a new enum - `ShouldInterceptResult` - which supports the ability to forward _and_ handle a result. Hyprland does this based on my testing.

Note that another significant use case as discussed in #1148 is binding to both press and release of a single bind; this PR doesn't currently handle that.

I understand if you have a better way to implement it in mind, so it's fine if you would prefer to do this yourself as you expressed elsewhere; hopefully this at least contributes some helpful tests.

Thank you so much for Niri! It's by far the most stable and reliable Wayland WM I've tested.

### DIFF
--- a/docs/wiki/Configuration:-Key-Bindings.md
+++ b/docs/wiki/Configuration:-Key-Bindings.md
@@ -81,7 +81,9 @@ binds {
 
 This is mostly useful for the scroll bindings.
 
-Binds can be set to trigger on release instead of on the initial key press. These will only trigger if no other binds were triggered between the press and release.
+### Release bindings
+
+Binds can be set to trigger on release instead of on the initial key press. These will only trigger if no other keys were pressed between their press and release.
 
 This should be set if you want to bind `Mod` to an action.
 
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -315,7 +315,7 @@ pub struct Niri {
 
     pub seat: Seat<State>,
     /// Scancodes and whether a release keybind is valid of the keys to suppress.
-    /// Release bindings only execute when no other binds are handled between press and release.
+    /// Release bindings only execute when no other keys are pressed between press and release.
     /// Therefore, we set the remaining scancodes to false when handling another keybind so we
     /// don't handle multiple release binds in a single interaction.
     pub suppressed_keys: HashMap<Keycode, bool>,


Glitch752:
Alright, I think this is ready for review? Hopefully I didn't miss anything huge.

Thanks again!

niri-config/src/lib.rs
```diff
@@ -3668,6 +3688,9 @@ where
                 "repeat" => {
                     repeat = knuffel::traits::DecodeScalar::decode(val, ctx)?;
                 }
+                "release" => {
+                    release = knuffel::traits::DecodeScalar::decode(val, ctx)?;
```
HigherOrderLogic: Since `release` is default to false, how do you think about replace this with [`check_flag_node`](https://docs.rs/knuffel/latest/knuffel/decode/fn.check_flag_node.html) so the syntax will be shorten to `Mod+A release { ... }`?

Glitch752: That makes sense, although I'm wary of having inconsistent behavior with the other flag properties. Do you know why `repeat`, `allow-when-locked`, etc. don't do this?

HigherOrderLogic: > Do you know why `repeat`, `allow-when-locked`, etc. don't do this?

I dont know the reason why since that's way long before I start using Niri. But it seems like it has been this way from the ever first PR/commit that add attributes to key binds.

34j:
#605 might be related

flowerysong:
Doesn't seem to be quite right, unfortunately. The bind triggers once the key is released, but after a brief pause it enters a weird loop where it constantly triggers. This persists until I press a different key.

Adding `repeat=false` does prevent the repeated triggering, but that feels like a workaround for a deeper problem in how the key events are being processed.

Glitch752:
Huh, I never saw that happening. I'll try to reproduce it and add some more tests when I get the chance; thanks!

flowerysong:
Oh, you might find this helpful if you want to pick this back up: before I discovered it wasn't working I rebased your work on top of the current `main` branch: https://github.com/flowerysong/niri/commits/release-keybinds

(To be clear I switched back to your actual branch after initial testing, to make sure it wasn't just a rebase problem. It wasn't.)

I might have missed a hunk or two (there was a lot of refactoring in this area since you branched), but it should be an easier starting point than doing the updates from scratch.

flowerysong:
Turned out to be pretty straightforward once I actually looked at the logic. `on_keyboard()` would previously return early if the key wasn't pressed. In order to handle binds for release events this early return was removed, so the calls to `start_key_repeat()` after the bind is handled need to be conditional.

https://github.com/flowerysong/niri/commit/4086b7bf4f98d795e8a656070fb6881156bdea28 

```diff
diff --git a/src/input/mod.rs b/src/input/mod.rs
index 143e682392..3680b35fdc 100644
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -478,11 +478,15 @@
                     mods_changed,
                 );
                 self.handle_bind(bind.clone());
-                self.start_key_repeat(bind);
+                if pressed {
+                    self.start_key_repeat(bind);
+                }
             }
             ShouldInterceptResult::InterceptAndHandle(bind) => {
                 self.handle_bind(bind.clone());
-                self.start_key_repeat(bind);
+                if pressed {
+                    self.start_key_repeat(bind);
+                }
             }
             ShouldInterceptResult::InterceptOnly => {}
```

Glitch752:
Oh, good catch! I'm not sure how I missed that. I'll rebase and put together some tests for repeat conditions later today. Thank you!

flowerysong:
Now that I've used it a bit I have a second thing, which is maybe controversial since you had a test to assert the opposite.

I have a lot of binds that are Mod+Ctrl+foo, and I often start to enter one and then change my mind. This sometimes (but only sometimes) results in unwanted triggering of the bind for Mod, depending on the exact order in which I press and release the keys. I prefer it to never trigger in the situation where another key has been pressed.

This also avoids triggering the Mod bind for things like hitting Mod+D when Mod+D is not a keybind, and I feel like it's overall a more predictable experience.

Implemented here: https://github.com/flowerysong/niri/commit/74073ee76f36fe62f7219c1a56babe35affdceba

Glitch752:
I'm not opposed to that change at all! I was just mimicking Hyprland's behavior when implementing. It should probably be made clear in the documentation, though.

Glitch752:
I suppose another question is whether something like `ctrl down, mod down, mod up, ctrl up` should trigger a mod bind. What are your thoughts?

flowerysong:
There's a modifier key pressed so a plain bind of `Mod` should not trigger. Theoretically `Mod+Ctrl release=true` could trigger in that situation.

Glitch752:
Whoops, I'm pretty bad at handling strange merge situations and probably didn't need to force-push; hopefully that was fine.

There's not a good way to test the core `State` logic like repeating without mocking a bunch of stuff, so I don't know how much I can improve that. Let me know if there's any other way I can test broader input behavior.

Everything looks good to me right now; I'm running this branch on my system and haven't had any input issues yet. Thanks for all your help, @flowerysong.

Glitch752:
Superseded by #2456; it's a much more complete implementation :)

PR #1924, opened 2025-06-28
Add `niri msg pointer` and `niri msg action set-pointer-location` commands

cornservant:
This PR implements the following two commands

- `niri msg pointer` which outputs information about the pointer. Currently that would only be the pointer coordinates, but more can be added later.
  - example usage
    ```
    $ niri msg -j pointer
    {"location":{"x":1916.0,"y":557.4905339686553}}

    $ niri msg pointer
    Pointer position: 1916 557.4905339686553
    ```
- `niri msg action set-pointer-location` which moves the pointer to the location specified by x and y coordinates
  - example usage
    ```
    $ niri msg set-pointer-location 1023.5 512
    ```

## Motivation

This feature also exists for Hyprland (`hyprctl cursorpos` and `hyprctl dispatch movecursor`).
It (more or less) solves the following issue of mine: https://github.com/YaLTeR/niri/discussions/1850

### DIFF
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -317,7 +317,7 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
 
             if let Some(pointer) = pointer {
                 println!(
-                    "Pointer position: {} {}",
+                    "Pointer location: {} {}",
                     pointer.location.x, pointer.location.y
                 );
             } else {
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -22,6 +22,7 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
         Msg::FocusedOutput => Request::FocusedOutput,
         Msg::PickWindow => Request::PickWindow,
         Msg::PickColor => Request::PickColor,
+        Msg::Pointer => Request::Pointer,
         Msg::Action { action } => Request::Action(action.clone()),
         Msg::Output { output, action } => Request::Output {
             output: output.clone(),
@@ -300,6 +301,27 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
                 println!("Hex: #{r:02x}{g:02x}{b:02x}");
             } else {
                 println!("No color was picked.");
+            }
+        }
+        Msg::Pointer => {
+            let Response::Pointer(pointer) = response else {
+                bail!("unexpected response: expected PointerPos, got {response:?}");
+            };
+
+            if json {
+                let pointer =
+                    serde_json::to_string(&pointer).context("error formatting response")?;
+                println!("{pointer}");
+                return Ok(());
+            }
+
+            if let Some(pointer) = pointer {
+                println!(
+                    "Pointer position: {} {}",
+                    pointer.location.x, pointer.location.y
+                );
+            } else {
+                println!("No pointer.");
             }
         }
         Msg::Action { .. } => {
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -16,8 +16,8 @@ use knuffel::Decode as _;
 use layer_rule::LayerRule;
 use miette::{miette, Context, IntoDiagnostic};
 use niri_ipc::{
-    ColumnDisplay, ConfiguredMode, LayoutSwitchTarget, PositionChange, SizeChange, Transform,
-    WorkspaceReferenceArg,
+    ColumnDisplay, ConfiguredMode, LayoutSwitchTarget, Point, PositionChange, SizeChange,
+    Transform, WorkspaceReferenceArg,
 };
 use smithay::backend::renderer::Color32F;
 use smithay::input::keyboard::keysyms::KEY_NoSymbol;
@@ -1926,6 +1926,8 @@ pub enum Action {
     UnsetWindowUrgent(u64),
     #[knuffel(skip)]
     LoadConfigFile,
+    #[knuffel(skip)]
+    SetPointerLocation(Point),
 }
 
 impl From<niri_ipc::Action> for Action {
@@ -2202,6 +2204,9 @@ impl From<niri_ipc::Action> for Action {
             niri_ipc::Action::SetWindowUrgent { id } => Self::SetWindowUrgent(id),
             niri_ipc::Action::UnsetWindowUrgent { id } => Self::UnsetWindowUrgent(id),
             niri_ipc::Action::LoadConfigFile {} => Self::LoadConfigFile,
+            niri_ipc::Action::SetPointerLocation { x, y } => {
+                Self::SetPointerLocation(Point { x, y })
+            }
         }
     }
 }
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -178,6 +178,15 @@ pub struct PickedColor {
     pub rgb: [f64; 3],
 }
 
+/// Location coordinates.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, PartialOrd)]
+pub struct Point {
+    /// x coordinate
+    pub x: f64,
+    /// y coordinate
+    pub y: f64,
+}
+
 /// Actions that niri can perform.
 // Variants in this enum should match the spelling of the ones in niri-config. Most, but not all,
 // variants from niri-config should be present here.
@@ -845,6 +854,13 @@ pub enum Action {
     /// Can be useful for scripts changing the config file, to avoid waiting the small duration for
     /// niri's config file watcher to notice the changes.
     LoadConfigFile {},
+    /// Set pointer location
+    SetPointerLocation {
+        /// Desired x coordinate
+        x: f64,
+        /// Desired y coordinate
+        y: f64,
+    },
 }
 
 /// Change in window or column size.
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -80,6 +80,8 @@ pub enum Request {
     FocusedOutput,
     /// Request information about the focused window.
     FocusedWindow,
+    /// Request information about the pointer.
+    Pointer,
     /// Request picking a window and get its information.
     PickWindow,
     /// Request picking a color from the screen.
@@ -156,6 +158,8 @@ pub enum Response {
     PickedWindow(Option<Window>),
     /// Information about the picked color.
     PickedColor(Option<PickedColor>),
+    /// Information about the pointer.
+    Pointer(Option<Pointer>),
     /// Output configuration change result.
     OutputConfigChanged(OutputConfigChanged),
     /// Information about the overview.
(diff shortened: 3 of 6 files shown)


cornservant:
inadvertently closed this PR,  guess I can't use the main branch on my fork for a bit now 😅

src/ipc/client.rs
```diff
@@ -302,6 +303,27 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+            if json {
+                let pointer =
+                    serde_json::to_string(&pointer).context("error formatting response")?;
+                println!("{pointer}");
+                return Ok(());
+            }
+
+            if let Some(pointer) = pointer {
+                println!(
+                    "Pointer position: {} {}",
```
miku4k:
```suggestion
                    "Pointer location: {} {}",
```
for consistency

miku4k:
`PointerHandle::set_position` doesnt update focus or send any window events. imo you should use `PointerHandle::motion` instead, which does those things.

the pointer location is not checked to be in bounds. setting it to something offscreen spawns the pointer back at the center of the first monitor once its moved. the only case where the pointer can be out of bounds currently is when the monitor its on is disconnected, so teleporting it somewhere else is reasonable. i think the movement should be clipped to the edges here instead.

you can reference `State::on_pointer_motion` for implementation details

cornservant:
@miku4k thanks for the review. I made some changes that address the issues you raised, though I'm not quite happy with how the PR is turning out atm because it touches the `on_pointer_motion` and `on_pointer_motion_absolute` methods for code reuse.

YaLTeR:
What's the usecase here? Like, something that *makes sense* to do using this.

For instance, someone on Discord asked me about this, and it turned out that their usecase was to change cursor when on top of a small Quickshell surface. What they actually needed was to just set the mouse input region for their surface, then it all just works without any hacks.

As another example, someone was trying to get Kando to work and they wanted this IPC. But actually that didn't make sense either because they already create a fullscreen overlay surface and get pointer events from there. What they wanted is a workaround to retrofit their X11/Windows logic to Wayland—not a valid usecase, won't even work properly in all cases.

> It (more or less) solves the following issue of mine: #1850

Please don't solve pointer warping by adding a function to warp it back immediately afterwards. As I mentioned in the issue, a flag to not warp the pointer is much better.

KapDecy:
I could really use this thing. 
My usecase: I want to create bindings that allow me to control windows (close, move, resize) using only the mouse. This feature would be very helpful with the first one.
I know that `pick-window` exists, but as far as I understand, it is not possible to immediately select the window under the cursor, only by clicking on it additionally.
Is there a chance that this pull request will be accepted? Or are there other ways to do what I want?

miku4k:
you can already interactively move and resize windows with mod+left and mod+right respectively, and you can bind actions (eg `close-window`) to mouse buttons https://yalter.github.io/niri/Configuration%3A-Key-Bindings.html#mouse-click-bindings

KapDecy:
> you can bind actions (eg `close-window`) to mouse buttons https://yalter.github.io/niri/Configuration%3A-Key-Bindings.html#mouse-click-bindings

Yes, but if I do that, I first need to activate the window I need and only then close it. I understand that it's just one extra click, but I believe that users should be able to customize niri to suit their needs.

YaLTeR:
> My usecase: I want to create bindings that allow me to control windows (close, move, resize) using only the mouse. This feature would be very helpful with the first one.

I don't see how these requests would help you with these usecases. In any case it sounds like the wrong feature to use for this.

skelitheprogrammer:
My usecase is to spawn context menu like fuzzel window with commands under cursor position to perform commands. Would love to have simple way to get position coordinates

Dakedres:
My usecase is to spawn `copyq menu` and move it near the cursor

Axlefublr:
for movement, you can use ydotool \
only downside is that you won't be able to bring back your cursor to its initial spot (since no way to get current cursor position without this pr)

AClon314:
my use case is read cursor position for obs `zoom in and follow mouse movement` script.

eliasiturri:
My use case is to programmatically switch my KVM Switch (I'm only interested in the usb peripherals) when i push at the edge of the monitor, which I struggle to do on the niri side (works just fine on my Mac's monitor).

You are doing a great job here and I see you are filled with issues and PR, but I think knowing the coordinates of your mouse or information on other input peripherals is always a nice feature. Keep up the good work regardless!

YaLTeR:
You can do it with a 1 px transparent overlay layer-shell surface.

eliasiturri:
> You can do it with a 1 px transparent overlay layer-shell surface.

Yes, that's my current approach when on niri. 

Like I said, niri is great, and I don’t want to imply that what you’ve built is anything less than awesome. That said, solutions like this feel more like workarounds than what I’d consider a standard feature.

Still, thanks for the suggestion!

YaLTeR:
I disagree; using something like `niri msg pointer` for this case would be a brittle hack (you'd have to manually compute monitor positions and sizes, account for rounding and so on). 1 px surface *is* the good solution for this case.

bczhc:
Surprised niri just doesn't have this ipc. My usecase is just to debug a cursor speed [issue](https://github.com/ValveSoftware/Proton/issues/7448#issuecomment-3991016787) with a proton game. I'm not a wayland dev and won't use some tricky approaches like 1px transparent layer etc etc - I'm just looking for a simple and direct way to get the mouse pointer position. With `hyprctl` it's easy to do this. Even I'm able to programmatically communicate with its socket to get a very high sample rate, as this [script](https://gist.githubusercontent.com/bczhc/8006ad0c8ea9410b8061271a57ee817a/raw/527f36bb00aada45c2d7cc80ec72f926fcb11457/hypr-cursor) writes.

YaLTeR:
> My usecase is just to debug a cursor speed https://github.com/ValveSoftware/Proton/issues/7448#issuecomment-3991016787

`sudo libinput debug-events`?

bczhc:
> `sudo libinput debug-events`?

Libinput and evtest don't do the thing because in my understanding they don't report the pointer location, but "raw" values which obviously are of no use because they are just constantly sent by my uinput automation program - i already know them.

alper-han:
KDE, GNOME, Hyprland, and Wayfire already expose a way to support absolute coordinates. COSMIC and niri do not yet expose the capabilities I need. To add absolute coordinate support for mouse recording to https://github.com/alper-han/CrossMacro , I need access to the current cursor position. In KDE, this is handled through a KWin script over D-Bus. In GNOME, it is handled through a Shell extension over D-Bus. Hyprland provides it through built-in IPC. Wayfire provides it through IPC/plugin support. I cannot say what the ideal interface for this should be, but on Linux there are real use cases where access to this kind of information is necessary.

YaLTeR:
Can you give an example?

Axlefublr:
macros. if you want to make a hotkey that clicks on a certain spot on the screen and then *comes back*, you necessarily need to be able to first check where you are

YaLTeR:
What's the usecase for that?

alper-han:
> Can you give an example?

CrossMacro can work without ABS support, but it currently has to rely on relative movement and effectively operate blindly.

For recording and playback, it first has to move the pointer to the 0x0 coordinate. This can cause undesirable behavior in some cases.
Also, because the movement is relative, if the user moves the mouse during playback, CrossMacro cannot recover the correct position. From that point onward, the rest of the playback becomes inaccurate.
Mouse acceleration can also make this unreliable if the user is not using a flat acceleration profile.

So ABS support is not required for every use case, but it is important for this kind of macro/recording use case.

If exposing this is not acceptable for security reasons, I understand. But I do not think that only reading the current pointer position is harmful by itself, since actually sending mouse input still requires input/uinput permissions.

YaLTeR:
I am asking for a good example use on niri for a macro that moves the mouse to a certain spot and then comes back, or otherwise *needs* access to mouse position

Axlefublr:
plants vs zombies. genuinely that game is why I forked niri in the first place, adding this pr in.
I hate using my mouse, especially vertical movement; and so I made macros to choose a plant to plant with my keyboard, so all I need to care about is choosing *where* to plant. much much more ergonomic, the game was literally unplayable for me before those macros

Axlefublr:
your response may reasonably be “so, cheating?” and you are only somewhat wrong, but it being on the compositor to be for or against cheating is kind of silly

YaLTeR:
Is it an X11 game? If so then you can use xdotool. In any case window-relative coords and input is more appropriate here than global niri coordinates

ElXreno:
My use case is [linux-wallpaperengine](https://github.com/Almamu/linux-wallpaperengine), which supports mouse interaction.

HigherOrderLogic:
> My use case is [linux-wallpaperengine](https://github.com/Almamu/linux-wallpaperengine?rgh-link-date=2026-06-06T11%3A11%3A50Z),

Isnt it a layer-shell? If so it can just consume the pointer event through the protocol.

ElXreno:
> > My use case is [linux-wallpaperengine](https://github.com/Almamu/linux-wallpaperengine?rgh-link-date=2026-06-06T11%3A11%3A50Z),
> 
> Isnt it a layer-shell? If so it can just consume the pointer event through the protocol.

It already consumes `wl_pointer`, but pointer focus only goes to the topmost surface, so once the cursor is over a window the background wallpaper can't see it - which is exactly the position parallax/eye-tracking needs. That's why it already uses a global query on Hyprland (`hyprctl cursorpos`); `niri msg pointer` is just the equivalent.

Refs: Almamu/linux-wallpaperengine#568 (added the global query for parallax), Almamu/linux-wallpaperengine#576 (the focus-vs-background-surface gap).

bczhc:
I saw the Hyprland cursor position in linux-wallpaperengine is fetched via socket communication: https://github.com/Gaiser147/linux-wallpaperengine/blob/acf7626f114b593e858d5d45cc4892ef5e1468b1/src/WallpaperEngine/Input/Drivers/WaylandMouseInput.cpp#L80-L145. It's a more efficient way and has a higher sample rate. For niri now but we don't have explicit niri socket message definition for 3rd tool use?. So if only using frequent `niri msg pointer` calls it will keep spawning lots of processes.

HigherOrderLogic:
> So if only using frequent `niri msg pointer` calls it will keep spawning lots of processes.

Niri would also need to implement messaging the pointer information through the socket if `niri msg pointer` is ever implemented.

bczhc:
It does. What i mean is one ipc cli call is a process creation and each call is a socket connection but if writing an external program which connects to the socket only once then fetching info from it repeatedly can be very efficient. For example in my comment https://github.com/niri-wm/niri/pull/1924#issuecomment-4122865353, the socket way is *way way more* efficient than a `while :; do hyprctl cursorpos; done | ts` approach from my experience before.

update: oops correction - hyprland socket is also "one shot" which means one info fetch is one socket connection. Their efficiency difference is mainly caused by 'process creation'.

YaLTeR:
tbh this maybe calls for a custom wayland protocol?

hxmbl:
so i have a project that kind of needs this...

cornservant:
@hxmbl here's an updated version of the patch based on the current main branch of niri if that's of any use to you: https://github.com/cornservant/niri/tree/pr/ipc-pointer-v2

hxmbl:
thanks, ill check it out

PR #1929, opened 2025-06-29
feat: add window alignment actions

b0o:
I use two relatively large monitors side-by-side, and I sit near the middle of them. I often want my primary working windows to be on the right side of the left monitor, or the left side of the right monitor. 

There was no easy way to do this previously, other than moving focus back-and-forth between adjacent windows to nudge the view into the desired position, or using the mouse+middle-click-dragging to move the view. And, if I wanted the first window to be on the right side of the screen, or the last window to be on the left side, there was no way to do that.

This PR adds actions for aligning windows to the left, center, and right edge of the screen:

- `AlignWindowLeft`: Aligns the focused window so it's at the left edge of the screen
- `AlignWindowCenter`: Aligns the focused window to the center of the screen (same as `CenterWindow`)
- `AlignWindowRight`: Aligns focused window to the right side of the screen

As mentioned, `AlignWindowCenter` does the exact same thing as the existing `CenterWindow`. I'm not sure how to handle this. Either:

1. Keep both for backward compatibility
2. Deprecate `CenterWindow` in favor of `AlignWindowCenter` 
3. Something else?

Also adds variants to do this by ID, but not sure how useful that really is. Happy to remove and just operate on the focused window. 

Final question, not sure if these should really be `AlignColumn{Left,Right,Center}`? Functionally, I think they would behave the same, whether it's operating on windows or columns.

<details>
  <summary>Demo Video</summary>

  https://github.com/user-attachments/assets/8607c704-f5a8-4a3b-b63f-11b36d4f0e78

</details>

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1778,6 +1778,15 @@ pub enum Action {
     CenterWindow,
     #[knuffel(skip)]
     CenterWindowById(u64),
+    AlignWindowLeft,
+    #[knuffel(skip)]
+    AlignWindowLeftById(u64),
+    AlignWindowCenter,
+    #[knuffel(skip)]
+    AlignWindowCenterById(u64),
+    AlignWindowRight,
+    #[knuffel(skip)]
+    AlignWindowRightById(u64),
     CenterVisibleColumns,
     FocusWorkspaceDown,
     #[knuffel(skip)]
@@ -2022,6 +2031,12 @@ impl From<niri_ipc::Action> for Action {
             niri_ipc::Action::CenterColumn {} => Self::CenterColumn,
             niri_ipc::Action::CenterWindow { id: None } => Self::CenterWindow,
             niri_ipc::Action::CenterWindow { id: Some(id) } => Self::CenterWindowById(id),
+            niri_ipc::Action::AlignWindowLeft { id: None } => Self::AlignWindowLeft,
+            niri_ipc::Action::AlignWindowLeft { id: Some(id) } => Self::AlignWindowLeftById(id),
+            niri_ipc::Action::AlignWindowCenter { id: None } => Self::AlignWindowCenter,
+            niri_ipc::Action::AlignWindowCenter { id: Some(id) } => Self::AlignWindowCenterById(id),
+            niri_ipc::Action::AlignWindowRight { id: None } => Self::AlignWindowRight,
+            niri_ipc::Action::AlignWindowRight { id: Some(id) } => Self::AlignWindowRightById(id),
             niri_ipc::Action::CenterVisibleColumns {} => Self::CenterVisibleColumns,
             niri_ipc::Action::FocusWorkspaceDown {} => Self::FocusWorkspaceDown,
             niri_ipc::Action::FocusWorkspaceUp {} => Self::FocusWorkspaceUp,
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -426,6 +426,42 @@ pub enum Action {
         #[cfg_attr(feature = "clap", arg(long))]
         id: Option<u64>,
     },
+    /// Align a window to the left edge of the screen.
+    #[cfg_attr(
+        feature = "clap",
+        clap(about = "Align the focused window to the left edge of the screen")
+    )]
+    AlignWindowLeft {
+        /// Id of the window to align.
+        ///
+        /// If `None`, uses the focused window.
+        #[cfg_attr(feature = "clap", arg(long))]
+        id: Option<u64>,
+    },
+    /// Align a window to the center of the screen.
+    #[cfg_attr(
+        feature = "clap",
+        clap(about = "Align the focused window to the center of the screen")
+    )]
+    AlignWindowCenter {
+        /// Id of the window to align.
+        ///
+        /// If `None`, uses the focused window.
+        #[cfg_attr(feature = "clap", arg(long))]
+        id: Option<u64>,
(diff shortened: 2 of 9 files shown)


YaLTeR:
Makes sense, and yeah I have similar bikeshedding concerns.

Regarding `AlignColumn...`: I think it makes sense to have them yeah, should be easy enough to add using the same logic.

Regarding `CenterWindow`: one thing that just came to my mind is that for floating windows, `CenterWindow` is both-axes center, whereas `AlignWindowLeft/Right` would be horizontal-only, and by this logic `AlignWindowCenter` would also be horizontal-only, giving a useful behavior difference. Though, in this case it would be natural to (later on) extend the actions with `AlignWindowTop/Bottom`, but how to call the vertical center? I think usually toolkits call it something like `Horizon` or whatever. But also, `AlignWindowCenter` doesn't obviously indicate by name that it is horizontal-only..

> Also adds variants to do this by ID, but not sure how useful that really is.

Yes, this is useful, keep. For `Column` not needed yet since we don't have column ids.

Haven't looked at the code in depth yet.

b0o:
> Regarding CenterWindow: one thing that just came to my mind is that for floating windows, CenterWindow is both-axes center, whereas AlignWindowLeft/Right would be horizontal-only, and by this logic AlignWindowCenter would also be horizontal-only, giving a useful behavior difference. 

Good thinking. Currently, for the floating layer, this PR always aligns the window to the center vertically AND the specified alignment. In tiling layer it doesn't do this, it just changes the horizontal position. I will fix this so it only changes the horizontal position in both cases.

> Though, in this case it would be natural to (later on) extend the actions with AlignWindowTop/Bottom, but how to call the vertical center? I think usually toolkits call it something like Horizon or whatever. But also, AlignWindowCenter doesn't obviously indicate by name that it is horizontal-only..

An option would be to call these commands `VerticalAlign` or `VAlign`. 

An alternative option would be to change this PR to have a single `AlignWindow` action which accepts the alignment as argument(s). Some ideas:

```kdl
// Explicit horizontal/vertical arguments
align-window horizontal="center"
align-window vertical="bottom"
align-window h="left" // maybe allow shorthand
align-window vert="top" // or slightly longer shorthand
align-window h="left" v="center" // multiple axes at once
// Allow not specifying the axis for unambiguous alignments
align-window "left"
align-window "right"
align-window "top"
align-window "top" "left" // multiple unambiguous axes at once
align-window "top" "bottom" // Invalid - can't use the same axis twice
align-window "center" // Centers on both axes
align-window vertical="center" // Centers vertically only
align-window "left" "center" // Infer that "center" must refer to the vertical axis
align-window "left" vertical="center" // Mixed named/unnamed args
```

I actually like this approach better as it introduces fewer new actions (especially if we plan to add AlignColumn in the future), and it opens up the possibility of specifying arbitrary alignment values like `align-window horizontal="30%"` in a future enhancement.

Thoughts?

YaLTeR:
I like this idea with horizontal= and vertical= yeah

tsdh:
I like the idea of the align commands.  I think they would allow me to use `layout.center-focused-column "always"` which I cannot use right now because in very rare cases I want to see as much as possible of the window left/right of the focused column.

Just as an addition: I'd love to have a toggle command switching between the alignment states left/centered/right.  That'd be awesome!

mastoca:
> I like this idea with horizontal= and vertical= yeah

I agree, setting like these would be wonderful. I'd advocate to have "center", "left", "right", "none" as valid options.

yebei199:
https://github.com/YaLTeR/niri/discussions/2464

That sounds closely related to an idea I’ve been thinking about—supporting precise control over column positioning. I’ve run into similar issues myself, especially when working across two monitors where the column alignment often needs fine-tuning.

IndiTheo:
There's a built-in feature in PaperWM that allows to change the focus mode with a keybind on **the compositor level** and define how windows snap/do not snap, there are three modes:
- Edge snapping (default niri)
- Free scrolling (no snapping)
- Always center

Niri allows to enlarge a window so it's bigger than a screen (continuously pressing the + button), and in this case it's impossible to scroll over the window, as it always snaps to **the left**. It would be nice to be able to change the default focus mode with a keybind, including free scrolling.

Related: https://github.com/YaLTeR/niri/discussions/1347

UPD:
In case a window is bigger than the screen, and there's only one screen/no other screen can pick the other part of window, the reasonable default is free scrolling for that specific window.

MintyDoggo:
Hello, I just saw this after making the column alignment PR: https://github.com/YaLTeR/niri/pull/3033

I would like these features as well so I could probably work on testing and trying to get these individual window alignment changes merged after getting the align column PR merged. Let me know if there's anything I can do to help

mushrowan:
hi! just noting that this feature would be great for my workflow too :)

artiepoole:
Hey @b0o , if you have this ready to go, please could you try this with columns wider than display width?

For context, I just posted https://github.com/niri-wm/niri/discussions/4269 with the intention to do similar work.

To note: I also asked Minty Doggo to test this on #3033 

Thanks in advance!

b0o:
I'm going to close this because I haven't been able to find time to work on it, unfortunately. If someone else would like to take a stab at this functionality, I'd be happy to give it a try!

artiepoole:
Hey @b0o, I wanted to let you know that I rebased and then modified this implementation into this draft: https://github.com/niri-wm/niri/pull/4417 
You're credited in the relevant commit, of course!
If you are interested, please take a look and provide any feedback before I make it "ready for review" ^.^

PR #1946, opened 2025-07-01
Register `org.freedesktop.ScreenSaver` at `/ScreenSaver`

HigherOrderLogic:
Found out about this while implementing an idle inhibitor.

Apparently, other tools bind this D-Bus protocol at both address, eg: [cosmic-idle](https://github.com/pop-os/cosmic-idle/blob/master/src/freedesktop_screensaver.rs#L87-L88), [hypridle](https://github.com/hyprwm/hypridle/blob/main/src/core/Hypridle.cpp#L545-L546). Some applications like VLC also try to call the protocol on this path instead of `/org/freedesktop/ScreenSaver`. Therefore I think it's sensible to register the interface at `/ScreenSaver` as well.


src/dbus/freedesktop_screensaver.rs
```diff
@@ -13,6 +13,7 @@ use zbus::{interface, Task};
 
 use super::Start;
 
+#[derive(Clone)]
```
YaLTeR, line 16: If you want this Cloned like that, you need to make the `counter` shared too, i.e. via `Arc<AtomicU32>` which you `fetch_add()` with `SeqCst` ordering just to be sure.

YaLTeR (commented):
Huh. That's weird but I see, alright, probably some legacy reason

src/dbus/freedesktop_screensaver.rs
```diff
@@ -44,15 +45,16 @@ impl ScreenSaver {
...
-            self.counter = self.counter.wrapping_add(1);
-            if self.counter == 0 {
-                self.counter += 1;
+            self.counter.fetch_add(1, Ordering::SeqCst);
+            if self.counter.load(Ordering::SeqCst) == 0 {
+                self.counter.fetch_add(1, Ordering::SeqCst);
             }
 
-            if let Entry::Vacant(entry) = inhibitors.entry(self.counter) {
+            let inhibitor_key = self.counter.load(Ordering::SeqCst);
```
YaLTeR: You shouldn't load atomics separately because another thread may be modifying them at the same time. This is why `fetch_add()` both adds and returns the previous value.

HigherOrderLogic: So I should do `if self.counter.fetch_add(...) == u32::MAX`?

YaLTeR: fetch_add returns the previous value, so I think something like `let mut key = counter.fetch_add(1); if key == 0 { key = counter.fetch_add(1); }`

HigherOrderLogic: > fetch_add returns the previous value, so I think something like `let mut key = counter.fetch_add(1); if key == 0 { key = counter.fetch_add(1); }`

If the previous value before adding 1 is 0, isnt the current value 1?

Since `.fetch_add` wrap on overflow, shouldn't `u32::MAX` be the correct one here?

HigherOrderLogic:
Sorry, close by mistake.

src/dbus/freedesktop_screensaver.rs
YaLTeR: The way it works is when you call `.fetch_add()`, you get the value that you will use, and the variable is also atomically incremented (so the next time someone calls it, they get their next value).

HigherOrderLogic: Pushed the change, here is the [code snippet](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8ef579da8ca2f5a28efeaca73fd9031c) that I use to validate my code, beside testing with the local build.

YaLTeR: This is what the code should be, please use this:

```rust
            let mut inhibitor_key = self.counter.fetch_add(1, Ordering::SeqCst);
            if inhibitor_key == 0 {
                // Clients don't like 0, add one more.
                inhibitor_key = self.counter.fetch_add(1, Ordering::SeqCst);
            }

            // Remove this line below:
            // let inhibitor_key = self.counter.load(Ordering::SeqCst);
```

YaLTeR:
Pushed some minor fixes, good to go now. Thanks!

PR #1951, opened 2025-07-01
make hot corner size configurable.

neio4:
I don't think my current approach should be the final solution and I am open to suggestions!
also my usage of `FloatOrInt<0, 1_000_000>` seems wrong to me. What would be an more appropriate limit than 1_000_000?
```kdl
gestures {
    hot-corners {
    	x 10
    	y 10
    }
}
```
any ideas how to structure the config are welcome!

I also think that the user should be able to configure 4 hot corners not just 1.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1341,6 +1341,10 @@ impl Default for DndEdgeWorkspaceSwitch {
 pub struct HotCorners {
     #[knuffel(child)]
     pub off: bool,
+    #[knuffel(child,unwrap(argument), default = FloatOrInt(1.))]
+    pub x: FloatOrInt<0, 1_000_000>,
+    #[knuffel(child,unwrap(argument), default = FloatOrInt(1.))]
+    pub y: FloatOrInt<0, 1_000_000>,
 }
 
 #[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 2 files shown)


Dlurak:
It seems like [this line](https://github.com/YaLTeR/niri/blob/37458d94b288945f6cfbd3c5c233f634d59f246c/src/niri.rs#L3081) also needs to be adjusted.

phisch:
I think if you want to make hot corners more configurable. You should make it so you can define a corner and a size.
`x` and `y` are typically used to reference a point, not a size. So something like that might be better:

```kdl

gestures {
    hot-corner LOCATION width=20 height=30  {
        action: "overview"
    }
}
```

And `LOCATION` would be `north-east | south-east | south-west | north-west`.

There are probably even better ways to do this, and it would be good to know what the standpoint of @YaLTeR is on this.

YaLTeR:
I agree that some configuration would be useful (especially which monitor they are on) but I don't have any concrete design yet. Some considerations:
- I'm not sure when you'd need to configure hot corner size per hot corner
- Is configuring hot corner size even needed? Aren't they hardcoded in GNOME Shell?
- For per-output hot corners it may make sense to put `hot-corners { off; }` into the `output {}` sections but idk
- `waycorners` covers many hot corner use cases, it doesn't have to all be in niri

phisch:
I think the main thing I want to configure for niris hot corner is which location to use. But then I could also use waycorners or quickshell instead.

Edit: idk how often I have accidentally triggered the top left hot corner, because the top left is often a corner of interest in most apps. It must be in the hundreds by now.

PR #1965, opened 2025-07-05
Implement a four-finger swipe gesture to do an interactive move

StratusFearMe21:
https://github.com/user-attachments/assets/485eb9de-71ee-44bd-95f8-1118c8b763e1

This PR implements a 4-finger swipe gesture which allows you to move the focused window with just your touchpad (no mod key). I've taken the same logic which decides the swipe direction for 3 finger swipes, and applied it to 4 finger swipes, this way the overview gesture can still be kept.

Please forgive me if the way I've implemented this isn't standard, I have very little experience with Smithay and Wayland Compositor design in general. That being said, there's couple of question marks in this PR.

- Is it possible for a focused window to be out of view? If so, my usage of `active_tile_visual_rectangle` in the `on_gesture_swipe_update` function is probably incorrect
- I *am* clamping the position of the window in the swipe gesture to the `global_bounding_rectangle`, but it's unclear why I need to shrink it by 1 logical pixel to avoid the window snapping to a weird position when it's all the way at the edge.

I've had so much fun using this window this window manager! Thanks for such an awesome project!

### DIFF
--- a/src/handlers/xdg_shell.rs
+++ b/src/handlers/xdg_shell.rs
@@ -153,7 +153,7 @@ impl XdgShellHandler for State {
 
         match start_data {
             PointerOrTouchStartData::Pointer(start_data) => {
-                let grab = MoveGrab::new(start_data, window, false);
+                let grab = MoveGrab::new(start_data, window, false, false);
                 pointer.set_grab(self, grab, serial, Focus::Clear);
             }
             PointerOrTouchStartData::Touch(start_data) => {
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -83,6 +83,61 @@ impl<D: SeatHandler> PointerOrTouchStartData<D> {
     }
 }
 
+#[derive(PartialEq, Eq, Clone, Copy)]
+enum SwipeDirection {
+    Horizontal,
+    Vertical,
+    Unknown,
+    Undecided,
+}
+
+pub struct SwipeDirectionDecider {
+    cx: f64,
+    cy: f64,
+    decision: SwipeDirection,
+}
+
+impl SwipeDirectionDecider {
+    pub fn decided() -> Self {
+        Self {
+            cx: 0.,
+            cy: 0.,
+            decision: SwipeDirection::Unknown,
+        }
+    }
+
+    fn undecided() -> Self {
+        Self {
+            cx: 0.,
+            cy: 0.,
+            decision: SwipeDirection::Undecided,
+        }
+    }
+
+    fn update_and_maybe_decide(&mut self, delta_x: f64, delta_y: f64) -> Option<SwipeDirection> {
+        if self.decision != SwipeDirection::Undecided {
+            return None;
+        }
+
+        self.cx += delta_x;
+        self.cy += delta_y;
+
+        // Check if the gesture moved far enough to decide. Threshold copied from GNOME Shell.
+        if self.cx * self.cx + self.cy * self.cy >= 16. * 16. {
+            self.decision = if self.cx.abs() > self.cy.abs() {
+                SwipeDirection::Horizontal
+            } else {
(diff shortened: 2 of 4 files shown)


YaLTeR:
Haven't looked at the code yet.

> Is it possible for a focused window to be out of view?

Yes, if you focus something off-screen and the animation hasn't caught up there yet. Also yes during gestures (view offset, workspace switch). Also maybe yes without monitors.

StratusFearMe21:
> Yes, if you focus something off-screen and the animation hasn't caught up there yet. Also yes during gestures (view offset, workspace switch). Also maybe yes without monitors.

Gotcha, in that case mapping `active_tile_visual_rectangle` to `pos_within_output` in this case should work just fine.

armerpunkt:
Took a little to get used to, but this seems to be working pretty well.  Just a couple comments:

1.  Sometimes a finger will come off and it'll think I let go. But I'm not sure if it's technically possible to keep it engaged as long as one finger remains?
2. It really is counterintuitive that it moves the focused window and not the window under the mouse. Would it be possible to have it focus on the window underneath the cursor (and if so maybe have that configurable?)

StratusFearMe21:
> Sometimes a finger will come off and it'll think I let go. But I'm not sure if it's technically possible to keep it engaged as long as one finger remains?

Yes, but only if I added a delay between you fully letting go and the window being set down. `libinput` is responsible for interpreting gestures like this, and if you let go of just one finger in the middle of a gesture, it'll output this
```
event20  GESTURE_SWIPE_UPDATE     26  +0.808s  4 -3.13/ 0.12 (-3.13/ 0.12 unaccelerated)
 event20  GESTURE_SWIPE_END            +0.920s  4 cancelled
 event20  GESTURE_SWIPE_BEGIN          +0.966s  3
 event20  GESTURE_SWIPE_UPDATE         +0.977s  3 -6.45/ 0.08 (-6.72/ 0.08 unaccelerated)
```
It'll fully end the gesture and begin a new one, so there's no way to differentiate between you just partially letting go or fully letting go.

> It really is counterintuitive that it moves the focused window and not the window under the mouse. Would it be possible to have it focus on the window underneath the cursor (and if so maybe have that configurable?)

You're right, I have focus follows mouse turned on, so that's probably why I like this gesture so much. I'll make sure to adjust that.

armerpunkt:
> It'll fully end the gesture and begin a new one, so there's no way to differentiate between you just partially letting go or fully letting go.

Hmm... I found a comment by one of their developers that makes it sounds like maybe it's a bug with libinput. So I'll see if I can get some feedback from them.

StratusFearMe21:
> Hmm... I found a comment by one of their developers that makes it sounds like maybe it's a bug with libinput. So I'll see if I can get some feedback from them.

Interesting, maybe also mention to them that I'm experiencing on an Apple Magic Trackpad in Bluetooth mode

mikenrafter:
> Yes, but only if I added a delay between you fully letting go and the window being set down. `libinput` is responsible for interpreting gestures like this, and if you let go of just one finger in the middle of a gesture, it'll output this
> ```
> event20  GESTURE_SWIPE_UPDATE     26  +0.808s  4 -3.13/ 0.12 (-3.13/ 0.12 unaccelerated)
>  event20  GESTURE_SWIPE_END            +0.920s  4 cancelled
>  event20  GESTURE_SWIPE_BEGIN          +0.966s  3
>  event20  GESTURE_SWIPE_UPDATE         +0.977s  3 -6.45/ 0.08 (-6.72/ 0.08 unaccelerated)
> ```
> It'll fully end the gesture and begin a new one, so there's no way to differentiate between you just partially letting go or fully letting go.


What about letting the user declare how long they want this period to be? From this excerpt you've shared, it seems like the timescales such libinput restarts happen on are rather short. 100ms of debounce would feel acceptable to me.

PR #1967, opened 2025-07-05
Allow NIRI_SOCKET to be derived

StratusFearMe21:
This PR allows the `niri` CLI to find a running instance of niri without the `NIRI_SOCKET` environment variable defined. Because I use S6 as my init system and run niri as well as my terminal emulator under S6 user services, `NIRI_SOCKET` isn't part of my terminal environment.

This PR finds the running `niri` instance by optionally using the `WAYLAND_DISPLAY` environment variable to filter instances of the IPC socket in the `socket_dir`. If there's more than one option for what the socket could be, it errors out and says that `NIRI_SOCKET ` isn't defined

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2330,6 +2330,7 @@ name = "niri-ipc"
 version = "25.5.1"
 dependencies = [
  "clap",
+ "directories",
  "schemars",
  "serde",
  "serde_json",
--- a/niri-ipc/Cargo.toml
+++ b/niri-ipc/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 
 [dependencies]
 clap = { workspace = true, optional = true }
+directories = "6.0.0"
 schemars = { version = "1.0.0", optional = true }
 serde.workspace = true
 serde_json.workspace = true
--- a/niri-ipc/src/socket.rs
+++ b/niri-ipc/src/socket.rs
@@ -1,10 +1,13 @@
 //! Helper for blocking communication over the niri socket.
 
 use std::env;
+use std::ffi::OsString;
 use std::io::{self, BufRead, BufReader, Write};
 use std::net::Shutdown;
-use std::os::unix::net::UnixStream;
-use std::path::Path;
+use std::os::unix::{ffi::OsStrExt, net::UnixStream};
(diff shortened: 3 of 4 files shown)


sodiboo:
Could this perhaps be implemented as an explicit action, rather than the default behaviour? Something like `NIRI_SOCKET="$(niri msg find-socket)" niri msg ...`, or `NIRI_SOCKET="auto" niri msg ...` (for example)

for that matter, does this need implementation in niri? `NIRI_SOCKET=$XDG_RUNTIME_DIR/niri.$WAYLAND_DISPLAY.*.sock niri msg ...` seems to work fine.

I don't like the idea that i'll forget to propagate `NIRI_SOCKET` somewhere, and then it'll "magically" work but only sometimes, and that "sometimes" being if and only if i'm not running any secondary instance of niri and there only exists one relevant socket.

Also, other apps may not necessarily use `niri-ipc`, so this change as-is will implicitly require work in other languages that don't depend on our rust crate, to match the behaviour. It changes the contract of `NIRI_SOCKET`, to one where the client apps should be aware of *how* that socket path is generated. If it's explicit like `NIRI_SOCKET="auto"`; that contract is still changed, but if we expose e.g. `niri msg find-socket`, then nothing needs to change in existing client applications.

As for your use case of the variable not being set in the S6 init system, it sounds like you should be using your init system properly? Why can't it propagate the environment variable? Does it need niri to do something? If so, you should patch niri to support S6, and maybe submit a PR for the init support. We already support systemd and dinit; and while i'm not sure how sustainable it is to support every init system out there, it is the case that at least for now, we *do* support several, so we might as well have another.

YaLTeR:
I agree with @sodiboo. The way you have things set up seems just wrong. `$NIRI_SOCKET` follows the standard pattern together with things like `$DISPLAY` and `$WAYLAND_DISPLAY` where the compositor sets environment variables for sockets it creates, then imports them into the global systemd/dinit environment, then also sets them for subprocesses. Instead of adding magic socket finding to niri msg, you need to fix your init system integration to import the relevant variables there.

StratusFearMe21:
I also agree with @sodiboo, I didn't think about that I could just use environment variables to derive `NIRI_SOCKET`. Thanks for the tip!

Issue #338, opened 2024-05-12
Add a way to set, rather than toggle, window fullscreen


YaLTeR:

Antiz96:
Just passing by to express my support and interest toward such a feature! 🤗 👼 

For what it's worth, it'd be nice to see a way to set/unset state not only for `fullscreen-window`, but also for others "maximizing" related actions such as `MaximizeColumn` and `MaximizeWindowToEdges`. That would improve flexibility and control, notably for IPC scripting!

Some related work were initially started in https://github.com/niri-wm/niri/pull/1968 but it has been stalled for a year by now. Any chance to see some this feature being added? I think I can speak confidently saying that it would be an appreciated addition! 😸

PR #1968, opened 2025-07-06
Exact fullscreen state

Dlurak:
As already proposed in #337 and #338 I've implemented an way to control the fullscreen mode more precisely.
There are now 3 actions: `fullscreen`, `normal` and `toggle`. Because `toggle` is the default it is fully downwards compatible.

## IPC-Controll

```bash
niri msg action fullscreen-window --action fullscreen
niri msg action fullscreen-window --action normal
niri msg action fullscreen-window --action toggle
niri msg action fullscreen-window # Equal to toggle
````

## Config

```kdl
// - "toggle", default behavior, toggling between fullscreen and normal
// - "normal", setting a window into its normal size independent of its current state
// - "fullscreen", setting a window into fullscreen independent of its current state
 Mod+Shift+F { fullscreen-window "toggle"; }
```

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -14,8 +14,8 @@ use knuffel::Decode as _;
 use layer_rule::LayerRule;
 use miette::{miette, Context, IntoDiagnostic};
 use niri_ipc::{
-    ColumnDisplay, ConfiguredMode, LayoutSwitchTarget, PositionChange, SizeChange, Transform,
-    WorkspaceReferenceArg,
+    ColumnDisplay, ConfiguredMode, FullscreenSetAction, LayoutSwitchTarget, PositionChange,
+    SizeChange, Transform, WorkspaceReferenceArg,
 };
 use smithay::backend::renderer::Color32F;
 use smithay::input::keyboard::keysyms::KEY_NoSymbol;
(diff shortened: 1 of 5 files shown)


Dlurak:
Maybe also allow these options for `MaximizeColumn`?

YaLTeR:
The way this works in other commands is "toggle" vs. "set", i.e. "toggle-window-urgent" vs. "un/set-window-urgent" or "toggle-window-floating" vs. "move-window-to-floating". So since we already have "fullscreen-window", I think it would make more sense to have a "un/set-window-fullscreen". And for maximize it would be "un/set-column-maximized".

Antiz96:
Hi,

Thanks for the work on this!

Allow me a friendly ping to ask if this change is still on its way? :angel:

The absence of a "maximized/fullscreen" state of a window that could be gathered from the IPC (as proposed in https://github.com/niri-wm/niri/pull/2836) and/or the lack of such a set/unset action for `MaximizeColumn` (as proposed in this PR) makes it really hard to do some IPC scripting around triggering window maximizing (e.g. to prevent toggling maximizing on a window that is already maximized).

As stated in https://github.com/niri-wm/niri/pull/2270#issue-3337717998, the workaround so far to determine if a window is *supposedly* maximized is to compare the window size with the screen output size, with an *arbitrary* pixel tolerance to *hopefully* take eventual gaps and exterior elements (like a panel, taskbar, dock) into account... Which is not elegant and reliable :sweat_smile:

A reliable way to get the maximized/fullscreen state from the EventStream would definitely be helpful, but such a set/unset option would obviously be the most straightforward and easy solution in that case.

For the context, I'm trying to implement https://github.com/niri-wm/niri/pull/2240 via an IPC script (and I need to make sure that I'm not toggling `MaximizeColumn` on a window / column that is already maximized).

Thanks again for the work! :pray:

PR #1984, opened 2025-07-08
Support Bluetooth media controls in default config

dev-nicolaos:
This PR adds a line to the default config to conditionally auto-start `mpris-proxy`. Running mpris-proxy allows Bluetooth devices such as headphones to [control media playback](https://wiki.archlinux.org/title/MPRIS#Bluetooth).

The motivation here is to reduce the amount of concepts a new user has to learn and manual setup they have to perform in order for their hardware to work as they expect.

I've tested this command on my machine and its working nicely. I've tried to be defensive enough in this script so that it will only run the command in a circumstance where it makes sense to do so as described in the comment. It does depend on systemd, but should fail silently without causing harm if systemd is not available.

All that said:

- If there's a better way to solve the problem of Bluetooth media controls not working by default rather than a `spawn-at-startup` in the default config, I'm all for it.
- If this feels like too specific a circumstance/too far outside Niri's scope to support, then I'll understand if this PR gets closed. At least this PR's existence should provide some search-ability for users trying to troubleshoot.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -266,6 +266,10 @@ layout {
 // This line starts waybar, a commonly used bar for Wayland compositors.
 spawn-at-startup "waybar"
 
+// This line starts mpris-proxy in order to handle bluetooth media controls if
+// 1) bluetooth is available, 2) mpris-proxy is installed, and 3) mpris-proxy is not already running.
+spawn-at-startup "bash" "-c" "which systemctl && systemctl -q is-active bluetooth && which mpris-proxy && ps -d | grep -q mpris-proxy || mpris-proxy"
+
 // Uncomment this line to ask the clients to omit their client-side decorations if possible.
 // If the client will specifically ask for CSD, the request will be honored.
 // Additionally, clients will be informed that they are tiled, removing some client-side rounded corners.


NickHastings:
The command you are using is systemd specific, so I think mpris-proxy should be started (automatically) by systemd using the user unit provided by bluez.

YaLTeR:
Hm, this does indeed feel fairly specific for a niri default config. I haven't seen something like this in other compositors' configs either. Sounds kinda more fitting for a distro spin integration.

How does it work on GNOME btw? Does something spawn mpris-proxy (if so, then what does)?

NickHastings:
> How does it work on GNOME btw? Does something spawn mpris-proxy (if so, then what does)?

Systemd with `After=dbus.socket dbus.service` and `WantedBy=default.target`
https://github.com/bluez/bluez/blob/master/tools/mpris-proxy.service.in

YaLTeR:
Sounds like this should work on niri too?

NickHastings:
> Sounds like this should work on niri too?

It does for me.

dev-nicolaos:
> Sounds like this should work on niri too?

Are you saying that bluetooth media controls should already be working in Niri without the user having to start `mpris-proxy`? Or that the approach Gnome uses could also be implemented in Niri?

NickHastings:
> > Sounds like this should work on niri too?
> 
> Are you saying that bluetooth media controls should already be working in Niri without the user having to start `mpris-proxy`?

That's what I'm saying.

dev-nicolaos:
If I remove the line in this PR from my config, bluetooth media controls do not work. I'm running Fedora 42 with the latest Niri release from COPR.

NickHastings:
> If I remove the line in this PR from my config, bluetooth media controls do not work. I'm running Fedora 42 with the latest Niri release from COPR.

Did you check the state of the systmed unit? Eg
```
systemctl --user status mpris-proxy
```

NickHastings:
IMHO, this PR should be closed and if you can't work out why mpris-proxy is not being automatically started, ask for help on Matrix or Discord.

dev-nicolaos:
Running `systemctl --user status mpris-proxy` yields

```bash
mpris-proxy.service - Bluetooth mpris proxy
     Loaded: loaded (/usr/lib/systemd/user/mpris-proxy.service; dis>
    Drop-In: /usr/lib/systemd/user/service.d
             └─10-timeout-abort.conf
     Active: inactive (dead)
       Docs: man:mpris-proxy(1)
```

> IMHO, this PR should be closed

That's fine with me. I mentioned in the original PR description that this may not be the best place for this.

> if you can't work out why mpris-proxy is not being automatically started, ask for help on Matrix or Discord.

For future reference, is the situation that Niri already includes code to start the service, or that that responsibility belongs to the distro?

nekopsykose:
> Loaded: loaded (/usr/lib/systemd/user/mpris-proxy.service; dis>
> Active: inactive (dead)  

this says that it's not enabled (e.g. `systemctl enable mpris-proxy`) and also wasn't started. you don't have to manually start it with spawn-at-startup every time, only enable it once.

i don't think gnome starts or enables this automatically either?

>> How does it work on GNOME btw? Does something spawn mpris-proxy (if so, then what does)?  

> Systemd with After=dbus.socket dbus.service and WantedBy=default.target

that is just a regular systemd unit like any other, and it doesn't say anything about what starts it for you (what this issue is about). it doesn't point to what enables it by default or similar.

NickHastings:
I can't say why the service is not enabled on OPs system. FYI on Debian it is enabled by the bluez package post install script. I don't use Fedora and don't plan to dig into the details of the way they package bluez. That's up to OP or other interested parties to determine, or just enable it and move on. Not a niri problem IMHO.

nekopsykose:
fedora seems to also enable it by default since some version of 41 https://bugzilla.redhat.com/show_bug.cgi?id=2323365 , only for new installs probably.

but yeah, not a niri issue :)

PR #1985, opened 2025-07-09
Add FileChooser portal endpoint to gtk exception list

itsjunetime:
I was trying to build something with [`rfd`](https://github.com/PolyMeilex/rfd) recently and couldn't get it to actually show the file chooser dialog in niri until I added this manual exception to the portals list.

The Gnome portal claims that it supports this endpoint, but I get 'this name is not activatable' errors in journalctl whenever the system tries to use Gnome for this endpoint, so I think the Gnome portal is just kinda lying.

I also guess I can't guarantee this is something that everyone needs or whatever, but it seems like the pretty clear and simple fix to what is probably also a few other people's problem.

### DIFF
--- a/resources/niri-portals.conf
+++ b/resources/niri-portals.conf
@@ -2,4 +2,5 @@
 default=gnome;gtk;
 org.freedesktop.impl.portal.Access=gtk;
 org.freedesktop.impl.portal.Notification=gtk;
+org.freedesktop.impl.portal.FileChooser=gtk;
 org.freedesktop.impl.portal.Secret=gnome-keyring;


Brisingr05:
The GNOME portal uses Nautilus as the file picker, so it needs to be installed.

itsjunetime:
Ah, gotcha. Well, regardless, I think we should still prefer the gtk portal over the gnome one in this case, since it works without an external dependency.

Stzx:
If the default value have been changed, it is recommended that the Wiki be updated.

https://github.com/YaLTeR/niri/wiki/Important-Software

itsjunetime:
Ah, I didn't notice that part in the wiki - I had read a different part about the portals and not seen this there. I guess since it's already documented and known, I'll just close this.

PR #2009, opened 2025-07-11
Add nushell completion support

artrixdotdev:
Adds `clap_complete_nushell` crate and implements it into the `niri completions` command.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,8 +2,7 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 
 use clap::{Parser, Subcommand};
-use clap_complete::{Generator, Shell};
-use clap_complete_nushell::Nushell;
+use clap_complete::Shell;
 use niri_ipc::{Action, OutputAction};
 
 use crate::utils::version;
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2,7 +2,8 @@ use std::ffi::OsString;
 use std::path::PathBuf;
 
 use clap::{Parser, Subcommand};
-use clap_complete::Shell;
+use clap_complete::{Generator, Shell};
+use clap_complete_nushell::Nushell;
 use niri_ipc::{Action, OutputAction};
 
 use crate::utils::version;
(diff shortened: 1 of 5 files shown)


src/cli.rs
```diff
@@ -56,7 +79,9 @@ pub enum Sub {
     /// Cause a panic to check if the backtraces are good.
     Panic,
     /// Generate shell completions.
-    Completions { shell: Shell },
+    Completions {
+        shell: CompletionShell, // Changed from Shell to CompletionShell
```
YaLTeR: you forgot to clean this up after the LLM

src/cli.rs
```diff
@@ -7,6 +7,29 @@ use niri_ipc::{Action, OutputAction};
...
+impl From<CompletionShell> for Shell {
+    fn from(shell: CompletionShell) -> Self {
+        match shell {
+            CompletionShell::Bash => Shell::Bash,
+            CompletionShell::Elvish => Shell::Elvish,
+            CompletionShell::Fish => Shell::Fish,
+            CompletionShell::PowerShell => Shell::PowerShell,
+            CompletionShell::Zsh => Shell::Zsh,
+            CompletionShell::Nushell => panic!("Nushell should be handled separately"),
+        }
```
YaLTeR: Is [the relevant trait from clap_complete](https://docs.rs/clap_complete/latest/clap_complete/aot/trait.Generator.html) dyn-compatible? If so, I think it would be cleaner to remove this From, instead inline the match into the code below, returning a `&dyn Shell::Bash` and so on, as well as `&dyn Nushell`, which would all fit into a single variable of type `&dyn Generator`. Then there would be only one call to `clap_complete::generate()`.

src/cli.rs
```diff
@@ -7,6 +7,31 @@ use niri_ipc::{Action, OutputAction};
...
+        match shell {
+            CompletionShell::Bash => Ok(Shell::Bash),
+            CompletionShell::Elvish => Ok(Shell::Elvish),
+            CompletionShell::Fish => Ok(Shell::Fish),
+            CompletionShell::PowerShell => Ok(Shell::PowerShell),
+            CompletionShell::Zsh => Ok(Shell::Zsh),
+            CompletionShell::Nushell => Err("Nushell should be handled separately"),
+        }
+    }
+}
```
YaLTeR: One last thing, could you move this to below other types in this file?

artrixdotdev: Done!

YaLTeR:
Thanks

PR #2014, opened 2025-07-12
Detect external session management

Vladimir-csp:
This should make `uwsm start niri.desktop` possible like with other compositors. Following discussion in #254.

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -1,5 +1,15 @@
 #!/bin/sh
 
+# Detect if being run as a user service, which implies external session management,
+# exec compositor directly
+if [ -n "${MANAGERPID:-}" ] && [ "${SYSTEMD_EXEC_PID:-}" = "$$" ]; then
+    case "$(ps -p "$MANAGERPID" -o cmd=)" in
+    *systemd*--user*)
+        exec niri --session
+        ;;
+    esac
+fi
+
 if [ -n "$SHELL" ] &&
    grep -q "$SHELL" /etc/shells &&
    ! (echo "$SHELL" | grep -q "false") &&


YaLTeR:
Does any other compositor's systemd integration do this? I don't think I've seen this elsewhere.

Vladimir-csp:
Other compositor's desktop entries just launch compositors themselves ([sway](https://github.com/swaywm/sway/edit/master/sway.desktop), [labwc](https://github.com/labwc/labwc/edit/master/data/labwc.desktop), [wayfire](https://github.com/WayfireWM/wayfire/edit/master/wayfire.desktop)), except Hyprland's which uses `uwsm` specifically in the [default entry](https://github.com/hyprwm/Hyprland/blob/main/systemd/hyprland-uwsm.desktop), but also has a plain compositor launch in an [example](https://github.com/hyprwm/Hyprland/blob/main/example/hyprland.desktop).

This change should not affect anything, since the only way those conditions are met is when the script is launched as a main exec of the user unit. And this should only happen if something external manages the compositor, which means that internal session management is not needed.

YaLTeR:
> This change should not affect anything, since the only way those conditions are met is when the script is launched as a main exec of the user unit. And this should only happen if something external manages the compositor, which means that internal session management is not needed.

Display managers and greeters don't usually do this? (Idk what they usually do, just there's a lot of weird issues around some display managers)

Vladimir-csp:
> Display managers and greeters don't usually do this? (Idk what they usually do, just there's a lot of weird issues around some display managers)

In theory, they start what is in the `Exec` key. In practice, a bit more than that. For example sddm does shell/shell profile wrapping, and I suppose other DMs except GDM also do that.

But that is not what this change is about. It's about skipping own unit/session management and starting compositor right away when definitely being launched in a unit.

YaLTeR:
Well let's try I guess

PR #2026, opened 2025-07-13
handle `SIGINT`, `SIGTERM`, `SIGHUP`

sodiboo:
Currently, if you start `niri` from the command line and then press  `Ctrl+C`, or it is `kill`-ed, niri will not exit gracefully and it will leave some garbage runtime files (namely, `$NIRI_SOCKET` is not cleaned up). This leads to quite the mess if you're repeatedly starting and stopping niri to test things.

too many niri IPC sockets

This PR adds a simple signal handler using the `signals` feature of `calloop`. No new transitive dependencies are added to niri, but `calloop` now depends on `nix`.

---

I also changed the import of `smithay::reexports::calloop::EventLoop` to one of `calloop::EventLoop` in `main.rs`. This way, the imports don't have two separate instances of `calloop`. We already reference toplevel `calloop` by name in the file anyways; there is no good reason to use the smithay reexport.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::{env, mem};
 
+use calloop::signals::{Signal, Signals};
 use calloop::EventLoop;
 use clap::{CommandFactory, Parser};
 use directories::ProjectDirs;
@@ -172,8 +173,21 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     store_and_increase_nofile_rlimit();
 
+    // Create the main event loop.
+    let mut event_loop = EventLoop::<State>::try_new().unwrap();
+
+    // Handle Ctrl+C and other signals.
+    event_loop
+        .handle()
+        .insert_source(
+            Signals::new(&[Signal::SIGINT, Signal::SIGTERM, Signal::SIGHUP]).unwrap(),
+            |_, _, state| {
+                state.niri.stop_signal.stop();
+            },
+        )
+        .unwrap();
+
     // Create the compositor.
-    let mut event_loop = EventLoop::try_new().unwrap();
     let display = Display::new().unwrap();
     let mut state = State::new(
         config,
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,6 +9,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::{env, mem};
 
+use calloop::EventLoop;
 use clap::{CommandFactory, Parser};
 use directories::ProjectDirs;
 use niri::cli::{Cli, Sub};
@@ -26,7 +27,6 @@ use niri_config::Config;
 use niri_ipc::socket::SOCKET_PATH_ENV;
 use portable_atomic::Ordering;
 use sd_notify::NotifyState;
-use smithay::reexports::calloop::EventLoop;
 use smithay::reexports::wayland_server::Display;
 use tracing_subscriber::EnvFilter;
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -494,6 +494,7 @@ dependencies = [
  "async-task",
  "bitflags 2.9.1",
  "futures-io",
+ "nix 0.29.0",
  "polling",
  "rustix 0.38.44",
  "slab",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ async-io = { version = "2.4.1", optional = true }
 atomic = "0.6.1"
 bitflags.workspace = true
 bytemuck = { version = "1.23.1", features = ["derive"] }
-calloop = { version = "0.14.2", features = ["executor", "futures-io"] }
+calloop = { version = "0.14.2", features = ["executor", "futures-io", "signals"] }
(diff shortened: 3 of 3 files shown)


src/main.rs
```diff
@@ -172,8 +173,21 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     store_and_increase_nofile_rlimit();
 
+    // Create the main event loop.
+    let mut event_loop = EventLoop::<State>::try_new().unwrap();
+
+    // Handle Ctrl+C and other signals.
+    event_loop
+        .handle()
+        .insert_source(
+            Signals::new(&[Signal::SIGINT, Signal::SIGTERM, Signal::SIGHUP]).unwrap(),
```
YaLTeR, line 183: Isn't SIGHUP usually "reload config"? Or that's the only one sent if you close a terminal window running "niri" and therefore needs to clean up too?

sway seems to only handle SIGINT and SIGTERM: https://github.com/swaywm/sway/blob/f57c82a6f7cbd0c6499def867fc98459fb72a1d5/sway/main.c#L157

Other than this question, looks good

sodiboo: > Isn't SIGHUP usually "reload config"? Or that's the only one sent if you close a terminal window running "niri" and therefore needs to clean up too?

"reload config" is typically implemented using a user-defined signal: `pkill -SIGUSR2 waybar` to reload the waybar config.

---

I used these three signals in particular, because i initially implemented it in terms of `ctrlc` [as seen in `wpaperd`, another calloop-based application](https://github.com/danyspin97/wpaperd/commit/7865433233ff58f33ca4af94fd5031e26f409f52) (the only code result i could easily dig up when searching `"calloop" "sigint"`); and [the `ctrlc` crate implements those three](https://github.com/Detegr/rust-ctrlc/blob/ac79af3262199bce508257771b61608df4db22c4/src/platform/unix/mod.rs#L113-L150), but then i realized that `calloop` actually exposes a signal handler (which is better anyways because handling it outside of the event loop has a race condition of ignoring signals received before `EventLoop::run()` as it will *reset* the stop signal), and i preserved the set of signals to catch.

---

[Referencing Wikipedia](https://en.wikipedia.org/wiki/Signal_(IPC)#POSIX_signals), these are the signals that normally result in a regular process termination (without core dump):

- `SIGALRM`
- `SIGHUP`
- `SIGINT`
- `SIGKILL`
- `SIGPIPE`
- `SIGTERM`
- `SIGUSR1`
- `SIGUSR2`
- `SIGVTALRM`

Of these:
- `SIGALRM`/`SIGVTALRM` are related to libc timer functions we don't use
- `SIGUSR1`/`SIGUSR2` are not used by niri so should not be handled here
- `SIGPIPE` relates to a fault condition when performing I/O (and should not result in a "regular graceful shutdown")
- `SIGKILL` cannot be handled

leaving only `SIGINT`, `SIGTERM`, `SIGHUP` as termination signals we should handle.

YaLTeR: Fair enough. Thanks for the links

YaLTeR:
Thanks

YaLTeR:
So, it doesn't seem to actually ever trigger for me. Not on Ctrl+C, not on closing the terminal. I added this:
```
            |event, _, state| {
                info!("quitting due to receiving signal {:?}", event.signal());
```
and it doesn't print on Ctrl+C.

sodiboo:
> So, it doesn't seem to actually ever trigger for me. Not on Ctrl+C, not on closing the terminal. I added this:
> 
> ```
>             |event, _, state| {
>                 info!("quitting due to receiving signal {:?}", event.signal());
> ```
> 
> and it doesn't print on Ctrl+C.

how odd..? it Works On My Machine™
image

YaLTeR:
Okay, so apparently it does work, unless I build with Tracy integration. Idk what's going on there, probably it sets its own handler or something?

shellhazard:
Just bumped `niri-flake` to use the latest `niri-unstable` and `Ctrl+C` no longer works for me to cut to a new line in terminals running within niri. Confirmed it's an issue with both `alacritty` and `ghostty`, but is _not_ an issue with `kitty`. Reverted to stable and no issues - commenting here as I suspect this change is related.

```
Added packages:
[A+]  #1  niri-unstable  2025-07-15-007d355, 2025-07-15-007d355-doc
```

```json
{
  "lastModified": 1752569955,
  "narHash": "sha256-a21pjNhJYJ+OTQmBJ3NluU65PvMb54/mA7aEWJh5s/4=",
  "owner": "sodiboo",
  "repo": "niri-flake",
  "rev": "8fc18813bf6ceaabb3063050819a20807e11279b",
  "type": "github"
}
```

YaLTeR:
Appears to work here on Fedora in alacritty

YaLTeR:
Found this in the calloop Signals docs (who would've thought):
```
//! The source will take care of masking and unmasking signals for the thread it runs on,
//! but you are responsible for masking them on other threads if you run them. The simplest
//! way to ensure that is to setup the signal event source before spawning any thread, as
//! they'll inherit their parent signal mask.
```
Tracy does in fact spawn threads. So perhaps creating Signals earlier will make the Ctrl+C override work even with Tracy?

PR #2027, opened 2025-07-13
Simplify pointer handling in constraint check

abmantis:
Minor change so that `get_pointer()` (which has a lock) does not get called twice. Also moved the call to `current_location()` to the scope where it is needed.

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -5769,18 +5769,16 @@ impl Niri {
     ///
     /// Make sure the pointer location and contents are up to date before calling this.
     pub fn maybe_activate_pointer_constraint(&self) {
-        let pointer = self.seat.get_pointer().unwrap();
-        let pointer_pos = pointer.current_location();
-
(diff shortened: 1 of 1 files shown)


YaLTeR:
Seems reasonable; did you test it with some client that has a constraint with a region? E.g. https://github.com/Smithay/client-toolkit/blob/master/examples/relative_pointer.rs

abmantis:
Closed by mistake.
I tested it with a browser game that grabs the mouse, but I'll test with that example too.

abmantis:
> Seems reasonable; did you test it with some client that has a constraint with a region? E.g. https://github.com/Smithay/client-toolkit/blob/master/examples/relative_pointer.rs

Tested just now with that example, and all modes seem to work as expected. 
I had to comment out the text rendering because it was panicking due to null fonts (not sure why, but didn't seem too relevant for the test anyway).

YaLTeR:
Thanks

PR #2032, opened 2025-07-13
nix: update flake inputs, drop nix-filter

bloxx12:
- Drop nix-filter and instead opts for nixpkgs' `lib.fileSet`
- Update flake inputs

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,6 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
-    nix-filter.url = "github:numtide/nix-filter";
 
     # NOTE: This is not necessary for end users
     # You can omit it with `inputs.rust-overlay.follows = ""`
@@ -18,7 +17,6 @@
     {
       self,
       nixpkgs,
-      nix-filter,
       rust-overlay,
     }:
     let
@@ -50,16 +48,16 @@
           pname = "niri";
           version = self.shortRev or self.dirtyShortRev or "unknown";
 
-          src = nix-filter.lib.filter {
-            root = self;
-            include = [
-              "niri-config"
-              "niri-ipc"
-              "niri-visual-tests"
-              "resources"
-              "src"
-              ./Cargo.lock
+          src = lib.fileset.toSource {
+            root = ./.;
+            fileset = lib.fileset.unions [
+              ./niri-config
+              ./niri-ipc
+              ./niri-visual-tests
+              ./resources
+              ./src
               ./Cargo.toml
+              ./Cargo.lock
             ];
           };
 
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1742707865,
-        "narHash": "sha256-RVQQZy38O3Zb8yoRJhuFgWo/iDIDj0hEdRTVfhOtzRk=",
+        "lastModified": 1752077645,
+        "narHash": "sha256-HM791ZQtXV93xtCY+ZxG1REzhQenSQO020cu6rHtAPk=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "dd613136ee91f67e5dba3f3f41ac99ae89c5406b",
+        "rev": "be9e214982e20b8310878ac2baa063a961c1bdf6",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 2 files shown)


YaLTeR:
cc @sodiboo

flake.nix
```diff
@@ -1,10 +1,8 @@
-# This flake file is community maintained
```
sodiboo: this comment is supposed to mean that YaLTeR does not maintain this file or necessarily test it locally; it may break at times and it's up to us to fix it.

it maybe isn't the clearest in intent? it could probably be improved? but I don't think it's supposed to be "unnecessary" like that commit suggests. and probably shouldn't be removed without further thought.

if YaLTeR is fine with this omission then it's whatever; doesn't really bother me, it can go for all i care.

so, @YaLTeR thoughts?

flake.nix
```diff
@@ -50,18 +47,24 @@
...
+                  ./niri-config
+                  ./niri-ipc
+                  ./niri-visual-tests
+                  ./resources
+                  ./src
+                  ./Cargo.toml
+                  ./Cargo.lock
+                ]
+              );
+            };
```
sodiboo: what's with the `lib.sources.cleanSource`? this removes VCS noise; but we're in a git flake; `./.` is already "cleaned" and we actually *can't* even access the files that normally would have been cleaned by that function. it does *nothing*.


```suggestion
          src = lib.fileset.toSource {
            root = ./.;
            fileset = lib.fileset.unions [
              ./niri-config
              ./niri-ipc
              ./niri-visual-tests
              ./resources
              ./src
              ./Cargo.toml
              ./Cargo.lock
            ];
          };
```

both the original `nix-filter`, the version intersecting `cleanSource`, and this one with just `unions`, produce the *exact same store path*.

flake.nix
YaLTeR: > this comment is supposed to mean that YaLTeR does not maintain this file or necessarily test it locally; it may break at times and it's up to us to fix it.

Yes, this is exactly the intention. I think the comment signals it fine enough

sodiboo: yeah. so it should stay, right?

YaLTeR: Yeah let's keep it

sodiboo (approved):
ya it seems good now

faukah:
Added you as a co-author, I hope that's fine.

YaLTeR:
The nix check in the github action thinks that it failed?

faukah:
Huh, that is indeeed weird... Maybe @sodiboo has an Idea, I'm pretty clueless.

sodiboo:
wtf. how dare it fail

sodiboo:
sometimes Nix builds fail erratically in CI if the CI version is too old.

the Nix version in CI is like, *old*. we have `DeterminateSystems/nix-installer-action@v3` but the latest is `v18`.

how has this happened??? [dependabot is supposed to keep that up to date](https://github.com/YaLTeR/niri/blob/56901eed5dbb51d9d669e8ea5a53868896ddeba4/.github/dependabot.yml#L17-L20) but it has seemingly [never made a CI-related PR](https://github.com/YaLTeR/niri/pulls?q=author%3Aapp%2Fdependabot+-label%3Arust+-serde)...

my hypothesis is that this is the cause of that build failure

faukah:
> sometimes Nix builds fail erratically in CI if the CI version is too old.
> 
> the Nix version in CI is like, _old_. we have `DeterminateSystems/nix-installer-action@v3` but the latest is `v18`.
> 
> how has this happened??? [dependabot is supposed to keep that up to date](https://github.com/YaLTeR/niri/blob/56901eed5dbb51d9d669e8ea5a53868896ddeba4/.github/dependabot.yml#L17-L20) but it has seemingly [never made a CI-related PR](https://github.com/YaLTeR/niri/pulls?q=author%3Aapp%2Fdependabot+-label%3Arust+-serde)...
> 
> my hypothesis is that this is the cause of that build failure

*lol*

YaLTeR:
Dependabot is in some half broken state on the niri repo, I'm surprised it even works for Cargo. I tried all kinds of toggling it back and forth and nothing helped, and my issue in the Dependabot repo was ignored

terlar:
Perhaps switching to renovate could be an option? In my experience it works much better and is much more flexible.

sodiboo:
> Perhaps switching to renovate could be an option? In my experience it works much better and is much more flexible.

i would have rather prescribed one dose of just manually updating the action, and dealing with the dependency bots at a later date.

But it seems the problem actually has nothing to do with the outdated action; it's still working, but our builds are failing because [someone added a nonexistent nushell completion to the nix package](https://github.com/YaLTeR/niri/issues/2047).

sodiboo:
Baffled at how i didn't discover this locally, because the package *is* indeed failing now on my machine. Earlier, i tried to build it and it worked fine?? That's why i assumed it was CI related (since i've had similar issues in e.g. niri-flake with spurious build failures caused by outdated Nix)

YaLTeR:
Seems good now, thanks!

PR #2036, opened 2025-07-14
Add CONTRIBUTING.md

YaLTeR:


CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+### Reviewing
+
+Reviewing pull requests is something I need the most help with since there are a lot of them, and it's quite time-consuming.
+Anyone with code accepted into niri is welcome, but this is not a requirement; even if you aren't familiar with Rust you may find some logic problems.
+
+Pick a pull request, then review its code.
+
+- Check that everything looks good, check various conditions for edge cases.
+- See if there are any scenarios the author forgot to handle.
+- Check that the code fits well into the rest of niri, follows its design and code style.
```
titaniumtraveler, line 61: I feel it might be helpful to expand on this.
I often feel you have a very specific vision for `niri`, but it feels sometimes relatively hard to deduce that from the existing code.

Maybe something equivalent to https://github.com/YaLTeR/niri/wiki/Development:-Design-Principles, but in terms of code.

Like when to use functions vs methods, how and when to break them up, what kind of complexity is acceptable. What and where do you want to maintain compatibility guaranties etc.

For some modules like the hotkey-overlay it might be also useful to have an short ~2-3 paragraph explanation of code structure, purpose and implementation. Though honestly that could maybe something that people can contribute and you just check whether the description is correct. (As a matter of optimizing time usage and all that)

CONTRIBUTING.md
```diff
@@ -0,0 +1,87 @@
...
+## Writing pull requests
+
+When creating pull requests, please keep the following in mind.
+
+- Make sure new features align with niri's design directions. Ideally, there should be an existing issue or discussion where we settled on that solution.
+- Keep pull requests focused on a single feature or bug fix with no unrelated changes.
+- Try to split your changes into small, self-contained commits. Every commit should build and pass tests. This makes it much easier to review your PR, and bisect for regressions in the future.
+    - When addressing PR comments, try to squash the changes straight into the relevant commits.
+    - When working on bigger features, I usually start with a big messy commit, then gradually split out smaller self-contained changes from it as the code gets into shape.
+    - [git-rebase.io](https://git-rebase.io/) is a helpful guide for splitting commits and cleaning up history in git.
```
titaniumtraveler: This section specifically is extremely helpful :slightly_smiling_face:
When contributing to a new project it is often hard to ascertain the expectations on pull-requests so this very much is making that easier.

titaniumtraveler (approved):
All in all a very helpful `CONTRIBUTING.md` and I will be very happy to point people at it in the future in case they need some additional context how something should work.

CONTRIBUTING.md
YaLTeR: I agree that it's vague, but I'm not sure I can quite form it into words myself. The main idea here is to look at the surrounding code for whatever you're changing, get the vibe, and try to follow it (which is just a general good advice for any project).

There are certainly some module-specific rules (like, layout actions should be in the randomized tests); would be good to write either module comments or a wiki page for those at some point.

CONTRIBUTING.md
YaLTeR: I see. This is more or less my ideal pull request; I understand that rebasing is weird and complicated and many people won't do it properly, hence me using Squash and Merge a lot. But yeah, hopefully more people start doing this

CONTRIBUTING.md
titaniumtraveler: Ye. Fair. Maybe include module documentation in the list of possible contributions?

I would prefer to have the module documentation in-line. Less likely to get out of sync that way. (Also easier to find)

At most have the wiki link to the rustdoc.

CONTRIBUTING.md
titaniumtraveler: That is very good to know explicitly.

As you might remember with the `niri validate --json` PR I was unsure if I should rather rebase now, or should just `--fixup` and automerge later etc.
Force-pushing might be annoying to deal with to other people, so it kinda depends on the project.

Really happy it is going to clearly documented now :green_heart:

HigherOrderLogic: I also want to add my opinion on rebasing and force-pushing a PR. While it may produce a clean PR that is easy to manage and merge later on, some review and comment on PR files maybe marked as outdated on GitHub, even though the changes was not made by the PR, but by rebasing on other commits. Thus, maybe adding some tips about when to rebase might be helpful too.

YaLTeR: I went more in-depth on rebasing and force pushing

YaLTeR:
Thanks for feedback

Issue #1904, opened 2025-06-25
Investigate screen reader support


YaLTeR:
Disclaimer: I have very little idea about how any of this works. I'm testing on Fedora 42 Silverblue.

As of now, apparently, just running `orca` is enough to get in-app screen reading and, for some toolkits (GTK 3), key announce and Orca shortcuts. However, there are a number of shortcomings that I noticed.

1. Apps that don't support the screen reader (e.g. Alacritty) don't say anything. In GNOME, they do announce keys. (But not always for some reason?)
2. Key announce and Orca shortcuts don't work in GTK 4. In GNOME, they do.
3. Orca doesn't override shortcuts properly, e.g. pressing Insert on niri will trigger the Insert key, even in GTK 3 where the Orca shortcuts otherwise work. In GNOME, it correctly triggers Insert only on double-press, regardless of the target app.
4. niri itself doesn't announce anything to Orca. Things like:
    - dialog contents, i.e. the exit confirmation dialog
    - built-in notifications, i.e. the config validation fail notification
    - workspace and monitor focus changes
    - window titles in the overview

    Can we use [AccessKit](https://accesskit.dev/) for exposing this maybe?
5. The default config should provide the standard shortcut for toggling Orca: <kbd>Super</kbd><kbd>Alt</kbd><kbd>S</kbd>. Need to find what handles this in GNOME, especially how to turn it off properly (assuming turning on is just spawning `orca`, not sure how to turn off properly with a "Screen reader off" announcement).
    - Or is this handled elsewhere? How does GNOME remember whether Orca was on or off across restarts? Maybe there's some Orca daemon running that listens for this shortcut on its own, and remembers its own state?
6. I've also heard that users rely on an audio cue for "the desktop has finished loading and you can now turn on the screen reader".
    - I guess for niri this can be handled by distros adding a spawn-at-startup line playing an audio cue.

DataTriny:
Hello @YaLTeR, AccessKit dev here.

> Key announce and Orca shortcuts don't work in GTK 4. In GNOME, they do.

[...]

PR #2042, opened 2025-07-14
Basic AccessKit support for exit confirmation dialog

DataTriny:
Part of #1904 

Start exposing niri's UI to assistive technologies. This is gated behind the more generic `accessibility` feature (enabled by default and tied to the `dbus` feature since the AT-SPI protocol uses it). No bounds information areprovided for now because it still isn't clear how to do this under Wayland.

This can be tested by starting the Orca screen reader inside a niri session and then pressing `Super`+`Shift`+`E`. Orca will announce the alert and its content. One big issue though is that these dialogs seem to not steal the focus to other apps, so Orca won't notify the user if they are dismissed. Orca users expect to hear which widget had the focus in the app before the modal dialog showed up.

### DIFF
--- a/src/ui/exit_confirm_dialog.rs
+++ b/src/ui/exit_confirm_dialog.rs
@@ -15,8 +15,7 @@ use crate::render_helpers::renderer::NiriRenderer;
 use crate::render_helpers::texture::{TextureBuffer, TextureRenderElement};
 use crate::utils::{output_size, to_physical_precise_round};
 
-const TEXT: &str = "Are you sure you want to exit niri?\n\n\
-                    Press <span face='mono' bgcolor='#2C2C2C'> Enter </span> to confirm.";
+const KEY_NAME: &str = "Enter";
 const PADDING: i32 = 16;
 const FONT: &str = "sans 14px";
 const BORDER: i32 = 8;
@@ -24,6 +23,8 @@ const BORDER: i32 = 8;
 pub struct ExitConfirmDialog {
     is_open: bool,
     buffers: RefCell<HashMap<NotNan<f64>, Option<MemoryBuffer>>>,
+    #[cfg(feature = "dbus")]
+    accesskit_adapter: Option<accesskit_unix::Adapter>,
 }
 
 impl ExitConfirmDialog {
@@ -34,12 +35,26 @@ impl ExitConfirmDialog {
                 NotNan::new(1.).unwrap(),
                 Some(render(1.)?),
             )])),
+            #[cfg(feature = "dbus")]
+            accesskit_adapter: None,
         })
     }
 
     pub fn show(&mut self) -> bool {
         if !self.is_open {
             self.is_open = true;
+
+            #[cfg(feature = "dbus")]
+            {
+                let mut adapter = accesskit_unix::Adapter::new(
+                    ActivationHandler::default(),
+                    super::a11y::ActionHandler::default(),
+                    super::a11y::DeactivationHandler::default(),
+                );
+                adapter.update_window_focus_state(true);
+                self.accesskit_adapter = Some(adapter);
+            }
+
             true
         } else {
             false
@@ -49,6 +64,12 @@ impl ExitConfirmDialog {
     pub fn hide(&mut self) -> bool {
         if self.is_open {
             self.is_open = false;
+
+            #[cfg(feature = "dbus")]
+            {
+                self.accesskit_adapter = None;
+            }
+
             true
         } else {
             false
@@ -106,13 +127,15 @@ fn render(scale: f64) -> anyhow::Result<MemoryBuffer> {
     let mut font = FontDescription::from_string(FONT);
     font.set_absolute_size(to_physical_precise_round(scale, font.size()));
 
+    let markup = text(true);
+
     let surface = ImageSurface::create(cairo::Format::ARgb32, 0, 0)?;
     let cr = cairo::Context::new(&surface)?;
     let layout = pangocairo::functions::create_layout(&cr);
     layout.context().set_round_glyph_positions(false);
     layout.set_font_description(Some(&font));
     layout.set_alignment(Alignment::Center);
-    layout.set_markup(TEXT);
+    layout.set_markup(&markup);
 
     let (mut width, mut height) = layout.pixel_size();
     width += padding * 2;
@@ -128,7 +151,7 @@ fn render(scale: f64) -> anyhow::Result<MemoryBuffer> {
     layout.context().set_round_glyph_positions(false);
     layout.set_font_description(Some(&font));
     layout.set_alignment(Alignment::Center);
-    layout.set_markup(TEXT);
+    layout.set_markup(&markup);
 
     cr.set_source_rgb(1., 1., 1.);
     pangocairo::functions::show_layout(&cr, &layout);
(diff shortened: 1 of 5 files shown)


src/ui/exit_confirm_dialog.rs
```diff
@@ -15,15 +15,16 @@ use crate::render_helpers::renderer::NiriRenderer;
...
+const KEY_NAME: &str = "Enter";
 const PADDING: i32 = 16;
 const FONT: &str = "sans 14px";
 const BORDER: i32 = 8;
 
 pub struct ExitConfirmDialog {
     is_open: bool,
     buffers: RefCell<HashMap<NotNan<f64>, Option<MemoryBuffer>>>,
+    #[cfg(feature = "accessibility")]
+    accesskit_adapter: Option<accesskit_unix::Adapter>,
```
YaLTeR, line 27: I guess this should ideally be long-lived on the full `Niri` state struct, as the global accessibility provider object?

src/ui/exit_confirm_dialog.rs
```diff
@@ -34,12 +35,26 @@ impl ExitConfirmDialog {
...
             self.is_open = true;
+
+            #[cfg(feature = "accessibility")]
+            {
+                let mut adapter = accesskit_unix::Adapter::new(
+                    ActivationHandler::default(),
+                    super::a11y::ActionHandler::default(),
+                    super::a11y::DeactivationHandler::default(),
+                );
+                adapter.update_window_focus_state(true);
```
YaLTeR, line 54: And I guess that this should be `true` only when some niri UI has keyboard focus, rather than a window or a layer surface?

DataTriny:
> Why do you think this is necessary? Is there some nontrivial overhead, or does it not build on musl or FreeBSD?

Due to how AT-SPI works, it brings non trivial overhead but only when an assistive technology is actually running. Do you think we should simply use the `dbus` feature instead?

> Perhaps make it have proper focus, but only when a screen reader is attached? Is that something we know?

[`ActivationHandler::initial_tree_requested`](https://docs.rs/accesskit/latest/accesskit/trait.ActivationHandler.html#tymethod.request_initial_tree) could be used to enable a global flag somewhere since it will only be called if an assistive technology is running. Similarely, the deactivation handler could be used to do the oppositve.

But I'd suggest avoiding special paths for assistive technology users if possible, as this is more stuff to test. The reason why I didn't include the config error UI here is because it is clear to me that it is a notification and that it will require a different accessibility tree. But the exit confirmation dialog clearly is modal since it forces the user to interact with it by pressing either Enter to accept or anything else to dismiss.

src/ui/exit_confirm_dialog.rs
DataTriny: Roughly speaking, one AccessKit unix adapter equals one top-level window. It sends events to assistive technologies letting them know that the window was closed when the adapter is dropped. When we'll have a main background window, it will make sense to store its adapter in a more central state.

src/ui/exit_confirm_dialog.rs
DataTriny: If I understand correctly, this is not an issue here because the focus should immediately be put on this dialog when it is open, and there is no way for something else to steal it since any action which is not pressing the Enter key will close the dialog. Am I right?

YaLTeR:
> Due to how AT-SPI works, it brings non trivial overhead but only when an assistive technology is actually running. Do you think we should simply use the `dbus` feature instead?

I think that's fine. Anyone who doesn't need a11y won't enable anything from what I understand, and anyone who needs a11y probably doesn't want to find out that their compositor has it compiled out.

> [`ActivationHandler::initial_tree_requested`](https://docs.rs/accesskit/latest/accesskit/trait.ActivationHandler.html#tymethod.request_initial_tree) could be used to enable a global flag somewhere since it will only be called if an assistive technology is running. Similarely, the deactivation handler could be used to do the oppositve.

I see, sounds good.

> But I'd suggest avoiding special paths for assistive technology users if possible, as this is more stuff to test.

That is a fair point yeah, I'll consider it when this is more in shape.

> The reason why I didn't include the config error UI here is because it is clear to me that it is a notification and that it will require a different accessibility tree.

What would be good documentation on how these different UI components should be represented in the accessibility tree? I'm guessing web ARIA stuff is documented well somewhere, but it probably doesn't include a lot of desktop-specific concepts like notifications?

src/ui/exit_confirm_dialog.rs
YaLTeR: That is correct; I'm thinking more about the future general case that handles several niri UI elements, rather than code specific to this one dialog.

DataTriny:
> I think that's fine. Anyone who doesn't need a11y won't enable anything from what I understand, and anyone who needs a11y probably doesn't want to find out that their compositor has it compiled out

Totally agree, I'll move everything to the `dbus` feature then. If taking away the keyboard focus isn't too much work, I can add it in here as well (just for the exit confirmation dialog I mean).

> What would be good documentation on how these different UI components should be represented in the accessibility tree?

Documentation is something we still need to work on with AccessKit. Right now I can't tell you how to properly implement notification. We'll have to figure it out, and niri can be a good test bed.

YaLTeR:
> If taking away the keyboard focus isn't too much work, I can add it in here as well (just for the exit confirmation dialog I mean).

I suspect it might be a bit involved. We don't have a UI toolkit in niri, so these things are kind of hand-coded in places. So it's better if I look into it (I don't think it should be difficult, just don't recall off the bat what exactly to change).

> Documentation is something we still need to work on with AccessKit. Right now I can't tell you how to properly implement notification. We'll have to figure it out, and niri can be a good test bed.

I see, alright!

YaLTeR:
Do I understand correctly that the reason why we need to take away keyboard focus from a window is because giving it focus *back* (after dismissing the dialog) is what causes Orca to re-announce the widget?

YaLTeR:
Try this patch for the focus:
```diff
diff --git i/src/handlers/xdg_shell.rs w/src/handlers/xdg_shell.rs
index a9eceb67..d10bd106 100644
--- i/src/handlers/xdg_shell.rs
+++ w/src/handlers/xdg_shell.rs
@@ -303,7 +303,16 @@ impl XdgShellHandler for State {
         // We need to hand out the grab in a way consistent with what update_keyboard_focus()
         // thinks the current focus is, otherwise it will desync and cause weird issues with
         // keyboard focus being at the wrong place.
-        if self.niri.is_locked() {
+        if self
+            .niri
+            .exit_confirm_dialog
+            .as_ref()
+            .is_some_and(|d| d.is_open())
+        {
+            trace!("ignoring popup grab because the exit confirm dialog is open");
+            let _ = PopupManager::dismiss_popup(&root, &popup);
+            return;
+        } else if self.niri.is_locked() {
             if Some(&root) != self.niri.lock_surface_focus().as_ref() {
                 trace!("ignoring popup grab because the session is locked");
                 let _ = PopupManager::dismiss_popup(&root, &popup);
diff --git i/src/niri.rs w/src/niri.rs
index 1d82ba6e..17596de5 100644
--- i/src/niri.rs
+++ w/src/niri.rs
@@ -509,6 +509,7 @@ pub enum KeyboardFocus {
     LayerShell { surface: WlSurface },
     LockScreen { surface: Option<WlSurface> },
     ScreenshotUi,
+    ExitConfirmDialog,
     Overview,
 }
 
@@ -606,6 +607,7 @@ impl KeyboardFocus {
             KeyboardFocus::LayerShell { surface } => Some(surface),
             KeyboardFocus::LockScreen { surface } => surface.as_ref(),
             KeyboardFocus::ScreenshotUi => None,
+            KeyboardFocus::ExitConfirmDialog => None,
             KeyboardFocus::Overview => None,
         }
     }
@@ -616,6 +618,7 @@ impl KeyboardFocus {
             KeyboardFocus::LayerShell { surface } => Some(surface),
             KeyboardFocus::LockScreen { surface } => surface,
             KeyboardFocus::ScreenshotUi => None,
+            KeyboardFocus::ExitConfirmDialog => None,
             KeyboardFocus::Overview => None,
         }
     }
@@ -1058,7 +1061,14 @@ impl State {
         }
 
         // Compute the current focus.
-        let focus = if self.niri.is_locked() {
+        let focus = if self
+            .niri
+            .exit_confirm_dialog
+            .as_ref()
+            .is_some_and(|d| d.is_open())
+        {
+            KeyboardFocus::ExitConfirmDialog
+        } else if self.niri.is_locked() {
             KeyboardFocus::LockScreen {
                 surface: self.niri.lock_surface_focus(),
             }
@@ -3918,6 +3928,7 @@ impl Niri {
             // layer-shell, the layout will briefly draw as active, despite never having focus.
             KeyboardFocus::LockScreen { .. } => true,
             KeyboardFocus::ScreenshotUi => true,
+            KeyboardFocus::ExitConfirmDialog => true,
             KeyboardFocus::Overview => true,
         };
 ```
One behavioral difference from this is that now the exit confirm dialog will close popups and some layer-shell launchers (they close upon losing focus), but I guess it's fine.

DataTriny:
Closing in favor of #2273.

PR #2044, opened 2025-07-14
implement ext-foreign-toplevel-list-v1 by augmenting our existing implementation of foreign toplevels

sodiboo:
- https://github.com/sodiboo/niri/commit/d2d51bcf0ceb6d47fe5b8985268cf0902a2e921d but brought back from the archives.
- duplicate of #1450 but taking a different approach

i originally didn't submit this as a PR, because i didn't realize there were any client apps that depended on *just toplevel list*: i thought this protocol was useless on its own. but apparently cosmic needs it?

Besides technical reasons why one approach is better than the other, this implementation has one clear concrete difference over #1450: the `identifier` of a foreign toplevel is stable and can be used to manipulate a toplevel with `niri msg`, they are the same identifier. It's not actually clear that this is a *desirable* trait: do we want to allow clients to parse this identifier? In #1450, the foreign toplevel identifier is random bytes.

### DIFF
--- a/src/protocols/foreign_toplevel.rs
+++ b/src/protocols/foreign_toplevel.rs
@@ -1,31 +1,37 @@
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
 use arrayvec::ArrayVec;
 use smithay::output::Output;
+use smithay::reexports::wayland_protocols::ext::foreign_toplevel_list::v1::server::{
+    ext_foreign_toplevel_handle_v1::{self, ExtForeignToplevelHandleV1}, ext_foreign_toplevel_list_v1::{self, ExtForeignToplevelListV1},
+};
 use smithay::reexports::wayland_protocols::xdg::shell::server::xdg_toplevel;
-use smithay::reexports::wayland_protocols_wlr;
+use smithay::reexports::wayland_protocols_wlr::foreign_toplevel::v1::server::{
+    zwlr_foreign_toplevel_handle_v1::{self, ZwlrForeignToplevelHandleV1}, zwlr_foreign_toplevel_manager_v1::{self, ZwlrForeignToplevelManagerV1},
+};
 use smithay::reexports::wayland_server::backend::ClientId;
 use smithay::reexports::wayland_server::protocol::wl_output::WlOutput;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
 use smithay::reexports::wayland_server::{
     Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource,
 };
-use smithay::wayland::shell::xdg::{ToplevelStateSet, XdgToplevelSurfaceRoleAttributes};
-use wayland_protocols_wlr::foreign_toplevel::v1::server::{
-    zwlr_foreign_toplevel_handle_v1, zwlr_foreign_toplevel_manager_v1,
+use smithay::wayland::shell::xdg::{
+    ToplevelStateSet, XdgToplevelSurfaceRoleAttributes,
 };
-use zwlr_foreign_toplevel_handle_v1::ZwlrForeignToplevelHandleV1;
-use zwlr_foreign_toplevel_manager_v1::ZwlrForeignToplevelManagerV1;
 
 use crate::niri::State;
+use crate::window::mapped::MappedId;
 use crate::utils::with_toplevel_role;
 
-const VERSION: u32 = 3;
+const EXT_VERSION: u32 = 1;
+const WLR_VERSION: u32 = 3;
 
 pub struct ForeignToplevelManagerState {
     display: DisplayHandle,
-    instances: Vec<ZwlrForeignToplevelManagerV1>,
+    ext_instances: Vec<ExtForeignToplevelListV1>,
+    wlr_instances: Vec<ZwlrForeignToplevelManagerV1>,
     toplevels: HashMap<WlSurface, ToplevelData>,
 }
 
@@ -38,33 +44,41 @@ pub trait ForeignToplevelHandler {
 }
 
 struct ToplevelData {
+    identifier: MappedId,
     title: Option<String>,
     app_id: Option<String>,
     states: ArrayVec<u32, 3>,
     output: Option<Output>,
-    instances: HashMap<ZwlrForeignToplevelHandleV1, Vec<WlOutput>>,
+
+    ext_instances: HashSet<ExtForeignToplevelHandleV1>,
+    wlr_instances: HashMap<ZwlrForeignToplevelHandleV1, Vec<WlOutput>>,
     // FIXME: parent.
 }
 
+#[derive(Clone)]
 pub struct ForeignToplevelGlobalData {
-    filter: Box<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
+    filter: Arc<dyn for<'c> Fn(&'c Client) -> bool + Send + Sync>,
 }
 
 impl ForeignToplevelManagerState {
     pub fn new<D, F>(display: &DisplayHandle, filter: F) -> Self
     where
         D: GlobalDispatch<ZwlrForeignToplevelManagerV1, ForeignToplevelGlobalData>,
+        D: GlobalDispatch<ExtForeignToplevelListV1, ForeignToplevelGlobalData>,
         D: Dispatch<ZwlrForeignToplevelManagerV1, ()>,
+        D: Dispatch<ExtForeignToplevelListV1, ()>,
         D: 'static,
         F: for<'c> Fn(&'c Client) -> bool + Send + Sync + 'static,
     {
         let global_data = ForeignToplevelGlobalData {
-            filter: Box::new(filter),
+            filter: Arc::new(filter),
         };
-        display.create_global::<D, ZwlrForeignToplevelManagerV1, _>(VERSION, global_data);
+        display.create_global::<D, ExtForeignToplevelListV1, _>(EXT_VERSION, global_data.clone());
+        display.create_global::<D, ZwlrForeignToplevelManagerV1, _>(WLR_VERSION, global_data);
         Self {
             display: display.clone(),
-            instances: Vec::new(),
+            ext_instances: Vec::new(),
+            wlr_instances: Vec::new(),
             toplevels: HashMap::new(),
         }
     }
@@ -81,7 +95,11 @@ pub fn refresh(state: &mut State) {
             return true;
         }
 
-        for instance in data.instances.keys() {
+        for instance in data.ext_instances.iter() {
+            instance.closed();
+        }
+
+        for instance in data.wlr_instances.keys() {
             instance.closed();
         }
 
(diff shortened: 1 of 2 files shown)


src/protocols/foreign_toplevel.rs
```diff
@@ -277,7 +349,111 @@ impl ToplevelData {
...
+    ) {
+        match request {
+            ext_foreign_toplevel_list_v1::Request::Stop => {
+                resource.finished();
+
+                let state = state.foreign_toplevel_manager_state();
+                state.ext_instances.retain(|x| x != resource);
+            }
+            ext_foreign_toplevel_list_v1::Request::Destroy => {}
+            _ => todo!(),
```
YaLTeR:
```suggestion
            _ => unreachable!(),
```

src/protocols/foreign_toplevel.rs
```diff
@@ -244,14 +291,39 @@ fn refresh_toplevel(
...
+    }
+    fn add_wlr_instance<D>(
         &mut self,
         handle: &DisplayHandle,
         client: &Client,
         manager: &ZwlrForeignToplevelManagerV1,
     ) where
-        D: Dispatch<ZwlrForeignToplevelHandleV1, ()>,
-        D: 'static,
+        D: Dispatch<ZwlrForeignToplevelHandleV1, ()> + 'static,
```
YaLTeR, line 327: These are on separate lines canonically to be able to add more `Dispatch`es easily

src/protocols/foreign_toplevel.rs
```diff
@@ -244,14 +291,39 @@ fn refresh_toplevel(
...
+        }
+        if let Some(app_id) = &self.app_id {
+            toplevel.app_id(app_id.clone());
+        }
+
+        toplevel.done();
+
+        self.ext_instances.insert(toplevel);
+    }
+    fn add_wlr_instance<D>(
```
YaLTeR, line 321: Blank lines between functions pls

YaLTeR (commented):
I think I also prefer this approach to the other one. Though, I must note that ext-foreign-toplevel-list won't necessarily grow toplevel management functionality (rather, separate protocols may be added). See:
- https://github.com/quickshell-mirror/shell-protocols/pull/1
- https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/156
- https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/196

That is to say, the naming here could be somewhat adjusted, in particular the `WLR_VERSION`-`EXT_VERSION`-like pairs, which imply it's ~the same protocol (like, say, wlr/ext data control), but it isn't. (I don't want the names to get too unwieldy, but I also don't want to rename everything in the future when one more of these protocols arrives.)

The code looks ok at a glance, but I didn't read through it in-depth, which I will still need to do before merging.

> Besides technical reasons why one approach is better than the other, this implementation has one clear concrete difference over #1450: the `identifier` of a foreign toplevel is stable and can be used to manipulate a toplevel with `niri msg`, they are the same identifier. It's not actually clear that this is a _desirable_ trait: do we want to allow clients to parse this identifier? In #1450, the foreign toplevel identifier is random bytes.

I don't have an opinion on this; while it's useful, there's nothing that says other compositors should also match their IPC identifiers with this, so it ends up being niri-specific either way.

src/window/mapped.rs
```diff
@@ -183,6 +183,27 @@ impl MappedId {
...
+    /// We use the decimal representation of the ID, which is up to 20 characters long for u64::MAX.
+    /// This is within the 32-character limit, and is nice because it matches up with how `niri msg`
+    /// prints the IDs to the console.
+    ///
+    /// We also prepend "mapped-" prefix in case we want to extend this namespace in the future.
+    /// (Xwayland maybe? it'd have the same ID counter, but i figure it's better to futureproof it).
+    ///
+    /// This brings the total length to 27 characters for u64::MAX.
+    pub fn to_protocol_string(self) -> String {
+        format!("mapped-{}", self.0)
```
sodiboo: Realistically, i think this `mapped-` prefix adds nothing: that is only because it derives from a struct called `MappedId`, but that's kind of an internal naming? We could get away with just the numeric ID as a string here: if we need other namespaces for any reason, adding a prefix *is* fully backwards compatible. But when would we ever need another namespace?

src/protocols/foreign_toplevel.rs
sodiboo: In these functions, `self` represents a toplevel, and these functions specifically serve to initialize a protocol-specific toplevel handle. All other relevant objects are already initialized outside of this function. They canonically only have one `Dispatch` bound.

src/protocols/foreign_toplevel.rs
```diff
@@ -38,33 +44,41 @@ pub trait ForeignToplevelHandler {
...
         D: GlobalDispatch<ZwlrForeignToplevelManagerV1, ForeignToplevelGlobalData>,
+        D: GlobalDispatch<ExtForeignToplevelListV1, ForeignToplevelGlobalData>,
         D: Dispatch<ZwlrForeignToplevelManagerV1, ()>,
+        D: Dispatch<ExtForeignToplevelListV1, ()>,
         D: 'static,
         F: for<'c> Fn(&'c Client) -> bool + Send + Sync + 'static,
     {
         let global_data = ForeignToplevelGlobalData {
-            filter: Box::new(filter),
+            filter: Arc::new(filter),
```
sodiboo, line 74: This filter is reused for both protocols, but that is probably wrong. `ext-foreign-toplevel-list-v1` should probably be less privileged than `zwlr-foreign-toplevel-manager-v1`. Do we want to expose `ext-foreign-toplevel-list-v1` for restricted clients too? Whatever the answer is, that's probably a policy decision to be made at a higher level in `src/niri.rs`, and we should maybe accept two filters to this constructor.
I'd imagine at least for now, though, niri is just going to apply its same filter to both protocols. So it doesn't matter(?)

src/protocols/foreign_toplevel.rs
```diff
@@ -1,31 +1,37 @@
...
 
-const VERSION: u32 = 3;
+const EXT_VERSION: u32 = 1;
+const WLR_VERSION: u32 = 3;
 
 pub struct ForeignToplevelManagerState {
     display: DisplayHandle,
-    instances: Vec<ZwlrForeignToplevelManagerV1>,
+    ext_instances: Vec<ExtForeignToplevelListV1>,
+    wlr_instances: Vec<ZwlrForeignToplevelManagerV1>,
```
sodiboo: We use a hashset for tracking handles for each toplevel; this should probably also be a hashset. Makes `Stop` implementation more efficient, as it's a hashset removal instead of `Vec::retain`

src/protocols/foreign_toplevel.rs
YaLTeR: Same filter works yeah

src/window/mapped.rs
sodiboo: I've gone ahead and changed it to just be the ID string, without a prefix.

Like i said, we can add prefixes in the future (or a "don't parse" me random bytes implementation also cannot ever overlap, if it is fixed 32 length).

sodiboo:
> I don't have an opinion on this; while it's useful, there's nothing that says other compositors should also match their IPC identifiers with this, so it ends up being niri-specific either way.

Well, yes. Of course. Even if this *was* the common practice and encouraged in protocol documentation, it would *still* be niri-specific, because for this property to be useful, a client needs to know how to make use of the niri IPC interface.

Although, reading through the documentation, i notice now one minor concern with my current implementation:

> It is recommended that a compositor includes an opaque generation value in identifiers

This is mainly related to compositor handoff, but can even apply to us. Multiple-invocation clients with "one-shot" connections, such as many CLI apps, are allowed to *store* an identifier and persist it.

If it's a "dumb" implementation like i have, exposing an internal identifier, and the session is handed off to another process (e.g. session restore: we are still in the *same session*), then there's no guarantee that the new compositor allocates toplevels in the same order, and existing identifiers may now refer to a different toplevel.

Of course, niri doesn't support any such session restore. But it can also be an issue if such an app simply thinks it is the same session, but niri has restarted in the time since. Then, the identifiers may be reused with no notice; of course they can! it's a new session!

The "opaque generation value" can be based on e.g. the startup time of the compositor.

# Other compositors

- COSMIC ([usage](https://github.com/pop-os/cosmic-comp/blob/a74c90f14fa32e53f7abdaaadf3774b3da86cb19/src/wayland/protocols/toplevel_info.rs#L337)), as well as #1450 ([usage](https://github.com/bbb651/niri/blob/b9c08ab00c517965c2d073917f4962f82201f9ed/src/handlers/compositor.rs#L169)) both use the default provided by Smithay ([implementation](https://github.com/Smithay/smithay/blob/ace98eb20dfffc2f24ea5681bfc6112b5ccc8956/src/wayland/foreign_toplevel_list/mod.rs#L325-L335)) which does random bytes. Smithay also exposes a way to provide an identifier, but nobody seems to be using it.

- wlroots doesn't seem to support specifying a stable identifier. They have a generic [`token`](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/c14aa1d0b82b5e1853a800534f76b90bc55df392/util/token.c) util which does random bytes, and [it's always used for the identifier no matter what](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/c14aa1d0b82b5e1853a800534f76b90bc55df392/types/wlr_ext_foreign_toplevel_list_v1.c#L158-169). So, Sway ([usage](https://github.com/swaywm/sway/blob/f57c82a6f7cbd0c6499def867fc98459fb72a1d5/sway/tree/view.c#L822-L823)) and labwc ([usage](https://github.com/labwc/labwc/blob/359dd8f3c7a400a89613d013e5b9ba5647cdd476/src/foreign-toplevel/ext-foreign.c#L86-L87)) both do random bytes.

- Jay does something similar to wlroots: they have a [`ToplevelIdentifier`](https://github.com/mahkoh/jay/blob/5375dee0078b3060fb3fcd7cf8cab2430bd2fc84/src/utils/toplevel_identifier.rs#L14) newtype which is generated on object creation, and wraps a similar generic [`Opaque`](https://github.com/mahkoh/jay/blob/5375dee0078b3060fb3fcd7cf8cab2430bd2fc84/src/utils/opaque.rs#L19-L25) util.

- Hyprland exposes. uh. [the low bits of the client-specific `foreign_toplevel_list` object as well as the pointer to the window data](https://github.com/hyprwm/Hyprland/blob/bc764f7065c8e46b4e62cbb511b419034712c509/src/protocols/ForeignToplevel.cpp#L58). What? Isn't that just wrong? Yeah! It is. I tested it on my machine; two different clients on the same Hyprland will receive different identifiers for the same toplevel. It also doesn't list existing toplevels when a new list is created. Hyprland's implementation is just broken.

- Louvre has [a similar abstraction to that in this PR](https://github.com/CuarzoSoftware/Louvre/blob/f50fa742ad2cb87e64fe70a1f35c547c502b657c/src/lib/core/roles/LToplevelRole.h#L817-L827), and it works [exactly the same way](https://github.com/CuarzoSoftware/Louvre/blob/f50fa742ad2cb87e64fe70a1f35c547c502b657c/src/lib/core/roles/LToplevelRole.cpp#L121-L125): it's just a `u64` incremented for each mapped surface.

Also, [the initial implementation in COSMIC](https://github.com/i509VCB/cosmic-comp/blob/acba22544df7fac3bfe2b07cd9c0068d49da30d3/src/wayland/protocols/ext_foreign_toplevel_list.rs#L98) (which wasn't merged, but is the compositor linked on the protocol merge request) did the exact same random byte scheme as in present day COSMIC.

---

So, the opaque generational salt seems to be mostly a hypothetical. The main implementations sidestep it by relying on probability to guarantee two random identifiers aren't identical. Louvre *doesn't* address it. We already have the issue in `niri-ipc`. Nobody actually does a generation value. Niri has *absolutely zero* support for session handoff or anything else that would make this an issue in any other case than if the compositor crashes and some clients are left confused by that.

I reckon we just *don't* put a generation value. I think it's way nicer that the output of `lswt` and `niri msg windows` have the same readable identifier.

If we find in the future that such a generation value *needs to be included after all* (e.g. if we implement handoff?), then changing it does absolutely nothing to generic wayland clients (it's already opaque to them), and the niche case of any client parsing it and associating with `niri-ipc` will likely only affect a handful of clients who read our release notes, and the change on their end is fairly trivial.

sodiboo:
I've just noticed that [Hyprland has its own protocol to correlate foreign toplevel handles with IPC identifiers](https://wayland.app/protocols/hyprland-toplevel-mapping-v1). This is the only compositor that does so.

We could potentially offer a similar protocol. Is there any need to do so yet?

YaLTeR:
Thanks for the overview of the identifiers. Guess let's hope nobody starts relying on the identifier being a number mapped ID like that.

> I've just noticed that [Hyprland has its own protocol to correlate foreign toplevel handles with IPC identifiers](https://wayland.app/protocols/hyprland-toplevel-mapping-v1). This is the only compositor that does so.
> 
> We could potentially offer a similar protocol. Is there any need to do so yet?

I don't think it's needed yet, no

src/protocols/foreign_toplevel.rs
YaLTeR, line 1: I actually think that `Vec` with its `retain()` is faster than a `HashSet` for the small expected number of elements here (linear scan in a local memory region is very fast, and `HashSet` has to hash). But also it doesn't really matter and let's leave it as is; it's less code and it's clearer code this way.

YaLTeR (commented):
My remaining concern is the naming one that I mentioned above (`EXT_VERSION`-`WLR_VERSION`), other than that need to test it and then it's good

PR #2050, opened 2025-07-15
revert nushell completion for flake.nix

sodiboo:
- #2009 modified the nix package in an invalid way

theoretically we could install a nushell completion here, but `installShellCompletion` isn't it (see https://github.com/NixOS/nixpkgs/issues/384084). [we already support all the shells it supports](https://github.com/NixOS/nixpkgs/blob/8f4200d165b73db63dc277e9984a480984415593/pkgs/build-support/install-shell-files/setup-hook.sh#L104).

for now, just remove it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -122,7 +122,6 @@
               installShellCompletion --cmd niri \
                 --bash <($out/bin/niri completions bash) \
                 --fish <($out/bin/niri completions fish) \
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks

RossSmyth:
NixOS/nixpkgs#428168
Once staging is merged

Issue #1588, opened 2025-05-17
Implement support for listening and reconfiguring the keyboard settings from locale1


Conan-Kudo:
Niri currently only supports statically configuring the keyboard in configuration, but nowadays there is a good source for keyboard settings that Niri could listen to: [`org.freedesktop.locale1`](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.locale1.html).

Please add support for it so that things like installers, login managers, and keyboard layout configuration tools that set the keyboard through `org.freedesktop.locale1` would be able to configure Niri dynamically.

Here are a few example implementations for other compositors:

* KWin: https://invent.kde.org/plasma/kwin/-/merge_requests/2921
* GNOME Kiosk: https://gitlab.gnome.org/GNOME/gnome-kiosk/-/merge_requests/39
* COSMIC: https://github.com/pop-os/cosmic-settings-daemon/pull/35
* Sway: https://github.com/alebastr/sway-systemd/blob/main/src/locale1-xkb-config

YaLTeR:
Alright, so I suppose the way this should interact with niri's own config file is similar to the [proposal for GNOME Shell](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7761#note_2181736):

- If niri xkb options are unset, then use the keymap from locale1, failing that (e.g. no dbus) keep the current behavior of passing empty strings to XKB, causing it to fetch from some environment variables.
- If some niri xkb options is set, use that and ignore locale1.
- Don't sync niri config back to locale1 (it feels kinda wrong to sync back since we're not a DE like Cosmic).

PR #2052, opened 2025-07-15
Implement fetching xkb options from org.freedesktop.locale1

YaLTeR:
Implements #1588 

cc @Conan-Kudo

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -15,6 +15,9 @@ input {
             // For example:
             // layout "us,ru"
             // options "grp:win_space_toggle,compose:ralt,ctrl:nocaps"
+
+            // If this section is empty, niri will fetch xkb settings
+            // from org.freedesktop.locale1.
         }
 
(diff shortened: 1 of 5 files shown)


YaLTeR:
This may be of interest to LXQt too: @stefonarch

stefonarch:
> This may be of interest to LXQt too: @stefonarch

Thanks, we'll looking for it. Actually `lxqt-config-input` is stripped down for wayland and doesn't include layouts.

YaLTeR:
With this merged, you can change keyboard layout with `localectl set-x11-keymap` as long as you don't have it in the niri config. (Or manually via D-Bus)

alicealysia:
Heck yes!!!!!

stefonarch:
Nice, it looks like the hack to set the layout based on `$LANG` at first login isn't needed anymore now for niri

https://github.com/lxqt/lxqt-wayland-session/blob/4094fb13ea40a233ce6329fa32674a02a7095c0b/startlxqtwayland.in#L118

PR #2055, opened 2025-07-16
clear sigmask before spawning 

sodiboo:
- Follow-up to #2026: @shellhazard noticed that it broke their terminals within niri.

My previous PR did not take proper care for how children are spawned: i didn't realize a high-level-ish API like `calloop::signals` would have such a footgun. So, niri would propagate a sigmask that blocks `SIGINT`, `SIGTERM`, and `SIGHUP` to all children. It also doesn't help that i use `kitty`, which also sets its own signal handler and therefore unsets the sigmask for its children; so i could basically never have found this organically.

---

The most straightforward way to implement this was with the `nix` crate, which is what calloop uses to implement the `Signals` event source. No new dependencies are added to niri as a whole, but niri does now directly depend on `nix`. (that is, the unixy Rust crate. not that *other* nix which i love so dearly)


src/utils/spawning.rs
```diff
@@ -141,6 +141,24 @@ fn spawn_sync(
...
+    // When spawning children, the signalfd is trivial: `Signals::new` creates it with CLOEXEC.
+    //
+    // But, the sigmask is always inherited, so we want to clear it before spawning children.
+    // That way, we don't affect their normal signal handling. In particular, if a child doesn't
+    // care about signals, we must not block it from receiving them.
+    unsafe {
+        process.pre_exec(|| {
+            nix::sys::signalfd::SigSet::empty()
+                .thread_set_mask()
+                .map_err(Into::into)
```
sodiboo: I'm handling the error here instead of discarding it, because as far as i can tell, the only errors it can ever return is from incorrect usage. So, there is no error case that *can* occur here which we would even want to ignore.

YaLTeR:
I was under the impression that `Command::spawn()` already cleared signal-related stuff, but perhaps it was signal handlers and not sigmask.

Either way, could you use rustix instead of nix? The smithay ecosystem is using that in most places by now, and elsewhere in niri we use that too.

YaLTeR:
Guess I was mistaken, this seems to only reset SIGPIPE under some conditions: https://github.com/rust-lang/rust/blob/3014e79f9c8d5510ea7b3a3b70d171d0948b1e96/library/std/src/sys/process/unix/unix.rs#L366

sodiboo:
> I was under the impression that `Command::spawn()` already cleared signal-related stuff, but perhaps it was signal handlers and not sigmask.

Yeah! I assumed it would as well, which is why I didn't even think about it in my first PR. But then, I tested it, and it really does propagate the sigmask to children!

> could you use rustix instead of nix?

rustix does not support the relevant function: https://docs.rs/rustix/1.0.8/rustix/not_implemented/libc_internals/fn.pthread_sigmask.html

so I'm not really sure how I'd do it in rustix other than just diving down to libc and directly calling the relevant unsafe function. why do that when `nix` has a much nicer Rust API.

does it really matter? `nix` is already a transitive dependency via `pipewire`, `libspa`, and `zbus` (as well as `calloop` now but that's only for the signal handling feature and can be eliminated if we did the signals another way)

sodiboo:
also I forgot to mention this in my original comment: I set the sigmask to empty instead of unblocking just the ones we listen on because (a) I observed that the initial sigmask in my session is empty anyways, (b) there's no clean way to propagate the initial sigmask to the spawns (just clearing the ones we listen to would have the effect of unblocking those in children if they were initially blocked when niri starts), and (c) I don't think there's any scenario where you'd *want* the spawned processes to have blocked signals by default; e.g. `kitty` handles signals and therefore clears sigmask in its children, and that's why it was *not* broken. so if this same bug happens one level up, it probably makes sense for niri to make "not broken" children (like `kitty`), even if the environment was "broken".

YaLTeR:
Alright, all of this makes sense. Regarding nix, it's just that everything was moving towards rustix, and I don't particularly like having this extra dep for the equivalence of 1 simple unsafe call... Especially considering that it's an extra version number that I need to be careful with updating so as not to introduce further duplicate nix crate versions into the dependency graph (increasing the binary size).

Perhaps you could replace it with the corresponding unsafe call from the `libc` crate? I think it's fine in this case to ignore errors. Also while we're at it, maybe extract this sigmask clearing into a function, and put the comment on that function, thereby avoiding the need for the "see other comment" second comment.

sodiboo:
If I'm going to be reimplementing the sigmask logic by hand, and also [setting the sigmask far earlier](https://github.com/YaLTeR/niri/pull/2026#issuecomment-3078092118) (before we create an event loop, before we initialize tracy), would it make sense to also reimplement the signalfd handling, declaring the calloop version "insufficient"? That way, instead of pointing at the calloop helper saying "here's what it does and why we need to clean up after it", it'd be somewhat cleaner to just do it all in one codebase, and have the set/clear sigmask actions being handled similarly to one another, instead of one implicitly through calloop and the other in our code. and, the sigmask can be handled separately from the signalfd.

This would also eliminate the `nix` dependency in `calloop`, leaving just `zbus` and pipewire-rs crates depending on it.

> I think it's fine in this case to ignore errors

I disagree: the only error cases for `pthread_sigmask` is if *our code* uses it incorrectly. There is no documented error it can return that is ever fine to ignore. That said, if I do manually call the libc function, it's like 2 lines of code to get the errno, so "not using `nix`" doesn't mean "not handling errors". It's not like error handling here is difficult in any way or a burden, so there is no reason not to do it.

YaLTeR:
> If I'm going to be reimplementing the sigmask logic by hand

This sounds dramatic, isn't it just a single call to empty the mask?

> would it make sense to also reimplement the signalfd handling, declaring the calloop version "insufficient"?

No, using calloop Signals is fine. As long as we avoid an explicit nix dep in niri itself. As far as I can see from the Cargo.lock update in the original PR, this calloop feature used some nix version already present in the dep tree. (It would be better to do this work in calloop itself eventually.)

> That said, if I do manually call the libc function, it's like 2 lines of code to get the errno

Alright that works

cornservant:
Using the rustix's *experimental* runtime feature, you could write something like this, no?

```rust
use rustix::{
    io,
    runtime::{How, KernelSigSet, kernel_sigprocmask},
};

pub fn reset_sigmask() -> io::Result<KernelSigSet> {
    unsafe { kernel_sigprocmask(How::SETMASK, Some(&KernelSigSet::empty())) }
}
```

YaLTeR:
Well it's just one libc call (IIUC), it doesn't need nix or experimental rustix

sodiboo:
> > If I'm going to be reimplementing the sigmask logic by hand
> 
> This sounds dramatic, isn't it just a single call to empty the mask?

Well, it's two calls, one to create an empty mask and one to set use it. But yes. That description was indeed a bit more dramatic than necessary.

I was more thinking "also enabling the sigmask" (+1 call, woe is me), if we're doing something to fix it when tracy is enabled, but then i realized that this doesn't feel like it makes sense really (why would tracy forking off prevent the main thread from blocking signals??), so i have no idea how to fix it with tracy, so i didn't change that at all.

sodiboo:
[alpine :(](https://github.com/YaLTeR/niri/actions/runs/16374444906/job/46270735604?pr=2055)

YaLTeR:
paging in @zimward

YaLTeR:
> I was more thinking "also enabling the sigmask" (+1 call, woe is me), if we're doing something to fix it when tracy is enabled, but then i realized that this doesn't feel like it makes sense really (why would tracy forking off prevent the main thread from blocking signals??), so i have no idea how to fix it with tracy, so i didn't change that at all.

I think you just need to move `Signals::new()` to above `tracy_client::Client::start();`? Which is the point where Tracy spawns some threads. That's what I'm reading from the Signals docs explanation at least.

src/utils/signals.rs
```diff
@@ -0,0 +1,57 @@
...
+            },
+        )
+        .unwrap();
+}
+
+pub fn unblock_all() -> io::Result<()> {
+    set_sigmask(&empty_siget()?)
+}
+
+pub fn empty_siget() -> io::Result<libc::sigset_t> {
```
YaLTeR:
```suggestion
pub fn empty_sigset() -> io::Result<libc::sigset_t> {
```

src/utils/signals.rs
YaLTeR, line 1: Good call extracting both of these into a module

zimward:
> [alpine :(](https://github.com/YaLTeR/niri/actions/runs/16374444906/job/46270735604?pr=2055)

seems like the gitlab is down. gonna change the ci to a alpine container ig.

sodiboo:
> I think you just need to move `Signals::new()` to above `tracy_client::Client::start();`? Which is the point where Tracy spawns some threads. That's what I'm reading from the Signals docs explanation at least.

It doesn't make sense to me that this *would* fix it, but i have added logic to block the signals early, now, which should be correct if this is the fix.

src/utils/signals.rs
sodiboo: lol

sodiboo:
> It doesn't make sense to me that this _would_ fix it.

That did fix it.

YaLTeR:
Thanks!

shellhazard:
Thanks all!

PR #2059, opened 2025-07-16
feat: bezier curve for animation

HigherOrderLogic:
This PR add support for bezier curve usage in animation curve.

While Niri has support for spring animation, I find it really unintuitive and it isnt exactly easy to achieve the result I desire. Bezier curve offer better fine-grained control and more intuitive to control things such as acceleration and overshoot.

<details>
<summary>Demo</summary>

https://github.com/user-attachments/assets/b9f305fa-45da-4009-84d8-74dd261d491b

</details>

I will add documentation after the configuration format is finalised.


Babkock:
Sounds like a good idea to me

niri-config/src/lib.rs
```diff
@@ -1309,12 +1309,13 @@ pub struct EasingParams {
...
 }
 
-#[derive(knuffel::DecodeScalar, Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub enum AnimationCurve {
     Linear,
     EaseOutQuad,
     EaseOutCubic,
     EaseOutExpo,
+    Bezier(f64, f64, f64, f64),
```
YaLTeR: Please add this to the config parse test below

docs/wiki/Configuration:-Animations.md
```diff
@@ -80,12 +80,21 @@ animations {
...
 
-Currently, niri only supports four curves:
+Currently, niri only supports five curves:
 
 - `ease-out-quad` <sup>Since: 0.1.5</sup>
 - `ease-out-cubic`
 - `ease-out-expo`
 - `linear` <sup>Since: 0.1.6</sup>
+- `bezier` <sup>Since: next release</sup>
+    Beside the curve name, you will also need to specify 4 float number, which corresponds to 4 control points of the curve, for example:
```
YaLTeR:
```suggestion
- `bezier` <sup>Since: next release</sup> 
    A custom [cubic Bézier curve](https://www.w3.org/TR/css-easing-1/#cubic-bezier-easing-functions). You need to set 4 numbers defining the control points of the curve, for example:
```

docs/wiki/Configuration:-Animations.md
```diff
@@ -80,12 +80,21 @@ animations {
...
 ```
 
-Currently, niri only supports four curves:
+Currently, niri only supports five curves:
 
 - `ease-out-quad` <sup>Since: 0.1.5</sup>
 - `ease-out-cubic`
 - `ease-out-expo`
 - `linear` <sup>Since: 0.1.6</sup>
+- `bezier` <sup>Since: next release</sup>
```
YaLTeR: I say we name this cubic-bezier to more closely match CSS

niri-config/src/lib.rs
```diff
@@ -3466,7 +3467,90 @@ impl Animation {
...
+                            let y2: FloatOrInt<{ i32::MIN }, { i32::MAX }> =
+                                knuffel::traits::DecodeScalar::decode(val, ctx)?;
+
+                            Some(AnimationCurve::Bezier(x1.0, y1.0, x2.0, y2.0))
+                        }
+                        unknown_curve => {
+                            ctx.emit_error(DecodeError::unexpected(
+                                &val.literal,
+                                "argument",
+                                format!("unexpected argument `{unknown_curve}`"),
```
YaLTeR: Can you list supported types here in this msg?

niri-config/src/lib.rs
```diff
@@ -3466,7 +3467,90 @@ impl Animation {
...
+                        "ease-out-expo" => Some(AnimationCurve::EaseOutExpo),
+                        "bezier" => {
+                            let val = iter_args.next().ok_or_else(|| {
+                                DecodeError::missing(
+                                    child,
+                                    "missing x1 coordinate for control point",
+                                )
+                            })?;
+                            // the X axis represents time so it cannot be negative
+                            let x1: FloatOrInt<0, { i32::MAX }> =
```
YaLTeR: CSS says x1 and x2 are also <= 1

YaLTeR (changes requested):

YaLTeR:
Your example cubic-bezier looks very broken and not what you see on https://cubic-bezier.com/#.05,.7,.1,1 any idea what's up with that?

https://github.com/user-attachments/assets/40e4602b-a5c5-47e1-8041-6f212e8443a0

HigherOrderLogic:
I got that example curve from Material 3 design page, and it seems that the weird bounce back issue only happens with some specific curves but not all of them.

I'll look into this more.

HigherOrderLogic:
I ran [keyframe's cubic Bezier curve reference implementation](https://github.com/gre/bezier-easing) and compared against the crate itself, and it seems like the crate's kinda borked so I switched to another crate, Kurbo.

Tested the Material 3 curve above and worked properly.

YaLTeR:
Uhh I don't really want to pull in the whole kurbo (along with euclid) just for this. Here let's copy the impl from libadwaita: https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-easing.c?ref_type=heads#L469-L531

src/animation/bezier.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn y_for_t(&self, t: f64) -> f64 {
+        let omt = 1. - t;
+        3. * omt * omt * t * self.y1 + 3. * omt * t * t * self.y2 + t * t * t
+    }
+
+    fn t_for_x(&self, x: f64) -> f64 {
+        let mut min_t = 0.;
+        let mut max_t = 1.;
+
+        for _ in 0..=30 {
```
HigherOrderLogic, line 33: Not sure if it's a good idea to iterate this much.

Also since this is binary search we can probably use a while loop that check for `abs(max_t - min_t) > 2^-30`.

YaLTeR:
Thanks

PR #2060, opened 2025-07-16
Implement org.fd.a11y KeyboardMonitor (better Orca screen reader support)

YaLTeR:
Required for screen readers to announce and handle keys.

Currently, only keyboard watching is implemented, but no grabs and no double modifier. I'm also not sure if niri should synthesize and send key repeat events (key repeat is handled client-side on Wayland, though we do have server-side repeat for binds, and there will be optional server-side repeat on Wayland in the future).

Related: #1904

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -365,7 +365,8 @@ binds {
 
     // You can also use a shell. Do this if you need pipes, multiple commands, etc.
     // Note: the entire command goes as a single argument in the end.
-    // Mod+T { spawn "bash" "-c" "notify-send hello && exec alacritty"; }
+    // For example, this is a standard bind to toggle the screen reader (orca).
+    Super+Alt+S hotkey-overlay-title=null { spawn "sh" "-c" "pkill orca || exec orca"; }
 
     // Example volume keys mappings for PipeWire & WirePlumber.
     // The allow-when-locked=true property makes them work even when the session is locked.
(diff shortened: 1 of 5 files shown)


YaLTeR:
Alright, this should be ready. Also added an example bind to toggle orca.

Some additional thoughts:
- I didn't add an access checker for `org.gnome.Orca.KeyboardMonitor` [here](https://gitlab.gnome.org/GNOME/mutter/-/blob/cbb7295ac1f93a2dfd55a7c0544688e7e5c4d2e2/src/backends/meta-a11y-manager.c#L338-339). Is it necessary? If this D-Bus call is filtered from Flatpak apps, then it probably isn't necessary? (It's kind of annoying to implement.)
    - => it's not visible from Flatpak. In Mutter this check is done so apps don't get the idea they can abuse this to grab keys. Sounds like not a concern for niri.
- Key-repeated keys won't be sent (and announced) like in Mutter, but apparently that's not super important. (Implementing this would require synthesizing key repeat in niri which is not something we currently do.)
- The blocking send in the middle of keyboard processing is kind of ugly; if it turns out to be a problem, we could probably spawn a thread for sending key events and communicate with it over a channel.

YaLTeR:
cc @DataTriny: would you like to take a test of this?

DataTriny:
I did and it seems to work well. Now obviously without key grabbing this is not very useful on its own but that is a good start!

The key bind for Orca is a nice addition too.

YaLTeR:
I didn't edit the PR description, but the current version fully implements grabbing, including the double-press to passthrough the modifier

DataTriny:
It didn't work on my side. Let me try again.

DataTriny:
NVM, it doesn't work on my install of GNOME under Wayland either.

YaLTeR:
Huh, weird. Maybe need newer orca? Apparently this dbus interface is very recent

DataTriny:
Config issue on my side. I got it working and haven't found issues with it so far.

YaLTeR:
Great, thanks for testing

YaLTeR:
Tested on another PC myself, seems to work fine.

Issue #2048, opened 2025-07-15
open-on-output / open-on-workspace doesn't apply to named workspaces on multiple outputs on startup


iynaix:
window-rules involving named workspaces assigned to different outputs doesn't seem to work, everything seems to just open on a single output

Relevant section of config:

```kdl
workspace "01" { open-on-output "DP-2"; }
workspace "02" { open-on-output "DP-2"; }
workspace "03" { open-on-output "DP-2"; }
workspace "04" { open-on-output "DP-2"; }
workspace "05" { open-on-output "DP-2"; }
workspace "06" { open-on-output "DP-1"; }
workspace "07" { open-on-output "DP-1"; }
workspace "08" { open-on-output "HDMI-A-1"; }
workspace "09" { open-on-output "HDMI-A-1"; }
workspace "010" { open-on-output "HDMI-A-1"; }

spawn-at-startup "librewolf"
spawn-at-startup "brave" "--incognito"

window-rule {
    match app-id="^librewolf$"
    open-maximized true
}
window-rule {
    match at-startup=true
    match app-id="^librewolf$"
    open-on-output "HDMI-A-1"
    open-on-workspace "09"
}
window-rule {
    match at-startup=true
    match app-id="^brave$"
    open-on-output "DP-2"
    open-on-workspace "01"
}
```

Upon restart (and opening a terminal)

```console
> niri msg windows
Window ID 2:
  Title: "(4) Discord | Friends — LibreWolf"
  App ID: "librewolf"
  Is floating: no
  PID: 4457
  Workspace ID: 1

Window ID 3:
  Title: "New Private Tab - Brave"
  App ID: "brave-browser"
  Is floating: no
  PID: 4473
  Workspace ID: 1

Window ID 4: (focused)
  Title: "~"
  App ID: "com.mitchellh.ghostty"
  Is floating: no
  PID: 6043
  Workspace ID: 1
```

### System Information

* niri version: niri unstable 2025-06-24 (commit 37458d94b288945f6cfbd3c5c233f634d59f246c)

* Distro: NixOS Unstable

* GPU: Nvidia RTX 3080Ti

* CPU: AMD Ryzen 7 7800X3D

armerpunkt:
[...]

PR #2061, opened 2025-07-16
wiki/Window-Rules: Clarify how multiple directives work in a rule

iynaix:
Clarify that directives, unlike matchers are OR-ed together instead of AND, which led to confusion such as https://github.com/YaLTeR/niri/issues/2048

### DIFF
--- a/wiki/Configuration:-Window-Rules.md
+++ b/wiki/Configuration:-Window-Rules.md
@@ -125,7 +125,8 @@ window-rule {
 ```
 
 Match and exclude directives have the same syntax.
-There can be multiple *matchers* in one directive, then the window should match all of them for the directive to apply.
+There can be multiple *matchers* in one directive, then the window must match all of them for the directive to apply.
+When there are multiple *directives* in a rule, any of the directives can be matched to be applied.
 
 ```kdl
 window-rule {


YaLTeR:
Hm, but it says above?

> In order for the rule to apply, a window needs to match *any* of the `match` directives, and *none* of the `exclude` directives.

And in the examples?

iynaix:
Well, all the examples just show the single matcher, which was partly why I was confused and thought multiple matchers were needed.

cornservant:
My attempt to reverse engineer the confusion:

Let's say you read the Overview section of the "Configuration: Window Rules" page.
Given that directives and matchers are not clearly defined you couldn't be blamed for assuming that those are kind of interchangeable terms.
Furthermore the phrasing ["Here are all matchers [...] that a window rule could have"](https://github.com/YaLTeR/niri/pull/2061/files#diff-3d1e696839d568c4102163cec75dd64a4cc4e24d60ef550ab4526e903d70d9c2R27-R39) could suggest that

1. the following list of directives is comprehensive
2. every directive has exactly one matcher (going by the given examples)
3. you can get away with the impression that all match directives have to be satisfied for the window rule to apply unless you read further. But why would you do that if the overview example gave you all the information you thought you needed?

I'm not sure what's the best fix that also keeps the theme of "the example is the documentation/reference"

iynaix:
So how I thought it works (yes, reading is hard and I skimmed :laughing: )

```kdl
window-rule {
	match CONDITION_1
	match CONDITION_2
	open-maximized true
}
```

I thought all the conditions would just be AND together; if you wanted to do open-maximized for another program, you would create another separate window-rule.

Viruzaum:
I was going to suggest a more clear example, but i think this part explains it very well already:

```kdl
window-rule {
    // Match Firefox, but only when it is active...
    match app-id="firefox" is-active=true

    // ...or match Telegram...
    match app-id=r#"^org\.telegram\.desktop$"
    // ...
}
```

It already shows that this rule matches firefox **and** active, **or** Telegram.

armerpunkt:
I think what threw me off is that one of the first things you see is:
```kdl
window-rule {
    match title="Firefox"
    match app-id="Alacritty"
    match is-active=true
    match is-focused=false
    match is-active-in-column=true
    match is-floating=true
    match is-window-cast-target=true
    match is-urgent=true
    match at-startup=true
[...]
```
Which makes it seem like every match should go on a separate line.  And unless you scroll all the way down to the "Window Matching" section you don't see the note about matches being combined on one line. 

I think maybe if you changed the example from the "Overview" to something like:
```kdl
// Set open-maximized to true for all windows.
window-rule {
    open-maximized true
}

// Then, for Alacritty, set open-maximized back to false.
window-rule {
    match app-id="Alacritty"
    open-maximized false
}

// At startup put Alacritty on specific workspace
window-rule {
    match app-id="Alacritty" at-startup=true
    open-on-workspace "chat"
}
```
that would help people understand it sooner.

Issue #2045, opened 2025-07-14
Musl CI


YaLTeR:
People are using niri on musl-based systems, but we don't test building with it on the CI. There are some missing functions like `close_range()` which we may accidentally use and break the musl build without knowing it.

I guess an Alpine container should do the job for this, similar to how we have a Fedora build? There's a niri package in Alpine for reference: https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/community/niri

PR #2064, opened 2025-07-17
ensure that niri builds on musl (alpine)

chinatsu:
thought i'd try my hand at #2045

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,8 +222,11 @@ jobs:
 
       - name: Install dependencies
         run: |
-          apk add --no-cache ${{ env.DEPS_APK }} libadwaita-dev tar
+          apk add --no-cache ${{ env.DEPS_APK }}
 
+      - name: Install tar (for caching)
+        run: |
+          apk add --no-cache tar
       - uses: Swatinem/rust-cache@v2
       - run: cargo build --no-default-features --features dbus,xdp-gnome-screencast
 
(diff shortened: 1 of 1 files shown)


YaLTeR:
cc @zimward

On musl, you want to build with something like `--no-default-features --features dbus,xdp-gnome-screencast`. Check what the Alpine pkg does

chinatsu:
> On musl, you want to build with something like --no-default-features --features dbus,xdp-gnome-screencast

ah, right. i thought i could get away with not doing that, but i'll include this as well. :)

edit: another thing, i'm noticing the [apkbuild](https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/niri/APKBUILD) doesn't have libadwaita-dev as a dependency. is this not necessary when building with only dbus and xdp-gnome-screencast?

zimward:
> cc @zimward
> 
> On musl, you want to build with something like `--no-default-features --features dbus,xdp-gnome-screencast`. Check what the Alpine pkg does

Thats what my current impl is doing. all thats left till i PR it is to implement the cacheing

chinatsu:
oh oops, i see now you've got a feature branch in the works, @zimward. we could close this if you want to get yours in instead :)

the caching used in this pr isn't working currently anyways

zimward:
either way is fine by me as i will get time to work on it in the evening at the earliest. (You don't need libadwaita btw)

chinatsu:
i'll close this then, for the sake of having someone contribute for the first time! :saluting_face:

PR #2070, opened 2025-07-17
fix: retain relative position inside column when switching

mageOfstructs:
The current behavior when switching to a column is to just focus its first tile. This PR changes this behavior to try to retain the index of the focused window in the previous column. Draft because I realize this is something that should probably be configurable, though I am unsure which section should receive the field.

First PR here so please let me know if there's anything I can change.


YaLTeR:
> The current behavior when switching to a column is to just focus its first tile.

The current behavior is that every column remembers its focused tile and you get back to that tile. This is intentional and mirrors how workspaces work. This also enables tabbed columns to just work.

PR #2109, opened 2025-07-22
Add per-axis scroll speed config for input devices

bkuri:
This came out of the need to have `natural-scrolling` enabled on my vertical wheel, but disabled in my horizontal wheel. Additionally, my mouse's horizontal wheel requires a much higher scroll factor in order to be effective. This solves both issues.

- Adds `horizontal` and `vertical` parameters to `scroll-factor`
- Accepts negative values to inverse scroll direction
- Properly complements/overrides global `scroll-factor` setting if present
- Includes docs and tests

Of course feel free to tweak as needed. :+1:


wiki/Configuration:-Input.md
```diff
@@ -248,9 +258,13 @@ Settings specific to `touchpad`s:
...
 - `disabled-on-external-mouse`: do not send events while external pointer device is plugged in.
 
-Settings specific to `touchpad` and `mouse`:
+Settings specific to `touchpad`, `mouse`, `trackpoint`, and `trackball`:
 
 - `scroll-factor`: <sup>Since: 0.1.10</sup> scales the scrolling speed by this value.
+- `scroll-factor-horizontal`: <sup>Since: next release</sup> scales the horizontal scrolling speed by this value. If not set, falls back to `scroll-factor`. Negative values invert the scrolling direction.
+- `scroll-factor-vertical`: <sup>Since: next release</sup> scales the vertical scrolling speed by this value. If not set, falls back to `scroll-factor`. Negative values invert the scrolling direction.
+
+**Note**: While `trackpoint` and `trackball` devices have their own scroll factor config fields, scroll events from these devices currently fall back to default values (1.0) at runtime due to how libinput reports their axis sources. Trackpoints use [button scrolling](https://wayland.freedesktop.org/libinput/doc/latest/trackpoints.html#button-scrolling-on-trackpoints) by default, which generates different axis source events than mouse wheels or touchpad finger scrolling.
```
YaLTeR: Then what's the point of adding those options? Let's not add options that don't work (yet?)

niri-config/src/lib.rs
```diff
@@ -225,6 +225,10 @@ pub struct Touchpad {
     pub middle_emulation: bool,
     #[knuffel(child, unwrap(argument))]
     pub scroll_factor: Option<FloatOrInt<0, 100>>,
+    #[knuffel(child, unwrap(argument))]
+    pub scroll_factor_horizontal: Option<FloatOrInt<-100, 100>>,
+    #[knuffel(child, unwrap(argument))]
+    pub scroll_factor_vertical: Option<FloatOrInt<-100, 100>>,
```
YaLTeR: It would be more consistent with other options to have these as properties on `scroll-factor`, i.e.:
```kdl
// The combined version
scroll-factor 2
// Split version
scroll-factor horizontal=3 vertical=-1
// Override just one
scroll-factor 1 vertical=-2
```
Should be doable by adding a new `struct ScrollFactor` with one argument and two property fields.

niri-config/src/lib.rs
```diff
@@ -3937,6 +3957,139 @@ mod tests {
...
+                }
+            }
+            "#,
+        );
+
+        let touchpad = &parsed.input.touchpad;
+        assert_eq!(touchpad.natural_scroll, true);
+        assert_eq!(touchpad.scroll_factor, Some(FloatOrInt(1.5)));
+        assert_eq!(touchpad.scroll_factor_horizontal, Some(FloatOrInt(2.0)));
+        assert_eq!(touchpad.scroll_factor_vertical, Some(FloatOrInt(-1.5)));
```
YaLTeR: I prefer `assert_debug_snapshot!()` tests because they can update automatically when changing the config structure. Please convert these to just call `assert_debug_snapshot!()` like the main parsing test. Then call `cargo insta test --accept --all` to fill it in.

src/input/mod.rs
```diff
@@ -3070,31 +3070,50 @@ impl State {
...
+                    .or_else(|| $cfg.scroll_factor.map(|x| x.0))
+                    .unwrap_or(1.);
+                let v_factor = $cfg
+                    .scroll_factor_vertical
+                    .map(|y| y.0)
+                    .or_else(|| $cfg.scroll_factor.map(|y| y.0))
+                    .unwrap_or(1.);
+                (h_factor, v_factor)
+            }};
+        }
```
YaLTeR: After you put the scroll-factor config into a `struct ScrollFactor`, this can be a proper method on that struct rather than a macro.

src/tests/scroll_factor.rs
```diff
@@ -0,0 +1,315 @@
+use niri_config::{FloatOrInt, Mouse, Touchpad};
+
+// Reproduce the macro from input/mod.rs for testing
+macro_rules! extract_scroll_factors {
```
YaLTeR: After you put the scroll-factor config into a `struct ScrollFactor` and add a method on that struct, these tests can and should move to niri-config.

src/tests/scroll_factor.rs
```diff
@@ -0,0 +1,315 @@
...
+        left_handed: false,
+        disabled_on_external_mouse: false,
+        middle_emulation: false,
+        scroll_factor: scroll_factor.map(FloatOrInt),
+        scroll_factor_horizontal: scroll_factor_horizontal.map(FloatOrInt),
+        scroll_factor_vertical: scroll_factor_vertical.map(FloatOrInt),
+    }
+}
+
+#[test]
```
YaLTeR: I very much appreciate the testing, but I'm not sure this fairly simple logic needs such an array of tests?

wiki/Configuration:-Input.md
```diff
@@ -248,9 +258,13 @@ Settings specific to `touchpad`s:
 - `click-method`: can be `button-areas` or `clickfinger`, changes the [click method](https://wayland.freedesktop.org/libinput/doc/latest/clickpad-softbuttons.html).
 - `disabled-on-external-mouse`: do not send events while external pointer device is plugged in.
 
-Settings specific to `touchpad` and `mouse`:
+Settings specific to `touchpad`, `mouse`, `trackpoint`, and `trackball`:
 
 - `scroll-factor`: <sup>Since: 0.1.10</sup> scales the scrolling speed by this value.
+- `scroll-factor-horizontal`: <sup>Since: next release</sup> scales the horizontal scrolling speed by this value. If not set, falls back to `scroll-factor`. Negative values invert the scrolling direction.
```
YaLTeR: Question: is it more natural to have scroll-factor-horizontal and vertical *override* the main one, or *multiply* it? E.g. `scroll-factor 2 horizontal=-1` will currently result in horizontal=-1, but maybe it's more natural for it to result in -2? I'm on the fence myself, I think I'm leaning towards override.

YaLTeR (commented):
Thanks, some comments

wiki/Configuration:-Input.md
bkuri: Noted. Just wanted to clarify the scope, but I'll make it more concise.

niri-config/src/lib.rs
bkuri: Yeah, that does look much clearer.

- Implemented `ScrollFactor` struct with combined and split configuration support
- Added support for your suggested syntax: `scroll-factor 2`, `scroll-factor horizontal=3 vertical=-1`, `scroll-factor 2 vertical=-1` (base + override)

niri-config/src/lib.rs
bkuri: I've added `assert_debug_snapshot!()` tests as suggested.

src/input/mod.rs
bkuri: Good call. I've since replaced the macro with `ScrollFactor::get_factors()` method, which is definitely cleaner and more maintainable.

src/tests/scroll_factor.rs
bkuri: No problem; I've simplified the battery of tests to just the most important ones. :smile:

wiki/Configuration:-Input.md
bkuri: Yeah, I do think override is the most intuitive as well.

bkuri:
@YaLTeR I've updated the PR per your comments and everything works well in my local build. Looking forward to your confirmation. Cheers!

docs/wiki/Configuration:-Input.md
```diff
@@ -250,7 +252,15 @@ Settings specific to `touchpad`s:
 
 Settings specific to `touchpad` and `mouse`:
 
-- `scroll-factor`: <sup>Since: 0.1.10</sup> scales the scrolling speed by this value.
+- `scroll-factor`: <sup>Since: 0.1.10, enhanced in next release</sup> scales the scrolling speed. Can be set in multiple ways:
```
YaLTeR: We only support `Since: version` format, nothing extra

docs/wiki/Configuration:-Input.md
```diff
@@ -250,7 +252,15 @@ Settings specific to `touchpad`s:
...
+  - Combined: `scroll-factor 2.0` sets both horizontal and vertical to the same value (multiplies with window-specific scroll factors)
+  - Split: `scroll-factor horizontal=2.0 vertical=-1.0` sets each axis independently (overrides window-specific scroll factors)
+  - Mixed: `scroll-factor 2.0 vertical=-1.0` uses 2.0 as base for both axes, but overrides vertical to -1.0 (overrides window-specific scroll factors)
+  - Negative values invert the scrolling direction for that axis
+  - Per-axis values (split/mixed forms) take precedence over window-specific scroll factors
+  - Combined values multiply with window-specific scroll factors for compatibility
```
YaLTeR: Was this LLM-written? Because it's quite a bit wordier than I'd write it... Don't need this many details. Compare to the original line.

docs/wiki/Configuration:-Input.md
```diff
@@ -250,7 +252,15 @@ Settings specific to `touchpad`s:
...
+
+**Note**: `trackpoint` and `trackball` devices use mouse settings for scroll factors due to how libinput reports their axis sources. These devices use [button scrolling](https://wayland.freedesktop.org/libinput/doc/latest/trackpoints.html#button-scrolling-on-trackpoints) by default.
```
YaLTeR: Do they actually? Also what does button scrolling have to do with this part of the docs?

docs/wiki/Configuration:-Input.md
YaLTeR:
```
- `scroll-factor`: <sup>Since: 0.1.10</sup> scales the scrolling speed by this value.

    <sup>Since: next release</sup> You can also override horizontal and vertical scroll factor separately like so: `scroll-factor horizontal=2.0 vertical=-1.0`
```
Don't need much else on top of this

niri-config/src/lib.rs
```diff
@@ -191,6 +191,42 @@ pub enum TrackLayout {
...
+        let base_value = self.base.map(|f| f.0).unwrap_or(default);
+        let h = self.horizontal.map(|f| f.0).unwrap_or(base_value);
+        let v = self.vertical.map(|f| f.0).unwrap_or(base_value);
+        (h, v)
+    }
+    
+    pub fn has_per_axis_override(&self) -> bool {
+        // Per-axis override means:
+        // 1. We have properties set (horizontal/vertical)
+        // 2. They differ from base, or only one is set
```
YaLTeR: The comment does not match the code, there's no case 2. Have you checked Claude output before pushing?..

niri-config/src/lib.rs
```diff
@@ -4054,6 +4090,286 @@ mod tests {
...
+                }
+                touchpad {
+                    scroll-factor 1.5
+                }
+            }
+            "#,
+        );
+
+        let mouse_sf = parsed.input.mouse.scroll_factor.as_ref().unwrap();
+        assert_eq!(mouse_sf.base, Some(FloatOrInt(2.0)));
```
YaLTeR: There tests should also use assert_debug_snapshot

niri-config/src/lib.rs
```diff
@@ -4054,6 +4090,286 @@ mod tests {
...
+                    scroll-factor 2.0
+                }
+                touchpad {
+                    scroll-factor horizontal=1.5 vertical=-1.0
+                }
+            }
+            "#,
+        );
+
+        assert_debug_snapshot!(parsed.input.mouse.scroll_factor);
```
YaLTeR: Please use inline snapshots for these tests instead of files

niri-config/src/lib.rs
```diff
@@ -4054,6 +4090,286 @@ mod tests {
...
+
+        // Touchpad: base=1.5, horizontal overridden to 3.0
+        let touchpad_sf = parsed.input.touchpad.scroll_factor.as_ref().unwrap();
+        assert_eq!(touchpad_sf.base, Some(FloatOrInt(1.5))); // Base value
+        assert_eq!(touchpad_sf.horizontal, Some(FloatOrInt(3.0))); // Overridden
+        assert_eq!(touchpad_sf.vertical, None); // Not overridden
+        assert!(touchpad_sf.has_per_axis_override()); // Should be considered per-axis
+    }
+
+    // === ScrollFactor behavior tests (moved from src/tests/scroll_factor.rs as requested by YaLTeR) ===
```
YaLTeR: This once again looks like a line left over by Claude that you didn't notice and clean up

niri-config/src/lib.rs
```diff
@@ -4054,6 +4090,286 @@ mod tests {
...
+
+    // Helper function to extract scroll factors from ScrollFactor, matching runtime implementation
+    fn extract_scroll_factors(sf: Option<&ScrollFactor>) -> (f64, f64) {
+        sf.map(|s| s.get_factors(1.0)).unwrap_or((1.0, 1.0))
+    }
+
+    fn create_mouse_config(scroll_factor: Option<ScrollFactor>) -> Mouse {
```
YaLTeR: We don't need tests that create a full mouse config. It is sufficient to test ScrollFactor itself

niri-config/src/lib.rs
```diff
@@ -191,6 +191,42 @@ pub enum TrackLayout {
...
+}
+
+impl ScrollFactor {
+    pub fn new(value: f64) -> Self {
+        Self {
+            base: Some(FloatOrInt(value)),
+            horizontal: None,
+            vertical: None,
+        }
+    }
```
YaLTeR: Not needed

niri-config/src/lib.rs
```diff
@@ -191,6 +191,42 @@ pub enum TrackLayout {
...
+
+    pub fn get_factors(&self, default: f64) -> (f64, f64) {
```
YaLTeR: Default is always 1, don't pass it in

Rename to `h_v_factors` so it's clearer what the return tuple means

niri-config/src/lib.rs
```diff
@@ -4054,6 +4090,286 @@ mod tests {
...
+        let sf = ScrollFactor {
+            base: None,
+            horizontal: Some(FloatOrInt(100.0)),
+            vertical: Some(FloatOrInt(-100.0)),
+        };
+        let config = create_touchpad_config(Some(sf));
+        let (h_factor, v_factor) = extract_scroll_factors(config.scroll_factor.as_ref());
+        assert_eq!(h_factor, 100.0);
+        assert_eq!(v_factor, -100.0);
+    }
```
YaLTeR: You can cut down these tests to like 1 or 2. Once again the logic is very simple, don't need to have 20 tests for two unwrap_or conditions

src/input/mod.rs
```diff
@@ -3086,31 +3086,50 @@ impl State {
...
             .and_then(|root| self.niri.layout.find_window_and_output(&root).unzip().0)
-            .and_then(|window| window.rules().scroll_factor);
-        let scroll_factor = scroll_factor * window_scroll_factor.unwrap_or(1.);
+            .and_then(|window| window.rules().scroll_factor)
+            .unwrap_or(1.);
+
+        // Determine final scroll factors based on configuration
+        let (horizontal_factor, vertical_factor) = if let Some(ref sf) = device_scroll_factor {
+            if sf.has_per_axis_override() {
+                // Per-axis settings override everything, including window factor
```
YaLTeR: Huh? No, they don't override the window factor..

src/input/mod.rs
```diff
@@ -3086,31 +3086,50 @@ impl State {
...
 
-        let scroll_factor = match source {
-            AxisSource::Wheel => self.niri.config.borrow().input.mouse.scroll_factor,
-            AxisSource::Finger => self.niri.config.borrow().input.touchpad.scroll_factor,
-            _ => None,
+        let device_scroll_factor = {
+            let config = self.niri.config.borrow();
+            match source {
+                AxisSource::Wheel => config.input.mouse.scroll_factor.clone(),
+                AxisSource::Finger => config.input.touchpad.scroll_factor.clone(),
```
YaLTeR: It's Copy

src/input/mod.rs
bkuri: Understood. I've since replaced it because earlier they did override the window factor (they didn't multiply with it).

bkuri:
@YaLTeR I am indeed using claude since I don't have a lot of free time to work on this, although I should have done a more thorough sweep, so I certainly apologize for that.

I've now carefully reviewed/fixed every one of your comments, so I'm confident that the PR is now ready to go. :+1:

docs/wiki/Configuration:-Input.md
bkuri: that applied to the old parameters, so I've removed it.

YaLTeR:
> I am indeed using claude since I don't have a lot of free time to work on this

So you're trading your free time for my free time? I don't think that's fair towards either myself or the other 30 PRs waiting for my reviews.

bkuri:
@YaLTeR ok, so should I have just asked for a feature request, then? I thought a PR would've been more effective precisely because I respect your time and mine. Again I apologize for not being more thorough, and I am 100% sure that the PR is ready to go. Hopefully you can do one last check. Otherwise feel free to discard the PR.

YaLTeR:
> @YaLTeR ok, so should I have just asked for a feature request, then? I thought a PR would've been more effective precisely because I respect your time and mine.

I appreciate PRs for features or fixes I consider appropriate for niri (no problem here in this case). However, PR review is quite time and energy consuming. By now I'd say I'm close to having spent more time and energy reviewing this PR than I would've spent implementing the feature myself.

Normally this wouldn't be a big deal since the PR author will presumably consider the feedback, learn and get more experience for future contributions. I'm getting the impression that this is not what's happening here though. Especially considering your lack of free time. So the effectiveness and time saving equation doesn't work out as you can see.

> I am 100% sure that the PR is ready to go. Hopefully you can do one last check.

Okay, let me take another look.

niri-config/src/lib.rs
YaLTeR: This comment was not addressed, the tests still use manual assert_eq instead of assert_debug_snapshot

bkuri:
@YaLTeR I always consider people's feedback and appreciate your patience and candor.
I've replaced the remaining `assert_eq` calls with `assert_debug_snapshot`.
Lmk if I can help with those minor tweaks that you mentioned.
Thanks again! :+1:

niri-config/src/lib.rs
```diff
@@ -4054,6 +4073,237 @@ mod tests {
...
+                }
+                touchpad {
+                    scroll-factor 1.5 horizontal=3.0
+                }
+            }
+            "#,
+        );
+
+        let mouse_sf = parsed.input.mouse.scroll_factor.as_ref().unwrap();
+        assert_eq!(mouse_sf.base, Some(FloatOrInt(2.0)));
```
YaLTeR: These are still not snapshot tests... At this point you'd save *yourself* time if you did it manually you know

niri-config/src/lib.rs
```diff
@@ -4054,6 +4073,237 @@ mod tests {
...
+                horizontal: Some(
+                    FloatOrInt(
+                        2.0,
+                    ),
+                ),
+                vertical: None,
+            },
+        )
+        "###);
+        assert_debug_snapshot!(parsed.input.touchpad.scroll_factor, @r###"
```
YaLTeR: Here and in other places, only one # is needed, not ###

niri-config/src/lib.rs
bkuri: I see. The problem was that I mistakenly added some extra files and then reset HEAD, which must've also reset my latest changes. Sorry about that. :disappointed:

bkuri:
@YaLTeR I do apologize for all the back-and-forth.

I had already changed all `assert_eq` tests to `assert_debug_snapshot`, but then noticed that some extra files were added to the commit, so I did a `reset HEAD` and didn't notice that the changes were lost in the meantime. :angry:

I have now personally checked each and every change and only used claude for the commit messages/pushes, which I hope is okay.

Sorry for the wasted time and I will do my best to avoid this situation in the future.

bkuri:
BTW, I noticed a `r##` in `niri-config/src/lib.rs`, which I didn't change since it appears to come from `main`. Just thought you should know. :+1:

YaLTeR:
Thanks

PR #2126, opened 2025-07-25
Add --path argument for niri msg screenshot* commands

iynaix:
Adds a `-f` path argument to niri msg action screenshot / screenshot-window / screenshot-screen. Specifying an output path makes it easier to process the screenshot with editors such as swappy than watching the output directory for new files, as discussed in #1057 and #719.

I don't know if the way I passed the path through the screenshot ui is ok?

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -223,6 +223,8 @@ pub enum Action {
 
         /// Path to save the screenshot to.
         ///
+        /// The path must be absolute, otherwise an error is returned.
+        ///
         /// If `None`, the screenshot is saved according to the `screenshot-path` config setting.
         #[cfg_attr(feature = "clap", arg(long, action = clap::ArgAction::Set))]
         path: Option<String>,
@@ -240,6 +242,8 @@ pub enum Action {
         show_pointer: bool,
 
         /// Path to save the screenshot to.
+        ///
+        /// The path must be absolute, otherwise an error is returned.
         ///
         /// If `None`, the screenshot is saved according to the `screenshot-path` config setting.
         #[cfg_attr(feature = "clap", arg(long, action = clap::ArgAction::Set))]
@@ -260,6 +264,8 @@ pub enum Action {
         write_to_disk: bool,
 
         /// Path to save the screenshot to.
+        ///
+        /// The path must be absolute, otherwise an error is returned.
         ///
         /// If `None`, the screenshot is saved according to the `screenshot-path` config setting.
         #[cfg_attr(feature = "clap", arg(long, action = clap::ArgAction::Set))]
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -120,16 +120,19 @@ pub enum Action {
     ScreenshotTogglePointer,
     Screenshot(
         #[knuffel(property(name = "show-pointer"), default = true)] bool,
-        #[knuffel(property(name = "path"))] Option<String>,
+        // Path; not settable from knuffel
+        Option<String>,
     ),
     ScreenshotScreen(
         #[knuffel(property(name = "write-to-disk"), default = true)] bool,
         #[knuffel(property(name = "show-pointer"), default = true)] bool,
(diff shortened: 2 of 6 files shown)


Vortriz:
would love to have this!

wind-mask:
Consider supporting output to `stdout` to facilitate direct output?

iynaix:
> Consider supporting output to `stdout` to facilitate direct output?

While not directly, you could do that by doing `-f /dev/stdout`

niri-ipc/src/lib.rs
```diff
@@ -220,6 +220,12 @@ pub enum Action {
         ///  Whether to show the mouse pointer by default in the screenshot UI.
         #[cfg_attr(feature = "clap", arg(short = 'p', long, action = clap::ArgAction::Set, default_value_t = true))]
         show_pointer: bool,
+
+        /// Path to save the screenshot to.
+        ///
+        /// If `None`, the screenshot is saved according to the `screenshot-path` config setting.
+        #[cfg_attr(feature = "clap", arg(short = 'f', long, action = clap::ArgAction::Set))]
```
YaLTeR: Maybe we don't need the short flag for these, especially since it doesn't match the first letter

src/ui/screenshot_ui.rs
```diff
@@ -775,7 +778,12 @@ impl ScreenshotUi {
             return None;
         }
 
-        action(raw, mods)
+        let path = if let Self::Open { path, .. } = self {
+            path.clone()
```
YaLTeR: I'm not a big fan of how this clone()s regardless of whether the action would actually need it. You could probably first call action(), then inspect if it's ConfirmScreenshot with write_to_disk true, then set the path before returning.

src/ui/screenshot_ui.rs
```diff
@@ -775,7 +778,12 @@ impl ScreenshotUi {
...
+        } else {
+            None
+        };
+        action(raw, mods, path)
```
YaLTeR: Actually, I don't think you need to thread `path` through Action::ConfirmScreenshot at all. Just read it from the screenshot UI inside Niri::confirm_screenshot(). That will also fix the fact that confirming by pointer loses the path with your current code.

YaLTeR (changes requested):

niri-ipc/src/lib.rs
iynaix: Yea, i did `-f` for file since `-p` was already taken, I've removed it.

iynaix:
I've made the suggested changes. Took a while to figure it out, but the path is now read directly from within `confirm_screenshot()`.

YaLTeR:
Pushed a small fix. Some thoughts have occurred to me:
1. This explicit path probably shouldn't go through strftime and `~` expansion to avoid surprises. Whoever calls the IPC can do it themselves if needed.
2. `niri msg action screenshot --path path` should probably resolve `path` against its (the CLI's) working directory and pass through the absolute path. Right now it's resolved against the niri compositor's working directory, which can be surprising.
3. The parent folder creation logic should also probably be skipped for the explicit path.

iynaix:
I've managed to get (1) & (3) done, but I don't know how to resolve (2).

Using an absolute path like `/tmp/abc.jpg` or one with a tilde `~/abc.jpg` works, but a relative path such as `./abc.jpg` or `abc.jpg` always creates the file at the top of my home directory.

YaLTeR:
I did the following changes:
- small clean-up
- added making the path absolute in the niri CLI
- added a check that the path is absolute in the niri server
- made path not settable from knuffel config (don't think it's very useful + don't want to bother with these checks for it)

Is this ok? If so then it's good to merge

iynaix:
That looks good to me! Thanks for the help, I'm not familiar enough with the codebase to have come up with the absolute path fix in the IPC. :heart:

YaLTeR:
Thanks.

> While not directly, you could do that by doing `-f /dev/stdout`

This won't work with what this PR does and needs a different approach.

Issue #511, opened 2024-07-03
Screencast doesn't work on some NVIDIA setups


calvinbaart:
Whenever I try to start streaming in Vesktop I get the gnome-portal selection window but after selecting and pressing share nothing happens. The niri log does show some errors:

```
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.644137Z DEBUG niri::dbus::mutter_screen_cast: record_monitor connector="DP-3" properties=RecordMonitorProperties { cursor_mode: Some(Metadata), _is_recording: None }
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.644798Z DEBUG niri::dbus::mutter_screen_cast: start
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.644819Z DEBUG niri::niri: StartCast session_id=0
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.644850Z DEBUG niri::pw_utils: pw stream: state changed: Unconnected -> Connecting
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.645864Z DEBUG niri::pw_utils: pw stream: state changed: Connecting -> Paused
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.645874Z DEBUG niri::pw_utils: pw stream: sending signal with 130
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.688523Z  WARN niri::pw_utils: error creating GBM buffer object: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.688542Z  WARN niri::pw_utils: error creating GBM buffer object: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.688546Z  WARN niri::pw_utils: error creating GBM buffer object: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
Jul 03 11:17:00 niri[2068]: 2024-07-03T09:17:00.688550Z  WARN niri::pw_utils: error creating GBM buffer object: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
[...]

PR #2140, opened 2025-07-27
Implement cursor metadata in window screencast

abmantis:


src/niri.rs
```diff
@@ -5085,11 +5087,34 @@ impl Niri {
...
+                    let (tile, tile_offset, _) = ws
+                        .tiles_with_render_positions()
+                        .find(|(tile, _, _)| tile.window().id().get() == id)
+                        .unwrap();
+                    let pointer_location = self.seat.get_pointer().unwrap().current_location();
+                    let window_offset = tile_offset + tile.window_loc() + mapped.buf_loc().to_f64();
+                    Some((pointer_location - window_offset).upscale(scale))
+                }
+                _ => None,
+            };
```
YaLTeR: Unfortunately the logic here is insufficient. For instance:
- Screencasting a window on a different workspace shouldn't show the cursor, currently it does
- Screencasting a window hidden in a tab shouldn't show the cursor
- Similarly, screencasting a window when the overview is open probably shouldn't show the cursor, currently it will, and at the incorrect location as if the overview wasn't open
- When in the middle of a workspace switch, the location will be incorrect as it doesn't take into account the workspace position
- When the cursor doesn't intersect the window buffer (fully outside), PipeWire probably shouldn't receive this outside location

src/pw_utils.rs
```diff
@@ -96,6 +97,13 @@ impl Drop for SyncobjMap {
     }
 }
 
+const CURSOR_FORMAT: spa_video_format = SPA_VIDEO_FORMAT_ARGB;
+const CURSOR_BPP: u32 = 4;
+const CURSOR_WIDTH: u32 = 24;
+const CURSOR_HEIGHT: u32 = 24;
```
YaLTeR: Why is this hardcoded? You can change the cursor size, and it depends on the cursor theme too. Moreover, clients attach arbitrary buffers as the cursor.

src/pw_utils.rs
```diff
@@ -936,16 +957,106 @@ impl Cast {
...
+                        bitmap[index + 1] = 0xFF;
+                        bitmap[index + 2] = 0xFF;
+                        bitmap[index + 3] = 0xAA;
+                    } else if dist <= radius_sq {
+                        bitmap[index] = 0x00;
+                        bitmap[index + 1] = 0x00;
+                        bitmap[index + 2] = 0x00;
+                        bitmap[index + 3] = 0xAA;
+                    }
+                }
```
YaLTeR: We should draw the cursor normally through niri, through `pointer_element()`. Probably with a separate damage tracker for just the cursor bitmap.

src/pw_utils.rs
```diff
@@ -936,16 +957,106 @@ impl Cast {
...
+        trace!("writing cursor metadata");
+        let cursor_meta: &mut spa_meta_cursor = unsafe { &mut *cursor_meta_ptr };
+        cursor_meta.id = 1;
+        cursor_meta.position.x = pointer_location.x.round() as i32;
+        cursor_meta.position.y = pointer_location.y.round() as i32;
+        cursor_meta.hotspot.x = (CURSOR_WIDTH / 2) as i32;
+        cursor_meta.hotspot.y = (CURSOR_HEIGHT / 2) as i32;
+
+        // TODO: Can the bitmap be initialized previously in start_cast?
+        // Alternativelly, initialize on first pass. For this, bitmap_offset must be 0 on
```
YaLTeR: Since the cursor changes at runtime, it should be tracked separately with its own `OutputDamageTracker`, sent as a `spa_meta_bitmap` whenever it changes, and otherwise don't send a new bitmap.

src/pw_utils.rs
```diff
@@ -975,9 +1087,12 @@ impl Cast {
             let pw_buffer = pw_buffer.as_ptr();
 
             let spa_buffer = (*pw_buffer).buffer;
+            if let Some(pointer) = pointer_location {
+                Self::add_cursor_metadata(spa_buffer, &pointer);
+            }
+
             let fd = (*(*spa_buffer).datas).fd;
             let dmabuf = &self.dmabufs.borrow()[&fd];
-
```
YaLTeR, line 1006: Unrelated change

src/niri.rs
```diff
@@ -1941,6 +1941,7 @@ impl State {
                         bbox.size,
                         scale,
                         wait_for_sync,
+                        None,
```
YaLTeR: Output screencasts should also send cursor as metadata rather than baking it in (when cursor-as-metadata is requested). So we need a small refactor to be able to get pointer and the rest of the render separately when needed.

This would also make it easy to support other two cursor modes (baked-in and no pointer) which would also be good for this PR.

abmantis:
> Please rebase onto main, it has the necessary spa_buffer access.

I did rebase to main yesterday, but it now seems to have some issue with the cursor bitmap buffer which I haven't figured out yet. Probably some bad pointer offset calculation. I should take a look at it back later today.

src/pw_utils.rs
abmantis: The intent was to make this PR simpler/smaller and easier to merge, and then work on a follow-up to add proper cursor support. 
The current dot indicator in this PR  works quite nicely and is an improvement over having no indicator. But if you prefer to have the full thing before merging this, I'm fine with that too.

src/niri.rs
abmantis: Same reason as for https://github.com/YaLTeR/niri/pull/2140#discussion_r2260803397 . I was trying to keep this PR small and iterate later with followups, instead of a big bang. But I can try adding that too in this one if you prefer.

src/pw_utils.rs
YaLTeR: I would prefer the full thing yeah. Don't think adding a circle like this separately is worth it

abmantis:
For some reason, without https://github.com/YaLTeR/niri/commit/15b4acc17ee2c8ac31f7e6093df57aa66e6bc2de#diff-a517890ab4c703e485977a1943ceb154a6bcb3cc4d83135097a2841d569e7326L930-L931 the cast just stops after a second:

I've printed the stride and offset from the dmabuf and the current ones on the chunks, and for some reason the chunk ones overflow:

```
2025-08-08T14:18:18.137093Z DEBUG niri::pw_utils: pw stream: moving to ready state
2025-08-08T14:18:18.139977Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Streaming
2025-08-08T14:18:18.140313Z DEBUG niri::pw_utils: pw buffer: fd = 183, stride = 7168, offset = 0
2025-08-08T14:18:18.140324Z DEBUG niri::pw_utils: pw buffer: chunk stride = 7168, chunk offset = 0
2025-08-08T14:18:18.140329Z DEBUG niri::pw_utils: pw buffer: fd = 184, stride = 2048, offset = 10092544
2025-08-08T14:18:18.140333Z DEBUG niri::pw_utils: pw buffer: chunk stride = 2048, chunk offset = 10092544
2025-08-08T14:18:18.152639Z DEBUG niri::pw_utils: pw buffer: fd = 185, stride = 7168, offset = 0
2025-08-08T14:18:18.152673Z DEBUG niri::pw_utils: pw buffer: chunk stride = 7168, chunk offset = 0
2025-08-08T14:18:18.152681Z DEBUG niri::pw_utils: pw buffer: fd = 186, stride = 2048, offset = 10092544
2025-08-08T14:18:18.152688Z DEBUG niri::pw_utils: pw buffer: chunk stride = 2048, chunk offset = 10092544
2025-08-08T14:18:18.742885Z DEBUG niri::pw_utils: pw buffer: fd = 187, stride = 7168, offset = 0
2025-08-08T14:18:18.742918Z DEBUG niri::pw_utils: pw buffer: chunk stride = -1426063361, chunk offset = 2868903935
2025-08-08T14:18:18.742921Z DEBUG niri::pw_utils: pw buffer: fd = 188, stride = 2048, offset = 10092544
2025-08-08T14:18:18.742924Z DEBUG niri::pw_utils: pw buffer: chunk stride = -1442840576, chunk offset = 2852126720
2025-08-08T14:18:18.745123Z DEBUG niri::pw_utils: pw stream: state changed: Streaming -> Paused
```

They only overflow when I write to the bitmap array. I don't see anything wrong with the bitmap, so maybe something is failing to take the extra bitmap size into account?

YaLTeR:
Hmm, idk, I looked through this PR's diff and [the code in mutter](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/backends/meta-screen-cast-stream-src.c) and couldn't see a problem off the bat. Try filling the bitmap with sentinel values (like 0x01, 0x02, 0x03 and so on) to see where it overwrites subsequent data? Maybe check with gdb?

src/pw_utils.rs
```diff
@@ -619,31 +653,18 @@ impl PipeWire {
                         let plane_count = dmabuf.num_planes();
                         assert_eq!((*spa_buffer).n_datas as usize, plane_count);
 
-                        for (i, (fd, (stride, offset))) in
-                            zip(dmabuf.handles(), zip(dmabuf.strides(), dmabuf.offsets()))
-                                .enumerate()
-                        {
+                        for (i, fd) in dmabuf.handles().enumerate() {
```
abmantis: I'll revert this (and the related changes) back once I find why it causes issues

abmantis:
> Hmm, idk, I looked through this PR's diff and [the code in mutter](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/backends/meta-screen-cast-stream-src.c) and couldn't see a problem off the bat. Try filling the bitmap with sentinel values (like 0x01, 0x02, 0x03 and so on) to see where it overwrites subsequent data? Maybe check with gdb?

I'm struggling to find the cause of this. The cursor_meta pointer is slightly different on every frame, which makes it so that I can't just watch that memory space for changes. 
I'll keep trying, but I've pushed the code here in case some one wants to try it too. This is now based on `main`, but contains a revert of https://github.com/YaLTeR/niri/commit/15b4acc17ee2c8ac31f7e6093df57aa66e6bc2de

abmantis:
@YaLTeR looking at https://docs.pipewire.org/video-src_8c-example.html#a57 the example sets the stride/offset on every frame. Are you sure that is not required? 
Also, I noticed that the chunk address changes on every frame, so how can it have the stride set only at start?

abmantis:
I'm logging the `bitmap_data` address and `bitmap_data.add(MAX_CURSOR_BITMAP_SIZE as usize)`, and also the address on the spa_data `chunk`:
image

Notice that the address pairs pointed by the green (frame 1) and blue (frame 2) arrows match, which means that the bitmap_data is in the proper memory space since the data chunk comes right after. 
Also notice that the address pointed by the red arrow (frame 2) is in the middle of the address range of `bitmap_data` of frame 1. That is why it gets a corrupted value when I write the bitmap vs. when I don't.

YaLTeR:
From Matrix:

> well on main there are only dmabufs, and for them stride and offset are data about the dmabuf itself, not about pipewire structures
> since dmabufs are allocated once and then don't change, there's no need to change their stride and offset
> (dmabuf is like a handle to GPU memory)
> That said, at least Mutter sets them every frame, like niri before that commit. It's not impossible that pipewire itself has some issue where it automatically changes those values even for dmabufs and nobody has noticed so far. (This is just speculation)
> The screencast codepaths there don't get a lot of various testing
> Have you tried debugging with gdb and watching the memory location corresponding to the stride field and seeing what changes it?

I'm open to changing it back to set stride/offset every frame, but if it's indeed a bug in PW, it'd be good to get it found and fixed

YaLTeR:
Non-dmabuf stride and offset do not fall under what I described above of course

abmantis:
Opened an issue on pipewire: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4884

YaLTeR:
Thanks, subscribed

abmantis:
I was able to confirm that the PR linked to that issue fixes the problem! :partying_face: 
I can finally get back to implementing this.

YaLTeR:
Awesome! I guess you should add a runtime check for pipewire version with the fix though.

src/niri.rs
abmantis: I think all of these are solved now, except:

> When in the middle of a workspace switch, the location will be incorrect as it doesn't take into account the workspace position

I think it would look weird to see the pointer "run" across the window and then disappear when the workspace switches. Currently it stays still (if the mouse is not moving) and then disappears when the mouse is not focusing the window anymore.

YaLTeR: Could it somehow use the same logic as pointer focus? Currently, pointer focus [won't update during animations](https://github.com/YaLTeR/niri/blob/e837e39623457dc5ad29c34a5ce4d4616e5fbf1e/src/niri.rs#L966) (which includes workspace switches and such), but it will update if you input something during an animation, so you can still click things while hovered in-between two workspaces (mouse while hovering with a touchpad for instance). Could the screencast logic reuse the same point contents somehow? Then it would always be accurate without unnecessary movement during animations.

src/pw_utils.rs
abmantis: For some reason (pipewire acting up?) if I do that sometimes it does not update the cursor when it changes on the screencast, despite it being sent. For now I'll leave that optimization out and try to look into it later.

src/niri.rs
abmantis: Done now.

src/pw_utils.rs
```diff
@@ -995,6 +1148,16 @@ impl Cast {
...
+                    renderer,
+                    spa_buffer,
+                    scale,
+                    pointer_elements,
+                    &pointer_location,
+                    &pointer_hotspot,
+                );
+            }
 
             let fd = (*(*spa_buffer).datas).fd;
```
abmantis, line 1142: we should now be able to also optimize the screencast code to not render a new frame when there is no damage to it, and send only cursor metadata. But that appears to require some extra refactoring, so I suggest leaving it for a follow-up PR instead.

ASong5:
Is there more work that needs to be done on this for cursor data to be sent for screencasting? I have built with these changes locally but the cursor still does not appear in screencasts. Would love an update on this. Thanks.

YaLTeR:
fwiw I haven't had a chance to look at this again yet; I'll try to before the next release if I find time. Testing is very welcome

abmantis:
> Is there more work that needs to be done on this for cursor data to be sent for screencasting? I have built with these changes locally but the cursor still does not appear in screencasts. Would love an update on this. Thanks.

No, it should be done. What version of pipewire do you have installed?

ASong5:
> > Is there more work that needs to be done on this for cursor data to be sent for screencasting? I have built with these changes locally but the cursor still does not appear in screencasts. Would love an update on this. Thanks.
> 
> No, it should be done. What version of pipewire do you have installed?

I am on 1.2.7 which is what comes with Ubuntu 25.04. Will I need to build pipewire from source?

abmantis:
> > > Is there more work that needs to be done on this for cursor data to be sent for screencasting? I have built with these changes locally but the cursor still does not appear in screencasts. Would love an update on this. Thanks.
> > 
> > No, it should be done. What version of pipewire do you have installed?
> 
> I am on 1.2.7 which is what comes with Ubuntu 25.04. Will I need to build pipewire from source?

Yeah, you need ~~1.2.8~~ at least. The previous versions had a buffer memory bug. 
EDIT: meant 1.4.8.

ASong5:
> > > > Is there more work that needs to be done on this for cursor data to be sent for screencasting? I have built with these changes locally but the cursor still does not appear in screencasts. Would love an update on this. Thanks.
> > > 
> > > 
> > > No, it should be done. What version of pipewire do you have installed?
> > 
> > 
> > I am on 1.2.7 which is what comes with Ubuntu 25.04. Will I need to build pipewire from source?
> 
> Yeah, you need 1.2.8 at least. The previous versions had a buffer memory bug.

Updated to pipewire 1.4.7 and the cursor still does not appear in screencasts. In the journal it says that ```WARN niri::pw_utils: cursor metadata mode requires PipeWire >= 1.4.8```. I'm guessing you meant 1.4.8 not 1.2.8.

ASong5:
After compiling pipewire 1.5.8 from source, I get the following in journal:

```WARN niri::pw_utils: error rendering cursor, using default bitmap```

How do I further troubleshoot this? Have you tested on later version of pipewire or only 1.4.8?

File: src/pw_utils.rs [hunk 9]
```diff
@@ -956,35 +996,148 @@ impl Cast {
...
+            renderer,
+            size,
+            render_scale,
+            Transform::Normal,
+            Fourcc::Argb8888,
+            pointer_elements.iter(),
+        ) {
+            Ok(pointer_vec) => pointer_vec,
+            Err(_) => {
+                warn!("error rendering cursor, using default bitmap");
```
abmantis:
```suggestion
                error!("error rendering cursor");
```

abmantis:
> After compiling pipewire 1.5.8 from source, I get the following in journal:
> 
> `WARN niri::pw_utils: error rendering cursor, using default bitmap`
> 
> How do I further troubleshoot this? Have you tested on later version of pipewire or only 1.4.8?

That log message is a bit misleading since there is no default bitmap. I've just updated.

Anyway, can you enable debug logs and see if something else pops up? I don't know why the rendering of the cursor elements is failing.

ASong5:
> > After compiling pipewire 1.5.8 from source, I get the following in journal:
> > `WARN niri::pw_utils: error rendering cursor, using default bitmap`
> > How do I further troubleshoot this? Have you tested on later version of pipewire or only 1.4.8?
> 
> That log message is a bit misleading since there is no default bitmap. I've just updated.
> 
> Anyway, can you enable debug logs and see if something else pops up? I don't know why the rendering of the cursor elements is failing.

```
Nov 24 22:13:29 pundrew-desktop niri[3200758]: 2025-11-25T03:13:29.065236Z ERROR niri::pw_utils: error rendering cursor
Nov 24 22:13:29 pundrew-desktop niri[3200758]: 2025-11-25T03:13:29.235871Z ERROR niri::pw_utils: error rendering cursor
Nov 24 22:13:29 pundrew-desktop niri[3200758]: 2025-11-25T03:13:29.732085Z ERROR niri::pw_utils: error rendering cursor
Nov 24 22:13:29 pundrew-desktop niri[3200758]: 2025-11-25T03:13:29.943374Z ERROR niri::pw_utils: error rendering cursor
Nov 24 22:13:39 pundrew-desktop niri[3200758]: 2025-11-25T03:13:39.099299Z DEBUG niri::pw_utils: pw stream: state changed: Streaming -> Paused stream_id=1
Nov 24 22:13:39 pundrew-desktop niri[3200758]: 2025-11-25T03:13:39.105405Z DEBUG niri::dbus::mutter_screen_cast: stop
Nov 24 22:13:39 pundrew-desktop niri[3200758]: 2025-11-25T03:13:39.105442Z DEBUG niri::niri: StopCast session_id=1
Nov 24 22:13:39 pundrew-desktop niri[3200758]: 2025-11-25T03:13:39.105472Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Unconnected stream_id=1
Nov 24 22:13:46 pundrew-desktop xdg-desktop-portal-gnome[3201079]: Failed to associate portal window with parent window 
Nov 24 22:13:46 pundrew-desktop xdg-desktop-portal-gnome[3201079]: vkAcquireNextImageKHR(): A swapchain no longer matches the surface properties exactly, but can still be used to present to the surface successfully. (VK_SUBOPTIMAL_KHR) (1000001003)
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.281916Z DEBUG niri::dbus::mutter_screen_cast: record_monitor connector="DP-1" properties=RecordMonitorProperties { cursor_mode: Some(Metadata), _is_recording: None }
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.282368Z DEBUG niri::dbus::mutter_screen_cast: start
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.282385Z DEBUG niri::niri: StartCast session_id=4 stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.282415Z DEBUG niri::pw_utils: pw stream: state changed: Unconnected -> Connecting stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.284338Z DEBUG niri::pw_utils: pw stream: state changed: Connecting -> Paused stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.284347Z DEBUG niri::pw_utils: pw stream: sending signal with 68 stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.319507Z DEBUG niri::pw_utils: pw stream: got format = VideoInfoRaw { format: VideoFormat::BGRx, flags: VideoFlags(0xc), modifier: 0, size: spa_rectangle { width: 2560, height: 1440 }, framerate: spa_fraction { num: 0, denom: 1 }, max_framerate: spa_fraction { num: 165080, denom: 1000 }, views: 0, interlace_mode: 0, pixel_aspect_ratio: spa_fraction { num: 0, denom: 0 }, multiview_mode: 0, multiview_flags: 0, chroma_site: 0, color_range: 0, color_matrix: 0, transfer_function: 0, color_primaries: 0 } stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.319521Z DEBUG niri::pw_utils: pw stream: fixating the modifier stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.319527Z DEBUG niri::pw_utils: find_preferred_modifier: size=Size<smithay::utils::geometry::Physical> { w: 2560, h: 1440 }, fourcc=XR24, modifiers=[144115188621556484, 144115188621556484, 144115188621556484, 144115188621556484, 144115188621540100, 144115188621343492, 144115188621122308, 144115188621557508, 144115188621541124, 144115188621344516, 144115188621123332, 144115188075858436, 0, 72057594037927935]
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.319735Z DEBUG niri::pw_utils: pw stream: allocation successful (modifier=Unrecognized(144115188621556484), plane_count=2), moving to confirmation pending stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.320575Z DEBUG niri::pw_utils: pw stream: got format = VideoInfoRaw { format: VideoFormat::BGRx, flags: VideoFlags(0x4), modifier: 144115188621556484, size: spa_rectangle { width: 2560, height: 1440 }, framerate: spa_fraction { num: 0, denom: 1 }, max_framerate: spa_fraction { num: 165080, denom: 1000 }, views: 0, interlace_mode: 0, pixel_aspect_ratio: spa_fraction { num: 0, denom: 0 }, multiview_mode: 0, multiview_flags: 0, chroma_site: 0, color_range: 0, color_matrix: 0, transfer_function: 0, color_primaries: 0 } stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.320584Z DEBUG niri::pw_utils: pw stream: moving to ready state stream_id=4
Nov 24 22:13:48 pundrew-desktop niri[3200758]: 2025-11-25T03:13:48.323409Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Streaming stream_id=4
Nov 24 22:13:55 pundrew-desktop niri[3200758]: 2025-11-25T03:13:55.182715Z ERROR niri::pw_utils: error rendering cursor
Nov 24 22:13:55 pundrew-desktop niri[3200758]: 2025-11-25T03:13:55.184213Z ERROR niri::pw_utils: error rendering cursor
```


I don't know if these logs help but it's all I get in journal which is still flooded with these "error rendering cursor" messages.

ASong5:
Along with the cursor issue, niri crashed completely when I started a screencast:

```
v 27 12:08:21 pundrew-desktop waybar[3516842]: Error reading events from display: Connection reset by peer
Nov 27 12:08:21 pundrew-desktop xdg-desktop-portal-gtk[3517570]: Error reading events from display: Broken pipe
Nov 27 12:08:21 pundrew-desktop easyeffects[3517200]: Lost connection to Wayland compositor.
Nov 27 12:08:21 pundrew-desktop org.gnome.Nautilus[581213]: Lost connection to Wayland compositor.
Nov 27 12:08:21 pundrew-desktop niri[3516722]: 2025-11-27T17:08:21.132051Z DEBUG niri::utils::watcher: exiting watcher thread for Regular { user_path: "/home/pundrew/.config/niri/config.kdl", system_path: "/etc/niri/config.kdl" }
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.741268Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Unconnected stream_id=5
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.741164Z DEBUG niri::pw_utils: pw stream: state changed: Streaming -> Paused stream_id=5
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.741060Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Unconnected stream_id=4
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.741012Z DEBUG niri::pw_utils: pw stream: state changed: Streaming -> Paused stream_id=4
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.740910Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Unconnected stream_id=2
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.740797Z DEBUG niri::pw_utils: pw stream: state changed: Streaming -> Paused stream_id=2
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.740690Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Unconnected stream_id=1
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.739668Z DEBUG niri::pw_utils: pw stream: state changed: Paused -> Unconnected stream_id=0
Nov 27 12:08:20 pundrew-desktop niri[3516722]: 2025-11-27T17:08:20.739275Z DEBUG niri::pw_utils: pw stream: state changed: Streaming -> Paused stream_id=0
Nov 27 12:08:20 pundrew-desktop niri[3516722]: note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at /rustc/595b9a498bc55fcd30111e430d8e4290ed833b4c/library/core/src/ops/function.rs:250:5
Nov 27 12:08:20 pundrew-desktop niri[3516722]:   15: call_once<fn() -> core::result::Result<(), alloc::boxed::Box<dyn core::error::Error, alloc::alloc::Global>>, ()>
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/main.rs:265:10
Nov 27 12:08:20 pundrew-desktop niri[3516722]:   14: main
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/calloop-0.14.3/src/loop_logic.rs:670:13
Nov 27 12:08:20 pundrew-desktop niri[3516722]:   13: run<niri::niri::State, niri::main::{closure_env#3}, core::option::Option<core::time::Duration>>
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/main.rs:265:46
Nov 27 12:08:20 pundrew-desktop niri[3516722]:   12: {closure#3}
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/niri.rs:715:19
Nov 27 12:08:20 pundrew-desktop niri[3516722]:   11: refresh_and_flush_clients
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/niri.rs:3816:18
Nov 27 12:08:20 pundrew-desktop niri[3516722]:   10: redraw_queued_outputs
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/niri.rs:4635:17
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    9: redraw
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/backend/mod.rs:79:38
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    8: with_primary_renderer<(), niri::niri::{impl#5}::redraw::{closure_env#3}>
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/backend/tty.rs:1462:14
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    7: with_primary_renderer<(), niri::niri::{impl#5}::redraw::{closure_env#3}>
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/niri.rs:4645:22
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    6: {closure#3}
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at ./git/niri/src/niri.rs:5325:26
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    5: render_windows_for_screen_cast
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    4: get_window_pointer_location
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at /rustc/595b9a498bc55fcd30111e430d8e4290ed833b4c/library/core/src/option.rs:2169:5
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    3: core::option::unwrap_failed
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at /rustc/595b9a498bc55fcd30111e430d8e4290ed833b4c/library/core/src/panicking.rs:148:5
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    2: core::panicking::panic
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at /rustc/595b9a498bc55fcd30111e430d8e4290ed833b4c/library/core/src/panicking.rs:78:14
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    1: core::panicking::panic_fmt
Nov 27 12:08:20 pundrew-desktop niri[3516722]:              at /rustc/595b9a498bc55fcd30111e430d8e4290ed833b4c/library/std/src/panicking.rs:698:5
Nov 27 12:08:20 pundrew-desktop niri[3516722]:    0: __rustc::rust_begin_unwind
Nov 27 12:08:20 pundrew-desktop niri[3516722]: stack backtrace:
Nov 27 12:08:20 pundrew-desktop niri[3516722]: called `Option::unwrap()` on a `None` value
Nov 27 12:08:20 pundrew-desktop niri[3516722]: thread 'main' (3516722) panicked at src/niri.rs:5228:14:

```
Will update when I have reproduced it with verbose logs. I never experienced this prior to building with these changes.

ASong5:
Was able to reproduce consistently by starting a screencast and then dragging the screencasted window via Mod+MB1. @abmantis are you able to reproduce this on your system?

abmantis:
> Was able to reproduce consistently by starting a screencast and then dragging the screencasted window via Mod+MB1. @abmantis are you able to reproduce this on your system?

I've just tried that and cannot reproduce the crash. But I'll look at the unwraps and see if I can make them safer.

abmantis:
@ASong5 can you try again and check the logs? The `error rendering cursor` message should now log the error to.
Also the crashes when dragging should now be fixed. :crossed_fingers: Can you test that out too since I was not able to reproduce it?

ASong5:
I just tested and the crashes are gone! However, cursor data is still not being sent properly for me:

```
pundrew@pundrew-desktop:~$ journalctl -r -b -0
Dec 01 21:30:25 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:25 pundrew-desktop niri[952124]:     0: error binding texture
Dec 01 21:30:25 pundrew-desktop niri[952124]: Caused by:
Dec 01 21:30:25 pundrew-desktop niri[952124]: 2025-12-02T02:30:25.197522Z ERROR niri::pw_utils: error rendering cursor: error rendering
Dec 01 21:30:25 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:25 pundrew-desktop niri[952124]:     0: error binding texture
Dec 01 21:30:25 pundrew-desktop niri[952124]: Caused by:
Dec 01 21:30:25 pundrew-desktop niri[952124]: 2025-12-02T02:30:25.008598Z ERROR niri::pw_utils: error rendering cursor: error rendering
Dec 01 21:30:24 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:24 pundrew-desktop niri[952124]:     0: error binding texture
Dec 01 21:30:24 pundrew-desktop niri[952124]: Caused by:
Dec 01 21:30:24 pundrew-desktop niri[952124]: 2025-12-02T02:30:24.873792Z ERROR niri::pw_utils: error rendering cursor: error rendering
Dec 01 21:30:24 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:24 pundrew-desktop niri[952124]:     0: error binding texture
Dec 01 21:30:24 pundrew-desktop niri[952124]: Caused by:
Dec 01 21:30:24 pundrew-desktop niri[952124]: 2025-12-02T02:30:24.737332Z ERROR niri::pw_utils: error rendering cursor: error rendering
Dec 01 21:30:24 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:24 pundrew-desktop niri[952124]:     0: error binding texture
Dec 01 21:30:24 pundrew-desktop niri[952124]: Caused by:
Dec 01 21:30:24 pundrew-desktop niri[952124]: 2025-12-02T02:30:24.580520Z ERROR niri::pw_utils: error rendering cursor: error rendering
Dec 01 21:30:24 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:24 pundrew-desktop niri[952124]:     0: error binding texture
Dec 01 21:30:24 pundrew-desktop niri[952124]: Caused by:
Dec 01 21:30:24 pundrew-desktop niri[952124]: 2025-12-02T02:30:24.238514Z ERROR niri::pw_utils: error rendering cursor: error rendering
Dec 01 21:30:24 pundrew-desktop niri[952124]:     1: Failed to bind Framebuffer
Dec 01 21:30:24 pundrew-desktop niri[952124]:     0: error binding texture
```
https://github.com/YaLTeR/niri/issues/511 is the most related issue I can find, but it seems to be fixed already and mainly affected nvidia cards whereas I use amd: 
```
pundrew@pundrew-desktop:~$ glxinfo | grep "OpenGL renderer"
OpenGL renderer string: AMD Radeon RX 7800 XT (radeonsi, navi32, LLVM 20.1.8, DRM 3.64, 6.17.0-6-generic)
```

I'm not sure if you had a chance to test with newer versions of pipewire than 1.4.8 but I will downgrade to 1.4.8 and report back in case it is an issue with newer versions of pipewire.

abmantis:
I've just tested it with pipewire 1.4.9 since that is now available on NixOS.
Regarding the error, I'll try to look a bit into it. How are you running niri?

ASong5:
Tested with pipewire version 1.4.8 and the issue persists.

I'm running niri via SDDM

busyoGG:
Same here. It also breaks the original cursor capture function.

The error occurred at  ` let pointer_vec = match render_to_vec`

### edited:

I found that `pointer_elements.len()` is 0 when starting screencast if you didn't move the cursor, which is why it reports an error.

But it's not the reason why the cursor disappears.

abmantis:
What app are you testing with @ASong5 @busyoGG ? 
I was not able to reproduce the errors.

But I've found a bug when using a second display where the cursor is missing. I'll look into that.

busyoGG:
OBS with multiple monitors

ASong5:
OBS with multiple displays here as well. I haven't checked if the errors happen during something like a discord screenshare, but I am guessing it does since the cursor does not appear there either

abmantis:
Ok, I am able to reproduce it with OBS! I was testing it with https://www.webrtc-experiment.com/Pluginfree-Screen-Sharing/ and it works fine there. I'll take a look at fixing it now that I can reproduce it.

abmantis:
Everything seems to be working now, even with OBS! :partying_face: Don't forget to enable the option to capture the cursor in OBS!
I was also able to test it properly with multi monitors and scales (had some bugs on that)!

Hopefully we can now get this moving :)

YaLTeR:
I was planning to look at this and I went ahead and added pointer to screenshot-window in preparation: https://github.com/YaLTeR/niri/commit/9f8eadc5bc04994f395832e7135693f87aa899a9

If I'm not mistaken, for this PR the logic should be similar to that commit. I think I'll look and correct it myself if that's ok

abmantis:
> I was planning to look at this and I went ahead and added pointer to screenshot-window in preparation: [9f8eadc](https://github.com/YaLTeR/niri/commit/9f8eadc5bc04994f395832e7135693f87aa899a9)
> 
> If I'm not mistaken, for this PR the logic should be similar to that commit. I think I'll look and correct it myself if that's ok

Yes, feel free to update the PR as you see fit!

YaLTeR:
I'm like 80% done with refactoring this PR now, everything that worked before should keep working, and even some things that didn't (all hidden/embedded/metadata modes should work fine). Still need to fix a few TODOs.

YaLTeR:
Alright I think I finished it, will need to test it out more

abmantis:
> Alright I think I finished it, will need to test it out more

Nice! I hope it wasn't too bad, since I am familiar with C++, but a Rust noob.
I'll try to test it out during the weekend too.

YaLTeR:
Thanks again for your work and for figuring out the PipeWire bug and getting it fixed!

abmantis:
> Thanks again for your work and for figuring out the PipeWire bug and getting it fixed!

Happy to contribute and see this getting merged!
Thanks for reviewing.

PR #2147, opened 2025-07-30
github wiki replacement / mkdocs-docs

chinatsu:
okay, i think i'm at a good point to open a draft PR to keep track of the rest of the process.

TLDR; a demo of the docs can be found here https://chinatsu.github.io/niri/
and the existing rustdocs can be found here https://chinatsu.github.io/niri/niri_ipc

the context of this PR is this discussion thread: https://github.com/YaLTeR/niri/discussions/1830

as such, 
- the main goal is to generate a set of html files that look nice and make it easier for users to browse the documentation.
  - the real goal is to cover all of the desires presented in the discussion thread, which should be covered decently here.
- a subgoal for me has been to touch the source markdown as little as possible, 
    - but it has been necessary to move the files a little deeper into a directory structure to keep the mkdocs parts from floating out to the rest of the project as well. this is the source of the vast majority of changed files
    - the other changes i've to the markdown files have been simple improvements that are beneficial regardless.
- the markdown-files used to generate the [github wiki](https://github.com/YaLTeR/niri/wiki) are the same used to generate the mkdocs docs. (point 6 in the discussion thread)
- the existing rustdoc documentation is also still available, in a subdirectory (`/niri/rustdoc/`)
    - this can be changed in the pipeline, e.g. so that `/niri/` leads to the rustdocs, but `/niri/docs/`  leads to the mkdocs docs

i believe i've covered these points reasonably well with this PR.
as such, github wiki, markdown previews in github, and the mkdocs docs can coexist, probably indefinitely.

---

one thing to note: the pipeline for deploying the github pages does a lot of things now.
it generates the rustdoc, and generates the mkdocs site, then it collects those various things and puts it into a publish directory, then that directory gets published to github pages.

---

a few things remain in order for me to feel like it's truly ready to merge.

- [x] address the `TODO`s in `docs/mkdocs.yaml` (they remain configured to my name/repo so that i can present a reasonable demo page, this is the easy step)
- [ ] wait for pygments to approve and merge my PR to support KDL: https://github.com/pygments/pygments/pull/2936 (TODO in docs/pyproject.toml`, also, no idea how long this could take)
- [ ] ... maybe configure a proper icon? shouldn't block this PR, but it would be nice to see our own favicon and such.

### DIFF
--- a/docs/wiki/Example-systemd-Setup.md
+++ b/docs/wiki/Example-systemd-Setup.md
@@ -18,7 +18,7 @@ Unlike [`spawn-at-startup`](./Configuration:-Miscellaneous.md#spawn-at-startup),
     So we will make our own.
     Create `~/.config/systemd/user/swaybg.service` with the following contents:
 
-    ```
+    ```systemd
     [Unit]
     PartOf=graphical-session.target
     After=graphical-session.target
@@ -43,7 +43,7 @@ Unlike [`spawn-at-startup`](./Configuration:-Miscellaneous.md#spawn-at-startup),
 4. `swayidle` similarly does not provide a service, so we will also make our own.
     Create `~/.config/systemd/user/swayidle.service` with the following contents:
 
-    ```
+    ```systemd
     [Unit]
     PartOf=graphical-session.target
     After=graphical-session.target


chinatsu:
also lol, we should really squash this

.github/workflows/ci.yml
```diff
@@ -249,21 +249,10 @@ jobs:
       - run: nix flake check
         continue-on-error: true
 
-  check-links:
-    runs-on: ubuntu-24.04
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          show-progress: false
-      - uses: lycheeverse/lychee-action@v2.0.2 # later versions break fragment checks. don't bump until this is fixed: https://github.com/lycheeverse/lychee/issues/1574
```
YaLTeR, line 300: What about this btw? Failing the CI if links are broken is useful

.github/workflows/ci.yml
```diff
@@ -273,28 +262,46 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
         with:
-          show-progress: false
-
+          lfs: true
+      - name: Configure Git Credentials
+        run: |
+          git config user.name github-actions[bot]
+          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
```
YaLTeR: Is this needed?

docs/mkdocs.yaml
```diff
@@ -0,0 +1,90 @@
...
+      anchor_linenums: true
+      line_spans: __span
+      pygments_lang_class: true
+  - pymdownx.inlinehilite
+  - pymdownx.snippets
+  - pymdownx.superfences
+  - admonition
+  - pymdownx.details
+  - pymdownx.keys
+  - attr_list
```
YaLTeR: What are all these other extensions for?

docs/mkdocs.yaml
```diff
@@ -0,0 +1,90 @@
...
+plugins:
+  - search
+hooks:
+  - hooks/shortcodes.py
+  - hooks/remove-must-fail.py
```
YaLTeR, line 50: Do we use must-fail anywhere? If not maybe can remove

docs/mkdocs.yaml
```diff
@@ -0,0 +1,90 @@
...
+  links:
+    not_found: warn
+    anchors: warn
+    absolute_links: relative_to_docs
+    unrecognized_links: warn
+not_in_nav: |
+  _Sidebar.md
+  Configuration:-Overview.md
+  index.md
+# ah, wouldn't it be nice if we could autogenerate this with wiki/_Sidebar.md
```
YaLTeR, line 68: Can't we?

docs/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[project]
+name = "niri"
+version = "0.1.0"
+description = "Add your description here"
```
YaLTeR: Is this field required?

docs/hooks/shortcodes.py
```diff
@@ -0,0 +1,39 @@
...
+        return _badge_for_version(args)
+
+    return re.sub(
+        r"<sup>Since: (.*?)</sup>",
+        replace, markdown, flags = re.I | re.M
+    )
+
+def _badge_for_version(version: str):
+    path = f"https://github.com/YaLTeR/niri/releases/tag/v{version}"
+    return f"<span class=\"badge\">[Since {version}]({path})</span>"
```
YaLTeR:
```suggestion
    return f"<span class=\"badge\">[Since: {version}]({path})</span>"
```

.github/workflows/ci.yml
```diff
@@ -273,28 +262,46 @@ jobs:
...
-
-      - name: Generate documentation
+      - name: Generate rustdoc documentation
         run: cargo doc --no-deps -p niri-ipc
-
       - run: cp ./resources/rustdoc-index.html ./target/doc/index.html
-
+      - name: Make publish directory
+        run: mkdir -p publish/niri_ipc
+      - name: Copy rustdoc documentation to publish
```
YaLTeR: These steps don't really need names

.github/workflows/ci.yml
```diff
@@ -273,28 +262,46 @@ jobs:
...
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    needs:
+      - build
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
         with:
-          show-progress: false
-
+          lfs: true
```
YaLTeR, line 318: `show-progress: false` to avoid spam

YaLTeR (commented):
Thanks a lot again for working on this!

I think, apart from the other stuff, I might want to tweak the theme. Or maybe there are some preset themes I could choose from? With the current one I don't quite like some dark mode colors and contrast.

Speaking of theme, can it switch light/dark automatically depending on the desktop preference?

Also, would you mind writing a development wiki page with how to build/serve the docs locally for testing?

> and the existing rustdocs can be found here https://chinatsu.github.io/niri/rustdoc

It can not in fact be found there

docs/mkdocs.yaml
chinatsu: We use must-fail once. Removing it made a rust test fail, so I decided to keep it and instead remove it with this script.

docs/mkdocs.yaml
chinatsu: We can! Surely! I just don't feel like diving into that rabbit hole (just yet?)

docs/pyproject.toml
chinatsu: Unsure, probably not. I'll look at it soontm

.github/workflows/ci.yml
chinatsu: Probably not, things should run fine without.

.github/workflows/ci.yml
chinatsu: I have enabled strict mode in the mkdocs config. This will fail the build if links are broken by itself, rendering lychee superfluous imo :)

chinatsu:
> It can not in fact be found there

Sorry! I have been moving it around to different names to see what feels better. It's now at https://chinatsu.github.io/niri/niri_ipc and I have edited the pr to reflect that

docs/mkdocs.yaml
YaLTeR: Right, I see

docs/mkdocs.yaml
YaLTeR: Should be a simple Python script. Don't need to parse full MD or anything, just autogenerate one from the other the simplest way possible

YaLTeR:
> Sorry! I have been moving it around to different names to see what feels better. It's now at https://chinatsu.github.io/niri/niri_ipc and I have edited the pr to reflect that

As far as I can tell, this even preserves existing links to the current rustdoc

docs/mkdocs.yaml
chinatsu: [Code blocks](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/) desire 
```yaml
markdown_extensions:
  - pymdownx.highlight:
      anchor_linenums: true
      line_spans: __span
      pygments_lang_class: true
  - pymdownx.inlinehilite
  - pymdownx.snippets
  - pymdownx.superfences
```

`admonition` is strictly not necessary, it seems, as `github-callouts` sorts out admonitions on its own

`pymdownx.keys` makes these.. key-looking things (https://chinatsu.github.io/niri/Getting-Started.html#main-default-hotkeys)

`attr_list` might not be necessary anymore, but it lets us add css classes to markdown elements: https://python-markdown.github.io/extensions/attr_list/
i believe this is needed for.. one of the other extensions? but i'm frankly not quite sure.

~~regardless, i'm doing a little bit of investigation to see which ones i can remove right now.~~

edit: i've removed the ones i think i can remove

docs/hooks/shortcodes.py
chinatsu: i found [another thing that should be fixed in the shortcodes script](https://chinatsu.github.io/niri/Configuration%3A-Debug-Options.html#deactivate-unfocused-windows), so i'll sort this out at the same time

chinatsu:
> As far as I can tell, this even preserves existing links to the current rustdoc

that's good! it's a little annoying that the index file in `/niri/niri_ipc` redirects to `/niri/niri_ipc/niri_ipc` though (doubling the name :thinking: )

> I think, apart from the other stuff, I might want to tweak the theme. Or maybe there are some preset themes I could choose from? With the current one I don't quite like some dark mode colors and contrast.

i have been thinking about this. it is totally possible to tweak the theme, either [within the confines of mkdocs-material](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#custom-colors) (which i do prefer because it comes with a lot of other very useful features), or with a [different mkdocs theme](https://github.com/mkdocs/catalog#-theming)

edit: my idea for colors, at least, is that i'm interested in seeing where the logo development ends up. it would be nice to reference some of the colors used there in a theme more specific to the project. but since it's all kind of up in the air at the moment, i haven't felt a need to make a draft for something like this yet.

> Speaking of theme, can it switch light/dark automatically depending on the desktop preference?

yep! i haven't configured it yet though. https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#color-palette-toggle

> Also, would you mind writing a development wiki page with how to build/serve the docs locally for testing?

certainly :)

docs/mkdocs.yaml
chinatsu: probably, yeah. but for now i think i'd prefer to leave the navigation configured like this, even though it means we'd have to change it in two places if we do decide to move pages around. 

(ps: i'm leaving conversations answered with my opinion answer unresolved, so if you feel my answer's sufficient feel free to hit that resolve button!)

chinatsu:
it's not perfect just yet, since the pygments PR is still open, and the logo discussion is (though pretty stable at this point) still not 100% decided.

that said, as a compromise i suggest we can be happy with using my pygments fork with KDL support tentatively, until the PR has been merged.

i have also chosen.. _one_ of the logos, but i'm not sure if it's the preferred one at this moment.

YaLTeR:
Let's go!!

Fireye04:
Woooo!!!

PR #2163, opened 2025-08-01
Add the `LoadConfigFile` action

vanderlokken:
The new action forces config file reloading, allowing to reduce lag between file updates and receiving visual feedback from the compositor.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1924,6 +1924,8 @@ pub enum Action {
     SetWindowUrgent(u64),
     #[knuffel(skip)]
     UnsetWindowUrgent(u64),
+    #[knuffel(skip)]
+    LoadConfigFile,
 }
 
 impl From<niri_ipc::Action> for Action {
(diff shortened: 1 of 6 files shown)


niri-config/src/lib.rs
```diff
@@ -1922,6 +1922,7 @@ pub enum Action {
     SetWindowUrgent(u64),
     #[knuffel(skip)]
     UnsetWindowUrgent(u64),
+    LoadConfigFile,
```
YaLTeR, line 1928: I don't think it's needed as a bindable action, at least for now

niri-ipc/src/lib.rs
```diff
@@ -840,6 +840,8 @@ pub enum Action {
         #[cfg_attr(feature = "clap", arg(long))]
         id: u64,
     },
+    /// Load config file.
```
```suggestion
    /// Reload the config file.
```

src/main.rs
```diff
@@ -269,27 +268,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        event_loop
-            .handle()
-            .insert_source(rx, |event, _, state| match event {
-                calloop::channel::Event::Msg(config) => state.reload_config(config),
-                calloop::channel::Event::Closed => (),
-            })
-            .unwrap();
-        watcher
-    };
+    state.setup_config_file_watcher(&watch_path);
```
YaLTeR: We have `dbus::start()` and `xwayland::satellite::setup()` so perhaps this would be more consistent as a `watcher::setup()`

YaLTeR:
Thanks!

PR #2168, opened 2025-08-02
niri-config: support merging particular fields

bbedward:
This is my attempt at "part 1" of supporting `!include` or `!import` directives. Since this got so massive, I wanted to get it up for review because merging alone is still probably valuable.

Two main changes:
1) Because the existing `niri-config/lib.rs` was very monolithic and hard to work with (~5700 lines) - I tried to break it down into logical modules mostly representing  top-level config paths, but also some other files for common things. e.g. `animation, appearance, binding, debug, gestures, input...`.
2) Add merging, for some but not all config directives (animations, gestures, environment, bindings, layout) - at least ones that I think would be more commonly desired to merge. The approach is to loop them all and combine using merge_with()

I know #1 makes it much tougher to review so sorry about that, but they are separate commits at least so they can be reviewed semi-separately.


YaLTeR:
Did the config split on main (slightly differently than in this PR but same idea)

bbedward:
I'm just gonna close it and come back to the merging later, since it needs re-impl anyway

PR #2195, opened 2025-08-06
niri-config: add disable-set-bpc option

bbedward:
The setting of max bpc to 8 bricks some oled panels driven by amdgpu, the fix once it occurs is to kill niri and reset the max BPC value `modeset -M amdgpu -w <CONNECTOR_ID>:"max bpc":16` This of course requires an external display (driven by the other gpu, if it exists), as the framebuffer remains gone indefinitely after niri is started.

At least, it's true for:
Asus Zephrys G16 GA605WI laptop
AMD Radeon 890M
Samsung ATNA60DL04 16" OLED 240Hz display

At least one other user reported this issue but not sure of the hardware.

```
niri::backend::tty: set max bpc to 8
Aug 05 15:59:59 archy niri[1050]: 2025-08-05T19:59:59.943422Z  WARN niri::backend::tty: error creating DRM compositor, will >
Aug 05 15:59:59 archy niri[1050]: 2025-08-05T19:59:59.944527Z  WARN niri::backend::tty: error connecting connector: error cr>
Aug 05 15:59:59 archy niri[1050]: Caused by:
Aug 05 15:59:59 archy niri[1050]:     0: The underlying drm surface encountered an error: DRM access error: Error testing st>
Aug 05 15:59:59 archy niri[1050]:     1: DRM access error: Error testing state on device Some("/dev/dri/card0") (Cannot al>
Aug 05 15:59:59 archy niri[1050]:     2: Cannot allocate memory (os error 12)
```

This does not occur on any other compositor I tried: mutter, sway, hyprland, cosmic

---

the debug flag at least allows me and others to use niri on these systems where bpc bricks the displays.

a real fix path may look like:
- Determine if setting max bpc to 8 is having the desired effect or if its necessary to override it at all
- Determine if there's a way to detect if driver supports it before setting it
- If setting max bpc to 8 is a necessary step, then may have to report to `amdgpu` and await a fix, in either case something like this config is necessary at least because that may take 1 to multiple linux releases before a fix in the driver came to fruition - and OLED+AMD is a pretty common combo in modern laptops, so if this affects a decent percentage of them then a significant amount of users would be effected - so makes sense to give them some path that works.

Maybe relevant, when the display is driven my the nvidia gpu (an the particular asus laptop it supports muxing, which means it can drive the integrated display by the NVIdia gpu) - niri works, but it doesn't set bpc anyway with:
```
error setting max bpc: couldn't find max bpc property
```

So maybe this does point to an issue in the `amdgpu` driver or just an implementation difference between nvidia proprietary, not sure.

### DIFF
--- a/wiki/Configuration:-Debug-Options.md
+++ b/wiki/Configuration:-Debug-Options.md
@@ -297,11 +297,12 @@ debug {
 
 <sup>Since: next release</sup>
 
-Skips setting max bpc on startup.
+When connecting monitors, niri sets their max bpc to 8 in order to reduce display bandwidth and to potentially allow more monitors to be connected at once.
+Restricting bpc to 8 is not a problem since we don't support HDR or color management yet and can't really make use of higher bpc.
 
-Some displays driven by amdgpu encounter a broken DRM state when setting max bpc to 8. Niri sets this on startup for increases display compatibility. See [this](https://gitlab.freedesktop.org/drm/amd/-/issues/4487) bug report for more details.
-
-Setting this disables this behavior. It may cause issues with some displays.
+Apparently, setting max bpc to 8 breaks some displays driven by AMDGPU.
+If this happens to you, set this debug flag, which will prevent niri from changing max bpc.
+AMDGPU bug report: https://gitlab.freedesktop.org/drm/amd/-/issues/4487.
 
 ```kdl
 debug {


HigherOrderLogic:
Did you have your monitor in HDR mode by any chance?

niri-config/src/lib.rs
```diff
@@ -2339,6 +2339,8 @@ pub struct DebugConfig {
     #[knuffel(child)]
     pub disable_direct_scanout: bool,
     #[knuffel(child)]
+    pub disable_set_bpc: bool,
```
YaLTeR: Maybe `keep_max_bpc_unchanged`?

Also needs to be mentioned in the debug options wiki page.

YaLTeR (commented):
> * Determine if setting max bpc to 8 is having the desired effect or if its necessary to override it at all
> 
> * Determine if there's a way to detect if driver supports it before setting it

The driver already reports ranges for all properties and is free to refuse setting a property to a particular value. So I think this should be reported to amdgpu ([this issue tracker](https://gitlab.freedesktop.org/drm/amd/-/issues) I guess).

niri-config/src/lib.rs
bbedward: Added and renamed

bbedward:
> > * Determine if setting max bpc to 8 is having the desired effect or if its necessary to override it at all
> > * Determine if there's a way to detect if driver supports it before setting it
> 
> The driver already reports ranges for all properties and is free to refuse setting a property to a particular value. So I think this should be reported to amdgpu ([this issue tracker](https://gitlab.freedesktop.org/drm/amd/-/issues) I guess).

Created [this issue](https://gitlab.freedesktop.org/drm/amd/-/issues/4487)

> Did you have your monitor in HDR mode by any chance?

It's an integrated laptop display without an OSD so I can't explicitly override the mode that it's in. It supports HDR but its not active within niri.

wiki/Configuration:-Debug-Options.md
```diff
@@ -292,6 +293,22 @@ debug {
...
 
+### `keep-max-bpc-unchanged`
+
+<sup>Since: next release</sup>
+
+Skips setting max bpc on startup.
+
+Some displays driven by amdgpu encounter a broken DRM state when setting max bpc to 8. Niri sets this on startup for increases display compatibility. See [this](https://gitlab.freedesktop.org/drm/amd/-/issues/4487) bug report for more details.
+
+Setting this disables this behavior. It may cause issues with some displays.
```
YaLTeR:
```suggestion
When connecting monitors, niri sets their max bpc to 8 in order to reduce display bandwidth and to potentially allow more monitors to be connected at once.
Restricting bpc to 8 is not a problem since we don't support HDR or color management yet and can't really make use of higher bpc.

Apparently, setting max bpc to 8 breaks some displays driven by AMDGPU.
If this happens to you, set this debug flag, which will prevent niri from changing max bpc.
AMDGPU bug report: https://gitlab.freedesktop.org/drm/amd/-/issues/4487.
```

YaLTeR:
Run cargo +nightly fmt pls

bbedward:
> Run cargo +nightly fmt pls

Done, sorry

wiki/Configuration:-Debug-Options.md
YaLTeR:
```suggestion
When connecting monitors, niri sets their max bpc to 8 in order to reduce display bandwidth and to potentially allow more monitors to be connected at once.
Restricting bpc to 8 is not a problem since we don't support HDR or color management yet and can't really make use of higher bpc.

Apparently, setting max bpc to 8 breaks some displays driven by AMDGPU.
If this happens to you, set this debug flag, which will prevent niri from changing max bpc.
AMDGPU bug report: https://gitlab.freedesktop.org/drm/amd/-/issues/4487.
```

YaLTeR:
Thanks. Test failure is unrelated

PR #2207, opened 2025-08-09
WIP: allow passing skip-animation for focus actions in config and IPC

maxbol:
> Nb: This is more of a conversation starter than a full-fledged, ready commit.

I want animations in niri, especially when navigating with my touchpad or mouse. I think turning off animations for focus actions when doing gestures based navigation looks and feels weird, because the snapping becomes to hard.

BUT

I find myself often having a few full-sized text editors open when working on code, and want to be able to switch between them super quickly, without loosing mental bandwidth to watching an animation.

This change makes it possible to provide an argument to *some* focus actions called `skip-animation`, that allows you to accomplish both of these things. I can configure my keyboard shortcuts to allow for instantaneous jumps between columns and workspaces, while at the same time preserving the smooth gestures based navigation that I want to have. I'm dogfooding it right now and it works perfectly based on my needs. The change is backwards compatible in that both config and IPC still accepts a parameter-less call to the actions in question. 

If you like this direction, I can pull together a full PR with this applied to every action where it makes sense. 

Before doing that, maybe some discussion?
* Is this something anyone else wants? Can you see the use for it?
* Is there some other, more non-intrusive way to do it? (Maybe being able to set a global config option that turns off all animations that would have been triggered by keyboard hotkeys?)
* Should the approach be more generalized? (Ie the possibility of passing a full animation config to the action instead of simply a `skip_animation` boolean?)
* I have never written a line of Rust in my life before this, so let me know if I messed something up!


FullStackAlex:
I would like to disable animations on certain actions / key binds too.

PandorasFox:
I'm a bit interested in the emitted IPC events for workspace/window changes including whether it's a skip-animation change or not - I'm glueing a parallax wallpaper daemon to the event ipc stream, so knowing if a a workspace change was animated is relevant there.

YaLTeR:
That will break with touchpad/screen/mouse gestures. I'm not prepared to emit an IPC event every frame for those

maxbol:
> I'm a bit interested in the emitted IPC events for workspace/window changes including whether it's a skip-animation change or not - I'm glueing a parallax wallpaper daemon to the event ipc stream, so knowing if a a workspace change was animated is relevant there.

I'm not entirely sure that I grasp this or the relevance to this PR. Could there be an issue where other plugins/applications are expecting to be able to go to the config for the exact animation timings etc for a specific transition, and now because of `skip_animation` that might not be the case anymore?

Atm this PR does not change anything about emitted events, nor was I intending to.

PandorasFox:
> > I'm a bit interested in the emitted IPC events for workspace/window changes including whether it's a skip-animation change or not - I'm glueing a parallax wallpaper daemon to the event ipc stream, so knowing if a a workspace change was animated is relevant there.
> 
> I'm not entirely sure that I grasp this or the relevance to this PR. Could there be an issue where other plugins/applications are expecting to be able to go to the config for the exact animation timings etc for a specific transition, and now because of `skip_animation` that might not be the case anymore?
> 
> Atm this PR does not change anything about emitted events, nor was I intending to.

It's mostly for handling how the wallpaper state transitions - if you're doing a wallpaper-per-workspace, and sometimes skip-animation when switching workspaces, it's necessary to know whether the wallpaper should crossfade or just insta-switch to the new buffer.

I don't think that granularly tracking the animation states is necessary here, since parallax animations in a different z-layer are only loosely coupled (their durations and interps can differ) - it's just the singular bit of "this is an (animated, interpolated) state change" vs "this is an instant state change" that's desired

maxbol:
> > > I'm a bit interested in the emitted IPC events for workspace/window changes including whether it's a skip-animation change or not - I'm glueing a parallax wallpaper daemon to the event ipc stream, so knowing if a a workspace change was animated is relevant there.
> > 
> > 
> > I'm not entirely sure that I grasp this or the relevance to this PR. Could there be an issue where other plugins/applications are expecting to be able to go to the config for the exact animation timings etc for a specific transition, and now because of `skip_animation` that might not be the case anymore?
> > Atm this PR does not change anything about emitted events, nor was I intending to.
> 
> It's mostly for handling how the wallpaper state transitions - if you're doing a wallpaper-per-workspace, and sometimes skip-animation when switching workspaces, it's necessary to know whether the wallpaper should crossfade or just insta-switch to the new buffer.
> 
> I don't think that granularly tracking the animation states is necessary here, since parallax animations in a different z-layer are only loosely coupled (their durations and interps can differ) - it's just the singular bit of "this is an (animated, interpolated) state change" vs "this is an instant state change" that's desired

Ok! Well I'd be happy to accommodate that if people think the overall direction of this is good. So if that's the case I'd happily spend some time soon making it cover all cases/actions, and if that means adding something to the event stream as well I'd happily do that too.

@YaLTeR  - this is *currently* doing everything *I* need it to do, but ofcourse I'd love to upstream it. Do you have any hesitations to the direction of this before I spend some time making it merge-ready?

PandorasFox:
I'd also be happy to hack on this and work in the bool to the events afterwards - I didn't mean to come across as asking for this to-be-added-before-merge, I'm just thinking about a tangent here :)

YaLTeR:
fwiw I think per-workspace wallpapers should be compositor-side because you can do it properly there. In niri, even without image loading so far, you can imagine a workspace option to place a specific layer surface on its background (similar to how place-within-backdrop can take a specific layer surface and place it in a different location).

For this present PR, I haven't got around to it yet sry

MintyDoggo:
I feel that this is slightly related:

I was looking for a way to decouple the `WorkspaceSwitchAnim` so you can have granular control of the animations between changing workspaces via gestures, keybinds, etc. Currently, it looks like they all share `WorkspaceSwitchAnim`

For me personally I like to have no animations when switching workspaces via keybinds, but i want animations when using gestures (similar to original post). The fix to add `skip-animations` would solve my use case, but it would be cool to have granular control over each animation

I'm not familiar with the project source but would this be something I could try contributing to? I’d be happy to give it a shot if @YaLTeR think it’s a useful direction

PR #2210, opened 2025-08-09
Workspace switch overview shadow

IsseW:
This enables rendering of workspace overview shadows while in a workspace switch. Best visible when `mod + middle mouse` dragging, or three finger drag gesture on a trackpad.

I also extended the rectangle that's checked when checking if a workspace's shadow should be drawn. Without this the shadow pops into view. This could be split into a separate PR if wanted though.

### DIFF
--- a/src/layout/monitor.rs
+++ b/src/layout/monitor.rs
@@ -1570,7 +1570,11 @@ impl<W: LayoutElement> Monitor<W> {
 
         let scale = self.scale.fractional_scale();
         let zoom = self.overview_zoom();
-        let overview_clamped_progress = self.overview_progress.as_ref().map(|p| p.clamped_value());
+        let overview_clamped_progress = self
+            .workspace_switch
+            .is_some()
+            .then_some(1.0)
+            .or(self.overview_progress.as_ref().map(|p| p.clamped_value()));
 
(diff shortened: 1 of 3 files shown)


YaLTeR:
It's intentionally not drawn while not in the overview to avoid a costly shader during normal operation

IsseW:
> It's intentionally not drawn while not in the overview to avoid a costly shader during normal operation

That makes sense, this PR only adds drawing them if there is a workspace switch in progress. Where the space between workspaces is shown.

YaLTeR:
Yes, I'm talking about when a workspace switch is in progress. When it isn't, it doesn't matter because it would get culled by the damage tracker anyway

IsseW:
> Yes, I'm talking about when a workspace switch is in progress. When it isn't, it doesn't matter because it would get culled by the damage tracker anyway

Okay fair, I didn't see it as that much of a cost. Would this be something you'd consider merging if it was a config? On my setup I was annoyed by it not rendering at least.

YaLTeR:
I'm not a big fan of adding config options for minor things like these tbh. So far I don't think I'm convinced that this is useful enough to add an option. GNOME also doesn't have a shadow during workspace switch fwiw

PR #2225, opened 2025-08-13
use ScanoutCandidate kind for surfaces

cmeissl:
this allows to get direct scan-out on surfaces again


YaLTeR:
Primary plane scanout does work, but overlay planes started do nothing without this change, yeah. Thanks

PR #2240, opened 2025-08-15
feat: `open-maximized-if-alone` window rule

equals03:
This PR adds a new `open-maximized-if-alone` window rule. 

As you can imagine It makes a window open maximized only if it's the only one on the workspace.
This was a feature I missed when I made the move from Hyprland and the only real mention I could find was here https://github.com/YaLTeR/niri/discussions/859, so I gave it a stab.

Set `open-maximized-if-alone = true`, and the window maximizes only if it's alone on the workspace. If other windows are around, it follows other rules or defaults as usual.

```kdl
window-rule {
    match app-id="zen$"
    match app-id="firefox$"
    
    open-maximized-if-alone true
}
```

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1423,6 +1423,8 @@ pub struct WindowRule {
     #[knuffel(child, unwrap(argument))]
     pub open_maximized: Option<bool>,
     #[knuffel(child, unwrap(argument))]
+    pub open_maximized_if_alone: Option<bool>,
+    #[knuffel(child, unwrap(argument))]
     pub open_fullscreen: Option<bool>,
     #[knuffel(child, unwrap(argument))]
     pub open_floating: Option<bool>,
@@ -4249,6 +4251,7 @@ mod tests {
 
                 open-on-output "eDP-1"
                 open-maximized true
+                open-maximized-if-alone true
                 open-fullscreen false
                 open-floating false
                 open-focused true
@@ -5027,6 +5030,9 @@ mod tests {
                     ),
                     open_on_workspace: None,
                     open_maximized: Some(
+                        true,
+                    ),
+                    open_maximized_if_alone: Some(
                         true,
                     ),
                     open_fullscreen: Some(
--- a/src/handlers/xdg_shell.rs
+++ b/src/handlers/xdg_shell.rs
@@ -917,7 +917,8 @@ impl State {
         let mut floating_width = None;
         let mut height = None;
         let mut floating_height = None;
-        let is_full_width = rules.open_maximized.unwrap_or(false);
+        let wants_full_width = rules.open_maximized.unwrap_or(false);
+        let wants_full_width_if_alone = rules.open_maximized_if_alone.unwrap_or(false);
         let is_floating = rules.compute_open_floating(toplevel);
 
         // Tell the surface the preferred size and bounds for its likely output.
@@ -930,7 +931,7 @@ impl State {
                     .or_else(|| self.niri.layout.active_workspace())
             });
 
-        if let Some(ws) = ws {
+        let is_full_width = if let Some(ws) = ws {
             // Set a fullscreen state based on window request and window rule.
             if (wants_fullscreen.is_some() && rules.open_fullscreen.is_none())
                 || rules.open_fullscreen == Some(true)
@@ -944,6 +945,9 @@ impl State {
             floating_width = ws.resolve_default_width(rules.default_width, true);
             height = ws.resolve_default_height(rules.default_height, false);
             floating_height = ws.resolve_default_height(rules.default_height, true);
+
+            let is_full_width =
+                wants_full_width || (wants_full_width_if_alone && !ws.has_windows());
 
             let configure_width = if is_floating {
                 floating_width
@@ -960,7 +964,11 @@ impl State {
                 is_floating,
                 &rules,
             );
-        }
+
+            is_full_width
+        } else {
+            wants_full_width
+        };
 
         // Set the tiled state for the initial configure.
         update_tiled_state(toplevel, config.prefer_no_csd, rules.tiled_state);
--- a/src/tests/window_opening.rs
+++ b/src/tests/window_opening.rs
@@ -415,6 +415,7 @@ fn target_size() {
     // * want fullscreen
     // * open-fullscreen
     // * open-maximized
+    // * open-maximized-if-alone
     // * open-floating
     // * default-column-width
     // * border
@@ -430,6 +431,7 @@ fn target_size() {
         WantFullscreen::AfterInitial(None),
     ];
     let open_maximized = [None, Some("true")];
+    let open_maximized_if_alone = [None, Some("true")];
     let open_floating = [None, Some("true")];
     let default_column_width = [
         None,
@@ -450,12 +452,14 @@ fn target_size() {
     for fs in open_fullscreen {
         for wfs in want_fullscreen {
             for om in open_maximized {
-                for of in open_floating {
-                    for dw in default_column_width {
-                        for dh in default_window_height {
-                            for b in border {
-                                for t in tabbed {
-                                    powerset.push((fs, wfs, om, of, dw, dh, b, t));
+                for omia in open_maximized_if_alone {
+                    for of in open_floating {
+                        for dw in default_column_width {
+                            for dh in default_window_height {
+                                for b in border {
+                                    for t in tabbed {
+                                        powerset.push((fs, wfs, om, omia, of, dw, dh, b, t));
+                                    }
                                 }
                             }
                         }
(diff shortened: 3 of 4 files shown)


equals03:
rats.. I realise that I have a merge commit from when I updated the fork. 

if you are even interested in this change, let me know and I'll rebase over latest and submit again.

HigherOrderLogic:
Can this be extended into something more generic, eg: a match rule that matches the first window open in a workspace?

equals03:
> Can this be extended into something more generic, eg: a match rule that matches the first window open in a workspace?

Initially I was thinking about implementing something like the Hyprland workspace rules `t[1-3]` etc. but in terms of the Niri rule resolution it gets a little messy as the workspace doesn't seem to be available at resolution time (I could be wrong!). 

I'll give it another look though - I wasn't even sure anyone other than me was interested in something like this. :+1:

YaLTeR:
Does this make sense as a window rule? Rather than, say, a workspace rule. We don't have workspace rules yet but that doesn't mean that we should shove things that would make more sense on a workspace into window rules

YaLTeR:
Also yeah idk about this in general because it kinda violates the opening new windows doesn't affect other window sizes principle

equals03:
> Also yeah idk about this in general because it kinda violates the opening new windows doesn't affect other window sizes principle

In my mind it doesn't actually violate it. it only has an affect on its match. ie. `open-maximized` opens maximized always, this just has an extra condition and it doesn't influence the size/position of any other window. Your call though.

> Does this make sense as a window rule? Rather than, say, a workspace rule. We don't have workspace rules yet but that doesn't mean that we should shove things that would make more sense on a workspace into window rules

I think, in this particular case, it makes sense on a window. I know I probably have a different use-case but if I'm opening vscode/firefox and there is nothing else on the workspace, I want it to use the entire space. If there is something else there, I want it to behave as normal. At the same time, I don't mind my terminals not expanding. So enforcing this on a workspace level would be all or nothing unless, ofc, there were excludes... but then i would have to opt out... this seems simpler and in my rules setup quite easy to manage.

However, I will concede that it may be convenient to match this behaviour to specific workspaces only and have a different set of matches/excludes. 

Either way, this looked like a very easy implementation for a feature I miss :) 

again, your project, your call.

YaLTeR:
FWIW I'm using at-startup=true open-maximized true rule for Firefox so it opens fullscreen when autostarted and as a half split otherwise

equals03:
> FWIW I'm using at-startup=true open-maximized true rule for Firefox so it opens fullscreen when autostarted and as a half split otherwise

great its working for you :) my use-case is completely different but so be it. 
I think if you aren't interested/not convinced go ahead and close the PR.

anyways, regardless, thanks for an epic project - appreciate all the effort!

Bodify:
I was scouring the niri wiki looking for a window rule like this. I hope you reconsider and add it or something similar eventually. I use half width for everything except browser. I would love for it to be able to check my workspace isnt empty and thus open at half width and fit perfectly. This would be lovely for my use case. I've only been maining niri for a couple of days now though and I just wanted to share my enthusiasm for this rule.

YaLTeR:
Btw yeah you can script this using the IPC. Check if window is alone on a workspace and matches the rule, then maximize it

Bodify:
> Btw yeah you can script this using the IPC. Check if window is alone on a workspace and matches the rule, then maximize it

Thank you for the info, I will go try do that. For a newbie like me though I am still eyeing the simple window rule. Not too sure how all this works, especially through an App Launcher but I will definitely spend some time to figure it out and learn some things along the way. Thank you!

Zenkazio:
Hey, whats up with this? I was about to suggest the same functionality, but a bit different, as a match.

```
window-rule {
    match is_first_on_workspace=true

    open-maximized true
} 
```

Anyway I think until this is decided. I will write a background script to do it with `niri msg event-stream`

heyoeyo:
I have an [IPC listener script](https://github.com/heyoeyo/niri_tweaks?tab=readme-ov-file#niri_tile_to_npy) that can do this solo-window maximization, for anyone interested. The script is actually for making niri act like a auto-tiler up to 'N' windows, but if you run it with the `-n 1` flag, it will just do solo-window maximization.

I'll throw in my 2 cents and vouch for the workspace rules approach to handle this more generally, specifically with a `match` option on window counts, something like:
```bash
workspace-rule {
  match tile-count=1

  open-maximized true
}
# Other possibilities?:
# match float-count
# match maximized-count
```

Workspace rules also seem like a good spot for a 'window spawning strategy' (like #2115), if that's ever considered. For example having: 'spawn-left', 'spawn-right' (the current default), 'stack-left', 'stack-right' (though stacking conflicts with the design principles tbf) and matching on window counts would allow for a wide range of customization to the behavior. It might also generalize to handling other cases, like #1929, which might be covered by 'spawn-right' on one monitor/workspace, 'spawn-left' on another.

og900aero:
I am also eagerly awaiting this feature. After the missing swallow feature, this would be an important feature for me too.

1Naim:
I'm currently running this on top of Niri v25.11, and it seems it doesn't work correctly for discord because of the splash window. Are there any workarounds planned for such software? (i imagine steam would suffer the same way)

Antiz96:
Just my 2 cents, but I think it would indeed be nice to allow such a rule to be generic to every windows, and not only to specific `app-id` matches (so that, globally speaking, the first window opened on a workspace gets maximized).

For the record, I'm **not** vouching for the next opened windows after that to automagically resize the first one (like it's done on automatic tillers), which indeed wouldn't make much sense on a scroller and would violate the overall principle / design. This is solely about having the first / only opened window being maximized automatically, nothing more.
Taking my own workflow as an example, most of the time I'm *intentionally* only opening a single window on a workspace with the expectation to dedicate the whole "screen space" to it *(I'm aware that "screen space" is a bit of a blurry notion in the context of a scroller, but you get the point I guess)*, specifically on a multi monitors setup with each window being maximized and spread across the different monitors. 
Therefore, more often than not, I find myself having to press the "toggle maximize window" keybind about 90% of the time, in order to maximize the first windows I open on my different workspaces. Assuming my workflow is legitimate, I guess it would make sense to have an option that allows for the opposite, namely having to press the "toggle maximize windows" only for the 10% of the time where I actually don't want the first window to be maximized. :smile:

Hopefully, such an option can be considered.  
In any case, thanks a lot for your wonderful work with Niri!

Antiz96:
> Btw yeah you can script this using the IPC. Check if window is alone on a workspace and matches the rule, then maximize it

That's a fair point actually, I just went ahead and did that.

For anyone interested, the tool is available at https://github.com/Antiz96/oniri.  
You can just install it, start it from your niri config (`spawn-at-startup "oniri"`) and it will automatically maximize the only window of a workspace, whether it's the first one opened or the last remaining one after all the other windows got closed (unless it is started with the `--first-only` option, in which case, only the first opened window gets maximized).

---

> Check if window is alone on a workspace and matches the rule, then maximize it

Allow me just a side note about the above quote: Due to current limitations of the niri IPC, it's unfortunately **not** as easy as "checking if window is alone on a workspace, then maximize it".  
Indeed, the IPC currently does not provide a way to *set* the maximize state of a window, it only provides a way to *toggle* it (see https://github.com/niri-wm/niri/pull/1968). That means we need to verify if a window is not already maximized before sending the "toggle maximize" action (otherwise we might end up minimizing a window unexpectedly), but the IPC doesn't expose the maximized state of a window either (see https://github.com/niri-wm/niri/pull/2836)...

As stated in [this comment](https://github.com/niri-wm/niri/pull/2270#issue-3337717998), for the lack of a better option, the workaround to determine if a window is *supposedly* maximized so far is to compare the window size with the output (monitor) logical size. However, simply expecting an equality between both is not an option because there are multiple things that may have an influence on a window size despite being maximized. The obvious one to mention here is [gaps](https://github.com/niri-wm/niri/wiki/Configuration:-Layout#gaps), but also things like panel / taskbar, which occupy some screen space. A certain "size difference tolerance" should therefore be applied to *hopefully* take those elements into account.  
This is obviously a pretty fragile and not elegant solution, which will not work as expected for specific setups (https://github.com/Antiz96/oniri/issues/3).

Exposing the "maximized" state of a window via the IPC or, better yet, providing a way to "set / unset" it rather than just toggling it (as respectively proposed in the two pull requests linked below) would significantly ease IPC scripting here.

- https://github.com/niri-wm/niri/pull/2836
- https://github.com/niri-wm/niri/pull/1968

I hope those two changes can make their way through the IPC at some point :pray: 
In any case, as always, thanks a lot for your great work with niri :hugs:

gruvw:
What about something like #3708: "Allow for multiple default-column-width node values".

Something like:

```yaml
default-column-width {
    proportion 1.0
    proportion 0.5
}
```

Would let us have the first window of a workspace open in full screen while keeping the principle at the core of Niri of "Opening a new window never causes existing windows to resize." true.

og900aero:
And would be great an option for maximize_to_edges not simple maximized.

PR #2244, opened 2025-08-15
flake: skip tests that require a valid EGL display

cole-h:
Otherwise:

    niri> failures:
    niri>
    niri> ---- tests::animations::clientside_height_change_doesnt_animate stdout ----
    niri>
    niri> thread 'tests::animations::clientside_height_change_doesnt_animate' panicked at src/tests/animations.rs:87:54:
    niri> called `Result::unwrap()` on an `Err` value: error creating EGL display
    niri>
    niri> Caused by:
    niri>     Unable to obtain a valid EGL Display.
    niri> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    niri>
    niri> ---- tests::animations::height_resize_animates_next_y stdout ----
    niri>
    niri> thread 'tests::animations::height_resize_animates_next_y' panicked at src/tests/animations.rs:87:54:
    niri> called `Result::unwrap()` on an `Err` value: error creating EGL display
    niri>
    niri> Caused by:
    niri>     Unable to obtain a valid EGL Display.
    niri>
    niri>
    niri> failures:
    niri>     tests::animations::clientside_height_change_doesnt_animate
    niri>     tests::animations::height_resize_animates_next_y
    niri>
    niri> test result: FAILED. 147 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.45s

---

Closes https://github.com/YaLTeR/niri/issues/2245.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -118,8 +118,7 @@
           checkFlags = [
             # These tests require the ability to access a "valid EGL Display", but that won't work
             # inside the Nix sandbox
-            "--skip=tests::animations::clientside_height_change_doesnt_animate"
-            "--skip=tests::animations::height_resize_animates_next_y"
+            "--skip=tests::animations
           ];
 
           postInstall =
--- a/flake.nix
+++ b/flake.nix
@@ -118,7 +118,7 @@
           checkFlags = [
             # These tests require the ability to access a "valid EGL Display", but that won't work
             # inside the Nix sandbox
-            "--skip=tests::animations
+            "--skip=tests::animations"
           ];
 
           postInstall =
--- a/flake.nix
+++ b/flake.nix
@@ -115,6 +115,13 @@
             export XDG_RUNTIME_DIR="$(mktemp -d)"
           '';
 
+          checkFlags = [
+            # These tests require the ability to access a "valid EGL Display", but that won't work
+            # inside the Nix sandbox
+            "--skip=tests::animations::clientside_height_change_doesnt_animate"
+            "--skip=tests::animations::height_resize_animates_next_y"
+          ];
+
           postInstall =
             ''
               installShellCompletion --cmd niri \


sodiboo (approved):
banged my head against this for like 2 hours, trying to get those tests to work in the Nix sandbox. tried a bunch of things.

this is what Smithay logs during the failure:

```
DEBUG smithay::backend::egl::display: Supported EGL client extensions: [""]
DEBUG smithay::backend::egl::display: Trying EGL platform: PLATFORM_SURFACELESS_MESA
 INFO smithay::backend::egl::display: Skipping EGL platform because one or more required extensions are not supported. Missing extensions: ["EGL_MESA_platform_surfaceless"]
ERROR smithay::backend::egl::display: Unable to find suitable EGL platform
```

i think it's not actually possible to do in the sandbox, because the decision whether to use mesa or nvidia OpenGL drivers is deferred until runtime; and in fact they may not even exist on all build machines (such as in CI).

it feels wrong that we can't have GL libraries in the Nix sandblox? like, yes, *GPU drivers* can't exist there, but why doesn't it like, fall back to LLVMpipe? is that even possible? am i missing something?

in the end, i gave up and just skip the tests like you do:

https://github.com/sodiboo/niri-flake/commit/083135e6d0c3d7af636c34602b85cc429aacf93c

so, this looks good to me, i guess. but i'm not happy about needing to do this.

sodiboo:
- closes #2245

YaLTeR:
@sodiboo have you checked how tests for Mutter work? I'm fairly sure it also wants EGL at test time (and it may even need the real one, not surfaceless like niri wants here).

flake.nix
```diff
@@ -115,6 +115,13 @@
             export XDG_RUNTIME_DIR="$(mktemp -d)"
           '';
 
+          checkFlags = [
+            # These tests require the ability to access a "valid EGL Display", but that won't work
+            # inside the Nix sandbox
+            "--skip=tests::animations::clientside_height_change_doesnt_animate"
+            "--skip=tests::animations::height_resize_animates_next_y"
```
YaLTeR:
```suggestion
            "--skip=tests::animations
```
Let's just do this if we do this perhaps

flake.nix
```diff
@@ -115,6 +115,12 @@
             export XDG_RUNTIME_DIR="$(mktemp -d)"
           '';
 
+          checkFlags = [
+            # These tests require the ability to access a "valid EGL Display", but that won't work
+            # inside the Nix sandbox
+            "--skip=tests::animations
```
cole-h:
```suggestion
            "--skip=tests::animations"
```

Oops that's what I get for applying from mobile.

sodiboo:
> have you checked how tests for Mutter work? I'm fairly sure it also wants EGL at test time.

[`-Dtests=disabled`](https://github.com/NixOS/nixpkgs/blob/3b0feaf8c2160a10a974c890e93f5f5387728784/pkgs/by-name/mu/mutter/package.nix#L92)?

though, this commit message https://github.com/NixOS/nixpkgs/commit/755387543f459d6edc31595a98f8d998ffb7ccca suggests they disabled the tests to get rid of a dependency on gtk3, and not because the tests were failing.

> it may even need the real one, not surfaceless like niri wants here

as far as i understand it, surfaceless is in fact "real" GL? just doesn't intrinsically output to anything. it's still using GPU resources.

providing a headless Wayland or Xorg server in the build environment is inconvenient, but doable.  
providing GPU resources at all, i believe is not possible by design.

YaLTeR:
Thanks

YaLTeR:
> as far as i understand it, surfaceless is in fact "real" GL? just doesn't intrinsically output to anything. it's still using GPU resources.

It's real GL, but it doesn't spawn a window (which comes with its own annoying requirements of having some windowing system to spawn it into)

PR #2249, opened 2025-08-16
niri: fix gpu hotplugging

athul-krishna-kr:
For the most part, `device_removed()` closes file descriptors opened on /dev/dri/card*, except the one
opened by `session.open()`.

* b2d061c: in this commit, session fd is closed on drm remove event
* a34c2cc: skip render node (/dev/dri/render*) on drm add event.

If render node is not skipped, it'll be opened in the udev event triggered via driver unbind and rebind
`seatd      633           root  18u   CHR 226,128      0t0 1503 /dev/dri/renderD128`.
And it cannot be closed by
`udevadm trigger --action=remove /dev/dri/renderD128` 
as `device_removed()` returns with `niri::backend::tty: no such device` error.

And so with the next driver unbind and rebind `device_added()` fails with
`device added: 57856 "/dev/dri/card0"`
`error adding device: no render node`.

wlroots also implements similar check (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/session/session.c) in `is_drm_card()`.

## hotplug test
* sudo udevadm trigger --action=remove /dev/dri/card0
* echo "0000:03:00.0" | sudo tee /sys/bus/pci/drivers/amdgpu/unbind
* echo "0000:03:00.0" | sudo tee /sys/bus/pci/drivers/amdgpu/bind

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,7 +97,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
 dependencies = [
  "android-properties",
- "bitflags 2.9.3",
+ "bitflags 2.9.4",
  "cc",
  "cesu8",
  "jni",
@@ -144,9 +144,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.11"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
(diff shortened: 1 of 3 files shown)


src/backend/tty.rs
```diff
@@ -496,6 +497,19 @@ impl Tty {
...
+            Ok(str) => {
+                if str.contains("render") {
+                    return Ok(());
+                }
+            }
+            Err(err) => {
+                error!("invalid path: {}", err);
+                return Ok(());
+            }
+        }
```
HigherOrderLogic:
```suggestion
        if let Some(path_os_str) = path.file_name() and let Some(path_str) = path_os_str.to_str() {
        	if path_os_str.contains("render") {
        		do_things();
        	} else {
        		do_things();
        	}
        }
```

Isnt it better to do it this way instead of blindly checking if the whole string contains the word `render`?

athul-krishna-kr: @HigherOrderLogic the udev monitor only listens for drm events. So I think that makes the possible file paths 
`/dev/dri/card*` and `/dev/dri/render*` only.

Smithay:
`backend/udev.rs: let monitor = MonitorBuilder::new()?.match_subsystem("drm")?.listen()?;`

I can make the changes if strict filtering is needed.

athul-krishna-kr:
@YaLTeR FYI

YaLTeR:
Sorry I haven't had a chance to look at this yet

YaLTeR:
Closing device fd now exists on main

PR #2257, opened 2025-08-17
readme: use links to docs instead of gh wiki

chinatsu:
(i'm obsessed with making pull requests right in the browser)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks, guess will also need to update default-config.kdl

YaLTeR:
This also makes me think that the existence of the "open github editing current wiki page" button should be acknowledged on the Documenting niri page

chinatsu:
> Thanks, guess will also need to update default-config.kdl

ah, true. i'll leave it for now though because this website-only-workflow is stupid. i'll keep a mental note of it in case it's forgotten and put in a pr some day soon lol

> This also makes me think that the existence of the "open github editing current wiki page" button should be acknowledged on the Documenting niri page

also a good point! i do think forking it and making changes is much better though (changes often lead to the need to change things outside of that single file), so maybe it's best we don't point it out too heavily? lol

YaLTeR:
> so maybe it's best we don't point it out too heavily? lol

eh i think on the documenting niri page is appropriate, not too heavy

PR #2267, opened 2025-08-19
configure section permalinks

chinatsu:
i think this fixes https://github.com/YaLTeR/niri/discussions/2266

though, it's not consistent with github's permalinks that show up on the left side of a title.
this one shows up on the right side of the header.

it's visible when hovering over a title. it is always visible when the url matches the slug

image


YaLTeR:
Thanks, looks good

PR #2270, opened 2025-08-20
niri-ipc: Add window is_fullscreen and is_windowed_fullscreen

iynaix:
I'm trying to determine if a window is fullscreen using the IPC, but this information is not currently exposed. I've worked around it by:

1. Getting the workspace of the window
2. Getting the output the workspace is on
3. Comparing the size of the window to the logical dimensions of the output

This is rather awkward :sweat_smile:  and I think the fullscreen information should be exposed by the window.

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1155,6 +1155,10 @@ pub struct Window {
     pub is_floating: bool,
     /// Whether this window requests your attention.
     pub is_urgent: bool,
+    /// Whether this window is fullscreen
+    pub is_fullscreen: bool,
+    /// Whether this window is windowed fullscreen
+    pub is_windowed_fullscreen: bool,
     /// Position- and size-related properties of the window.
     pub layout: WindowLayout,
 }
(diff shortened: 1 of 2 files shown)


YaLTeR:
What do you want these for?

iynaix:
I was implementing https://github.com/YaLTeR/niri/pull/2240 via the IPC. I'm not too sure about the usefulness of `is_windowed_fullscreen` though.

YaLTeR:
Why do you need is_fullscreen for that?

iynaix:
To not do anything if there is a fullscreen window.

neunato:
This would be really useful to have for scripting.

It would make sense to include `is_maximized_to_edges` for completeness sake.

Keep in mind that currently it's not possible to distinguish between fullscreen and maximized (to edges) windows if there is no status bar, as the geometries will match.
Likewise, if there are no gaps nor borders, it's impossible to tell if a window is fullscreen/maximized to edges/maximized column.

My use cases:
1. I want to be able to switch from fullscreen to maximized-to-edges directly
2. I want to be able to automatically exit fullscreen by assigning window width
3. I have a script that toggles floating while restoring window tiling size and fullscreen/maximized state when unfloating

Using window geometry, besides being messy and slower due to additional IPC calls, is not reliable as it depends on a specific configuration.

iynaix:
Closing in favor of #2836

PR #2273, opened 2025-08-21
Implement basic screen reader announcements via AccessKit

YaLTeR:
Supersedes #2042.

Implemented screen reader support for:
- exit confirmation dialog
- screenshot UI, overview (it'll just say when these are focused)
- config parse error announcement
- workspace switch announcement
- important hotkeys (as an announcement for now)

AccessKit is only started on the main session instance, so to test it, you'll need to build and install the niri binary, then log into it from GDM.

@DataTriny hi, could you take a look?

I have some questions:

1. accesskit-unix sometimes deadlocks. Example bt: [in `update_window_focus_state()`](https://github.com/user-attachments/files/21917429/bt.log), [in `update_tree()`](https://github.com/user-attachments/files/21948049/bt2.log). I had to put it on a thread so that it didn't hang the entire compositor, but if it locks up, you have to restart niri to get the screen reader back.
2. Sometimes Orca will get stuck reading the Important Hotkeys announcement in full and not react to any key presses. The only way out I found was to `pkill orca`, but even that takes like 5 seconds to register.
3. Would be good to announce Important Hotkeys right away when Orca is started at compositor startup, but I guess for that to happen I need to refactor Important Hotkeys into a proper dialog that takes focus, which I'm not prepared to do for this PR.
4. I had to use a Live node for announcements, but it only emits the announcement when the node value changes. I had to work around that by conditionally appending a space character when the announcement text matches the previous one. This seems not ideal.
1. It always requests initial tree, even before starting Orca, and never deactivates on my (Fedora) machine. So there's no way to tell when an AT is connected, and I need to just always send updates?
5. I initially tried marking the exit confirm dialog and screenshot UI nodes as `set_hidden()` unless they were visible (focused), in hopes of recreating GNOME Shell's tree. However, that caused Orca and Accerciser to not see those nodes whatsoever. Not sure what's going on there, but not marking the nodes as hidden appears to work...
6. Orca doesn't read "Screenshot UI" or "Overview" the very first time after I start it. Closing and opening them again makes it read it. Not sure what's up with that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,54 @@
 version = 3
 
 [[package]]
+name = "accesskit"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9c0690ad6e6f9597b8439bd3c95e8c6df5cd043afd950c6d68f3b37df641e27c"
+
+[[package]]
+name = "accesskit_atspi_common"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8fb511e093896d3cae0efba40322087dff59ea322308a3e6edf70f28d22f2607"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "atspi-common",
+ "serde",
+ "thiserror 1.0.69",
+ "zvariant",
+]
+
+[[package]]
+name = "accesskit_consumer"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ec27574c1baeb7747c802a194566b46b602461e81dc4957949580ea8da695038"
+dependencies = [
+ "accesskit",
+ "hashbrown 0.15.5",
+]
+
+[[package]]
+name = "accesskit_unix"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2abbfb16144cca5bb2ea6acad5865b7c1e70d4fa171ceba1a52ea8e78a7515f4"
+dependencies = [
+ "accesskit",
+ "accesskit_atspi_common",
+ "async-channel",
+ "async-executor",
+ "async-task",
+ "atspi",
+ "futures-lite",
+ "futures-util",
+ "serde",
+ "zbus",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


YaLTeR:
Another thing: emulating announcements with Live nodes is kinda awkward because they are only emitted when the value changes, so there's no good way to emit the same announcement again.

DataTriny:
Hello @YaLTeR, I'll try to test this today.

> It always requests initial tree, even before starting Orca, and never deactivates on my (Fedora) machine.

I think I remember someone reporting this a while back and I'm pretty sure it was on Fedora as well. I wonder if some distributions force accessibility as soon as you have it enabled. Some UI toolkits have troubles lazily initializing accessibility so this might be a workaround.

> emulating announcements with Live nodes is kinda awkward because they are only emitted when the value changes, so there's no good way to emit the same announcement again.

What you can do is to use two live regions and constantly switch between them. The two regions don't even have to be in the tree at the same time, so you can just have one label node with the live property set but use two node IDs for this label. Whenever you want to announce something, just switch from one node ID to the other.

DataTriny:
My concern around announcements was more due to the fact that you wouldn't be able to announce something if the window containing the live region currently doesn't have the focus. I know that Orca has special code for some compositors or apps like notification daemons, but it would be better to avoid going this path if possible.

YaLTeR:
Hey! Just updated the PR and the description with up-to-date info and problems, also added a second deadlock bt. Otherwise it should be ready to test.

> My concern around announcements was more due to the fact that you wouldn't be able to announce something if the window containing the live region currently doesn't have the focus. I know that Orca has special code for some compositors or apps like notification daemons, but it would be better to avoid going this path if possible.

Well, it appears to work for me so far. I don't want to steal focus for the workspace switch announcement for instance because the focus instantly goes to the newly focused window then.

YaLTeR:
@DataTriny hey, have you had the chance to test this PR?

DataTriny:
@YaLTeR Sorry.

Have you been able to get Orca to read things? With your branch, if I press Super+Shift+E for instance I only hear Orca say "E" and not report the exit dialog, even though I see that it opens. Same goes for the important shortcuts. I don't get any announcements when switching between workspaces.

I had a very quick look at the first log file you provided but for now I can only say that a panic occurred somewhere, nothing more.

YaLTeR:
Yeah, Orca reads the dialogs and workspaces for me. Keep in mind that I made it connect to AccessKit only when running `niri --session` (because it marks itself as focused in a way that I think would conflict with the host compositor if running nested or as another instance).

YaLTeR:
Made the keys consumed by KeyboardMonitor not close the exit confirmation dialog, so you can e.g. use KP_Enter to repeat the exit confirmation dialog.

I'll merge this since I'm planning to make a new release on Saturday, and it's good to give the changes a few days of general stability testing from niri-git users. Especially if it's communicating over D-Bus regardless of whether Orca is running.

@DataTriny it should make them easier to test too by installing the niri-git COPR / AUR package, then logging out and back in. If you find the time to do it please write here how it goes, thanks!

DataTriny:
I just realized that I forgot to move the compiled binary to the right location yesterday evening... So running the proper version, I can indeed see that Orca announces everything.

You are right that it is best to ship something for users to test rather than not shipping anything. At least now we have something to improve upon, which I'll probably do.

Thank you for investing in accessibility.

YaLTeR:
Thanks for checking! Should I open an issue in accesskit about the deadlocks btw?

PR #2277, opened 2025-08-23
feat: reset HDR connector properties

### LINKED ISSUE #1841: Wrong colors on HDR enabled monitor
Hi there, I'm using 3 monitors and the middle one is a HDR one. There seems to be no option to enable/disable HDR in niri, but my monitor HUD says its HDR enabled. The colors are really messed up, its practically not usable.

Is there any way to fix the colors? On KDE, everything looks stellar. When not, am I able to disable HDR for the time being?

Thanks for your time!

### System Information

* Distro: Arch Linux

* GPU: NVIDIA 4090 RTX - Driver: 575.57.08

* CPU: Ryzen 9 7950X3D

my4ng:
This fixes #1841.

The is a stopgap solution until proper color management is implemented in niri. It resets the DRM connector properties `HDR_OUTPUT_METADATA` and `Colorspace` for outputs on connector connection and session resumption. This prevents HDR mode being carried over from another compositor, and thus making SDR content look wrong in niri.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -865,7 +865,7 @@ impl Tty {
 
         if let Ok(props) = ConnectorProperties::try_new(&device.drm, connector.handle()) {
             match reset_hdr(&props) {
-                Ok(()) => debug!("reset HDR properties"),
+                Ok(()) => (),
                 Err(err) => debug!("error resetting HDR properties: {err:?}"),
             }
 
@@ -873,7 +873,7 @@ impl Tty {
                 // We only use 8888 RGB formats, so set max bpc to 8 to allow more types of links to
                 // run.
                 match set_max_bpc(&props, 8) {
-                    Ok(bpc) => debug!("set max bpc to {bpc}"),
+                    Ok(_bpc) => (),
                     Err(err) => debug!("error setting max bpc: {err:?}"),
                 }
             }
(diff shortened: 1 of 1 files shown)


my4ng:
Certainly! I will make the changes for 1) and 2). 

The properties are taken from the [DRM Kernel Mode Setting](https://docs.kernel.org/gpu/drm-kms.html#standard-connector-properties) with the colorspace enum defined [here](https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/drm_connector.c#L1320). 

In KWin, a NULL HDR metadata is [created](https://github.com/KDE/kwin/blob/master/src/backends/drm/drm_pipeline.cpp#L727) for SDR, and subsequently a blob ID of 0 is [used](https://github.com/KDE/kwin/blob/master/src/backends/drm/drm_commit.cpp#L69) for the property.

YaLTeR:
Thanks. On my laptop, HDR doesn't seem to transfer over from GNOME, so I'll assume you tested this and it fixes the issue indeed.

my4ng:
> Thanks. On my laptop, HDR doesn't seem to transfer over from GNOME, so I'll assume you tested this and it fixes the issue indeed.

Yeah it works on my OLED monitor when switching from KWin with HDR enabled there, but it’s just my sample of one 😅

PR #2284, opened 2025-08-24
Support miette 7.4.0 without patching knuffel

FedericoCeratto:
Knuffel seems to be unmaintained and depends on an older version of miette but niri-config is able to build against miette 7.4.0 with a small code tweak.

### DIFF
--- a/niri-config/Cargo.toml
+++ b/niri-config/Cargo.toml
@@ -11,7 +11,7 @@ repository.workspace = true
 bitflags.workspace = true
 csscolorparser = "0.7.2"
 knuffel = "3.2.0"
-miette = { version = "5.10.0", features = ["fancy-no-backtrace"] }
+miette = { version = "7.4.0", features = ["fancy-no-backtrace"] }
 niri-ipc = { version = "25.5.1", path = "../niri-ipc" }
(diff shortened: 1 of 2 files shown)


YaLTeR:
Doesn't that still leave miette 5 in the dependency tree? That seems to defeat the purpose of this bump. We don't use miette anywhere else other than config parsing.

niri-config/Cargo.toml
```diff
@@ -11,7 +11,7 @@ repository.workspace = true
 bitflags.workspace = true
 csscolorparser = "0.7.2"
 knuffel = "3.2.0"
-miette = { version = "5.10.0", features = ["fancy-no-backtrace"] }
+miette = { version = "7.4.0", features = ["fancy-no-backtrace"] }
```
FedericoCeratto, line 14: Setting miette 7.4.0

FedericoCeratto:
@YaLTeR I'm confused: on my side `cargo clippy` `cargo check` and `cargo build` are passing and I'm setting miette `7.4.0` in  niri-config/Cargo.toml

YaLTeR:
Do you perhaps have a patched version of knuffel with miette 7.4.0?..

PR #2296, opened 2025-08-25
feat: add niri-uwsm.desktop

shimeoki:
adds a desktop entry for [uwsm](https://github.com/Vladimir-csp/uwsm).

if the user wants to use uwsm and a display or login manager, it's required to add this by hand, which is inconvenient. the entry requires root privileges for the creation and editing, and in the environment like nixos it's required to be done manually via a derivation or uwsm options in the configuration.

i think uwsm is pretty widely used at this point, so adding at least a desktop entry for the convenience would be great. the addition was approved from the uwsm author's side in this [comment](https://github.com/Vladimir-csp/uwsm/pull/162#issuecomment-3221185858).

`TryExec` directive should hide the entry if uwsm is not installed, so it wouldn't clutter display/login managers in the case of uwsm absence.

the issue is this entry needs to be installed to the system by the package manager. so in case of this PR's merge approval, something like a notice for packagers in the next release would be great.

### DIFF
--- a/resources/niri-uwsm.desktop
+++ b/resources/niri-uwsm.desktop
@@ -1,0 +1,7 @@
+[Desktop Entry]
+Name=Niri (uwsm-managed)
+Comment=A scrollable-tiling Wayland compositor
+Exec=uwsm start -- niri.desktop
+TryExec=uwsm
+Type=Application
+DesktopNames=niri


YaLTeR:
Meh. You don't need uwsm for niri. The only benefit AFAIK is if you're already using uwsm for another compositor and want to preserve your setup. So I don't think this should go into the niri repo.

shimeoki:
i started using uwsm pretty recently, but from my experience i noticed a few things:

- there is no ["import-environment is deprecated"](https://github.com/YaLTeR/niri/issues/254) message on startup, because uwsm imports variables in it's own way (as i think, i hope it's not just hiding the logs). it's not crucial, but still a some form of a relief that you are not using something deprecated.

- uwsm provides `uuctl` for controlling units. for example, it's pretty convenient to close discord this way (because you need to `pkill` it, otherwise trying to find a service manually is taking too much time).

- niri already launches apps as systemd units in app-graphical slice, but only if using uwsm to launch niri and the apps themselves, i could see that windows are being closed (visually) on `poweroff` or `reboot`. maybe niri shuts them down properly as well, but either it happens too fast or i am doing something wrong.

- while using niri before uwsm, i once or twice got the "stopping user session" for one a half minutes on shut down. as a counterpoint, i got something similar (?) while using niri with uwsm: all windows have been closed, but the niri haven't, and i couldn't do anything. i still don't know the cause of the problem, i haven't checked the logs then.

i started using uwsm after i started using niri, because i heard about uwsm from hyprland and wanted to try it. personally, i found some benefits while using uwsm, so i opened this PR in the first place.

i haven't found any information about it on the wiki and there is almost nothing about it in the issues, so maybe as a compromise it's worth it to add at least something to the wiki in the systemd section for the users of the other compositors?

this way, the entry can be added as an example, so it's not in the repository as a file and the packagers don't need to distribute it, but the user can easily create this entry from an example. also, this section can include some information about the usage of uwsm in niri as a whole.

YaLTeR:
> maybe as a compromise it's worth it to add at least something to the wiki in the systemd section for the users of the other compositors?

On the one hand, we can mention that niri works with uwsm. On the other hand, if you're already using uwsm then you probably already know what to try and how to run it?

shimeoki:
this still could be helpful to the users who just recently started using window compositors and hadn't knew about uwsm. like an additional application you can use and it's not required, but it works without any issues.

but, yeah, niri wiki is probably not the place to have this kind of documentation. uwsm readme itself could provide the information on usage by itself, and the user probably can find it from awesome lists or media.

sorry for taking your time. closing.

PR #2306, opened 2025-08-27
Use spawn-sh in examples in default-config.kdl

stefonarch:


YaLTeR:
But then the allow-when-locked comment appears after the first use of allow-when-locked, and also there are no examples on how to pass multiple arguments to spawn... Though, it does look nicer admittedly... Not sure, I'm conflicted

stefonarch:
Moved  up the "allow-when-locked".
You mean to document the legacy `spawn "command" "option" "argument"` style? It could be mentioned, stating that it's still supported (reading old configs).

YaLTeR:
It's not legacy though. spawn-sh is just a simpler version. If you know what you're doing, spawn is the preferred way

stefonarch:
Ok, on my side I'm glad not to have to use more than two doublequotes anymore ;)
So maybe this PR isn't needed.

YaLTeR:
Went for a bit of a compromise in https://github.com/YaLTeR/niri/commit/6e0b38050d4f0c8656b0ecee3140f4f07e2b895b, left two spawn examples and changed the rest

PR #2319, opened 2025-08-30
Feature: Add auto-confirm option to screenshot ui

gonengazit:
add an option to auto-confirm screenshots using the screenshot ui - i.e. - immediately take a screenshot once the user makes a selection without waiting for the user to confirm/press Enter

### DIFF
--- a/docs/wiki/Configuration:-Key-Bindings.md
+++ b/docs/wiki/Configuration:-Key-Bindings.md
@@ -382,6 +382,15 @@ binds {
 }
 ```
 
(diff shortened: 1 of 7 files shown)


YaLTeR:
1. Why is this needed?
2. Is it needed specifically on the action, or better just in the config?

gonengazit:
1. This is needed as more streamlined way to take screenshots, when you don't care much about accuracy. It's not a critical feature, but I prefer this type of ui
2. I think it's nice to have the option to bind both a "quick" screenshot, and an "accurate" screenshot, since they could be used for different things. However, It's probably ok as well as a "global" setting if you think that is better

gonengazit:
@YaLTeR what do you think? Are you in favour of merging this feature? If so I can fix the conflicts
Anything else that needs to be done before merging?

PR #2323, opened 2025-08-31
CI: automatically update flake.lock

ninelore:
Update Nix flake.lock weekly via GitHub Actions. This requires GitHub actions to have write access and to be allowed to create pull requests in the repository settings.

I've set the cron to every tuesday midnight GMT. This can be changed as desired.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,17 @@
+name: Update lockfile
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 2"
+
+jobs:
+  lockfile:
+    name: "Update lockfile"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - name: Install Nix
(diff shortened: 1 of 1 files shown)


YaLTeR:
Does this run nix flake check to ensure that it still builds?

> This requires GitHub actions to have write access and to be allowed to create pull requests in the repository settings.

Isn't this done using `permissions:`?

ninelore:
> Does this run nix flake check to ensure that it still builds?

Could be integrated, but it might be better placed in a separate action if desired

> Isn't this done using permissions:?

Would have to check on one of my repos since i haven't had a need or use for these yet

YaLTeR:
> Could be integrated, but it might be better placed in a separate action if desired

I mean, what's the point of trying to update if it won't work?

ninelore:
What I mean is nix flake check should be done on pull requests like this.

```yml
on:
  pull_request:
    paths: 
      - '**/*.nix'
      - 'flake.lock'
```

ninelore:
After some testing, GH Actions is definetly unable to create PRs without this checkbox
image

PR #2333, opened 2025-09-04
Refactor layout fullscreen handling, add un/fullscreen animations

YaLTeR:
The first part of the PR splits Column `is_fullscreen` into `is_pending_fullscreen` and `is_fullscreen()` like Tiles always were, and changes the relevant places to use the correct one of the two. This fixes a visual jump between requesting a window to fullscreen and it responding, clearly visible with animations off:

https://github.com/user-attachments/assets/f3316e87-6086-4608-ae2e-a6e202cf03f3

Afterward, a large commit implements everything necessary for un/fullscreen animations, which turns out to be quite a number of things. See the commit message.

https://github.com/user-attachments/assets/05b340b0-4e5e-4b66-93e3-bd42734f33c6

https://github.com/user-attachments/assets/d736341a-7e77-4c06-8fad-96857e2da93a

Testing wanted. If you see any new animation jank, or if anything broke, please report.


YaLTeR:
Okay so investigating a sometimes-glitchy fullscreen when pressing f in mpv with this PR led me to refactor a good chunk of core xdg-shell, layer-shell, and session-lock code in Smithay: https://github.com/Smithay/smithay/pull/1817

This PR now includes the Smithay refactor. It makes some Smithay logic more correct, including:
- fixes the aforementioned issue with f in mpv
- fixes a similar issue I found with spamming un/fullscreen on the Adwaita Demo sometimes forgetting the view position
- adds a check and protocol error for clients that try to map before acking the initial configure, instead of crashing niri (this is not allowed by the protocol)

However, it also means that this PR is in an even higher need of testing. I did a bunch of testing with things running on my desktop and various layer-shell clients, but more eyes won't hurt. Just check that everything you use still runs smoothly, and that there aren't any new ERROR messages in the niri output.

YaLTeR:
Updated to latest Smithay PR commit which also includes a fix for Qt layer-shell popup grabs.

alex-huff:
> Just check that everything you use still runs smoothly, and that there aren't any new ERROR messages in the niri output.

From my testing of kitty it seems these changes cause a configure to be sent to layer surfaces right after they are unmapped. Since the configure has different geometry, kitty resizes itself (while unmapped) and sends a SIGWINCH to the child program. When kitty remaps itself, it receives a configure from niri (with new geometry) and resizes itself accordingly and sends another SIGWINCH to the child program. The problem is the child program can take a while to resize itself causing an undesirable flicker when reshowing a hidden kitty panel.

https://github.com/user-attachments/assets/cd6a11df-292a-4f99-9dcf-15a422d4adb5

Is there a reason unmapping now causes niri to send a configure?

YaLTeR:
I don't know, what configure is it?

alex-huff:
> I don't know, what configure is it?

niri sends a `zwlr_layer_surface_v1.configure` after kitty commits will a null buffer.

YaLTeR:
I think I see the problem

YaLTeR:
I'll try to fix tomorrow

YaLTeR:
Fixed in ffb3030e36877aa346d476d717829e24bc7bfcc2

alex-huff:
> Fixed in [ffb3030](https://github.com/YaLTeR/niri/commit/ffb3030e36877aa346d476d717829e24bc7bfcc2)

Works great, Thank you!

PR #2340, opened 2025-09-05
niri-config: add support for include directives and merging

bbedward:
- Support position-aware `include "someotherfile.kdl"`
  - All the included files get watched too
  - They are basically just rendered into a singular config, then processed as normal.
- Add top-down merging for many sections
  - Add `MaybeSet` and `BoolFlag` so that default values can be differentiated between explicitly set values (so that later configs don't reset previously set values tod efaults)
  - Added a `niri-macros` for procedural `Mergeable` macro
    - It was just an idea to prevent needing to define `merge_with` logic for each configuration struct
  - Not all fields can be merged - but `layout`, `input`, `animations`, `gestures`, `binds`, `debug` and their children can be (if they have children)
  - `window-rule` and any named stuff like `output` or `workspace` remain un-changed logic wise.
- Added a doc page

~~Originally replaced the file watcher with [notify-debouncer-mini](https://docs.rs/notify-debouncer-mini) but reverted back to the original approach. That one remains in the history nonetheless as a potential future enhancement.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ serde_json = "1.0.143"
 tracing = { version = "0.1.41", features = ["max_level_trace", "release_max_level_debug"] }
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 tracy-client = { version = "0.18.2", default-features = false }
+xshell = "0.2.7"
 
 [workspace.dependencies.smithay]
 # version = "0.4.1"
@@ -123,7 +124,7 @@ proptest = "1.7.0"
 proptest-derive = { version = "0.6.0", features = ["boxed_union"] }
 rayon = "1.11.0"
 wayland-client = "0.31.11"
-xshell = "0.2.7"
+xshell.workspace = true
 
 [features]
 default = ["dbus", "systemd", "xdp-gnome-screencast"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
     "niri-config",
     "niri-ipc",
+    "niri-macros",
     "niri-visual-tests",
 ]
 
--- a/docs/wiki/Configuration:-Introduction.md
+++ b/docs/wiki/Configuration:-Introduction.md
@@ -78,32 +78,19 @@ input {
 
 #### Sections
 
-Most sections cannot be repeated. For example:
+Most sections that are repeated will be merged top-down in a where where "last one wins":
 
 ```kdl
-// This is valid: every section appears once.
 input {
-    keyboard {
-        // ...
-    }
-
     touchpad {
-        // ...
-    }
-}
-```
-
-```kdl,must-fail
-// This is NOT valid: input section appears twice.
-input {
-    keyboard {
-        // ...
+        scroll-factor 0.3
     }
 }
 
+// Scroll factor becomes 0.75, the previous value 0.3 is overridden
 input {
     touchpad {
-        // ...
+        scroll-factor 0.75
     }
 }
 ```
--- a/docs/wiki/Configuration:-Introduction.md
+++ b/docs/wiki/Configuration:-Introduction.md
@@ -78,7 +78,7 @@ input {
 
 #### Sections
 
-Most sections that are repeated will be merged top-down in a where where "last one wins":
+Most sections that are repeated will be merged top-down so that the "last one wins":
 
 ```kdl
 input {
@@ -92,6 +92,19 @@ input {
     touchpad {
         scroll-factor 0.75
     }
+}
+```
+
+Some configurations are not mergeable, for example environment:
+
+```kdl,must-fail
+// This is NOT valid: environment cannot appear twice
+environment {
+    DISPLAY ":0"
+}
+
+environment {
+    DISPLAY: "1"
 }
 ```
 
--- a/docs/wiki/Configuration:-Modularization.md
+++ b/docs/wiki/Configuration:-Modularization.md
@@ -1,0 +1,185 @@
+### Modularization
+
+<sup>Since: next release</sup>
+
+You can use `include` directives to resolve configuration from multiple files.
+
+```kdl
+// gestures.kdl
+gestures {
+    dnd-edge-view-scroll {
+        trigger-width 30
+        delay-ms 100
+        max-speed 1500
+    }
+
+    dnd-edge-workspace-switch {
+        trigger-height 50
+        delay-ms 100
+        max-speed 1500
+    }
+
+    hot-corners {
+        // off
+    }
+}
+```
+
+```kdl
+// binds.kdl
+binds {
+    Mod+Left { focus-column-left; }
+    Super+Alt+L { spawn "swaylock"; }
+}
+```
+
+```kdl
+// layout.kdl
+layout {
+    gaps 16
+}
+```
+
+```kdl
+// config.kdl
+include "layout.kdl"
+
+layer-rule {
+    match namespace="waybar"
+    match at-startup=true
+
+    // Properties that apply continuously.
+    opacity 0.5
+}
+
+include "gestures.kdl"
+include "binds.kdl"
+```
+
+This will result in a final effective configuration of:
+
+```kdl
+layout {
+    gaps 16
+}
+
+
+layer-rule {
+    match namespace="waybar"
+    match at-startup=true
+
+    // Properties that apply continuously.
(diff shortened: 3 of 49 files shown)


niri-config/src/include.rs
```diff
@@ -0,0 +1,572 @@
...
+
+fn parse_include_directive(line: &str) -> miette::Result<Option<String>> {
+    let trimmed = line.trim();
+
+    // Match: include "path/to/file.kdl"
+    if let Some(rest) = trimmed.strip_prefix("include") {
+        let rest = rest.trim();
+
+        // Extract quoted string
+        if rest.starts_with('"') && rest.ends_with('"') && rest.len() >= 2 {
```
HigherOrderLogic: `#"foo/bar"#` is also a valid string in Kdl.

src/utils/watcher.rs
```diff
@@ -106,20 +111,101 @@ fn see(config_path: &ConfigPath) -> io::Result<(SystemTime, PathBuf)> {
...
+                for file_path in &self.included_files {
+                    if let Ok((mtime, canon)) = see_path(file_path) {
+                        new_props.insert(file_path.clone(), (mtime, canon));
+                    }
+                }
+                self.last_props = new_props;
+            }
+            Err(_) => {
+                // If we can't load config, just watch the main file(s)
+                self.included_files.clear();
```
HigherOrderLogic: This isnt really ideal.

What if I'm editing an included file, and then make a typo before saving, do I have to touch the main config file again for the changes to take effect?

bbedward: Updated it to preserve the state on load errors

niri-config/src/include.rs
bbedward: As far as I can tell, knuffel itself doesnt even support it (unless I'm missing something, raw strings fail everywhere in niri validate). So probably doesn't make sense to here either IMO. (not sure if the [the maintained fork of knuffel](https://github.com/TheLostLambda/knus) supports it though, didn't try it)

HigherOrderLogic: Yes, Knuffle doesnt support raw string, but I just mentioned it here for spec compatibility.

docs/wiki/Configuration:-Modularization.md
YaLTeR, line 1: Huh, this makes me think. Maybe merging should only be allowed across different files? As in, within a single file, you still wouldn't be able to have two `layout {}` sections. Maybe this would be less confusing than "some sections merge, some sections don't"? Not sure.

Cargo.toml
YaLTeR, line 1: Could you please extract this commit into a separate PR? Also it should have a very good justification since it's:
1. Quite a complex change
2. Adds a very nontrivial dependency, and we're trying not to introduce those if we can help it

niri-macros/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    let merge_body = match &input.data {
+        Data::Struct(data) => {
+            match &data.fields {
+                Fields::Named(fields) => {
+                    // Find all mutex fields
+                    let mutex_fields: Vec<&Ident> = fields
+                        .named
+                        .iter()
+                        .filter_map(|f| {
+                            let is_mutex = f.attrs.iter().any(|attr| {
```
YaLTeR, line 20: Huh? I don't think we have mutexes anywhere in the config

niri-config/src/maybe_set.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    pub fn get_mut(&mut self) -> &mut T {
+        &mut self.value
+    }
+
+    pub fn into_inner(self) -> T {
+        self.value
+    }
+
+    pub fn value(&self) -> &T {
```
YaLTeR, line 30: Why multiple methods for the same thing?

niri-config/src/include.rs
YaLTeR: Huh? I use raw strings all over the place in my config. They're all over the place in examples too: https://yalter.github.io/niri/Configuration:-Window-Rules#open-on-output

niri-config/src/include.rs
```diff
@@ -0,0 +1,572 @@
...
+    let mut result = String::new();
+    let current_dir = canonical_path
+        .parent()
+        .ok_or_else(|| miette::miette!("config file has no parent directory"))?;
+
+    for line in content.lines() {
+        let trimmed = line.trim();
+
+        // Check for include directive
+        if trimmed.starts_with("include") {
```
YaLTeR, line 91: Is manual string parsing really the best option here? Not sure how much knuffel exposes so maybe it is.

niri-config/src/include.rs
```diff
@@ -0,0 +1,572 @@
...
+            if let Some(include_path) = parse_include_directive(trimmed)? {
+                let resolved_path = resolve_include_path(&include_path, current_dir)?;
+                let included_content = resolve_includes_recursive(
```
YaLTeR, line 94: Needs some recursion limit bc if you make a cyclic include without it being apparent through the file system, it would be good not to infloop

docs/wiki/Configuration:-Introduction.md
```diff
@@ -95,7 +95,7 @@ input {
 }
 ```
 
-Some configurations are note mergeable, for example environment:
+Some configurations are not mergeable, for example environment:
```
YaLTeR, line 98: Please squash fixes like these into the relevant commits for easier review

niri-config/src/include.rs
bbedward: Yea, I guess I was having a brain fart and looking at `#"strings#"` and missing the `r#"r in front#"`. I accounted for it in https://github.com/YaLTeR/niri/pull/2340/commits/261d2375b44b6bc6ca6472edea639ce2a2b65854

docs/wiki/Configuration:-Modularization.md
bbedward: I mean, it's kinda that way because of the implementation - it would take a whole different approach to do that probably. Because this approach is pretty simplified:
1. Assemble a single config.kdl with all includes in their correct position
2.  Parse the final kdl

So it effectively turns all the kdl into a single kdl file, then parses. I can't really think of a nice way to enforce something like that with this particular approach at least (mergeable macro on the structs and the assembly approach just makes it so the merging is valid)

Could be documented better to be less confusing, maybe.

Cargo.toml
bbedward: This one was just left in the history but not part of the final PR, either way I rewrote the commit history so it's gone from it (though I do think inotify is probably more "correct" than polling it's not that big of a deal)

niri-macros/src/lib.rs
bbedward: This is pretty much only because of the on/off stuff in window rules.

```kdl
shadow {
  on // sets on
}

shadow {
  off // unsets on
}
```

those are the only fields using mutex

niri-config/src/maybe_set.rs
bbedward: Removed Deref in favor of value(), not sure what's more preferable but I think value() is more explicit.

niri-config/src/include.rs
bbedward: So, I did initially do proper parsing with `knuffel::ast` - but it was not that easy to follow and ended up pretty complex, so it made more sense to me to just keep a simple parser since all we're really interested in (with this implementation) is ripping the contents of includes into one file kdl, then parsing once.

To me, it seemed simpler and is probably less overhead than parsing all of the includes once, then parsing the final output config at the end.

But that's just me, this implementation is the kiss approach - pre-parse into one kdl config, then parse. Rather than parsing them all and trying to assemble them later. it just seems much simpler to me and makes it easy to make things position-aware and what not.

niri-config/src/maybe_set.rs
bbedward: this one https://github.com/YaLTeR/niri/pull/2340/commits/b391f34b518d081c9d78ff8dfe42654e72461fb6

niri-config/src/include.rs
bbedward: Added a limit https://github.com/YaLTeR/niri/pull/2340/commits/0d9926a4bf0de3b5bb6c969bf2ce215aff127eb0

BinaryQuantumSoul:
Lesgooo

YaLTeR:
Implemented in https://github.com/YaLTeR/niri/pull/2482

PR #2369, opened 2025-09-11
layout: new option `hide-floating-layout-if-unfocused`

yuukibarns:
Implement #1061

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2082,9 +2082,9 @@ checksum = "6a82ae493e598baaea5209805c49bbf2ea7de956d50d7da0da1164f9c6d28543"
 
 [[package]]
 name = "libdisplay-info"
-version = "0.2.2"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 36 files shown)


emenel:
i'd love to be able to hide/show floating windows in a workspace, but i would only use it as a key bind rather than automatically based on focus. i.e., i often want a calculator or music player that i can toggle, but sometimes i want it to stay visible when doing things in other windows.

YaLTeR:
Indeed, this does not look like a good enough solution to the problem

yuukibarns:
Yes, at first i did want to implement it as an action but i find there are some edge cases we have to cover, for example, when we turn a scrolling window into a floating one when floating layout is hidden, the window should be focused and then we have to show the floating layout.

My thoughts is that we can add an action similar to switch focus between scrolling and floating that only changes the FloatingActive Enum, maybe a new Hidden state like NoButRaised?

Anyway this pr has solved my own usecase and i proposed it just because someone in that thread reminded me to do so.

I think it is trivial for the owner to implement a new action and it will be more appropriate if he has time to look into it.

yuukibarns:
@YaLTeR, any advice or consideration?

YaLTeR:
No advice for now; if this works for you then great, keep using it

PR #2371, opened 2025-09-12
feat(move-floating-window): percentage change

nenikitov:
This PR adds new syntax to `move-floating-window` command to support proportional changes (ex: `+10%`, `-10%`, `10%`), similar to `set-window-width` and `set-window-height`.


YaLTeR:
Thanks!

PR #2373, opened 2025-09-12
Add ability to inhibit idle on windows

griccardos:
This adds the ability to inhibit idle on a window if configured.

There are a few discussions about the topic
#742 
#1855 

There are 3 options:
- `inhibit_idle "always"` - whenever this window is present. 
- `inhibit_idle "fullscreen"` - whenever this window is fullscreen
- `inhibit_idle  "never"` - don't inhibit on this window

Option 1 is useful, for example converting a video file that may take a couple of hours, but the window itself is not necessarily fullscreen
```kdl
window-rule {
    match title="HandBrake"
    inhibit-idle "always"
}
```
Option 2 is useful, for example when you are watching video, playing games with a controller and the window is fullscreen
```kdl
window-rule {
    match app-id="librewolf"
    inhibit-idle "fullscreen"
}
```
Option 3 is to disable inhibit on this window for example if you have enabled inhibit fullscreen on all windows, but you want this window as the exception not to inhibit when fullscreen
```kdl
window-rule {
    inhibit-idle "fullscreen"
}
window-rule {
    match app-id="Alacritty"
    inhibit-idle "never"
}
```

### DIFF
--- a/niri-config/src/appearance.rs
+++ b/niri-config/src/appearance.rs
@@ -622,6 +622,12 @@ pub enum BlockOutFrom {
     ScreenCapture,
 }
 
+#[derive(knuffel::DecodeScalar, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InhibitIdle {
+    Always,
+    Fullscreen,
+}
+
 #[derive(knuffel::Decode, Debug, Default, Clone, Copy, PartialEq)]
 pub struct BorderRule {
     #[knuffel(child)]
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -656,6 +656,36 @@ mod tests {
     }
 
     #[test]
+    fn inhibit() {
+        let parsed = do_parse(
+            r##"
+            window-rule {
+                inhibit-idle "fullscreen"                
+            }
+            window-rule {
+                inhibit-idle "always"
+            }
+            window-rule {
+                inhibit-idle null
+            }
+            window-rule {
+            }
+            "##,
+        );
+
+        assert!(matches!(
+            parsed.window_rules[0].inhibit_idle,
+            Some(Some(InhibitIdle::Fullscreen))
+        ));
+        assert!(matches!(
+            parsed.window_rules[1].inhibit_idle,
+            Some(Some(InhibitIdle::Always))
+        ));
+        assert!(matches!(parsed.window_rules[2].inhibit_idle, Some(None)));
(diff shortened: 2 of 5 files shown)


HigherOrderLogic:
`inhibit_idle  "never"` in this PR doesnt make sense. From a quick skim of your PR, seems like the client still can dispatch an idle inhibitor through the Screensaver Dbus and inhibit the compositor.

griccardos:
@HigherOrderLogic 
Perhaps `never` is the wrong term. The aim of `never`, is not to prevent inhibiting but to override a parent rule. Like in my example, I want to inhibit idle on all windows that are fullscreen with the exception of alacritty. 
perhaps renaming it to `none`?
Open to suggestions on the naming...

HigherOrderLogic:
> perhaps renaming it to none?

Yeah, `"none"` seems like a better name here.

YaLTeR:
> The aim of never, is not to prevent inhibiting but to override a parent rule.

I believe we use `null` for this

griccardos:
> > The aim of never, is not to prevent inhibiting but to override a parent rule.
> 
> I believe we use `null` for this

Yes, I can implement `null` for this case.

We will still need a 3rd variant as there are 4 cases:
- `"always"` - `Some(Always)`
- `"fullscreen"` - `Some(Fullscreen)`
- `null` - Use default irrespective of previous window rules - `Some(Default)`
- user leaves it out - Use previous window rules. - `None`

Pehaps we can call the 3rd variant `Default`

YaLTeR:
I think I used `Option<Option<T>>` elsewhere

griccardos:
Thanks for the tip, `Option<Option<InhibitIdle>>` worked nicely.
I have made the changes.
I've also done some testing and it works as expected.

nagisa:
Should a way to inspect idle inhibitors (and inhibition reasons) be added as well? Could be a field in `niri msg windows`.

YaLTeR:
It's tricky. There's Wayland idle inhibit, which has no reason. There's the D-Bus one that has no association to a window.

Crysthamus:
Hi, when is this PR expected to be merged?

Sempyos:
There is no timeline for when PRs get merged

Atilogit:
I'm loving the idea behind this PR but i think the property inhibit-idle shouldn't perform any matching (`"fullscreen"`). It would only need to be a boolean if a matcher `is-fullscreen` were to exist. I have described this in more detail in #3542.

PR #2312, opened 2025-08-29
Fall back to primary gpu for rendering

scottmckendry:
Continuing on from the wonderful work by @cmeissl on #1281, I've taken the patchset and caught it up to the main branch. I've also applied some small refactors to address the feedback from @YaLTeR. 

I'm currently [running this build](https://github.com/scottmckendry/nix/commit/fcb0c9921b41d53f5bf936609832de5e97f43ac4) as a daily driver on my main workstation and laptop. All seems very stable. 

Resolves #910

PR #2376, opened 2025-09-12
Add maximize-window-to-edges (true Wayland maximize)

YaLTeR:
Based on #2333, still missing window rules and some other minor stuff, and obviously proper commits.

This PR implements the true Wayland maximize, also known as the normal maximize that you have in other desktop environments. The "double click on titlebar", or the "maximize button next to the X button" maximize.

https://github.com/user-attachments/assets/52f6c978-4a25-45bd-ba09-2e8d89585d0f

Niri historically didn't implement true Wayland maximize. Our `maximize-column` makes the column full-width, preserving gaps, borders and struts, and works for multi-window columns. The true Wayland maximize, on the other hand, makes a single window occupy the entire working area, with no gaps or struts.

True maximize is bound to a Wayland xdg-toplevel state. Windows also have `set_maximized()` and `unset_maximized()` requests (triggered by their maximize button), and they also change the maximize button icon (to maximize or unmaximize) based on whether they are currently maximized or not. Windows will also remove their CSD shadows and rounded corners when they are in a Maximized state.

All this means that we couldn't attach Wayland maximize to our existing full-width column state. You *have* to set the Maximized state if you want the window un/maximize button to work correctly, but the Maximized state necessarily comes with square window corners, which we do not want to force in the full-width column state.

I also didn't want to add true maximize as a separate action because it's very similar to a full-width single-window column, yet subtly different, which could be confusing. The implementation side is also quite involved (as you can see from this PR).

However, with more feedback from different people lately, I reconsidered this decision. Primarily because double-click-to-maximize and the maximize button is something users (especially new users) expect to work (e.g. https://github.com/YaLTeR/niri/discussions/293, https://github.com/YaLTeR/niri/discussions/2216), and there's no other way to make it work. There's also been requests for variations of "maximize without borders" (e.g. https://github.com/YaLTeR/niri/discussions/508, https://github.com/YaLTeR/niri/discussions/820), which is exactly what true maximize provides. Finally, the fullscreen refactor cleared a lot of the way towards adding maximize, since these are very similar (both are serverside Wayland states with all the complexity that this entails).

True maximize is called "maximize-window-to-edges". Unfortunately, we can't call it just "maximize", because the existing column full-width action already has an "open-maximized" window rule, and I'm not sure I want to change its meaning. Besides, "maximize to edges" more or less describes well what it does—maximize to the entire working area, right up to screen edges or exclusive layer-shell surfaces if any.

Similarly to fullscreen, true maximize always puts the window into the scrolling layout (since otherwise it would easily obscure all windows below). Similarly to fullscreen, you can scroll left and right from it. Similarly to fullscreen, you can only have a single window in a maximized non-tabbed column. (See why I said maximize was very similar to fullscreen?) Similarly to fullscreen, maximizing and unmaximizing does not yet restore the window to the previous column, but it'll come when it comes to fullscreen.

Windowed-fullscreen also works as expected with maximized windows.

So yeah, feel free to give this some testing. I still need to play around with it and figure out what special behavior makes sense for actions (e.g. should `switch-preset-column-width` un-true-maximize? It probably should).

### DIFF
--- a/docs/mkdocs.yaml
+++ b/docs/mkdocs.yaml
@@ -84,6 +84,7 @@ nav:
     - Nvidia: Nvidia.md
     - Xwayland: Xwayland.md
     - Gestures: Gestures.md
+    - Fullscreen and Maximize: Fullscreen-and-Maximize.md
     - Packaging niri: Packaging-niri.md
     - Integrating niri: Integrating-niri.md
     - Accessibility: Accessibility.md
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -44,6 +44,7 @@ window-rule {
     open-on-output "Some Company CoolMonitor 1234"
     open-on-workspace "chat"
     open-maximized true
+    open-maximized-to-edges true
     open-fullscreen true
     open-floating true
     open-focused false
@@ -416,9 +417,27 @@ window-rule {
 }
 ```
 
+#### `open-maximized-to-edges`
+
+Make the window open [maximized to edges](./Fullscreen-and-Maximize.md).
+
+```kdl
+window-rule {
+    open-maximized-to-edges true
+}
+```
+
+You can also set this to `false` to *prevent* a window from opening maximized to edges.
+
+```kdl
+window-rule {
+    open-maximized-to-edges false
+}
+```
+
 #### `open-fullscreen`
 
-Make the window open fullscreen.
+Make the window open [fullscreen](./Fullscreen-and-Maximize.md).
 
 ```kdl
 window-rule {
--- a/docs/wiki/Fullscreen-and-Maximize.md
+++ b/docs/wiki/Fullscreen-and-Maximize.md
@@ -1,0 +1,67 @@
+There are several ways to make a window big on niri: maximizing the column, maximizing the window to edges, and fullscreening the window.
+Let's look at their differences.
+
+## Maximized (full-width) columns
+
+Maximizing the column via `maximize-column` (bound to <kbd>Mod</kbd><kbd>F</kbd> by default) expands its width to cover the whole screen.
+Maximized columns still leave space for [struts] and [gaps], and can contain multiple windows.
+The windows retain their borders.
+This is the simplest of the sizing modes, and is equivalent to `proportion 1.0` column width, or `set-column-width "100%"`.
+
+![Screenshot of a maximized column with two windows.](https://github.com/user-attachments/assets/a6c26f32-a712-4899-861c-d58a9a357e0e)
+
+You can make a window open in a maximized column with the [`open-maximized true`](./Configuration:-Window-Rules.md#open-maximized) window rule.
+
+## Windows maximized to edges
+
+You can maximize an individual window via `maximize-window-to-edges`.
+This is the same maximize as you can find on other desktop environments and operating systems: it expands a window to the edges of the available screen area.
+You will still see your bar, but not struts, gaps, or borders.
+
+Windows are aware of their maximized-to-edges status and generally respond by squaring their corners.
+Windows can also control maximizing-to-edges: when you click on the square icon in the window's titlebar, or double-click on the titlebar, the window will request niri to maximize or unmaximize itself.
+
+You can put multiple maximized windows into a [tabbed column](./Tabs.md), but not into a regular column.
+
+![Screenshot of a window maximized to edges.](https://github.com/user-attachments/assets/00eac78c-4ebe-4a29-88b8-e1047a5f09c6)
+
+You can make a window open maximized-to-edges, or prevent a window from maximizing upon opening, with the [`open-maximized-to-edges`](./Configuration:-Window-Rules.md#open-maximized-to-edges) window rule.
+
+## Fullscreen windows
+
+Windows can go fullscreen, usually seen with video players, presentations or games.
+You can also force a window to go fullscreen via `fullscreen-window` (bound to <kbd>Mod</kbd><kbd>Shift</kbd><kbd>F</kbd> by default).
+Fullscreen windows cover the entire screen.
+Similarly to maximize-to-edges, windows are aware of their fullscreen status, and can respond by hiding their titlebars or other parts of the UI.
+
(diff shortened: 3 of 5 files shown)


stefonarch:
Works nicely, just a observation: "Double click to maximize" seems to be implemented only on the GTK side, with the default (poor) Qt decoration happens nothing. Button shape changes fine.

YaLTeR:
> "Double click to maximize" seems to be implemented only on the GTK side, with the default (poor) Qt decoration happens nothing.

Yeah, it's a clientside thing. winit adwaita frame also does it, so does foot and electron, to name a few. So this is more of a Qt problem.

stefonarch:
I remember not even all x11 WM do it, example `xfwm4`.
Anyway, shortcuts will do it with any window :)

I see `lxqt-panel`  taskbar menu works out of the box with this feature, cool!

https://github.com/user-attachments/assets/f04cb7b7-30b0-4f92-a6eb-0646bcb544c6

Btw not sure if that's a good idea but I thought that the (on tiling WMs) useless "Minimize" button  as also the same menu item in taskbar buttons  could be reused in tiling WMs for "Send to scratchpad" (which would need a default scratchpad probably).

YaLTeR:
Nice! Yeah well if/when I get to the scratchpad I'll think about the minimize idea.

scottmckendry:
This is great @YaLTeR! just a note that the this branch doesn't build on nixos due to the pinned smithay commit being orphaned. I was able to work around this by pinning the rebased SHA from your open PR in smithay. 

I applied this patch on top of #2312 and will be testing it out further this week. So far so good.

YaLTeR:
Updated to latest Smithay PR commit which also includes a fix for Qt layer-shell popup grabs.

Philipp-M:
FWIW I'm testing this for roughly a week now, and it's working great and how I'd expect it to work and is probably 99% of what I need (compared to the `maximize-column` or `fullscreen-window` action).

I haven't had a good experience with other compositors like COSMIC or Hyprland (at least the most recent version) in this regard:
I like to use the right scrollbar with the mouse in e.g. browsers, and like to throw the mouse to the right edge of the screen, it doesn't hit the scrollbar there but misses the window (for COSMIC it's just an exclusive point-rectangle intersection... maybe a similar issue in Hyprland), which in niri on `main` only works with `fullscreen-window` (or disabled gaps probably) but then the status-bar is invisible...

YaLTeR:
Good to know. I've still got a bunch of TODOs for this branch but I've been running it myself for a while also

YaLTeR:
True maximize is now ready. Also see the new wiki page in the last commit.

Philipp-M:
Awesome!

I have one little suggestion (of not too much importance): When unmaximizing while the window was within a column, beneath or above another window, the column isn't restored, and instead a new column is used. Intuitively I think it should restore the layout of the previous column?
(Edit: I notice the same "issue" with `fullscreen-window`, so maybe intended?)

YaLTeR:
It's the same issue as fullscreen yeah and both will be fixed together at some point

LuckShiba:
Great, seems to work fine here!  I have a question though. 

Triggering `maximize-window-to-edges` works for toggling it off and on, but is it supposed to not be able to go to maximized mode when it's already in fullscreen mode? I'm not sure if it makes sense being able to go from fullscreen directly to maximized by using `maximize-window-to-edges`, but I think it could be cool, so it would be possible switching from fullscreen -> maximized rather than having to go fullscreen -> 'normal' -> maximized.

LuckShiba:
xwayland-satellite seems to panic when trying to open Steam on this branch
`WARN niri::utils::xwayland::satellite: xwayland-satellite exited with: exit status: 101` is the only log I could find, while it works on main

YaLTeR:
> Triggering `maximize-window-to-edges` works for toggling it off and on, but is it supposed to not be able to go to maximized mode when it's already in fullscreen mode? I'm not sure if it makes sense being able to go from fullscreen directly to maximized by using `maximize-window-to-edges`, but I think it could be cool, so it would be possible switching from fullscreen -> maximized rather than having to go fullscreen -> 'normal' -> maximized.

This is the [Wayland protocol](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_maximized) behavior at least:

> If the surface is in a fullscreen state, this request has no direct effect. It may alter the state the surface is returned to when unmaximized unless overridden by the compositor.

When you maximize during fullscreen, niri will remember this, so unfullscreening will get you to the maximized state right away.

We don't *have* to follow it for the `maximize-window-to-edges` action, but I thought it would be good if the action was consistent with the behavior of the CSD maximize button (which uses the Wayland request).

> xwayland-satellite seems to panic when trying to open Steam on this branch

Works fine for me.

LuckShiba:
> We don't _have_ to follow it for the `maximize-window-to-edges` action, but I thought it would be good if the action was consistent with the behavior of the CSD maximize button (which uses the Wayland request).

I see, guess that makes sense as the default behavior. Maybe this could be an parameter, like `force=true` in that case? I don't think it's currently possible to mimic this behavior via a script (it would need #2270 to be merged).
> 
> > xwayland-satellite seems to panic when trying to open Steam on this branch
> 
> Works fine for me.

Strange, I'll test a bit more but this panic happened exclusively in this branch. Steam itself wouldn't give error logs, but the xwayland-satellite panicked. Is there any way I can see xwayland-satellite logs to try to see an error or would it be shown on journal if it was the case? xeyes worked, though, so it wasn't xwayland in general.

YaLTeR:
> Strange, I'll test a bit more but this panic happened exclusively in this branch. Steam itself wouldn't give error logs, but the xwayland-satellite panicked. Is there any way I can see xwayland-satellite logs to try to see an error or would it be shown on journal if it was the case? xeyes worked, though, so it wasn't xwayland in general.

You can run xwayland-satellite manually and run Steam on its DISPLAY.

LuckShiba:
Can't reproduce anymore, now it's working. if I stumble to that problem again I'll send relevant logs, maybe it was not related to this PR or niri.

EDIT: now I isolated the problem as happening only when the clipboard is not empty, not only on this branch. So confirming now it's not related to this PR, it was just a coincidence.

YaLTeR:
Thanks everyone for testing, let's proceed on main!

PR #2379, opened 2025-09-13
feat(layout): trigger open animation when moving column or window between outputs

objz:
When moving a window or a whole column to another output, the open animation now plays on the destination. This makes it easier to see what actually moved, especially if there are multiple similar windows (e.g. terminals) on the target output. It just reuses `start_open_animation_for_window(..)` so it respects whatever animation config the user already set.

**Details**

* Single window moves => trigger open animation for that window.
* Column moves => trigger open animation for every window in the column.
* Same-output moves are unchanged.
* Floating handling unchanged.

**Why**
Before, windows/columns just appeared on the other output, which can be hard to spot in busy layouts. The animation makes the move obvious without adding any new config or custom effect.

---

**Before:**

https://github.com/user-attachments/assets/064390c2-7269-4ce3-afc7-5508867e1b09

**After:**

https://github.com/user-attachments/assets/80a456c4-b8c0-49ef-899f-0e793c672cd0


YaLTeR:
Eh, I'm not sure it should be the open animation

objz:
why not? I was trying to get something like hyprland where the window actually animates to the new state, but for that the animation system and other stuff would need changes and I wasn’t really sure how to do it yet. I just needed a quick fix for local use so I went with this, and thought maybe you’d also like it. any suggestions on how you’d approach this?

YaLTeR:
I haven't thought about it hard so far, maybe some other animation idk

objz:
a complete new ore just a different one?

YaLTeR:
Not sure yet

PR #2396, opened 2025-09-16
Add support for disabling DRM devices

shaunren:
Adds a debug option disable-drm-device that specifies DRM devices that niri is not allowed to bind to. This is useful for, e.g., VFIO GPU passthrough.

### DIFF
--- a/docs/wiki/Configuration:-Debug-Options.md
+++ b/docs/wiki/Configuration:-Debug-Options.md
@@ -18,6 +18,8 @@ debug {
     disable-direct-scanout
     restrict-primary-scanout-to-matching-format
     render-drm-device "/dev/dri/renderD129"
+    ignore-drm-device "/dev/dri/renderD128"
+    ignore-drm-device "/dev/dri/renderD130"
     force-pipewire-invalid-modifier
     dbus-interfaces-in-non-session-instances
     wait-for-frame-completion-before-queueing
@@ -112,6 +114,20 @@ You can set this to make niri use a different primary GPU than the default one.
 ```kdl
 debug {
     render-drm-device "/dev/dri/renderD129"
+}
+```
+
+### `ignore-drm-device`
+
+<sup>Since: next release</sup>
+
+List DRM devices that niri will ignore.
+Useful for GPU passthrough when you don't want niri to open a certain device.
+
+```kdl
+debug {
+    ignore-drm-device "/dev/dri/renderD128"
+    ignore-drm-device "/dev/dri/renderD130"
 }
 ```
 
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -83,6 +83,8 @@ pub struct Tty {
     primary_node: DrmNode,
     // DRM render node corresponding to the primary GPU.
     primary_render_node: DrmNode,
+    // Ignored DRM nodes.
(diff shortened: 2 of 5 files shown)


nekopsykose:
> This is useful for, e.g., VFIO GPU passthrough.

Can't you accomplish the same thing by explicitly binding to the device you want with `render-drm-device`? Disabling the "wrong" devices is just doing the inverse of what you can already do, isn't it?

docs/wiki/Configuration:-Debug-Options.md
```diff
@@ -18,6 +18,8 @@ debug {
     disable-direct-scanout
     restrict-primary-scanout-to-matching-format
     render-drm-device "/dev/dri/renderD129"
+    disable-drm-device "/dev/dri/renderD128"
+    disable-drm-device "/dev/dri/renderD130"
```
YaLTeR: Maybe it should say "ignore-drm-device"? Idk, what is it called in other compositors?

src/backend/tty.rs
```diff
@@ -320,6 +322,8 @@ impl Tty {
                 Ok::<_, anyhow::Error>((primary_node, primary_render_node))
             })?;
 
+        let disabled_nodes = disabled_nodes_from_config(&config.borrow());
```
YaLTeR: I feel like this setting should be live-reloaded. Since you find out at runtime that the device that you wanted to passthrough is taken by niri, and try to disable it.

shaunren:
> > This is useful for, e.g., VFIO GPU passthrough.
> 
> Can't you accomplish the same thing by explicitly binding to the device you want with `render-drm-device`? Disabling the "wrong" devices is just doing the inverse of what you can already do, isn't it?

`render-drm-device` only specifies the GPU used for rendering. Niri still binds to other GPUs by default, since monitors could possibly be connected to the other GPUs.

YaLTeR (commented):
fwiw cosmic-comp lazily creates the renderer only when some output is connected to the device. It's automatic but I don't know if that approach lets VFIO passthrough use it (since the drm node is still opened as I understand).

docs/wiki/Configuration:-Debug-Options.md
```diff
@@ -115,6 +117,17 @@ debug {
 }
 ```
 
+### `disable-drm-device`
+
+Disallow niri from binding to the the DRM device.
```
YaLTeR: typo and missing since next release

nekopsykose:
> Niri still binds to other GPUs by default, since monitors could possibly be connected to the other GPUs.

That makes sense, thanks :)

src/backend/tty.rs
shaunren: Implemented live-reloading.

docs/wiki/Configuration:-Debug-Options.md
shaunren: I'm not sure what's the best wording here. wlroots has the `WLR_DRM_DEVICES` environment var, which is a list of allowed DRM devices; and kwin has `KWIN_DRM_DEVICES`.

YaLTeR:
I did some fixes and rebased. I'll merge config includes, then push and merge this branch.

YaLTeR:
Thanks!

PR #2403, opened 2025-09-17
feat: expose keybinds through IPC

HigherOrderLogic:
Expose key binds list through the IPC, which maybe useful for shells to create a custom cheatsheet.

This PR builds but I havent test it yet, since I want to get some feedback on a few implementation details.

### DIFF
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -48,6 +48,7 @@ pub fn handle_msg(mut msg: Msg, json: bool) -> anyhow::Result<()> {
         Msg::EventStream => Request::EventStream,
         Msg::RequestError => Request::ReturnError,
         Msg::OverviewState => Request::OverviewState,
+        Msg::Binds => Request::Binds,
     };
 
     let mut socket = Socket::connect().context("error connecting to the niri socket")?;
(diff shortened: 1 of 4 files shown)


src/ipc/server.rs
```diff
@@ -446,6 +446,86 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
...
+                            hotkey_overlay_title,
+                            action,
+                            ..
+                        } = b;
+
+                        niri_ipc::Bind {
+                            trigger,
+                            modifiers,
+                            // TODO: get feedback on how to implement this
+                            action: format!("{action:?}"),
```
HigherOrderLogic: How should I format the action here? I want to avoid creating a duplicate struct as much as possible...

src/ipc/server.rs
```diff
@@ -446,6 +446,86 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
...
+                let binds = state
+                    .niri
+                    .config
+                    .borrow()
+                    .binds
+                    .0
+                    .iter()
+                    .map(|b| {
+                        let trigger = match b.key.trigger {
+                            niri_config::Trigger::Keysym(k) => k.name().map(String::from),
```
HigherOrderLogic: There should be a better way to format the trigger key here, instead of returning an `Option`.

src/ipc/client.rs
```diff
@@ -489,6 +490,52 @@ pub fn handle_msg(msg: Msg, json: bool) -> anyhow::Result<()> {
...
+                output.push_str(format!("Can repeat: {repeat}").as_ref());
+                if let Some(cd) = cooldown {
+                    let ms = cd.as_millis();
+                    output.push_str(format!("Cooldown: {ms}ms\n").as_ref());
+                } else {
+                    output.push_str("No cooldown");
+                }
+                output.push_str(format!("Allow when locked: {allow_when_locked}\n").as_ref());
+                output.push_str(format!("Allow inhibiting: {allow_inhibiting}\n").as_ref());
+                output.push_str("\n");
```
HigherOrderLogic: I dont really like how the string output is formatted here, any idea to improve it is welcome.

HigherOrderLogic (commented):
Here are the implementation details I would like to get feedback on.

PR #2436, opened 2025-09-24
layout/scrolling: window size limit fix in set_column_width

shettysach:
Solves #3116  

This change clamps width between min_width and max_width in `set_column_width` (src/layout/scrolling.rs). 
The implementation is similar to the one in `update_tile_sizes_with_transaction` (src/layout/scrolling.rs).


src/layout/scrolling.rs
```diff
@@ -4860,21 +4671,62 @@ impl<W: LayoutElement> Column<W> {
...
+        let min_size: Vec<_> = self
+            .tiles
+            .iter()
+            .map(Tile::min_size_nonfullscreen)
+            .map(|mut size| {
+                size.w = size.w.max(1.);
+                size.h = size.h.max(1.);
+                size
+            })
+            .collect();
```
nerdwave-nick: Could save allocations / space by just using the width, not the whole size.

The max also guarantees non-NaN value, so there's no need to check for that later

Lastly the unwrap (which comes later) could panic if tiles is empty I believe, so I've changed it to an `unwrap_or` with min-width `1.`;


```suggestion
        let min_width = self
            .tiles
            .iter()
            .map(Tile::min_size_nonfullscreen)
            .map(|size| {
                // gets out edge cases/NaN
                // limits to width only early
                size.w.max(1.);
            })
            .max()
            .unwrap_or(1);
```

src/layout/scrolling.rs
```diff
@@ -4860,21 +4671,62 @@ impl<W: LayoutElement> Column<W> {
...
+            .map(Tile::max_size_nonfullscreen)
+            .collect();
+
+        // Compute the column width.
+        let min_width = min_size
+            .iter()
+            .map(|size| NotNan::new(size.w).unwrap())
+            .max()
+            .map(NotNan::into_inner)
+            .unwrap();
```
nerdwave-nick: Would not be necessary with the change above
```suggestion
```

src/layout/scrolling.rs
```diff
@@ -4860,21 +4671,62 @@ impl<W: LayoutElement> Column<W> {
...
+                size.w = size.w.max(1.);
+                size.h = size.h.max(1.);
+                size
+            })
+            .collect();
+        let max_size: Vec<_> = self
+            .tiles
+            .iter()
+            .map(Tile::max_size_nonfullscreen)
+            .collect();
```
nerdwave-nick: This also limits to width early, then takes care of edge cases my using `max`, and uses *potentially* clearer semantics with the `INFINITY`
```suggestion
        let max_width = self
            .tiles
            .iter()
            .map(Tile::max_size_nonfullscreen)
            .filter_map(|size| {
                // gets rid of NaN and edge cases
                let w = size.w.max(0.);
                if w == 0. {
                    None
                } else {
                    Some(w)
                }
            })
            .min()
            .unwrap_or(f64::INFINITY); // better semantics?
```

src/layout/scrolling.rs
```diff
@@ -4860,21 +4671,62 @@ impl<W: LayoutElement> Column<W> {
...
+                let w = size.w;
+                if w == 0. {
+                    None
+                } else {
+                    Some(NotNan::new(w).unwrap())
+                }
+            })
+            .min()
+            .map(NotNan::into_inner)
+            .unwrap_or(f64::from(i32::MAX));
```
nerdwave-nick: Would be obsolete with above changes
```suggestion
```

nerdwave-nick (commented):
Now forgive me if I'm mistaken, I am not really a rust dev and went at this purely based off of what I'm seeing in this code and some googling, but I think this might be a more efficient AND more defensive way of doing this.

shettysach:
> Now forgive me if I'm mistaken, I am not really a rust dev and went at this purely based off of what I'm seeing in this code and some googling, but I think this might be a more efficient AND more defensive way of doing this.

Thank you. I think `update_tile_sizes_with_transaction` could also benefit from the same changes.

PR #2447, opened 2025-09-25
Remove non UTF-8 characters from Layer Shell Components filename

valerie-tar-gz:
The Layer Shell Components filename uses a non-UTF8 dash in its filename. This breaks Anaconda when installing a Fedora Atomic image that includes Niri.


YaLTeR:
Is this gonna break the title on GitHub Wiki and/or on our gh pages wiki?

image

image

YaLTeR:
Also, "non-UTF-8" does not make sense. I doubt that this an invalid UTF-8 character?

valerie-tar-gz:
> Is this gonna break the title on GitHub Wiki and/or on our gh pages wiki?

Possibly, I can look into it.

> Also, "non-UTF-8" does not make sense. I doubt that this an invalid UTF-8 character?

It uses a non-standard dash in the filename (I think it uses a hyphen instead), rpm-ostree fails because it can't "convert to current locale". There's been similar issues with non-english characters in other projects. 
image

YaLTeR:
> It uses a non-standard dash in the filename (I think it uses a hyphen instead), rpm-ostree fails because it can't "convert to current locale". There's been similar issues with non-english characters in other projects.

Well, some assorted thoughts on this:
- it cannot be converted *from* UTF-8 to "current locale" implying that it is in fact valid UTF-8
- why is it even converting UTF-8 to something? I though it was 2025 and everything was using UTF-8
- why is it treating a pathname as UTF-8 in the first place? Paths are famously not guaranteed to be UTF-8
- could probably also just rename the file in the Fedora package if it causes problems

tulilirockz:
The problem isn't with anaconda, but its directly with Ostree. For some reason, on the OSTree code it seems that a path having non-ascii characters just straight up breaks it, here is the upstream issue: https://github.com/bootc-dev/bootc/issues/975, they even have a lint for it now https://github.com/bootc-dev/bootc/pull/983

https://github.com/ostreedev/ostree/blob/6729096dc6d745881f81beaf9e001c2a03ddfa58/src/libotutil/ot-unix-utils.c#L53-L54 - This is a relevant line in the code

valerie-tar-gz:
Pretty much what Tulip said. I think I got UTF-8 and Ascii muddled (my bad).

Changing the Fedora package would work (and maybe COPRs too).

YaLTeR:
Anyway my main concern with renaming the file in the repo is that I'd like the GitHub Wiki page title to keep working

valerie-tar-gz:
Changing the packages is probably the better option then, yeah.

auri-f5bde6:
> Is this gonna break the title on GitHub Wiki and/or on our gh pages wiki?
> image image

From my limited understanding off mkdocs, it should be fine as the title is defined in https://github.com/YaLTeR/niri/blob/e837e39623457dc5ad29c34a5ce4d4616e5fbf1e/docs/mkdocs.yaml#L81C1-L81C56
That said, I'm pretty sure one have to find all occurrence of the old file name and replace it with the new one.
Edit: 
The only reference is in the file linked it seemes
```
diff --git a/docs/mkdocs.yaml b/docs/mkdocs.yaml
index 68253c04..bc56bef8 100644
--- a/docs/mkdocs.yaml
+++ b/docs/mkdocs.yaml
@@ -78,7 +78,7 @@ nav:
     - Tabs: Tabs.md
     - Overview: Overview.md
     - Screencasting: Screencasting.md
-    - Layer‐Shell Components: Layer‐Shell-Components.md
+    - Layer‐Shell Components: Layer-Shell-Components.md
     - IPC, niri msg: IPC.md
     - Application-Specific Issues: Application-Issues.md
     - Nvidia: Nvidia.md
```

bugaevc:
> The problem isn't with anaconda, but its directly with Ostree.

According to your links, bootc & OSTree are fine with characters that are (encoded as) valid UTF-8, but aren't valid ASCII. In other words: it's invalid UTF-8 that they would reject, not any non-ASCII characters.

noelmiller:
> Anyway my main concern with renaming the file in the repo is that I'd like the GitHub Wiki page title to keep working

All the other files are valid ASCII as far as I can tell. The way I tested if they were was using: 
`find . -print0 | perl -0 -nE 'if (my @m = /([^[:ascii:]])/g) { print "$_ -> Found: ", join(", ", @m), "\n" }'`

In my PR (https://github.com/YaLTeR/niri/pull/2735), you can see that it just renames the file, all the other files are valid ASCII, so it would be unlikely it would break the page title on Github Wiki. I could try testing it in my fork and see if that's the case.

noelmiller:
> > Is this gonna break the title on GitHub Wiki and/or on our gh pages wiki?
> > image image
> 
> From my limited understanding off mkdocs, it should be fine as the title is defined in https://github.com/YaLTeR/niri/blob/e837e39623457dc5ad29c34a5ce4d4616e5fbf1e/docs/mkdocs.yaml#L81C1-L81C56 That said, I'm pretty sure one have to find all occurrence of the old file name and replace it with the new one. Edit: The only reference is in the file linked it seemes
> 
> ```
> diff --git a/docs/mkdocs.yaml b/docs/mkdocs.yaml
> index 68253c04..bc56bef8 100644
> --- a/docs/mkdocs.yaml
> +++ b/docs/mkdocs.yaml
> @@ -78,7 +78,7 @@ nav:
>      - Tabs: Tabs.md
>      - Overview: Overview.md
>      - Screencasting: Screencasting.md
> -    - Layer‐Shell Components: Layer‐Shell-Components.md
> +    - Layer‐Shell Components: Layer-Shell-Components.md
>      - IPC, niri msg: IPC.md
>      - Application-Specific Issues: Application-Issues.md
>      - Nvidia: Nvidia.md
> ```

If it references it there, you would likely just want to make sure that all the hyphens in the name are valid ASCII as well. It's such a weird problem to have. 

Renaming it as part of packaging it for just Fedora wouldn't solve the problem for all bootc images since it is an ostree issue. It either needs to be fixed upstream in ostree or it needs to be fixed here. This file name is the only one with this issue. There are several other docs files that have hyphens in them that are using valid ASCII.

YaLTeR:
> All the other files are valid ASCII as far as I can tell

Yes, because no other pages needed to have a visible dash in the page title.

> In my PR (#2735), you can see that it just renames the file, all the other files are valid ASCII, so it would be unlikely it would break the page title on Github Wiki. I could try testing it in my fork and see if that's the case.

The name of the file is exactly what is used for the GitHub Wiki page title. ASCII dashes become spaces. I need a visible dash in that page though.

YaLTeR:
If it's possible to override the GitHub wiki page title with a first-level heading, then you can do that. But I don't know off the bat, please test.

YaLTeR:
Seems this shouldn't be necessary soon: https://github.com/YaLTeR/niri/pull/3307#issuecomment-3796662350

YaLTeR:
I think this will be fixed by https://github.com/ostreedev/ostree/pull/3559

Issue #1071, opened 2025-01-30
Vertical layout option for portrait monitors


YaLTeR:
Instead of infinite horizontal and fixed vertical space, for portrait monitors it makes sense to have infinite vertical and fixed horizontal space. So the same scrolling layout, but going down instead of to the right. And I guess workspaces go sideways then?

Originally requested in: https://github.com/YaLTeR/niri/discussions/757

Unfortunately, this would be quite an annoying and complex refactor in the code.

As an easier alternative, #401 will let you open windows as maximized by default on the portrait monitor.

Ornanovitch:
#401 seems to be a honest and functional compromise :)

chilledfrogs:
Both are useful for mobile devices

silvigon:
I would really love this! #401 would certainly help, but for my use case this would be the ideal solution.

slondr:
+1 this would be epic

WizardUli:
> As a dual monitor person with a portrait monitor, I would love to see [#401](https://github.com/YaLTeR/niri/issues/401) rather than this solution :)

For me it's completely opposite. I'm using my portrait monitor to either display a full-screen app like pdf viewer, webbrowser etc. on one workspace (currently in Sway) or on another workspace to display multiple consoles and file-browsers which I spawn **very** dynamically on-demand with count displayed being between 2-5 occasionally more having to use an additional workspace for them not to be squished too much. My use case would extremely well map to Niri's infinite strip if only the strip could be vertical. While I consider #401 to be very important I would not even consider it to be a workaround for the vertical strip.

danielytics:
I agree with @WizardUli that #401 is an important and useful enhancement, but not a replacement, at least in the long term. #401 would be an ok temporary "solution" to get portrait monitors to behave a bit better (and give users more control in multi monitor setups in general), so would be very welcome, but at least for me, it doesn't really "solve" what I hope for from a portrait monitor setup.

 I can work around it well enough through a mixture of fullscreen windows, workspaces, and consume-or-expel-window-left, so I wouldn't consider it "urgent" (for whatever urgent means in an open source request), but rather a nice to have.

Though, I wonder if `layout` could add an option to always spawn into the current column, instead of creating a new one? It wouldn't quite be the same as scrolling vertically, but it would at least mimic the portrait behavior on non-scrolling tiling window managers, and *I assume* its less work than supporting vertical scrolling?

YaLTeR:
[...]

PR #2449, opened 2025-09-26
Per-output and per-workspace layout

YaLTeR:
Alright, this is a big one, though most merge conflicts should be trivially resolvable.

In this PR we makes the layout config *mergeable*, as in, it can be combined from multiple parts. This is similar to how, say, border window rules can combine together and with the layout border options.

Then, this newfound merging ability is put to good use by introducing per-output and per-workspace layout overrides. In your config, you can now put a `layout {}` block inside `output {}` and `workspace {}` blocks:

```kdl
output "DP-2" {
    transform "90"

    // Overrides just for this output.
    layout {
        default-column-width { proportion 1.0; }
        // Anything else you want.
    }
}

workspace "aesthetic" {
    // Overrides just for this named workspace.
    layout {
        gaps 32

        struts {
            left 64
            right 64
            bottom 64
            top 64
        }

        border {
            on
            width 4
        }

        // Anything else you want.
    }
}
```

To do this, we split the config structs into a `Layout`, containing fully resolved values and no parsing, and a `LayoutPart`, containing only optional fields and parsing code. All parsing is done into `LayoutPart`s (which, among other things, means that base focus-ring and border get an `on` key like the window rules, resolving this small inconsistency), which are then merged together into the final resolved `Layout` (including the base options, the per-output overrides, and the per-workspace overrides).

In fact, this split improves other parts of the code, too. For example, default Layout option values are no longer duplicated in different parts of the code, and the layout code doesn't have to deal with as many parsing-only structs like `FloatOrInt` (since they all moved to `LayoutPart`).

The merging is written manually because different fields have different merging requirements. For example, we merge sections like `focus-ring` but override sections like `struts`. Color and gradient also have complex interdependent merging so that you can reliably override one with the other (this is not new, it existed in window rules, and actually didn't work in all the cases, which I fixed).

This layout merging also sets precedent and makes it obvious how to add merging to the rest of the config, which is a requirement for config includes. As such, this PR supersedes the merging parts of #2340 (which went with an automatic approach, which IMO cannot quite work due to the aforementioned differences in merging behavior).

Implements #401.

# Testing wanted

Mess around with custom layout overrides and try to break it, and look out for jank. One particular area of interest is interactive window moving: since interactively moved windows are "on top" of the entire layout (and particularly isn't "on" any workspace), I had to get a bit creative with config overrides, to make it work as expected (i.e. not flicker when you just want to drag and drop a window within a workspace with layout overrides).


Sempyos:
From my short testing, I haven't run into any issues with this while trying out all the layout options, along with swapping windows quickly between multiple configured workspaces. Neat PR, can't wait for merge
image

R0boter:
I tested on two monitor, it's worked,and i don't find any issues when i move window or column between two monitor
图片
图片
but i have one question, I set the vertical monitor like this
图片
set the horizontal monitor like this:
图片
when i move the column from horizontal to vertical monitor, the column width no automatic adaptation for vertical monitor
i don't know this is an issues or not?
for issues #1071 if we have default-window-height option to set window initial height，maybe we don't need vertical scroll
thanks for your job!!!

YaLTeR:
> when i move the column from horizontal to vertical monitor, the column width no automatic adaptation for vertical monitor

Not entirely sure what you meant, but I tried it and I think it works as intended? The window retains the proportion that it had. So if it was `proportion 1.0` on the vertical monitor, when you move it to the horizontal, it remains `proportion 1.0`, so it'll take up 100% of the width.

R0boter:
> > when i move the column from horizontal to vertical monitor, the column width no automatic adaptation for vertical monitor
> 
> Not entirely sure what you meant, but I tried it and I think it works as intended? The window retains the proportion that it had. So if it was `proportion 1.0` on the vertical monitor, when you move it to the horizontal, it remains `proportion 1.0`, so it'll take up 100% of the width.

I tested it again and it works fine, maybe I was looking at it wrong, thanks again

nickjj:
You mentioned this PR supersedes https://github.com/YaLTeR/niri/pull/2340 but that PR supported being able to include config files so you can situationally override specific config options in separate files.

Will that be supported as well with this PR?

Splitting things into separate configs allows much easier customizations per device, for example check out this comment thread: https://github.com/YaLTeR/niri/discussions/1322#discussioncomment-14215423

YaLTeR:
I mentioned that "this PR supersedes the merging parts of https://github.com/YaLTeR/niri/pull/2340". But I worked on includes on top of this PR's commits, it's still WIP in https://github.com/YaLTeR/niri/commits/wip/include

nickjj:
Awesome thanks, I'll keep an eye out for when both of these land in a release!

Really looking forward to switching to Niri!

YaLTeR:
Thanks everyone for testing!

PR #2456, opened 2025-09-26
Implement release keybinds and modifier-only binds

### LINKED ISSUE #605: Modifier-only keybinds exist

### LINKED ISSUE #1813: Cant set Keybind to `Mod`
I would like to bind `toggle-overview` to `Mod`. i seemingly cant do that, since niri requires a Modifier and an action Key for a keybind.

my niri config (using nixos niri-flake): https://codeberg.org/noofel/config/src/branch/main/programs/wm/compositors/niri/default.nix

### System Information

* niri version: `niri 25.05 (unknown commit)`

* Distro: `NixOS 25.11 (Unstable Branch)`

* GPU: 
(i have no idea sorry, also dont think its important)

* CPU: Intel i5-8350 @ 1.7GHz x8

flowerysong:
This builds on the work @Glitch752 did in #1923, but takes it in a slightly different direction.

Instead of tying bind validity to (notional, because `Mod` was never actually suppressed) key suppression status and whether any bind has already triggered, we track the most recently pressed key, make it the only one that can trigger a bind on release, and invalidate it when any key is released or any key or mouse button is pressed. This should prevent these binds from unexpectedly triggering when a user is trying to input something else, while allowing intentional binding of the same key to actions on press and release (e.g. for push to talk.)

I also made the implementation of modifier-only binds more generic. Non-`Mod` modifier binds are still configured with keysyms for the trigger (so `Ctrl release=true { toggle-overview; }` won't work but `Control_L release=true { toggle-overview; }` will.)

It seems to function as intended.

Closes #1923, fixes #1813, closes #1148, closes #605.


src/input/mod.rs
```diff
@@ -3999,76 +4079,124 @@ fn should_intercept_key(
...
+            // it was inhibited on press it wouldn't be in last_key.
+            ShouldInterceptResult::ForwardAndHandle(bind)
+        }
+        _ => ShouldInterceptResult::Forward,
     }
 }
 
+/// Finds the specified keybind. Note that release binds can be
+/// returned even when pressed=true because we still need to inhibit
+/// them.
```
Glitch752: I understand why this needs to be done, but I think this "probably returns a match" logic makes code using `find_bind` a lot more error-prone. I don't have any concrete suggestions at the moment, but maybe there's a way to restructure this into something like `find_bind_with_keysym` and `find_exact_bind`?

src/niri.rs
```diff
@@ -316,6 +316,7 @@ pub struct Niri {
     pub seat: Seat<State>,
     /// Scancodes of the keys to suppress.
     pub suppressed_keys: HashSet<Keycode>,
+    pub last_key: Option<Keycode>,
```
Glitch752: This approach is much clearer to me than the HashMap, so I like it. `last_key`, while descriptive of what it represents, doesn't do a very good job explaining this value's purpose in my opinion though. Maybe something like `active_release_key`?

src/input/mod.rs
```diff
@@ -569,11 +604,13 @@ impl State {
                 self.backend.change_vt(vt);
                 // Changing VT may not deliver the key releases, so clear the state.
                 self.niri.suppressed_keys.clear();
+                self.niri.last_key = None;
```
Glitch752: These feel easy to forget to put together, so maybe some method like `reset_held_key_state` would be better and clearer?

Glitch752:
I'm not great at code review, but I added a few of my thoughts. Otherwise, this looks good to me and seems like a more robust implementation strategy.
Thanks again for helping to push this feature forward a bit :)

Glitch752:
I would add: something to think about is that certain release binds may benefit from overlapping. For example, I would prefer if I could use `Mod release=true` while a push-to-talk key is held. I'm not sure how this implementation could be extended to handle that.

flowerysong:
> I'm not great at code review, but I added a few of my thoughts.

I think maybe you forgot to finish this? Once you have pending comments you can click on the button in the upper right to finish the review (with an optional additional comment) and make the comments visible.

GitHub screenshot showing the review interface

> For example, I would prefer if I could use `Mod release=true` while a push-to-talk key is held. I'm not sure how this implementation could be extended to handle that.

I don't think that's something you should want. 😁 If you want to perform keyboard input in between unmuting and muting, it would make more sense to me to set it up as a toggle: `Control_L release=true { spawn "wpctl" "set-mute" "@DEFAULT_AUDIO_SOURCE@" "toggle"; }`

In general I think making the behaviour predictable and easy for users to reason about is more important than trying to cover absolutely every way people might want it to work. Just being able to bind `Mod` probably accounts for 99% of what people want from this feature.

> Thanks again for helping to push this feature forward a bit :)

Oh, no, thank you. Without your PR I would still be begrudgingly hitting `Mod+D` to bring up the launcher.

Glitch752 (commented):
oops! So sorry haha, not sure how I missed that

Glitch752:
> I don't think that's something you should want

This is how I use push-to-talk on other OSes; sometimes I hold down the key and, while talking, press mod. I agree that this is somewhat unintuitive anyway though.

> In general I think making the behaviour predictable and easy for users to reason about is more important than trying to cover absolutely every way people might want it to work

100% agree with this. There's always the option to add more complexity later if it's needed (and in this case it probably isn't anyway). I just wanted to add that in case you hadn't considered it.

flowerysong:
> There's always the option to add more complexity later if it's needed (and in this case it probably isn't anyway). I just wanted to add that in case you hadn't considered it.

You need two things to achieve this: a second type of release bind that always triggers instead of caring about validity, and (if you want to support using modifier keys for this) a way to mask certain modifier keys after their bind triggers so they're treated as though they aren't pressed. I don't think the implementation of either part would be difficult.

src/input/mod.rs
flowerysong: Another way to look at it: it always returns a match, it just uses `pressed` as a hint as to which match is preferred when there are multiple. The better way to do this is probably to return a list, but that would be a more involved change.

src/input/mod.rs
flowerysong: If it were a complicated process or something replicated a half-dozen places a refactor might be worthwhile, but with only two places this is done and only two steps I think adding an abstraction would make the code harder to follow for no real benefit.

Glitch752:
Alright, that all makes sense; this looks good to me. I'll switch to this branch on my computer to get another tester, but I tried a lot of release bind combinations and all the behavior feels intuitive/correct.

flowerysong:
> a second type of release bind that always triggers instead of caring about validity

I went ahead and implemented this, mostly because of the other issue I ran into: we need to avoid triggering for `<modifier>+mouse`, and adding more invalidation highlighted the fact that PTT would be a terrible UX if the release bind is too easy to skip.

The modifier masking still feels like it would be an unnecessary complication for an initial feature merge, but I might find the time to futz around with it anyway.

hbeberman:
I've been using this patch for a few days as a global PTT mute/unmute and it's been working great for me.
```kdl
    KP_Insert { spawn-sh "wpctl set-mute @DEFAULT_AUDIO_SOURCE@ 0"; }
    KP_Insert release=true allow-invalidation=false  { spawn-sh "wpctl set-mute @DEFAULT_AUDIO_SOURCE@ 1"; }
```

JohnConnorNPC:
So I have Mod by itself assigned to fuzzel, and Mod + Scroll wheel to move workspaces up/down, and the mod by itself is being triggered after letting go of mod after using the scroll wheel.

flowerysong:
> So I have Mod by itself assigned to fuzzel, and Mod + Scroll wheel to move workspaces up/down, and the mod by itself is being triggered after letting go of mod after using the scroll wheel.

Should be fixed, but I didn't actually test it.

JohnConnorNPC:
> > So I have Mod by itself assigned to fuzzel, and Mod + Scroll wheel to move workspaces up/down, and the mod by itself is being triggered after letting go of mod after using the scroll wheel.
> 
> Should be fixed, but I didn't actually test it.

It did not work, but I made these changes and it works:



 src/input/mod.rs

  2854 +                  // Invalidate any pending release trigger, similar to mouse button presses
   2855 +                  self.niri.valid_release_trigger = None;

flowerysong:
That line number doesn't make any sense; are you sure you pulled down the latest version of this PR?

It would help if you provided the full context diff, so that we can see the lines surrounding your change.

JohnConnorNPC:
> That line number doesn't make any sense; are you sure you pulled down the latest version of this PR?
> 
> It would help if you provided the full context diff, so that we can see the lines surrounding your change.

I hope I got this right.   (It did solve my issue and has not shown any other issues yet)

     @@ -2851,6 +2851,9 @@ impl State {
                  let should_handle =
                      should_handle_in_overview || self.niri.mods_with_wheel_binds.contains(&modifiers);
                  if should_handle {
     +                // Invalidate any pending release trigger, similar to mouse button presses
     +                self.niri.valid_release_trigger = None;
     +
                      let horizontal = horizontal_amount_v120.unwrap_or(0.);
                      let ticks = self.niri.horizontal_wheel_tracker.accumulate(horizontal);
                      if ticks != 0 {

flowerysong:
Okay, that's (one of) the changes I made, so you just didn't update your local copy.

JohnConnorNPC:
> Okay, that's (one of) the changes I made, so you just didn't update your local copy.

ok I messed up getting the right patch. much apologies.  thankyou

YaLTeR:
I haven't looked at this PR in detail yet but the problem posed in the last commit piqued my interest. The special flag solution seems error-prone: the user needs to both know about and remember to set the flag, or risk a potentially privacy-sensitive release bind not triggering.

I have this alternative idea to solve this problem what if press+release binds were defined with this special syntax?

```kdl
binds {
    // Push to talk.
    Ctrl+Alt {
        press { spawn "unmute.sh"; }
        release { spawn "mute.sh"; }
    }
}
```

When handling this kind of bind, if niri runs the `press` action, then it will guarantee running the `release` action, regardless of whether any intermediate keys were pressed.

I would go even further and make this the only way to define release binds, instead of the `release=true` flag, so e.g. you would have:

```kdl
binds {
    // Toggle the Overview on Mod by itself.
    Super {
        release { toggle-overview; }
    }
}
```

This would be equivalent to a sole `release=True` bind under this PR.

By restricting release binds to this syntax, we can also avoid other awkwardness:
- No need for special handling of press/release duplicate key detection. You can only declare press and release on the same key together using this syntax.
- Press and release always stay together in the config, no way for them to get lost.
- Properties apply in an obvious way: `repeat` only to the press bind, `allow-when-locked` and `hotkey-overlay-title` to the combined press/release bind. No chance to mess up by setting `allow-when-locked` only on press and not release for example.

What do you think? Does this still cover all release bind usecases?

miku4k:
would that make `press` required for non-release binds?
```kdl
binds {
    Mod+Space {
        // spawn "rofi";
        press { spawn "rofi"; }
    }
}
```

YaLTeR:
No. We don't have any actions named press and release so there shouldn't be any conflict

b0o:
> I have this alternative idea to solve this problem what if press+release binds were defined with this special syntax?
> 
> ```kdl
> binds {
>     // Push to talk.
>     Ctrl+Alt {
>         press { spawn "unmute.sh"; }
>         release { spawn "mute.sh"; }
>     }
> }
> ```

This is exactly what I want! This would be beautiful.

VaclavC:
+1. Don't do flags, just use YaLTeR's syntax, please.

MaxistheSpy:
> I haven't looked at this PR in detail yet but the problem posed in the last commit piqued my interest. The special flag solution seems error-prone: the user needs to both know about and remember to set the flag, or risk a potentially privacy-sensitive release bind not triggering.
> 
> I have this alternative idea to solve this problem what if press+release binds were defined with this special syntax?
> 
> ```kdl
> binds {
>     // Push to talk.
>     Ctrl+Alt {
>         press { spawn "unmute.sh"; }
>         release { spawn "mute.sh"; }
>     }
> }
> ```
> 
> When handling this kind of bind, if niri runs the `press` action, then it will guarantee running the `release` action, regardless of whether any intermediate keys were pressed.
> 
> I would go even further and make this the only way to define release binds, instead of the `release=true` flag, so e.g. you would have:
> 
> ```kdl
> binds {
>     // Toggle the Overview on Mod by itself.
>     Super {
>         release { toggle-overview; }
>     }
> }
> ```
> 
> This would be equivalent to a sole `release=True` bind under this PR.
> 
> By restricting release binds to this syntax, we can also avoid other awkwardness:
> 
> * No need for special handling of press/release duplicate key detection. You can only declare press and release on the same key together using this syntax.
> * Press and release always stay together in the config, no way for them to get lost.
> * Properties apply in an obvious way: `repeat` only to the press bind, `allow-when-locked` and `hotkey-overlay-title` to the combined press/release bind. No chance to mess up by setting `allow-when-locked` only on press and not release for example.
> 
> What do you think? Does this still cover all release bind usecases?

Perhaps this is too much but it seems like possible are “press” “held” “other key pressed” “release”. Given release can kind of cover held so long as it is able to cancel whatever was triggered on press it’s probably safe to consider that covered. It would at a minimum be nice to have an option for basic handling if another key is pressed. In that case the user might want one of three things to happen. Either keep the action as described. Cancel it or additionally run a different action. The “run a different action” seems like a very edge case and It may be premature to support it as. That being said I know directly I’ve gotten value out of key binds that allow cancelation if any other key is pressed.

The simplest way I could see this being done is adding some concept of a “interruptible release” either with a flag, third keyword or my personal favorite making the syntax “release{ interrupt {} always {} }”. Where release {toggle overview} would be syntactic sugar for release {always{toggle overview}} although this would require the ability to do interrupt {cancel}; to allow for canceling the always block (naming might need work.)

(This is a path I don’t suggest going down but it’s worth noting The nested syntax also makes it sensical to potentially at some point have a press{interrupt {} always{}} where the default behavior(ie not declaring one) is syntactic sugar for interrupt. So default being if we do nothing for example super then T we do nothing on super but if always is specified it would activate immediately ignoring any future combos made off that press. I don’t see any reason to implement this and it seems like it would cause more problems then it would solve but making the release syntax nested blocks rather than flags or a keyword would make this extension much cleaner if there was demand to implement it)

z411:
Thanks for this, I've been using this patch to open overview with my Mod (set to Alt) key no problem:

```
Mod repeat=false release=true { toggle-overview; ]
```

That said, I found an issue when trying to bind it to a specific modifier key, in this case Super:

```
 397 │     Super repeat=false release=true { toggle-overview; }
     ·     ──┬──
     ·       ╰── invalid value
```

I tried Super+Super_L as mentioned by #605 but it does nothing.

flowerysong:
@z411 It's just `Super_L`.  Since the key has been released the modifier is no longer in effect. See the documentation added in the PR: https://github.com/YaLTeR/niri/pull/2456/files#diff-850fc2afe64e0bb69b9677090803b3571f4889fb3f8c7faa328606f472a28e68R98-R105

`Mod` has special handling so that the two binds can be the same, but trying to do that for the normal modifiers seems very fragile to me.

z411:
@flowerysong Works perfectly, thank you. This works great for my workflow.

bea231:
Thank you for this PR! Looking for its approval and merge

YaLTeR:
@flowerysong hey, are you interested in implementing my proposal in https://github.com/YaLTeR/niri/pull/2456#issuecomment-3575904270?

zstg:
Looking forward to seeing this merged!

JustinSpedding:
I implemented the suggested config syntax changes here: https://github.com/niri-wm/niri/pull/3621. I am running it locally, and it seems to work as intended. Let me know if it looks good and/or if there are things that could be improved

mezoahmedii:
Why was this closed?

flowerysong:
I initially lost interest in contributing to niri after some bad experiences with the community, and that decision was reinforced by the project's acceptance of LLM usage.

I kept the PR up to date for a bit anyway, but I have closed it to reflect the fact that lack of interest has persisted for many months now and the PR has been slowly rotting instead of being maintained. (The part of this change that I actually use is just 214 lines, and rebasing just those made my life easier.)

There is a newer version of this patch series by someone else, anyway. I haven't reviewed it but I skimmed a couple comments and people seem reasonably happy with the direction it's gone.

PR #2479, opened 2025-09-30
Pick up #1283, Add support for custom modes and modelines.

ToxicMushroom:
This now adds support for custom modes and custom modelines as discussed in #333.
Allowing users to run niri with a mode that their monitor does not advertise.
Users should be **CAUTIOUS** when using this, running monitors out of spec can result in damage. 

Current added config syntaxes:

**Custom mode**
```kdl
output "eDP-1" {
    mode custom=true "1920x1080@60"
}
```

**Modeline**
```kdl
output "eDP-1" {
    modeline "1920x1080@59.96" 173.00  1920 2048 2248 2576  1080 1083 1088 1120 "vertical" false
}
```

Please let me know anything you'd me to modify/fix/add/remove :), I tried incorporating the comments Yalter left on the preceding PR by cmeissl.


docs/wiki/Configuration:-Outputs.md
```diff
@@ -83,6 +83,32 @@ output "eDP-1" {
 }
 ```
 
+<sup>Since: next release</sup> You can configure a custom mode (not offered by the monitor) by setting `custom=true`.
+In this case, the refresh rate is mandatory.
```
YaLTeR, line 98:
```suggestion
In this case, the refresh rate is mandatory.

```

docs/wiki/Configuration:-Outputs.md
```diff
@@ -83,6 +83,32 @@ output "eDP-1" {
...
+> [!CAUTION]
+> Custom modes may damage your monitor, especially if it's a CRT.
+> Follow the maximum supported limits in your monitor's instructions.
```
YaLTeR, line 102:
```suggestion
> Follow the maximum supported limits in your monitor's instructions.

```

docs/wiki/Configuration:-Outputs.md
```diff
@@ -83,6 +83,32 @@ output "eDP-1" {
...
+```kdl
+// Use a custom mode for this display.
+output "HDMI-A-1" {
+    mode custom=true "2560x1440@143.912"
+}
+```
+
+### `modeline`
```
YaLTeR, line 111:
```suggestion
### `modeline`

<sup>Since: next release</sup>
```

docs/wiki/Configuration:-Outputs.md
```diff
@@ -83,6 +83,32 @@ output "eDP-1" {
...
+
+Directly configures the monitor's mode via a modeline. These can be calculated via utilities such as [cvt](https://man.archlinux.org/man/cvt.1.en) or [gtf](https://man.archlinux.org/man/gtf.1.en).
```
YaLTeR:
```suggestion
Directly configures the monitor's mode via a modeline.
These can be calculated via utilities such as [cvt](https://man.archlinux.org/man/cvt.1.en) or [gtf](https://man.archlinux.org/man/gtf.1.en).

```

YaLTeR: Should say that it overrides mode

niri-config/src/output.rs
```diff
@@ -1,11 +1,120 @@
...
 
 use crate::gestures::HotCorners;
 use crate::{Color, FloatOrInt};
 
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Outputs(pub Vec<Output>);
 
+#[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
+pub struct Mode {
+    #[knuffel(property, default)]
```
YaLTeR:
```suggestion
    #[knuffel(property, default = false)]
```

niri-config/src/output.rs
```diff
@@ -1,11 +1,120 @@
...
+
+    #[knuffel(argument, str)]
+    pub mode: ConfiguredMode,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum HSyncPolarity {
+    PHSync,
+    NHSync,
+}
```
YaLTeR: Just import these two enums from niri-ipc, we do it elsewhere like SizeChange

niri-ipc/src/lib.rs
```diff
@@ -964,6 +964,50 @@ pub enum OutputAction {
         #[cfg_attr(feature = "clap", arg())]
         mode: ModeToSet,
     },
+    /// Set a custom output mode (One not advertised by your monitor, use at your own risk)
```
YaLTeR: Make the style consistent with other comments please. No out-of-place title case letters, period at the end, etc.

niri-ipc/src/lib.rs
```diff
@@ -964,6 +964,50 @@ pub enum OutputAction {
...
+    CustomMode {
+        /// Custom mode to set
+        #[cfg_attr(feature = "clap", arg())]
+        mode: ConfiguredMode,
+    },
+    /// Set a custom VESA CVT modeline.
+    #[cfg_attr(feature = "clap", command(name = "modeline"))]
```
YaLTeR: Why is explicit name needed here?

niri-ipc/src/lib.rs
```diff
@@ -964,6 +964,50 @@ pub enum OutputAction {
...
+    Modeline {
+        /// The rate at which pixels are drawn (MHz)
+        #[cfg_attr(feature = "clap", arg())]
+        clock: f64,
+        /// Horizontal active pixels (pixels).
```
YaLTeR: Is this unit really useful here?

niri-ipc/src/lib.rs
```diff
@@ -964,6 +964,50 @@ pub enum OutputAction {
...
+        #[cfg_attr(feature = "clap", arg())]
+        hdisp: u16,
+        /// Horizontal Sync Pulse start position (pixels).
+        #[cfg_attr(feature = "clap", arg())]
+        hsync_start: u16,
+        /// Horizontal Sync Pulse end position (pixels).
+        #[cfg_attr(feature = "clap", arg())]
+        hsync_end: u16,
+        /// Total horizontal number of pixels before resetting to (pixels).
```
YaLTeR: Sentence missing word at the end?

niri-ipc/src/lib.rs
YaLTeR: Everywhere else in the file too.

src/backend/tty.rs
```diff
@@ -846,21 +848,27 @@ impl Tty {
...
-                } else {
-                    String::new()
-                },
-            );
-        }
+        let mode = if let Some(modeline) = &config.modeline {
+            calculate_drmmode_from_modeline(&modeline)
+        } else {
+            let opt_mode = pick_mode(&connector, config.mode);
+            let (mode, fallback) = opt_mode.ok_or_else(|| anyhow!("no mode"))?;
```
YaLTeR: Seems this extract-variable is unnecessary and can be left as it was

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
     output_state.redraw_state = RedrawState::WaitingForEstimatedVBlank(token);
 }
 
+pub fn calculate_drmmode_from_modeline(modeline: &Modeline) -> DrmMode {
```
YaLTeR:
```suggestion
pub fn calculate_drm_mode_from_modeline(modeline: &Modeline) -> DrmMode {
```

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+    let pixel_clock_kilo_hertz = modeline.clock * 1000.0;
+    // Calculated as documented in the CVT 1.2 standard
```
YaLTeR: Link to documentation or reference source code

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+    let vrefresh_hertz = (pixel_clock_kilo_hertz * 1000.0)
+        / (modeline.htotal as u64 * modeline.vtotal as u64) as f64;
```
YaLTeR, line 2599: This should use checked ops and I guess return an error if they fail, since that would mean the user provided invalid very large numbers. Actually, is there any other validation that could be done here?

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+        ModeFlags::from(&modeline.hsync_polarity) | ModeFlags::from(&modeline.vsync_polarity);
+
+    let mode_name = format!(
+        "{}x{}@{:.2}",
+        modeline.hdisp, modeline.vdisp, vrefresh_hertz
+    );
+    let name = modeinfo_name_slice_from_string(&mode_name);
+
+    DrmMode::from(drm_mode_modeinfo {
+        clock: pixel_clock_kilo_hertz as u32,
```
YaLTeR: Should this be .round()ed?

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+        hdisplay: modeline.hdisp,
```
YaLTeR: Should our field names here match drm mode info names?

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+        video_opt: false,
+        vblank: 0f64,
+        additional_hblank: 0,
+        early_vsync_rqd: false,
+        int_rqd: false,
+        margins_rqd: false,
+    };
+    let cvt_timing = libdisplay_info::cvt::Timing::compute(options);
+
+    let hsync_start = width + cvt_timing.h_front_porch as u16;
```
YaLTeR, line 2661: Link to reference code

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+
+    let htotal = hsync_end + cvt_timing.h_back_porch as u16;
+    let vtotal = vsync_end + cvt_timing.v_back_porch as u16;
+
+    let clock = f64::round(cvt_timing.act_pixel_freq * 1000f64) as u32;
+    let vrefresh = f64::round(cvt_timing.act_frame_rate) as u32;
+
+    let flags = drm_ffi::DRM_MODE_FLAG_NHSYNC | drm_ffi::DRM_MODE_FLAG_PVSYNC;
+
+    let mode_name = format!("{width}x{height}_{:.2}", cvt_timing.act_frame_rate);
```
YaLTeR: why | instead of @?

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+        // Defaults
+        hskew: 0,
+        vscan: 0,
+    };
+
+    control::Mode::from(drm_ffi_mode)
+}
+
+/// Returns a c-string of maximally 31 rust string chars + zero terminator. Excess characters are
+/// dropped.
```
YaLTeR:
```suggestion
// Returns a c-string of maximally 31 Rust string chars + null terminator. Excess characters are
// dropped.
```

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
+fn modeinfo_name_slice_from_string(mode_name: &String) -> [core::ffi::c_char; 32] {
+    let mode_name_bytes = unsafe {
+        std::slice::from_raw_parts(
+            mode_name.as_bytes() as *const _ as *const core::ffi::c_char,
+            mode_name.len(),
+        )
+    };
```
YaLTeR: I feel like there's a way to do this without unsafe

src/backend/tty.rs
```diff
@@ -2396,18 +2427,140 @@ fn queue_estimated_vblank_timer(
...
-        let refresh = target.refresh.map(|r| (r * 1000.).round() as i32);
+        let target_mode = target.mode;
+
+        if target.custom {
+            if let Some(refresh) = target_mode.refresh {
+                let custom_mode =
+                    calculate_mode_cvt(target_mode.width, target_mode.height, refresh);
+                return Some((custom_mode, false));
+            } else {
+                warn!("ignoring custom mode without refresh rate");
```
YaLTeR, line 2732: Can we check this at parse time somehow?

src/backend/tty.rs
```diff
@@ -2639,3 +2792,141 @@ fn make_output_name(
...
+        1177,
+    ),
+    hskew: 0,
+    vscan: 0,
+    vrefresh: 144,
+    mode_type: ModeTypeFlags(
+        USERDEF,
+    ),
+}");
+    }
```
YaLTeR, line 3055:
```suggestion
    }

```

src/ipc/client.rs
```diff
@@ -511,6 +512,25 @@ fn print_output(output: Output) -> anyhow::Result<()> {
     let serial = serial.as_deref().unwrap_or("Unknown");
     println!(r#"Output "{make} {model} {serial}" ({name})"#);
 
+    let print_qualifier = |is_preferred: bool, is_current: bool, is_custom_mode: bool| {
+        let mut qualifier = Vec::new();
+        if is_preferred {
+            qualifier.push("preferred");
+        };
```
YaLTeR: This changes behavior. It prints preferred before current, whereas the original code originally printed current first as it's more important.

YaLTeR (changes requested):

YaLTeR:
Also ofc question how did you test this, and can we get anyone else test it before merging.

ToxicMushroom:
I tested the custom mode and modeline on my laptop. (to render unsupported resolutions)

src/backend/tty.rs
ToxicMushroom: yes good catch

niri-ipc/src/lib.rs
ToxicMushroom: not useful since its a repeated word. I was trying to keep the docs of the modeline fields consistent.
I've reworded these, please check if it is better/worse

ToxicMushroom:
I'll continue fixing the issues tomorrow, would you like me to add a way to unset the modeline to the IPC as well ? 
Since modelines always override modes, I believe there is currently no way to go back to a normal mode once you set a modeline in your niri session.

YaLTeR:
The way it should work is an IPC mode or custom mode or modeline request should override any previous requests. Same with the wlr foreign toplevel management I think

ToxicMushroom:
There were a couple problems I ran into when trying to encode such a last-wins approach
1. Encoding it into the config via the `Either` seems difficult to me, I did not see knuffel support for it. + With the `Either` approach I'm not sure I'd be able to keep old configs working.
2. When keeping them 2 variables then there needs to be a lot of manual state tracking as to what was set last, or unsetting the other mode/modeline when the inverse got set. 

Both seem a bit cumbersome. I prefer 1 but it is probably not backwards compatible or I'm being skill issued by rust/kdl.

YaLTeR:
Not sure I understand the problem, just clear the stored modeline or custom_mode?

src/backend/tty.rs
ToxicMushroom: Would you prefer a duplicate `CustomMode` config struct that makes the refresh required ?

src/backend/tty.rs
ToxicMushroom: I can add asserts that hdisp < hsync_start < hsync_end < htotal.
And same for the verticals. Not sure about what else could be validated.

src/backend/tty.rs
ToxicMushroom: I've made them match now, I like longer variable names anyway, I think I got the short one from some other docs.

Duncaen:
> Also ofc question how did you test this, and can we get anyone else test it before merging.

I've been using this pr for a custom modeline for the past 4 days, haven't had any issues.

My use case is basically to use reduced blanking which allows my gpu to lower its memory clock speed and the power consumption (https://gitlab.freedesktop.org/drm/amd/-/issues/1403).

ToxicMushroom:
Nice! I actually have the same gpu on my desktop so I should look into that..

src/backend/tty.rs
YaLTeR: Panicking the compositor at runtime is not exactly a good way to validate user input.. Fixed it.

src/protocols/output_management.rs
```diff
@@ -640,25 +643,15 @@ where
...
-                new_config.mode = Some(niri_ipc::ConfiguredMode {
-                    width: mode.width,
-                    height: mode.height,
-                    refresh: Some(mode.refresh_rate as f64 / 1000.),
+                new_config.mode = Some(niri_config::output::Mode {
+                    custom: true,
+                    mode: niri_ipc::ConfiguredMode {
+                        width,
+                        height,
+                        refresh: Some(refresh as f64 / 1000.),
```
YaLTeR, line 684: The protocol [says](https://wayland.app/protocols/wlr-output-management-unstable-v1#zwlr_output_configuration_head_v1:request:set_custom_mode):

> If set to zero, the refresh rate is unspecified.

What do we do with it? I made it do nothing if refresh is 0 for now.

src/protocols/output_management.rs
YaLTeR, line 1: We started hitting the `output's old mode count doesn't match new modes` error message because a custom mode can appear or disappear. This needs to be handled somehow (how does wlroots do it?).

src/backend/tty.rs
YaLTeR: I can't think of a convenient way to go about it yet. The best way is probably to replace `knuffel::Decode` derive on the Mode with a hand-written version that verifies that the refresh rate is set.

YaLTeR:
Oh, and another idea: how hard would it be to call `calculate_drm_mode_from_modeline` to verify for errors:
- inside the niri IPC handler so that error is sent back to the IPC before it even hits the tty path
- check during config parsing

src/protocols/output_management.rs
ToxicMushroom: Sway seems to picks 60hz when unset, I think doing nothing is also a valid choice.
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_output.h
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/drm/drm.c#L675
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/drm/util.c#L259

ToxicMushroom: Or we could reply with https://wayland.app/protocols/wlr-output-management-unstable-v1#zwlr_output_configuration_head_v1:enum:error:entry:invalid_custom_mode ?

YaLTeR: > Or we could reply with https://wayland.app/protocols/wlr-output-management-unstable-v1#zwlr_output_configuration_head_v1:enum:error:entry:invalid_custom_mode ?

That's an error, it'll kill the client.

> Sway seems to picks 60hz when unset, I think doing nothing is also a valid choice.

Eh, let's do nothing then.

src/protocols/output_management.rs
ToxicMushroom: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_output_management_v1.c#L915 here they add new modes to the head. I'll need to read more code to understand it fully.

YaLTeR: The bigger problem is what to do when it's changed and removed. I guess when changing to a different custom mode we can reuse the mode, but when it's removed? Keep it around?

ToxicMushroom:
> Oh, and another idea: how hard would it be to call `calculate_drm_mode_from_modeline` to verify for errors:
> 
>     * inside the niri IPC handler so that error is sent back to the IPC before it even hits the tty path
> 
>     * check during config parsing

Yeah I think that's a good idea.
This would add drm-ffi, drm-sys and anyhow crates to niri-config though if that's okay ?

I'm planning to move `modeinfo_name_slice_from_string` and `calculate_drm_mode_from_modeline` to niri_config then

YaLTeR:
> Yeah I think that's a good idea. This would add drm-ffi, drm-sys and anyhow crates to niri-config though if that's okay ?
> 
> I'm planning to move `modeinfo_name_slice_from_string` and `calculate_drm_mode_from_modeline` to niri_config then

Hmm, maybe without those extra deps for niri-config? Either just do the same checks there (or at least the simple subset of them), or make a separate function that does the checks without extra deps? niri-config uses miette instead of anyhow, so in the first option you'd just use miette (then you can also attach proper spans, i.e. underline htotal in the error msg if it's less than hdisplay), for the second option one way is to return `Result<(), String>` from the check instead of anyhow

src/backend/tty.rs
ToxicMushroom: I've implemented this approach now, I still need to thoroughly test the new config parsing things though

src/protocols/output_management.rs
ToxicMushroom: I've pushed an untested draft of this for now, I'll find out how clients behave when I finish up the IPC parsing.

ToxicMushroom:
Alright with my basic tests I can now switch from modeline to custom-mode to mode with ipc. And cosic-randr list seems to reflect the current state correctly.

niri-config/src/output.rs
```diff
@@ -248,6 +248,7 @@ impl<S: ErrorSpan> Decode<S> for Modeline {
             hsync_end,
             htotal,
         );
+        ensure!(0u16 < htotal, ctx, htotal_value, "htotal {} > 0", htotal);
```
YaLTeR: Consistency with other messages please. `htotal {} must be > 0` Here and in other places

niri-ipc/src/lib.rs
```diff
@@ -1047,6 +1047,90 @@ pub enum OutputAction {
     },
 }
 
+macro_rules! ensure {
+    ($cond:expr, $fmt:literal $($arg:tt)* ) => {
+        if !$cond {
+            return Err(format!($fmt $($arg)*).into());
```
YaLTeR: Why the `.into()`, it's already a String

src/ipc/server.rs
```diff
@@ -406,6 +406,7 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
             } else {
                 OutputConfigChanged::OutputWasMissing
             };
+            action.validate()?;
```
YaLTeR: Move this to the very first line, before we lock ipc_outputs

niri-ipc/src/lib.rs
```diff
@@ -1047,6 +1047,90 @@ pub enum OutputAction {
...
+        }
+    };
+}
+
+impl OutputAction {
+    /// Validates extra constraints that were difficult to represent with clap.
```
YaLTeR: `Validates some required constraints on the modeline and custom mode.`

niri-ipc/src/lib.rs
```diff
@@ -1047,6 +1047,90 @@ pub enum OutputAction {
...
+                    0u16 < *vtotal,
+                    "vtotal {} > 0",
+                    vtotal,
+                );
+                Ok(())
+            }
+            OutputAction::CustomMode {
+                mode: ConfiguredMode { refresh, .. },
+            } => {
+                if refresh.is_none() {
```
YaLTeR: Should also check for `== Some(0)`?

niri-config/src/output.rs
```diff
@@ -1,56 +1,326 @@
...
 
-#[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Mode {
-    #[knuffel(property, default = false)]
+    // #[knuffel(property, default = false)]
     pub custom: bool,
 
-    #[knuffel(argument, str)]
+    // #[knuffel(argument, str)]
```
YaLTeR: Can remove these entirely

niri-config/src/output.rs
```diff
@@ -1,56 +1,326 @@
...
+                };
+
+                let width = width.parse().map_err(|_| "error parsing width")?;
+                let height = height.parse().map_err(|_| "error parsing height")?;
+                let refresh = refresh
+                    .map(|refresh| refresh.parse().map_err(|_| "error parsing refresh rate"));
+                let refresh = match refresh {
+                    Some(refresh) => Some(refresh?),
+                    None => None,
+                };
```
YaLTeR: `ConfiguredMode::from_str()`?

niri-config/src/output.rs
```diff
@@ -1,56 +1,326 @@
...
+                format!("unexpected node `{}`", span.escape_default()),
+            ));
+        }
+
+        let mut arguments = node.arguments.iter();
+
+        let clock_value = arguments
+            .next()
+            .ok_or_else(|| DecodeError::missing(node, "missing clock_value argument"))?;
+        let clock = knuffel::traits::DecodeScalar::decode(clock_value, ctx)?;
```
YaLTeR: For this stuff perhaps just write a simple inline macro kinda like I did here

https://github.com/YaLTeR/niri/blob/e6f3c538da0c646bda43fcde7ef7dc3b771e0c8b/niri-config/src/lib.rs#L170-L174

niri-config/src/output.rs
```diff
@@ -1,56 +1,326 @@
...
+        if !(hdisplay < hsync_start) {
+            ctx.emit_error(DecodeError::Conversion {
+                source: format!(
+                    "hdisplay {} must be < hsync_start {}",
+                    hdisplay, hsync_start
+                )
+                .into(),
+                span: hdisplay_value.literal.span().clone(),
+            });
+        }
```
YaLTeR: Duplicate check

niri-config/src/output.rs
```diff
@@ -1,56 +1,326 @@
...
+                let width = width.parse().map_err(|_| "error parsing width")?;
+                let height = height.parse().map_err(|_| "error parsing height")?;
+                let refresh = refresh
+                    .map(|refresh| refresh.parse().map_err(|_| "error parsing refresh rate"));
+                let refresh = match refresh {
+                    Some(refresh) => Some(refresh?),
+                    None => None,
+                };
+
+                Ok(ConfiguredMode {
```
YaLTeR: So what about checking for refresh = 0/missing with custom=true etc

src/protocols/output_management.rs
```diff
@@ -114,15 +114,41 @@ impl OutputManagementManagerState {
                 let modes_changed = old.modes != conf.modes;
```
YaLTeR, line 115: Need to update comment above this line saying that custom modes add/remove a mode on the TTY

src/protocols/output_management.rs
```diff
@@ -114,15 +114,41 @@ impl OutputManagementManagerState {
...
+                                        head.mode(&new_mode);
+                                        new_mode
+                                            .size(i32::from(mode.width), i32::from(mode.height));
+                                        if let Ok(refresh_rate) = mode.refresh_rate.try_into() {
+                                            new_mode.refresh(refresh_rate)
+                                        }
+                                    }
+                                } else if modes.len() > least_modes_len {
+                                    // One or more modes were removed
+                                    for mode in modes[least_modes_len..].iter() {
```
YaLTeR: Does this need to be a `.drain()` and similarly does the code above need to push the new mode to `modes`?

niri-ipc/src/lib.rs
ToxicMushroom: Is it okay to require > 0 inside the `FromStr for ConfiguredMode` impl ?

YaLTeR: Let's not (unless custom=true), that would be a breaking change. I believe 0 currently means pick any refresh rate

YaLTeR:
Well, let's do this. Thanks!

I tested dynamic mode addition/removal with wdisplays, seems to more or less work.

PR #2482, opened 2025-09-30
Config includes

YaLTeR:
Based on #2449 (will merge together), supersedes #2340.

Adds support for including other config files. Included files have the same structure as the main config file. Settings from included files will be merged with the settings from the main config file.

```kdl
// Some settings...

include "colors.kdl"

// Some more settings...
```

See the wiki page in the last commit for more details.

This PR is ready to go, but needs testing. There are a lot of config refactors. Existing configs should be unchanged, but testing is needed to verify this.

### DIFF
--- a/docs/wiki/Configuration:-Include.md
+++ b/docs/wiki/Configuration:-Include.md
@@ -241,3 +241,30 @@ include "struts.kdl"
 // Struts are not merged.
 // End result is only left and right struts.
 ```
+
+### Border special case
+
+There's one special case that differs between the main config and included configs.
+
+Writing `layout { border {} }` in an included config does nothing (since no properties are changed).
+However, writing the same in the main config will *enable* the border, i.e. it's equivalent to `layout { border { on; } }`.
+
+So, if you want to move your layout configuration from the main config to a separate file, remember to add `on` to the border section, for example:
+
+```kdl
+// separate.kdl
+layout {
+    border {
+        // Add this line:
+        on
+
+        width 4
+        active-color "#ffc87f"
+        inactive-color "#505050"
+    }
+}
+```
+
+The reason for this special case is that this is how it historically worked: back when I added borders, we didn't have any `on` flags, so I made writing the `border {}` section enable the border, with an explicit `off` to disable it.
+It wouldn't be too problematic to change it, however the default config always had a pre-filled `layout { border { off; } }` section with a note saying that commenting out the `off` is enough to enable the border.
+Many people likely have this part of the default config embedded in their configs now, so changing how it works would just cause a lot of confusion.


stefonarch:
Will something like `include "src/*.kdl` work?

Edit: answer is no ;)

But migration from niri-companion was successful. Path is relative to where the config is, fortunately.

immagine

YaLTeR:
> Will something like `include "src/*.kdl` work?
> 
> Edit: answer is no ;)

Yeah no, I'm not getting into globs. Not ready to dive into the file watching abyss.

> But migration from niri-companion was successful. Path is relative to where the config is, fortunately.

Yeah, though absolute paths will also work.

nickjj:
I haven't read all of the source code in this PR but how do you feel about supporting optional file paths with silent failures?

This allows you to define `include "src/custom.kdl.local"` and if the file exists, great use it, else ignore this statement entirely.

The benefit is you can post your dotfiles with that include statement and if you happen to add a potentially git ignored `src/custom.kdl.local` file, it will get used. It unlocks a very flexible system for customizations.

Both Ghostty and Hyprland support this idea. For example with Ghostty if the path starts with `?` it's considered optional. If you omit the `?` then it throws an error if the path isn't found.

Maybe a phase one implementation could be to make it optional by default so you don't have to think about extra parsing rules on the path?

YaLTeR:
I mean, should be simple enough, but is this the right place to do it? You can also just not commit some local changes. Or you can have an empty `config.kdl.local` that you gitignore and then change. Many tools won't have optional includes either way. Why add complexity to every tool when it's a concern outside the tool?

nickjj:
For folks who have dotfiles shared on either multiple devices or have them public with other people using them, the optional feature is handy because it reduces complexity for everyone involved at the cost of internal complexity for the tool.

You are right, you can touch an empty file as part of initial set up instructions for the dotfiles so it's not a deal breaker. I do this with other tools that don't support optional includes.

artrixdotdev:
> > Will something like `include "src/*.kdl` work?
> > Edit: answer is no ;)
> 
> Yeah no, I'm not getting into globs. Not ready to dive into the file watching abyss.


Should be relatively easy to implement with something like https://crates.io/crates/globset

stefonarch:
A nice thing is that  now `.desktop` files can be created to open specific settings by runners or menus:

immagine

YaLTeR:
Oh yeah that's a neat idea

YaLTeR:
Thanks everyone for testing!

johnrichardrinehart:
@nickjj based on the above interaction and #3012 I've created https://github.com/YaLTeR/niri/pull/3022. If you try that and it does(n't) work then please let me know in the PR.

PR #2499, opened 2025-10-01
Add option for mouse movement toggling DPMS

wu4:
I've got a very sensitive mouse, and I typically come back to find my monitor awake up any time anything happens in the house. I've been using a temporary solution of just flipping the mouse.  I thought to myself, surely other people would like to have it as an option, so I decided to draft a PR - `input.mouse.wake-monitor-on-movement` is required to be set in order for mouse movement to wake the monitor. I have it marked as a draft for these reasons:

- I'm not sure if `input.mouse` is the proper home for this option
- I haven't added anything to the docs yet because of the aforementioned
- I moved `should_activate_monitors` into `State` in order to be able to read the config
- I also added a small snippet to capture inputs that wake the monitor. I found it a bit annoying that a space press would pass through to whatever the currently-focused window was. Maybe this could be an option too?

### DIFF
--- a/niri-config/src/input.rs
+++ b/niri-config/src/input.rs
@@ -183,6 +183,8 @@ pub struct Mouse {
     pub middle_emulation: bool,
     #[knuffel(child)]
     pub scroll_factor: Option<ScrollFactor>,
+    #[knuffel(child)]
+    pub wake_monitor_on_movement: bool,
 }
 
 #[derive(knuffel::Decode, Debug, Default, PartialEq)]
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -100,7 +100,7 @@ impl State {
             }
         } else {
             // Power on monitors if they were off.
-            if should_activate_monitors(&event) {
+            if self.should_activate_monitors(&event) {
                 self.niri.activate_monitors(&mut self.backend);
 
                 // Notify the idle-notifier of activity only if we're also powering on the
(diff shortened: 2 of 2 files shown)


zaporter:
Thanks for doing this! I spent a long time trying to figure out how to disable this behavior in niri and it was only this post that finally convinced me that it was not possible.

In the meantime, I've swapped to this script (partially-ai-authored):

```sh
#!/usr/bin/env bash
# Display sleep script for niri that only wakes on keyboard input, not mouse movement.
# Grabs ALL input devices, waits for keyboard input, then releases and wakes.
#
# Requires evtest and that the user is in the "input" group

set -euo pipefail

GRAB_PIDS=()
FIFO="/tmp/display-sleep-$$"

cleanup() {
    for pid in "${GRAB_PIDS[@]}"; do
        kill "$pid" 2>/dev/null || true
    done
    rm -f "$FIFO"
    niri msg action power-on-monitors 2>/dev/null || true
}

trap cleanup EXIT INT TERM

mkfifo "$FIFO"

# Grab ALL input devices
for dev in /dev/input/event*; do
    props=$(udevadm info --query=property "$dev" 2>/dev/null || true)

    # Skip joysticks and switches
    if echo "$props" | grep -qE 'ID_INPUT_JOYSTICK=1|ID_INPUT_SWITCH=1'; then
        continue
    fi

    if echo "$props" | grep -q 'ID_INPUT=1'; then
        # For keyboards, pipe output to FIFO so we can detect keypresses
        if echo "$props" | grep -q 'ID_INPUT_KEYBOARD=1'; then
            evtest --grab "$dev" 2>/dev/null >"$FIFO" &
        else
            evtest --grab "$dev" >/dev/null 2>&1 &
        fi
        GRAB_PIDS+=($!)
    fi
done

if [ ${#GRAB_PIDS[@]} -eq 0 ]; then
    echo "No input devices found" >&2
    exit 1
fi

sleep 0.1
niri msg action power-off-monitors

# Wait for a key press (value 1) from the FIFO
grep -m 1 'EV_KEY.*value 1' <"$FIFO" >/dev/null

```

(Sorry if this is off-topic. Normally I would post in the associated issue but I cannot find one for this PR)

PR #2508, opened 2025-10-03
feat(config): include adds if-exists option

lonerOrz:
### PR Title

`feat(config): support optional include-if-exists directive`

---

### PR Description

Hi 👋,

This PR introduces a new config directive **`include-if-exists`**.
It behaves the same as the existing `include` directive, except that it does **not** fail if the target file is missing.

This makes configuration more flexible, for example:

* Local development overrides
* Optional themes or plugin configs

I’ve also added unit tests to:

* Verify the new `include-if-exists` behavior
* Ensure the existing `include` directive remains unchanged

The change is quite small, so if it doesn’t feel like a good fit, no worries at all 😊

Thanks!

### DIFF
--- a/docs/mkdocs.yaml
+++ b/docs/mkdocs.yaml
@@ -84,6 +84,7 @@ nav:
     - Nvidia: Nvidia.md
     - Xwayland: Xwayland.md
     - Gestures: Gestures.md
+    - Fullscreen and Maximize: Fullscreen-and-Maximize.md
     - Packaging niri: Packaging-niri.md
     - Integrating niri: Integrating-niri.md
     - Accessibility: Accessibility.md
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -44,6 +44,7 @@ window-rule {
     open-on-output "Some Company CoolMonitor 1234"
     open-on-workspace "chat"
     open-maximized true
+    open-maximized-to-edges true
     open-fullscreen true
     open-floating true
     open-focused false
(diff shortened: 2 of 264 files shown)


HigherOrderLogic:
Someone asked for similar thing in the original PR, and seems like [Yalter doesnt like the idea](https://github.com/YaLTeR/niri/pull/2482#issuecomment-3352291860).

lonerOrz:
> [Yalter doesnt like the idea](https://github.com/YaLTeR/niri/pull/2482#issuecomment-3352291860)

Thanks for raising that concern!

One important detail is that I did not change the existing `include` behavior — it remains strict and will fail if the file is missing, which is useful to ensure configs are complete.

The new directive `include-if-exists` is opt-in and makes the intent explicit. This way:

* Users who want strictness can keep using `include`
* Users who want flexibility can use `include-if-exists`

Compared to workarounds (like maintaining empty, gitignored files), this approach keeps the tool simple while improving developer experience. It doesn’t add hidden complexity — just another clear option.

HigherOrderLogic:
> One important detail is that I did not change the existing `include` behavior — it remains strict and will fail if the file is missing, which is useful to ensure configs are complete.

The original didnt change the functionality of the directive either.

It was "adding `?` to make the `include` directive fails silently if the file doesnt exist", which is what this PR is trying to achieve.

lonerOrz:
Thanks for clarifying!

For me, the reason I proposed this is because it makes my configuration a bit easier to manage in local/optional cases. But I totally understand that this is a subjective design choice and not every project would want to add it.

If you feel it doesn’t fit well with the project direction, please feel free to close this PR — no worries at all. I just thought it might be useful to share the idea. Thanks for taking the time to review it!

YaLTeR:
Yeah, so far I'm unconvinced. (And if I were convinced, it should be `include optional=true "..."`)

PR #2521, opened 2025-10-05
backend/tty: use pci device id to track ignored drm devices

athul-krishna-kr:
On my device (Asus GA402RJ) the drm primary node number of dgpu changes from card1 to card0 after unbind and rebind
of amdgpu, and this causes niri to panic ([niri_panic.log](https://github.com/user-attachments/files/22709607/niri_panic.log)).  

`udev_dispatcher.as_source_ref()` will panic when called from associated dispatching closure, i.e, while handling udev event. Because in this case, the initial ignored drm devices were:
> /dev/dri/card1 and /dev/dri/render128

after rebind it became:
> /dev/dri/card0 and /dev/dri/render128

Fix this issue by using pci device id of corresponding drm device to track ignored devices.
To allow for hot add/removal of gpus using udevadm trigger (`udevadm trigger --action=remove /dev/dri/card0`),
skip render nodes on `device_added()`

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -537,7 +537,7 @@ impl Tty {
         path: &Path,
         niri: &mut Niri,
     ) -> anyhow::Result<()> {
-        debug!("device added: {device_id} {path:?}");
+        debug!("adding device: {device_id} {path:?}");
 
         let node = DrmNode::from_dev_id(device_id)?;
 
@@ -687,8 +687,22 @@ impl Tty {
             return;
         };
 
+        if self.ignored_nodes.contains(&node) {
+            debug!("node is ignored, skipping");
+            return;
+        }
+
         let Some(device) = self.devices.get_mut(&node) else {
-            warn!("no such device");
+            if let Some(path) = node.dev_path() {
+                warn!("unknown device; trying to add");
+
+                if let Err(err) = self.device_added(device_id, &path, niri) {
+                    warn!("error adding device: {err:?}");
+                }
+            } else {
+                warn!("unknown device");
+            }
+
             return;
         };
 
@@ -784,7 +798,7 @@ impl Tty {
     }
 
     fn device_removed(&mut self, device_id: dev_t, niri: &mut Niri) {
-        debug!("device removed: {device_id}");
+        debug!("removing device: {device_id}");
 
         let Ok(node) = DrmNode::from_dev_id(device_id) else {
             warn!("error creating DrmNode");
@@ -792,7 +806,7 @@ impl Tty {
         };
 
         let Some(device) = self.devices.get_mut(&node) else {
-            warn!("no such device");
+            warn!("unknown device");
             return;
         };
 
--- a/docs/mkdocs.yaml
+++ b/docs/mkdocs.yaml
@@ -84,6 +84,7 @@ nav:
     - Nvidia: Nvidia.md
     - Xwayland: Xwayland.md
     - Gestures: Gestures.md
+    - Fullscreen and Maximize: Fullscreen-and-Maximize.md
     - Packaging niri: Packaging-niri.md
     - Integrating niri: Integrating-niri.md
     - Accessibility: Accessibility.md
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -44,6 +44,7 @@ window-rule {
     open-on-output "Some Company CoolMonitor 1234"
     open-on-workspace "chat"
     open-maximized true
+    open-maximized-to-edges true
     open-fullscreen true
     open-floating true
     open-focused false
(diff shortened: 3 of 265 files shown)


athul-krishna-kr:
Hi @YaLTeR , could you please review this PR?

YaLTeR:
I will review it when I will get to it

src/backend/tty.rs
```diff
@@ -512,9 +541,18 @@ impl Tty {
...
 
-        if self.ignored_nodes.contains(&node) {
-            debug!("node is ignored, skipping");
-            return Ok(());
+        match node.ty() {
+            NodeType::Render => {
+                return Ok(());
+            }
+            _ => {}
+        }
```
YaLTeR, line 549: Why is this necessary? It hasn't seemed to cause issues before. What systems does this affect and how?

src/backend/tty.rs
```diff
@@ -257,6 +257,33 @@ struct ConnectorProperties<'a> {
...
+
+impl PciID {
+    fn from_drm_node(node: &DrmNode) -> Option<PciID> {
+        use std::fs;
+
+        let device_id = fs::read_to_string(format!(
+            "/sys/dev/char/{}:{}/device/device",
+            DrmNode::major(node),
+            DrmNode::minor(node)
+        ));
```
YaLTeR, line 273: Will this always work properly? Is it supported on FreeBSD? Will this work for kmsro devices (scanout-only and render-only nodes)? Is there perhaps a helper in Smithay for this?

src/backend/tty.rs
```diff
@@ -257,6 +257,33 @@ struct ConnectorProperties<'a> {
...
+
+        match device_id {
+            Ok(device_id) => Some(PciID { device_id }),
+            Err(err) => {
+                error!(
```
YaLTeR, line 278: error! is for programming errors only: https://yalter.github.io/niri/Development%3A-Developing-niri.html#logging-levels

src/backend/tty.rs
```diff
@@ -1870,7 +1908,14 @@ impl Tty {
...
                 .filter(|node| {
-                    self.ignored_nodes.contains(node) || !device_list.contains_key(&node.dev_id())
+                    if let Some(device_id) = PciID::from_drm_node(&node) {
+                        if self.ignored_nodes.contains(&device_id) {
+                            return true;
+                        } else {
+                            return !device_list.contains_key(&node.dev_id());
+                        }
+                    }
+                    return !device_list.contains_key(&node.dev_id());
```
YaLTeR, line 1918: Kinda awkward writing this twice

src/backend/tty.rs
```diff
@@ -257,6 +257,33 @@ struct ConnectorProperties<'a> {
     connector: connector::Handle,
     properties: Vec<(property::Info, property::RawValue)>,
 }
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
YaLTeR, line 260: Formatting

```suggestion

#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```

src/backend/tty.rs
athul-krishna-kr: After looking through https://github.com/Smithay/drm-rs/blob/develop/src/node/mod.rs, this method is not supported on FreeBSD. I honestly have no idea whether this works for kmsro devices.
`cat /sys/dev/char/226:2/device/device` => primary node
`cat /sys/dev/char/226:129/device/device` => render node
always returns pci device id in linux, as long as they are in drm subsystem.

I couldn't find any Smithay helper functions to retrieve device ids for linux and FreeBSD. Plus is it appropriate to have platform specific code in niri, rather than Smithay?

With all things considered, I believe this pretty much invalidates this approach.

src/backend/tty.rs
athul-krishna-kr: This enables gpu hot-plugging through udev trigger since closing `device_fd` was merged in #2396. 
`sudo udevadm trigger --action=remove /dev/dri/card0`

My reference was how drm add event handled in wlroots 
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/session/session.c#L127

`ignore-drm-device` seems to have an issue when toggling it off after unbind & rebind of gpu ([niri_debug.log](https://github.com/user-attachments/files/22983394/niri_debug.log))
This also fixes it ([niri_skip_render.log](https://github.com/user-attachments/files/22983415/niri_skip_render.log))

I think this PR is invalid, shall I open another with this specific change?

YaLTeR: Add comment linking to https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/768fbaad54027f8dd027e7e015e8eeb93cb38c52

src/backend/tty.rs
YaLTeR: > Plus is it appropriate to have platform specific code in niri, rather than Smithay?

It's fine, though I'd rather limit it.

Is the main underlying problem that ignore-drm-device is a blocklist setting, rather than an allowlist in WLR, which prevents it from dealing with changing device names? What if you just ignore both card0 and card1 in the config, would that workaround for you?

athul-krishna-kr: > What if you just ignore both card0 and card1 in the config, would that workaround for you?

Nope, that did not work ([niri_panic.log](https://github.com/user-attachments/files/22984495/niri_panic.log))
Since card0 is not a valid node, it's not added to `ignored_nodes` list

YaLTeR: Uhh, panicking seems bad. That should be fixed regardless

YaLTeR: For the filtering, perhaps we could match by path in on_device_added rather than prepopulate node IDs? Like, check if the primary node path that we're given is in ignored paths; if not, create enough of it to get the render node path, then check that too

PR #2536, opened 2025-10-06
feat: expose TabletPadButton as a trigger for using commands

coleleavitt:


src/input/mod.rs
```diff
@@ -3386,6 +3387,36 @@ impl State {
...
+                mod_key,
+                Trigger::TabletPadButton(button_number),
+                mods,
+            )
+        };
+
+        if let Some(bind) = bind {
+            self.handle_bind(bind.clone());
+        } else {
+            debug!("Tablet pad button {} pressed (no binding)", button_number);
```
YaLTeR, line 3416: Shouldn't it get forwarded to the client normally?

PolyMeilex: The relevant Wayland code (`zwp_tablet_pad_v2`)  is not implemented in Smithay at this point.

PR #2559, opened 2025-10-09
add window-rule match is-maximized

evysgarden:
Adds support for matching a window on its maximized state.

Also exposes it in ipc.

With this it's possible to do the following for example:
```kdl
window-rule {
  // match on any window that is maximized
  match is-maximized=true

  // change the border
  border {
    active-color "#0000004c"
  }

  // .. or enable VRR
  variable-refresh-rate true
}
```

### DIFF
--- a/niri-config/src/window_rule.rs
+++ b/niri-config/src/window_rule.rs
@@ -83,6 +83,8 @@ pub struct Match {
     #[knuffel(property)]
     pub is_focused: Option<bool>,
     #[knuffel(property)]
+    pub is_maximized: Option<bool>,
+    #[knuffel(property)]
     pub is_active_in_column: Option<bool>,
     #[knuffel(property)]
     pub is_floating: Option<bool>,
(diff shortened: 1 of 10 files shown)


evysgarden:
my personal use for this was making the border color match my terminal opacity when maximized so there's no seem. I don't use gaps or struts so it just looks nicer.

Alternatively I might use it to emulate "smart borders" from sway.

YaLTeR:
Perhaps what you want instead is #2376?

evysgarden:
ah that would also be nice but it doesn't add an ability to match on full_width columns.

i could rename the match rule to make it more clear that it applies to all windows that are part of a full_width column.

maybe `is-full-width` would be more future-proof to not clash with the mentioned pr.

evysgarden:
also kind of a pain i didn't see that pr sooner.

now I've put so much work into understanding niri's codebase "^^

evysgarden:
@YaLTeR do you think it's worth to work on this or should I close it?

YaLTeR:
idk how i feel about it

evysgarden:
i'm gonna close this since #2376 is merged and this feature doesn't seem to be needed

PR #2562, opened 2025-10-10
Add BMP clipboard format for screenshots (with aliases)

remme123:
- freerdp clipboard only accepts BMP; add BMP improves compatibility
- Align with apps like eog that publish many image MIME types

Implementation:
- Encode PNG and BMP (32bpp BGRX, bottom-up) for screenshots
- Advertise: image/png, image/bmp, image/x-bmp, image/x-MS-bmp, image/x-ms-bmp, image/x-win-bitmap
- Send bytes by requested MIME
- Keep on-disk save as PNG (unchanged)

Compatibility:
- Fix negative height by using bottom-up BMP
- Use BGRX to improve support in older clients (e.g., FreeRDP)

Mime types set by niri now:
```bash
[hello@fc ~]$ wl-paste -l
image/png
image/bmp
image/x-bmp
image/x-MS-bmp
image/x-ms-bmp
image/x-win-bitmap
```

Mime types set by eog:
```bash
[hello@fc ~]3$ wl-paste --list-types
image/png
image/jpeg
image/avif
image/bmp
image/x-bmp
image/x-MS-bmp
image/x-icon
image/x-ico
image/x-win-bitmap
image/vnd.microsoft.icon
application/ico
image/ico
image/icon
text/ico
image/jxl
image/tiff
image/webp
audio/x-riff
text/plain;charset=utf-8
UTF8_STRING
COMPOUND_TEXT
TEXT
text/plain
STRING
text/plain;charset=utf-8
text/plain
text/uri-list
application/vnd.portal.filetransfer
application/vnd.portal.files
```

The code was written by Cursor. I’ve reviewed all the changes and didn’t find any issues.

I’ve already compiled, run, and tested it — now I can successfully paste Niri screenshots in FreeRDP.

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -293,7 +293,7 @@ impl ClipboardImageData {
     fn select_bytes_for_mime(&self, mime_type: &str) -> Arc<[u8]> {
         match mime_type {
             // Common BMP MIME types seen in the wild
-            "image/bmp" | "image/x-bmp" | "image/x-MS-bmp" | "image/x-win-bitmap" => {
+            "image/bmp" | "image/x-bmp" | "image/x-MS-bmp" | "image/x-ms-bmp" | "image/x-win-bitmap" => {
                 self.bmp.clone()
             }
             // Default to PNG for anything else we advertised
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -282,8 +282,28 @@ delegate_input_method_manager!(State);
 delegate_keyboard_shortcuts_inhibit!(State);
 delegate_virtual_keyboard_manager!(State);
 
+/// Clipboard payload supporting multiple MIME encodings for the same image.
+#[derive(Clone)]
+pub struct ClipboardImageData {
+    pub png: Arc<[u8]>,
+    pub bmp: Arc<[u8]>,
+}
+
+impl ClipboardImageData {
+    fn select_bytes_for_mime(&self, mime_type: &str) -> Arc<[u8]> {
+        match mime_type {
+            // Common BMP MIME types seen in the wild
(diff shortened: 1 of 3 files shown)


YaLTeR:
Is this really needed? GNOME Shell only advertises PNG. I think maybe FreeRDP should update instead?

remme123:
To maximize compatibility, and given that supporting BMP comes at very little cost, I think the benefits outweigh the overhead. The RDP protocol defines clipboard transfers to support only BMP format — no PNG or other formats — and Wayland doesn’t forbid using BMP for the clipboard. The remaining question is whether the memory overhead of storing BMP image data is acceptable. I think this should be left as a user-configurable option.

YaLTeR:
If GNOME Shell starts serving BMP we can also do that

AkechiShiro:
@YaLTeR Is there any reason why Niri is closely following what is done by GNOME Shell?

I mean you could also follow what KDE Plasma does as they are also quite widespread nowadays. 

And sure FreeRDP could also add a conversion step or attempt at supporting more formats, but I don't know if adding this step is easy into the current FreeRDP client code.

I believe it's better for users if Niri supports this feature because :  

- It will help improve the UX for new users who use daily RDP for various use cases and the behavior is similar to other mainstream DE.

If I would start using Niri tomorrow for instance, I would probably cherry pick this PR on top of the latest release as a patch.

This feature could also be behind a config flag if you believe there isn't any benefits for lots of users.

YaLTeR:
> Is there any reason why Niri is closely following what is done by GNOME Shell?

I think GNOME Shell has good defaults, plus it's the most widely used Linux DE, confirming that those defaults work.

> If I would start using Niri tomorrow for instance, I would probably cherry pick this PR on top of the latest release as a patch.

Sure, do that, that's like one of the main benefits of free software that you can do it.

PR #2564, opened 2025-10-10
feat: allow specifying `custom-shader` by path

HigherOrderLogic:
Allow user to specifying custom shader by path. The path is also tracked for hot reloading.

~~TODO: add docs.~~

### DIFF
--- a/docs/wiki/Configuration:-Animations.md
+++ b/docs/wiki/Configuration:-Animations.md
@@ -222,6 +222,16 @@ animations {
 }
 ```
 
+You can also refer to a shader file by specifying its path. <sup>Since: next release</sup>
+
+```kdl,must-fail
+animations {
+    window-open {
+        custom-shader path="./my-shader.glsl"
+    }
+}
+```
+
 #### `window-close`
 
 <sup>Since: 0.1.5</sup>
--- a/niri-config/src/animations.rs
+++ b/niri-config/src/animations.rs
@@ -1,8 +1,15 @@
+use std::cell::RefCell;
+use std::env::home_dir;
+use std::fs;
+use std::path::PathBuf;
+use std::rc::Rc;
+
+use knuffel::ast::SpannedNode;
 use knuffel::errors::DecodeError;
 use knuffel::Decode as _;
 
 use crate::utils::{expect_only_children, parse_arg_node, MergeWith};
-use crate::FloatOrInt;
+use crate::{BasePath, FloatOrInt, Includes};
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct Animations {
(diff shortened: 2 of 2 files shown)


docs/wiki/Configuration:-Animations.md
```diff
@@ -222,6 +222,16 @@ animations {
 }
 ```
 
+You can also refer to a shader file by specifying its path. <sup>Since: next release</sup>
```
YaLTeR, line 225: Since goes at the start of the paragraph

niri-config/src/animations.rs
```diff
@@ -151,6 +158,71 @@ impl Default for WorkspaceSwitchAnim {
...
+                            shader_path
+                        ),
+                    ));
+                    continue;
+                };
+                home.join(rest)
+            } else {
+                let base = ctx.get::<BasePath>().unwrap();
+                base.0.join(shader_path)
+            };
```
YaLTeR, line 186: Let's extract this into a function shared with include parsing

niri-config/src/animations.rs
```diff
@@ -151,6 +158,71 @@ impl Default for WorkspaceSwitchAnim {
...
+
+            let includes = ctx.get::<Rc<RefCell<Includes>>>().unwrap();
+            includes.borrow_mut().0.push(path.to_path_buf());
```
YaLTeR, line 189: This makes me want to rename this "includes" to something else everywhere which makes it clear it's not talking only about "include" config statements. But it should be done separately as it's a large change, and I haven't come up with the term to use instead yet.

At the very least could you mention in a comment on relevant fields that they encompass all included files and not just included config parts?

niri-config/src/animations.rs
```diff
@@ -383,7 +455,11 @@ where
         let mut custom_shader = None;
         let anim = Animation::decode_node(node, ctx, default, |child, ctx| {
             if &**child.node_name == "custom-shader" {
-                custom_shader = parse_arg_node("custom-shader", child, ctx)?;
+                custom_shader = if child.properties.is_empty() {
```
YaLTeR, line 458: I'd change these to `child.properties.contains_key("path")`

docs/wiki/Configuration:-Animations.md
YaLTeR, line 225: Move this part into a `###` section at the very end of the page so that it doesn't look like it only applies to window-open

YaLTeR (changes requested):

niri-config/src/lib.rs
```diff
@@ -348,20 +346,12 @@ where
...
-                    } else {
-                        // Otherwise, use the current include base dir
-                        let base = ctx.get::<BasePath>().unwrap();
-                        base.0.join(path)
+                    let Some(path) = expand_home_path(
+                        knuffel::traits::DecodeScalar::decode(path_val, ctx)?,
+                        node,
+                        ctx,
+                    ) else {
+                        continue;
```
YaLTeR, line 352: This commit, together with the expand_home_path() function, should come first. First you do the necessary refactors, then you add new code already using the simpler versions.

niri-config/src/lib.rs
```diff
@@ -348,20 +346,12 @@ where
...
-                            continue;
-                        };
-
-                        home.join(rest)
-                    } else {
-                        // Otherwise, use the current include base dir
-                        let base = ctx.get::<BasePath>().unwrap();
-                        base.0.join(path)
+                    let Some(path) = expand_home_path(
+                        knuffel::traits::DecodeScalar::decode(path_val, ctx)?,
```
YaLTeR: import DecodeScalar

niri-config/src/lib.rs
```diff
@@ -117,8 +117,12 @@ pub enum ConfigPath {
 struct BasePath(PathBuf);
 struct RootBase(PathBuf);
 struct Recursion(u8);
+
+// TODO: This is currently used to track all the files that needed to be watched. The name
```
YaLTeR: No TODOs in final commits, TODO is for work in progress so you don't forget, FIXME is for long term todos

YaLTeR: And add a similar comment to ConfigParseResult includes field

docs/wiki/Configuration:-Animations.md
```diff
@@ -440,6 +440,20 @@ animations {
 }
 ```
 
+### Specifying Custom Shader by Path
+
+
+For animations that accept custom shader, you can also refer to the file containing the shader by specifying its path.
```
YaLTeR:
```suggestion
### Specifying `custom-shader` by Path

<sup>Since: next release</sup>

For animations that accept a `custom-shader`, you can also attach the shader by path, rather than writing it out inline.
```

niri-config/src/animations.rs
```diff
@@ -151,6 +156,57 @@ impl Default for WorkspaceSwitchAnim {
...
 }
 
+fn parse_custom_shader_path<S: knuffel::traits::ErrorSpan>(
+    spanned: &SpannedNode<S>,
+    ctx: &mut knuffel::decode::Context<S>,
+) -> Result<String, DecodeError<S>> {
+    let mut shader_text = None;
+
+    for (name, val) in &spanned.properties {
+        if **name == "path".into() {
```
YaLTeR: If you take `&***name` you should be able to compare without `.into()` on the string

niri-config/src/animations.rs
```diff
@@ -151,6 +156,57 @@ impl Default for WorkspaceSwitchAnim {
...
+            let path_val = knuffel::traits::DecodeScalar::decode(val, ctx)?;
```
YaLTeR, line 167: One last thing, I think this should check for `shader_text.is_some()` for duplicate path. We don't do it everywhere but the derive macro does it at least, and would prevent watching unused paths in this case.

```
                    if shader_text.is_some() {
                        ctx.emit_error(DecodeError::unexpected(
                            name,
                            "property",
                            "unexpected duplicate property `path`",
                        ))
                    }
```

YaLTeR: Upon testing I discovered that I was wrong. `properties` is a map from Spanned, and Hash for Spanned only hashes the value (property name), so there can be no duplicates in this loop. Apparently, knuffel doesn't check for duplicates at all. So I removed this check.

YaLTeR:
Thanks

PR #2565, opened 2025-10-10
feat: event-stream event for when a screenshot is taken

ThatOneCalculator:
Useful for things like desktop shells that might want to show UI after a screenshot is taken :)

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1487,7 +1487,8 @@ pub enum Event {
     ScreenshotCaptured {
         /// The file path where the screenshot was saved, if it was written to disk.
         ///
-        /// If `None`, the screenshot was only copied to the clipboard.
+        /// If `None`, the screenshot was either only copied to the clipboard, or the path couldn't
+        /// be converted to a `String` (e.g. contained invalid UTF-8 bytes).
         path: Option<String>,
     },
 }
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -5696,11 +5696,11 @@ impl Niri {
             }
 
             #[cfg(feature = "dbus")]
-            if let Err(err) = crate::utils::show_screenshot_notification(image_path.as_ref()) {
+            if let Err(err) = crate::utils::show_screenshot_notification(image_path.as_deref()) {
                 warn!("error showing screenshot notification: {err:?}");
             }
 
-            // Send screenshot completion event
+            // Send screenshot completion event.
             let path_string = image_path
                 .as_ref()
                 .and_then(|p| p.to_str())
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -483,12 +483,12 @@ pub fn handle_msg(mut msg: Msg, json: bool) -> anyhow::Result<()> {
                     }
                     Event::ScreenshotCaptured { path } => {
                         let mut parts = vec![];
+                        parts.push("copied to clipboard".to_string());
                         if let Some(path) = &path {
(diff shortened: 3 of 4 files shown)


ThatOneCalculator:
Requesting review from @YaLTeR

src/niri.rs
```diff
@@ -5646,9 +5656,16 @@ impl Niri {
...
             #[cfg(feature = "dbus")]
-            if let Err(err) = crate::utils::show_screenshot_notification(image_path) {
+            if let Err(err) = crate::utils::show_screenshot_notification(image_path.clone()) {
                 warn!("error showing screenshot notification: {err:?}");
             }
+
+            // Send screenshot completion event
+            let path_string = image_path
+                .as_ref()
+                .map(|p| p.to_string_lossy().into_owned());
```
YaLTeR: Can we send a PathBuf over the IPC? Does that even roundtrip through JSON? If not, then I think sending no path is better than a lossy path. Then the comment on the IPC event should also say that `None` means that the path is not available (either not saved to disk, or couldn't be encoded as a string).

src/niri.rs
```diff
@@ -5646,9 +5656,16 @@ impl Niri {
             }
 
             #[cfg(feature = "dbus")]
-            if let Err(err) = crate::utils::show_screenshot_notification(image_path) {
+            if let Err(err) = crate::utils::show_screenshot_notification(image_path.clone()) {
```
YaLTeR: Maybe this can borrow?

src/niri.rs
```diff
@@ -5646,9 +5656,16 @@ impl Niri {
...
             }
+
+            // Send screenshot completion event
+            let path_string = image_path
+                .as_ref()
+                .map(|p| p.to_string_lossy().into_owned());
+            let _ = event_tx.send(path_string);
+
             #[cfg(not(feature = "dbus"))]
             drop(image_path);
```
YaLTeR: This would no longer be needed

niri-ipc/src/lib.rs
```diff
@@ -1455,6 +1447,13 @@ pub enum Event {
         /// For example, the config file couldn't be parsed.
         failed: bool,
     },
+    /// A screenshot was taken.
+    ScreenshotTaken {
```
YaLTeR: Bikeshedding: maybe ScreenshotCaptured is better?

ThatOneCalculator:
> But they already show the notification?

Yes, but that notification might not be enough. For example, in my QuickShell setup, I want to add an overlay that lets me click the image to open it in an editor like Satty, and also to look more like the macOS screenshot UI where it pops into the bottom right.

src/niri.rs
ThatOneCalculator: > Can we send a PathBuf over the IPC? Does that even roundtrip through JSON?

In my testing it just gets serialized as a string

> If not, then I think sending no path is better than a lossy path. Then the comment on the IPC event should also say that `None` means that the path is not available (either not saved to disk, or couldn't be encoded as a string).

How does this look?

```rs
let path_string = image_path
    .as_ref()
    .and_then(|p| p.to_str().map(|s| s.to_owned()));
```

YaLTeR: > In my testing it just gets serialized as a string

The problems may appear when there's an invalid utf-8 byte in the path. That might actually fail the serialization entirely, which would be bad, so yeah.

> How does this look?

Probably ok this way yeah. Can move the .map() outside too.

YaLTeR:
Thanks

PR #2572, opened 2025-10-11
Add ToggleForceShortcutsInhibit action

TheNeikos:
I use `Alt` as my Mod key, and some applications force you to use `Alt` for actions. With this feature, I can now disable global shortcuts and use them normally. 

This is mainly intended for long-running usage, eg. games or editing software.

I tried to follow the conventions in the repo and mainly reflected the 'toggle-inhibitors' action here. I'm open ofc to any change requests etc

### DIFF
--- a/docs/wiki/Configuration:-Key-Bindings.md
+++ b/docs/wiki/Configuration:-Key-Bindings.md
@@ -405,3 +405,18 @@ binds {
     Super+Alt+L allow-inhibiting=false { spawn "swaylock"; }
 }
 ```
+
+#### `toggle-force-shortcuts-inhibit`
+
+<sup>Since: next release</sup>
+
+Some applications require specific key combinations that might interfere with your choice of `Mod` key. For example if you use `Alt`.
+This action will stop _all_ shortcuts and instead forward them to the focused surface.
+
+Note: This shortcut can not be inhibited as otherwise it means that you could not toggle it back off.
+
+```kdl
+binds {
+    Mod+Escape { toggle-force-shortcuts-inhibit; }
+}
+```
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -129,6 +129,7 @@ pub enum Action {
         id: u64,
         write_to_disk: bool,
     },
+    ToggleForceShortcutsInhibit,
     ToggleKeyboardShortcutsInhibit,
     CloseWindow,
     #[knuffel(skip)]
@@ -374,6 +375,9 @@ impl From<niri_ipc::Action> for Action {
                 id: Some(id),
                 write_to_disk,
             } => Self::ScreenshotWindowById { id, write_to_disk },
+            niri_ipc::Action::ToggleForceShortcutsInhibit {} => {
+                Self::ToggleForceShortcutsInhibit
+            }
             niri_ipc::Action::ToggleKeyboardShortcutsInhibit {} => {
                 Self::ToggleKeyboardShortcutsInhibit
             }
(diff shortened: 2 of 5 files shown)


YaLTeR:
Maybe instead it would be better to extend the existing toggle keyboard inhibit action to also work on windows without their own keyboard inhibitor?

TheNeikos:
Sure! How would that work? There's 4 different states one could be in then:

| surface-inhibited | global-inhibited | result |
| ------- | ------ | ------ |
|         |        |   global   |
|         |    x   |   global-off   |
|    x    |        |    surface-off    |
|    x    |    x   |   surface-off, global-off     |

This seems to be sane?

----

I realized that I am also missing a feature that I would like to see, which is to also remove mouse interactions (aka no resize/drag and drop).

If we now decide to fold it all together into one action, I think that disabling mouse actions might be too far? Personally I consider it to be part of it, at least when globally disabled. But I could add it as a separate action, albeit that feels a bit weirder on its own.

YaLTeR:
> Sure! How would that work?

It would only enable the surface inhibit. And this won't disable the mouse interactions. 

Would this be insufficient for you?

The reason I'm heavily steering this towards surface inhibit is that it's more or less existing code and a natural extension of an existing action. Global inhibit is a novel thing, I'm not sure whether we want it yet, especially not sure if it should involve disabling mouse actions, etc etc... Especially considering that using Alt as the mod key is a bit of a self inflicted problem.

TheNeikos:
> Would this be insufficient for you?

In my situation yes. I also feel that you are reluctant of adding this, which is totally fine as well. It is definitely something that would add a cross-cutting concern across the input code, and should be added wisely.

The code as is right now, is more of a 'it works' proof of concept, and definitely needs to refined a bit more. Especially the mouse handling. But I did not have the time to dive further into the code.

I am not sure I am interested in the feature if its just inhibiting a single surface while mouse actions still work. Since for example alt+click in my situation is not passed through, removing crucial interactions from the apps I use.

I am fine with this just living in my fork if it goes against your vision of the project.

> Especially considering that using Alt as the mod key is a bit of a self inflicted problem.

Well, it is definitely a config option. But I mostly choose it because using the Super key gives me pretty bad hand cramps.

YaLTeR:
> Well, it is definitely a config option.

See, this is why it's so hard for me to approve adding new config options and stuff...

> But I mostly choose it because using the Super key gives me pretty bad hand cramps.

An alternative is something like switching super and alt with an xkb option.

SpyrosPsarras:
+1 for including the mouse interactions, please don't leave them out.

my use case is remote desktop: i run Moonlight fullscreen on a dedicated niri workspace, streaming my home PC which also runs niri. with `capturesyskeys` on + `toggle-keyboard-shortcuts-inhibit` the keyboard now goes fully to the host, which is great.

but the mouse is still trapped locally. Super is my mod key on both machines, so when i Super+left-drag to move a window *inside* the remote home PC, the laptop niri grabs it and moves the local Moonlight window instead. same with Super+right (resize) and Super+middle (scroll). it never reaches the guest.

for a remote desktop client that makes Super+drag basically unusable. so for me, disabling the mouse interactions while shortcuts are inhibited is not "too far" - it is the whole point. i want the inhibited surface to behave like the compositor isn't there at all: keyboard AND mouse both pass through.

so a big vote here for landing this with the mouse part included. thanks for working on it.

alk3pInjection:
upvote - this resolves a fairly common issue with vm, rdp, etc., since GlobalShortcuts portal won't always work as its still not adopted by some crappy software.

PR #2582, opened 2025-10-12
Add topic regarding managing external displays using Kanshi to FAQ

shartrec:
This simply adds some info to the wiki/FAQ about using Kanshi to manage external displays. It seems a topic that comes up often in discussions etc.

### DIFF
--- a/docs/wiki/FAQ.md
+++ b/docs/wiki/FAQ.md
@@ -124,3 +124,21 @@ binds {
     Super+Alt+L allow-when-locked=true { spawn "swaylock"; }
 }
 ```
+
+### How do I change output configuration based on connected monitors?
+
+If you require different output configurations depending on what outputs are connected then you can use [Kanshi](https://gitlab.freedesktop.org/emersion/kanshi).
+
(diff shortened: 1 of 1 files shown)


docs/wiki/FAQ.md
```diff
@@ -124,3 +124,9 @@ binds {
     Super+Alt+L allow-when-locked=true { spawn "swaylock"; }
 }
 ```
+
+### How do I manage external displays?
+
+The recommended way to manage connection and disconnection of external displays is to use [Kanshi](https://gitlab.freedesktop.org/emersion/kanshi).
```
YaLTeR: This is not really the question here, is it? Like, I don't use kanshi. Niri automatically connects external displays, and this behavior works for me. What does "manage" mean?

Should be something that talks about the actual problem that people have that they need kanshi for. Like, how do I change output configuration based on connected monitors? Or something like this. Not sure what the most common problems are.

shartrec: Yes. The real question is , as you say,
" how do I change output configuration based on connected monitors?

I'll update PR accordingly.

YaLTeR:
Thanks

PR #2584, opened 2025-10-12
Add IPC action to toggle animations

kevichi7:
- expose set-animations in the IPC API (defaulting to on)
  - update bindable action enum so configs can call it
  - toggle the runtime config, clock, and redraw when invoked

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -350,6 +350,10 @@ pub enum Action {
     UnsetWindowUrgent(u64),
     #[knuffel(skip)]
     LoadConfigFile,
+    #[knuffel(skip)]
+    SetAnimations {
+        off: bool,
+    },
 }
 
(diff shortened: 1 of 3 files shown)


YaLTeR:
Why?

kevichi7:
> Why?

I wanted to have a widget in my setup that can quickly disable all animations and transparency. Writing a script to regenerate and reload the config every time felt heavier than just exposing this as an IPC action, since IPC already covers other runtime toggles.

YaLTeR:
With includes you can now regenerate just a tiny config snippet, I think that's a better way.

> IPC already covers other runtime toggles

Not really, only a small subset of them

kevichi7:
> With includes you can now regenerate just a tiny config snippet, I think that's a better way.
> 
> > IPC already covers other runtime toggles
> 
> Not really, only a small subset of them

Got it. I'll try your approach instead and close the PR.

PR #2589, opened 2025-10-12
Remove uneeded window cast render

abmantis:
`render_windows_for_screen_cast` is already called on every redraw, so we don't need to call `dequeue_buffer_and_render` here.


YaLTeR:
The comment:

>                 // Use the cached output since it will be present even if the output was
>                // currently disconnected.

seems to point out the reason why redraw_cast() doesn't simply call queue_redraw for windows

PR #2598, opened 2025-10-14
Redraw cursor if hidden during screenshot

### LINKED ISSUE #2498: Cursor not visible in screenshots if cursor hide-when-typing is enabled
When taking a screenshot, that usually requires pressing a keyboard button. With the following config snippet:
```kdl
cursor {
    hide-when-typing
}
```

This will make the cursor immediately hide upon pressing the screenshot button, which makes the cursor invisible even when the cursor is enabled in the screenshot UI.

### System Information

* niri version: 25.08 (01be0e6)

* Distro: Gentoo Linux

* GPU: AMD Radeon 7800XT

* CPU: AMD Ryzen 7 7700X

ArijanJ:
Fix #2498

When using hide-when-typing or hide-after-inactive-ms, the cursor is completely gone from screenshots, and this seems to be the simplest fix for both cases

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -1826,6 +1826,12 @@ impl State {
             return;
         }
 
+        // Redraw the pointer if hidden through cursor{} options
+        if self.niri.pointer_visibility == PointerVisibility::Hidden {
+            self.niri.pointer_visibility = PointerVisibility::Visible;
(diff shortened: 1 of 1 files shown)


YaLTeR:
Makes sense, thanks!

PR #2609, opened 2025-10-16
feat: force render

busyoGG:
Allows the user to force a window to render even when it is offscreen or being captured by a screencast.

This behavior is controlled by the following window rule:

```
window-rule {
    // match rule
    force-render true    // not needed if the window is being cast
    force-render-fps 60 // if you need fps limit
}
```

https://github.com/user-attachments/assets/a9ed1da3-ed8c-4ef1-9e5c-a73ad2656a6b

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -573,7 +573,7 @@ struct SurfaceFrameThrottlingState {
     last_sent_at: RefCell<Option<(Output, u32)>>,
 }
 
-struct ForceRenderState {
+pub struct ForceRenderState {
     // To calculate time diff between current frame_callback_time and last render time.
     last_render_time: Duration,
     // To check if a force render frame is waiting to be rendered.
@@ -5090,8 +5090,8 @@ impl Niri {
 
                         if !force_render_state.is_waiting_for_render {
                             // Time diff since last frame callback.
-                            let time_diff: Duration =
-                                frame_callback_time.saturating_sub(force_render_state.last_render_time);
+                            let time_diff: Duration = frame_callback_time
+                                .saturating_sub(force_render_state.last_render_time);
 
                             // Calculate next frame callback time and push needed data to delayed_surfaces.
                             let delay_time = interval.saturating_sub(time_diff);
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -420,6 +420,7 @@ pub struct Niri {
     /// Window ID for the "dynamic cast" special window for the xdp-gnome picker.
     #[cfg(feature = "xdp-gnome-screencast")]
     pub dynamic_cast_id_for_portal: MappedId,
+    force_render_state: RefCell<HashMap<u32, ForceRenderState>>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
@@ -570,6 +571,20 @@ pub enum LockRenderState {
 struct SurfaceFrameThrottlingState {
     /// Output and sequence that the frame callback was last sent at.
     last_sent_at: RefCell<Option<(Output, u32)>>,
+}
+
+struct ForceRenderState {
+    last_time: Duration,
+    is_delay: bool,
+}
+
+impl Default for ForceRenderState {
+    fn default() -> Self {
+        Self {
+            last_time: Duration::ZERO,
+            is_delay: false,
+        }
+    }
 }
 
 pub enum CenterCoords {
@@ -2822,6 +2837,7 @@ impl Niri {
 
             #[cfg(feature = "xdp-gnome-screencast")]
             dynamic_cast_id_for_portal: MappedId::next(),
+            force_render_state: RefCell::new(HashMap::new()),
         };
 
         niri.reset_pointer_inactivity_timer();
@@ -4956,6 +4972,46 @@ impl Niri {
         }
     }
 
+    pub fn delay_to_send_frame_callbacks(
+        &mut self,
+        surface: WlSurface,
+        output: Output,
+        interval: Duration,
+    ) {
+        let timer = Timer::from_duration(interval);
+        self.event_loop
+            .insert_source(timer, move |_, _, state| {
+                state
+                    .niri
+                    .send_frame_callback_for_surface(surface.clone(), &output);
+                TimeoutAction::Drop
+            })
+            .unwrap();
+    }
+
+    fn send_frame_callback_for_surface(&mut self, surface: WlSurface, output: &Output) {
+        let frame_callback_time = get_monotonic_time();
+        let mut force_render_state_borrow = self.force_render_state.borrow_mut();
+
+        let force_render_state = force_render_state_borrow
+            .entry(surface.id().protocol_id())
+            .or_default();
+
+        // reset the delay state
+        force_render_state.last_time = frame_callback_time;
+        force_render_state.is_delay = false;
+
+        debug!("Sending frame callback for surface");
+
+        send_frames_surface_tree(
+            &surface,
+            output,
+            frame_callback_time,
+            FRAME_CALLBACK_THROTTLE,
+            |_, _| Some(output.clone()),
+        );
+    }
+
     pub fn send_frame_callbacks(&mut self, output: &Output) {
         let _span = tracy_client::span!("Niri::send_frame_callbacks");
 
@@ -4997,13 +5053,97 @@ impl Niri {
 
         let frame_callback_time = get_monotonic_time();
 
+        // Collect delayed surfaces
+        let delayed_surfaces = RefCell::new(HashMap::new());
+
         for mapped in self.layout.windows_for_output_mut(output) {
-            mapped.send_frame(
-                output,
-                frame_callback_time,
-                FRAME_CALLBACK_THROTTLE,
-                should_send,
-            );
+            // Check if the surface should be forced to render
+            if mapped.rules().force_render == Some(true) || mapped.is_window_cast_target() {
+                // Calculate delay time
+                let interval = if let Some(force_render_fps) = mapped.rules().force_render_fps {
+                    Duration::from_secs_f64(1.0 / force_render_fps as f64)
+                } else {
+                    Duration::ZERO
+                };
+
+                // Check if the surface should be forced to render
+                let should_force_render = |surface: &WlSurface, states: &SurfaceData| {
+                    // Check if the surface is on the primary output.
+                    let mut on_primary = true;
+                    let current_primary_output = surface_primary_scanout_output(surface, states);
+                    if current_primary_output.as_ref() != Some(output) {
+                        on_primary = false;
+                    }
+
+                    let mut force_render_state_borrow = self.force_render_state.borrow_mut();
+
+                    if !on_primary && !interval.is_zero() {
+                        //fps limited branch
+                        let force_render_state = force_render_state_borrow
+                            .entry(surface.id().protocol_id())
+                            .or_default();
+
+                        if !force_render_state.is_delay {
+                            // Time diff since last frame callback
+                            let time_diff: Duration =
+                                frame_callback_time.saturating_sub(force_render_state.last_time);
+
+                            // Calculate next frame callback time and push needed data to delayed_surfaces
+                            let delay_time = interval.saturating_sub(time_diff);
+                            let mut delayed = delayed_surfaces.borrow_mut();
+                            delayed
+                                .entry(surface.id())
+                                .or_insert((surface.clone(), delay_time));
+
+                            // Set force render state
+                            force_render_state.is_delay = true;
+                        }
+                        return None;
+                    } else {
+                        // Remove force render state if not on primary output
+                        if on_primary {
+                            force_render_state_borrow.remove(&surface.id().protocol_id());
+                        }
+                        //fps unlimited branch
+                        let frame_throttling_state = states
+                            .data_map
+                            .get_or_insert(SurfaceFrameThrottlingState::default);
+
+                        // Next, check the throttling status.
+                        let mut last_sent_at = frame_throttling_state.last_sent_at.borrow_mut();
+                        let mut send = true;
+
+                        // If we already sent a frame callback to this surface this output refresh
+                        // cycle, don't send one again to prevent empty-damage commit busy loops.
+                        if let Some((last_output, last_sequence)) = &*last_sent_at {
+                            if last_output == output && *last_sequence == sequence {
+                                send = false;
+                            }
+                        }
+
+                        if send {
+                            *last_sent_at = Some((output.clone(), sequence));
+                            Some(output.clone())
+                        } else {
+                            None
+                        }
+                    }
+                };
+
+                mapped.send_frame(
+                    output,
+                    frame_callback_time,
+                    FRAME_CALLBACK_THROTTLE,
+                    should_force_render,
+                );
+            } else {
+                mapped.send_frame(
+                    output,
+                    frame_callback_time,
+                    FRAME_CALLBACK_THROTTLE,
+                    should_send,
+                );
+            }
         }
 
         for surface in layer_map_for_output(output).layers() {
@@ -5043,6 +5183,12 @@ impl Niri {
                 FRAME_CALLBACK_THROTTLE,
                 should_send,
             );
+        }
+
+        // Call delay timer
+        let delayed_map = delayed_surfaces.into_inner();
+        for (_, (surface, interval)) in delayed_map {
+            self.delay_to_send_frame_callbacks(surface, output.clone(), interval);
         }
     }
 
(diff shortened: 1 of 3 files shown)


YaLTeR:
For screencast it should work out of the box; but why would this be needed outside a screencast?

busyoGG:
> For screencast it should work out of the box; but why would this be needed outside a screencast?

screenshot-2025-10-16_21-41-16

For screen recording, it already works out of the box without any configuration. 

Sometimes I run turn-based games with an auto-battle feature on another workspace, and if rendering doesn’t happen, the battle logic can’t complete.

HigherOrderLogic:
> Sometimes I run turn-based games with an auto-battle feature on another workspace, and if rendering doesn’t happen, the battle logic can’t complete.

Iirc Yalter said that clients should expect this and handle it properly instead.

busyoGG:
> > Sometimes I run turn-based games with an auto-battle feature on another workspace, and if rendering doesn’t happen, the battle logic can’t complete.
> 
> Iirc Yalter said that clients should expect this and handle it properly instead.

Hyprland also offers a similar feature.

I think providing a special solution to address this issue is also acceptable. 

If you don’t use it, it’s equivalent to not running this logic.

YaLTeR:
Yeah. That's how it works on GNOME etc

ykshek:
KDE's equivalent would be the window rule
Block Compositing: Force No 
image

busyoGG:
> KDE's equivalent would be the window ruleKDE  Block Compositing: Force No image

I don’t think it does the same thing.I just tried it and it doesn't force render the game on another workspace.

ykshek:
Nevertheless, in some retro games where game/network logic is bound to its fps, it'd be nice to have this window rule.

kamikaze211:
Oh I really need it lol
I play Honkai Star Rail, it has auto-battle mode, that I can run the game in back ground, but the background window frame limit makes it run very slow, so I have to wait for long time until the battle be completed.
When I use Plasma, there are some ways to resolve it, but I don't know how to do this on Niri.

phisch:
> > For screencast it should work out of the box; but why would this be needed outside a screencast?
> 
> For screen recording, it already works out of the box without any configuration.
> 
> Sometimes I run turn-based games with an auto-battle feature on another workspace, and if rendering doesn’t happen, the battle logic can’t complete.

I have faced the same issue in several games before, and would really appreciate an option like this window rule. Not sure if there is a better solution, but I personally like the window rule approach here.

ToxicMushroom:
My use case is FfXIV an mmo where you can sit for hours in a party finder and need to hear the game sounds to know when its filled. 
 
But ths game speed is messed up when on another workspace. Leading to missed sounds, jarring bgm, broken autowalk etc

YaLTeR:
How do they work on gnome, kde, sway?

alessiorapisarda:
> How do they work on gnome, kde, sway?

KDE works the same way as Niri currently, if a window (game) is on another workspace it will render at very low FPS/freeze
If it's hidden behind another windows (single workspace workflow) it will render as normal
In Niri the game won't render not only on another workspace but also if you scroll the window away in the same workspace (game must be fullscreen-maximized)

ToxicMushroom:
For me with FFXIV on another workspace: 
kde - fullspeed
gnome - fullspeed
sway - crawlspeed
niri - crawlspeed

idoric:
@ToxicMushroom With Niri, have you tried with the window in floating mode? Does it make a difference?

phisch:
> t should work out of the box; but why would this be needed outside a screencast?

I believe this isn't true for windows that are a dynamic screencast target. I just tested this out, and whenever the target of the dynamic screen cast is off-screen, the frames drop significantly.

YaLTeR:
> I believe this isn't true for windows that are a dynamic screencast target. I just tested this out, and whenever the target of the dynamic screen cast is off-screen, the frames drop significantly.

It isn't because it *currently* doesn't work (for screencasts).

busyoGG:
Perhaps I should reiterate that this feature **for screencasts** does not require explicit configuration and it's worked.

The frame rate drop could be due to a graphics card issue,idk.

https://github.com/user-attachments/assets/914c231b-a433-406f-94ae-e36617a6538a

YaLTeR:
> Perhaps I should reiterate that this feature **for screencasts** does not require explicit configuration.

Yes, that is how it is supposed to work and how this PR already does it. I was just clarifying for @phisch who misinterpreted my previous comment.

phisch:
Oh, I see. Sorry for misinterpreting! Very happy to see this in niri soon!

ToxicMushroom:
Thanks for keeling this uptodate, 9.5k line changes looks a bit much, did something go wrong with your rebase ?

busyoGG:
Yes, I'm trying to fix this. 😓

HigherOrderLogic:
Seems like you didnt fetch the new commits from upstream (this repo)?

MithicSpirit:
It looks like the changes from main were merged and squashed into this one commit without updating the parent commit, so now this one commit includes all changes since October.

src/niri.rs
```diff
@@ -4956,6 +4972,46 @@ impl Niri {
...
+                    .niri
+                    .send_frame_callback_for_surface(surface.clone(), &output);
+                TimeoutAction::Drop
+            })
+            .unwrap();
+    }
+
+    fn send_frame_callback_for_surface(&mut self, surface: WlSurface, output: &Output) {
+        let frame_callback_time = get_monotonic_time();
+        let mut force_render_state_borrow = self.force_render_state.borrow_mut();
```
ToxicMushroom, line 4996: I don't see the benefit for the RefCell and borrow_mut() there is already &mut self here ?

src/niri.rs
```diff
@@ -4997,13 +5053,97 @@ impl Niri {
...
-                output,
-                frame_callback_time,
-                FRAME_CALLBACK_THROTTLE,
-                should_send,
-            );
+            // Check if the surface should be forced to render
+            if mapped.rules().force_render == Some(true) || mapped.is_window_cast_target() {
+                // Calculate delay time
+                let interval = if let Some(force_render_fps) = mapped.rules().force_render_fps {
+                    Duration::from_secs_f64(1.0 / force_render_fps as f64)
```
ToxicMushroom: force_render_fps could be 0

src/niri.rs
```diff
@@ -572,6 +573,20 @@ struct SurfaceFrameThrottlingState {
     last_sent_at: RefCell<Option<(Output, u32)>>,
 }
 
+struct ForceRenderState {
+    last_time: Duration,
+    is_delay: bool,
```
ToxicMushroom: What does it mean to be delay ? 
The same can be said for last_time, why is it a duration.

These should probably be named better and/or have a doc describing the state.

src/niri.rs
```diff
@@ -420,6 +420,7 @@ pub struct Niri {
     /// Window ID for the "dynamic cast" special window for the xdp-gnome picker.
     #[cfg(feature = "xdp-gnome-screencast")]
     pub dynamic_cast_id_for_portal: MappedId,
+    force_render_state: RefCell<HashMap<u32, ForceRenderState>>,
```
ToxicMushroom: This is the only non-pub field.
Make it pub for consistency ?

src/niri.rs
```diff
@@ -4997,13 +5053,97 @@ impl Niri {
...
+            // Check if the surface should be forced to render
+            if mapped.rules().force_render == Some(true) || mapped.is_window_cast_target() {
+                // Calculate delay time
+                let interval = if let Some(force_render_fps) = mapped.rules().force_render_fps {
+                    Duration::from_secs_f64(1.0 / force_render_fps as f64)
+                } else {
+                    Duration::ZERO
+                };
+
+                // Check if the surface should be forced to render
```
ToxicMushroom: This seems to do more than simply check, it also modifies state.
And `should_...` implies a boolean return to me, but I do see that the closure above is named the same way.

ToxicMushroom (commented):
I've left some thoughts and things I thought Yalter would comment on given what was commented on during my PR review. 

Also make sure to construct your code comments as proper sentences (starting with capital and ending with .) to stay in line with the codebase.

I have no standing in this repo btw, just trying to help out.
Thanks for your work :)

src/niri.rs
busyoGG: I'm not very familiar with Rust lifetimes. I need to modify `self.force_render_state` in `should_force_render` closure, so I'm using a `RefCell`.

src/niri.rs
busyoGG: Yes, it's a problem. I will fix this.

src/niri.rs
busyoGG: You are right. I will change the variable names and add annotations.

src/niri.rs
busyoGG: I just followed the LSP's advice. I'll make it `pub` for consistency.

src/niri.rs
busyoGG: You're right, the annotation here is confusing. Since this function's logic is actually the same as `should_send`, I'll remove the annotation.

src/niri.rs
ToxicMushroom: Ah I see

src/niri.rs
```diff
@@ -572,6 +573,22 @@ struct SurfaceFrameThrottlingState {
     last_sent_at: RefCell<Option<(Output, u32)>>,
 }
 
+struct ForceRenderState {
```
MithicSpirit: I think this should be `pub` since `force_render_state` above is also `pub` (`cargo check` complains about this with warning `private_interfaces`).

src/niri.rs
MithicSpirit: Then you should make `struct ForceRenderState` also be `pub`. Otherwise, rust warns about the type being more private than the item.

busyoGG: I’m not sure why I overlooked it. Thanks for the reminder.

SanjayDevTech:
@busyoGG Thanks a lot for this change I was facing the same issue
seems like this pr is not active, just wanted to remind this.

zeroDivisible:
I was made aware of this PR through the discussion at https://github.com/niri-wm/niri/discussions/1525#discussioncomment-15940853 so I'll only add one more use case, described there: while watching youtube videos, in Firefox and under `niri`, if Firefox is not visible on the current workspace, the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) can be used by YT to decide to lower the quality of streamed content. Which in principle is a good optimization. However - yt is not very quick at switching back to higher quality for me so after flicking back to it, it takes minutes before the pixelated content gets replaced by proper high resolution stream. This seems to happen for me even with YouTube premium.

Now, there can be other plugins or settings which affect this in my setup but generally, I do believe that the problems being solved in this PR are sound.

mishl-dev:
Screen sharing any window to Discord on Niri is unusable when the window isn't focused. `force-render` with `is_window_cast_target()` would solve this without any config.

https://github.com/niri-wm/niri/blob/b07bde3ee82dd73115e6b949e4f3f63695da35ea/src/window/mod.rs#L168

src/niri.rs
```diff
@@ -4997,13 +5055,100 @@ impl Niri {
...
+                            let delay_time = interval.saturating_sub(time_diff);
+                            let mut delayed = delayed_surfaces.borrow_mut();
+                            delayed
+                                .entry(surface.id())
+                                .or_insert((surface.clone(), delay_time));
+
+                            // Set force render state.
+                            force_render_state.is_waiting_for_render = true;
+                        }
+                        return None;
```
MithicSpirit: Doesn't matter, but we might as well appease clippy.
```suggestion
                        None
```

src/niri.rs
```diff
@@ -4956,6 +4974,46 @@ impl Niri {
...
+
+        let force_render_state = force_render_state_borrow
+            .entry(surface.id().protocol_id())
+            .or_default();
+
+        // Reset the delay state.
+        force_render_state.last_render_time = frame_callback_time;
+        force_render_state.is_waiting_for_render = false;
+
+        debug!("Sending frame callback for surface");
```
MithicSpirit: Too spammy for `debug!` imo. See https://niri-wm.github.io/niri/Development%3A-Developing-niri.html#logging-levels.
```suggestion
        trace!("Sending frame callback for surface");
```

src/niri.rs
```diff
@@ -5044,6 +5189,12 @@ impl Niri {
                 should_send,
             );
         }
+
+        // Call delay timer
+        let delayed_map = delayed_surfaces.into_inner();
```
MithicSpirit, line 5195: Clippy complains about this because `ObjectId` technically has interior mutability. Ideally we wouldn't just disable the warning here, but I suspect that a proper fix would have to happen upstream.
```suggestion
        #[allow(clippy::mutable_key_type)]
        let delayed_map = delayed_surfaces.into_inner();
```

mishl-dev:
Entries are added to `force_render_state` when a window goes off its primary output, and removed when it comes back. That's the only cleanup path.

If a window closes while off its primary output the entry never gets removed and leaks in the map forever. Wayland recycles protocol IDs so eventually a new window gets the same ID as the dead one inherits the leftover state and silently skips its first frame callback.

Moortu:
from whomever wants a working patch for this:
[niri-force-render.patch](https://github.com/user-attachments/files/28681567/niri-force-render.patch)

WARNING: it's this patch but fixed up by claude. that's why I won't submit it as a commit for this.

MithicSpirit:
If anyone wants a not-vibed update I have a tree at https://github.com/mithicSpirit/niri/tree/force-render. I've been testing it for a few weeks, and it seems to work well. I'd have rebased to make the tree cleaner, but considering how messy the commit history is I expect that this will all be squashed down anyway, so it doesn't matter.

Moortu:
> If anyone wants a not-vibed update I have a tree at https://github.com/mithicSpirit/niri/tree/force-render. I've been testing it for a few weeks, and it seems to work well. I'd have rebased to make the tree cleaner, but considering how messy the commit history is I expect that this will all be squashed down anyway, so it doesn't matter.

My patch seems to have better cleanup and 2 guards? perhaps something to add to your patch? then I'll remove mine

MithicSpirit:
> My patch seems to have better cleanup and 2 guards? perhaps something to add to your patch? then I'll remove mine

It's been a while since I did this, but I'm pretty sure that my tree is literally just merging main into this PR and fixing compilation errors.

Torvosaure:
Yo,
any news on this one?

Sempyos:
No

justpav05:
So, has this feature been abandoned or what?

Sempyos:
YaLTeR is very busy, and that there is no timeline for PRs/reviews, if you want to like help, you can test the PR and review code.

Yilmaz4:
I've been using MithicSpirit's fork for over a month without any problems. I believe it's ready to merge.

src/niri.rs
```diff
@@ -5026,13 +5055,100 @@ impl Niri {
...
+                                .or_insert((surface.clone(), delay_time));
+
+                            // Set force render state.
+                            force_render_state.is_waiting_for_render = true;
+                        }
+                        None
+                    } else {
+                        // Remove force render state if not on primary output.
+                        if on_primary {
+                            force_render_state_borrow.remove(&surface.id().protocol_id());
```
cornservant, line 5110: This is the only place where force render states are being removed. I suspect the hash map will grow over time. Doesn't it make more sense to attach those states to the surface data?

(I'll try doing that on my own fork and see how it goes... https://github.com/cornservant/niri/commit/b0eddb98ab8853e980eabba3b8edc2c20406adaf)

justpav05:
> YaLTeR is very busy, and that there is no timeline for PRs/reviews, if you want to like help, you can test the PR and review code.

Okay, so how can I help? I mean, are there specific criteria or something like that?

MithicSpirit:
> Okay, so how can I help?

You can help YaLTeR by leaving well-analyzed reviews on PRs (including but not limited to this one). You can help yourself by compiling niri with this patch (and any others you want), rather than waiting on it being merged upstream.

Sempyos:
Read: https://github.com/niri-wm/niri?tab=contributing-ov-file#reviewing-and-testing-pull-requests.

busyoGG:
Hey guys, I've now switched to a different WM, so this PR won't have any new content. If anyone wants to implement the features, feel free to open another PR to help YaLTeR.

Ecamika:
The feat that added by this pr is very helpful to me. I tested MithicSpirit's fork. It worked successfully. But the author said he will no longer be maintained this pr. So may we should open a new pull request to continue?

MithicSpirit:
Yeah, I've been meaning to do that. I just want to address the memory leak mentioned above first.

justpav05:
Memory leak in Rust?)

veneficium42:
> Memory leak in Rust?)

Yes? Rust does not prevent memory leaks.

justpav05:
I'm mostly just joking)

MithicSpirit:
My reimplementation of this PR is available in #4392. I would appreciate it if people tested that PR (it's marked as a draft but only because it's missing docs; the feature should be functioning correctly now).

PR #2612, opened 2025-10-16
wiki: fix list in tab-indicator section

MageJohn:
Apology for the minor PR (this is not for hacktober). Just fixes a markdown error in the docs.

Before:
Screenshot from 2025-10-16 15-19-24

After:
Screenshot from 2025-10-16 15-19-43


Sempyos:
That seems fine,

YaLTeR:
Thanks.

> Apology for the minor PR (this is not for hacktober).

I mean, if it's a worthwhile minor PR, then those are only good because they are quick to review.

Issue #2399, opened 2025-09-16
Screencasting does not work by default, requires restarting desktop portals everytime.


enchanteddev:
When using OBS to do screenrecording, nothing is picked up and a black screen is shown. This is "fixed" by doing the following:

❯ systemctl --user stop xdg-desktop-portal.service xdg-desktop-portal-gnome.service xdg-desktop-portal-gtk.service
❯ systemctl --user start xdg-desktop-portal.service xdg-desktop-portal-gnome.service

This is not a typo, we need to kill all 3 services and start just 2 of them back.

And no, it doesn't work if I mask the gtk portal. It needs to exist and be stopped and be started again (by the gnome portal, I think, since I dont start it and it does get started as seen on btop at the same time I started the other 2 portals).

This seems like a problem with how gnome and gtk desktop portals interact.

My niri-portals.conf sets gnome for ScreenCast.

### System Information

* niri version: niri 25.08

* Distro: Fedora 42 Workstation

* GPU: Laptop with AMD iGPU and nvidia discrete.

* CPU: AMD Ryzen 7 5800H

I am really enjoying using niri, but this has become a pain point. I am unable to find any solutions and hope this gets fixed soon.

YaLTeR:
Idk it works out of the box for me and for many other people, so likely a problem with your setup

enchanteddev:
The way I have solved it currently is by using the wlr desktop portal, it kind of works right now.

I see, it could also be a gnome-gtk problem then given how weird and precise the stop and start commands need to be.

ASong5:
@enchanteddev sorry for the ping, but did you ever figure this out? I have the same exact issues and your workaround worked for me as well (except starting all three of them works for me). I'm not sure what the issue is...

EDIT: The issue seems to be fixed on my end by specifying gnome for the Settings interface. If your niri-portals.conf file is anything like mine and uses the gnome portal solely for screencasting with the gtk portal being used for everything else, e.g.,:

```
[preferred]
default=gtk;
org.freedesktop.impl.portal.Access=gtk;
org.freedesktop.impl.portal.Notification=gtk;
org.freedesktop.impl.portal.Secret=gnome-keyring;
org.freedesktop.impl.portal.FileChooser=gtk;
org.freedesktop.impl.portal.ScreenCast=gnome;
```
you will run into the issue.

The fix is to explicitly specify that it should use the gnome portal for the Settings interface where gtk fails:
```
[...]

PR #2616, opened 2025-10-16
Fix issue with screencasting when using gnome and gtk desktop portals in some cases. 

enchanteddev:
This is a very small PR that aims to solve the issue mentioned and solved here #2399 .

The solution here is to also explicitly mention the settings portal to be backed by gnome and gtk, which is probably what happens anyways for people with just one desktop portal installed, but messes up when multiple are installed (like KDE). 

This helps apps like OBS who need both the settings portal and the screencast portal to come from gnome for it to properly work.

As far as I am aware, this should not cause any regressions. The extra line merely mentions something explicitly that fresh installs of niri on systems with no other DE/Window manager would do anyways. But if this is not the case, and there is some regression, please let me know.

### DIFF
--- a/resources/niri-portals.conf
+++ b/resources/niri-portals.conf
@@ -3,3 +3,4 @@ default=gnome;gtk;
 org.freedesktop.impl.portal.Access=gtk;
 org.freedesktop.impl.portal.Notification=gtk;
 org.freedesktop.impl.portal.Secret=gnome-keyring;
+org.freedesktop.impl.portal.Settings=gtk;gnome;


YaLTeR:
This fix seems to prefer gtk to gnome. But this is not the expected default, it's expected that gnome is preferred (as it should already be).

What is actually happening, why does the problem occur? Let's perhaps get to the bottom of this somehow before figuring out what the correct change to make is?

enchanteddev:
Preferring gtk to gnome is merely in artifact of copy pasting from the issue thread; did not think about the order as much, but yes if gnome is expected default preference then the order must be switched. I will update the PR once I have tested that it works.

Regarding **why** the problem occurs, we might need to do some experimenting. 
If without that line it _only_ fails in OBS, then
- it could be an issue with OBS, or
- it could be that the desktop portal spec requires most (if not all of the portals) to be provided by the same service and OBS failing is perfectly well within that spec.

If it fails everywhere (or in most other software), then
- we might need to then enforce gnome defaults a bit more strictly than before (although I like mixing them because I prefer the KDE file picker), like in this PR.

I will try to get around with the results whenever I can.

enchanteddev:
Just checked, works with gnome;gtk; as well. I will update the PR after rest of the experimentation.

YaLTeR:
The default order is already gnome;gtk; though so how does this change anything?

enchanteddev:
In my portals.conf, I had the default set as just gtk. Now that I have changed it to be gnome;gtk; and removed the line about the settings portal (since it should now be unnecessary), it works!

So this was mostly an error at the user end. I may have not paid enough attention when writing my portals.conf file as in my head gnome portal is just a higher level wrapper over gtk portals as gnome portal relies on the gtk one - this was obviously not perfectly accurate.

Looking back at the issue #2399, the portals.conf file for @ASong5 (sorry for the ping, but I want to let you know what the real problem was) is also the same, with gtk (instead of gnome;gtk;) being the default. 

Thank you for your time. I am closing this PR.

ASong5:
In my case, prioritizing gtk over gnome as the default backend was intentional, as I do not know enough about portals - nor linux as a whole - and assumed it was in my best interest to use GTK over gnome wherever possible, as I heard gnome brings a bunch of unneeded dependencies and bloat into a non-DE compositor like niri. I'm fine with using gnome as the default backend if that is the "correct" way to do things, but it still doesn't really answer my original question from the issue thread about why the Settings interface also has to be backed by gnome (which you mentioned in this thread as well) for screencasting to work. Maybe the reason is self-evident to those more technical than I am, but I am genuinely curious.

enchanteddev:
From what I found from my experiments is that the Settings interface is not strictly required to be backed by the gnome portal, and browsers were able to screenshare just fine. Thus, it was only OBS that failed.

But around the same time I noticed the default being gtk in my portal.conf and changed to gnome;gtk and that fixed it so did not investigate deeper.

From my understanding, OBS is _probably_ not in the wrong here and situations like ours, where we have multiple interfaces backed by multiple portal backends is not something it is expecting.

PR #2617, opened 2025-10-16
Ability to toggle pointer on/off when screenshotting a window

cornservant:
This PR
- extends the `Niri::screenshot_window()` function by an `include_pointer` parameter.
- extends the IPC actions `ScreenshotWindow` and `ScreenshotWindowById` by `show_pointer` flag defaulting to `false`

Known issues
- I'm not sure if the calculated offset for the cursor elements accounts for all circumstances (e.g. presence of decorations, widths, window clipping the output rectangle, other ones I am not aware of, ...)
- Missing tests that check whether future changes break the offset calculation

Edit: Relevant Discussion
https://github.com/YaLTeR/niri/discussions/1885


YaLTeR:
This is now implemented on main

PR #2627, opened 2025-10-18
backend/tty: Consider only primary node in drm udev event

athul-krishna-kr:
Previously discussed in #2521.

Skip `/dev/dri/render*`  nodes in device_added() to use hot-plugging via udev trigger.
`sudo udevadm trigger --action=remove /dev/dri/card0`

Reference link:
https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/768fbaad54027f8dd027e7e015e8eeb93cb38c52

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -512,6 +512,13 @@ impl Tty {
 
         let node = DrmNode::from_dev_id(device_id)?;
 
+        // Only consider primary node on udev event
+        // https://gitlab.freedesktop.org/wlroots/wlroots/-/commit/768fbaad54027f8dd027e7e015e8eeb93cb38c52
+        if node.ty() != NodeType::Primary {
+            debug!("not a primary node, skipping");
+            return Ok(());
+        }
+
         if self.ignored_nodes.contains(&node) {
             debug!("node is ignored, skipping");
             return Ok(());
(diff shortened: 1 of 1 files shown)


YaLTeR:
- Should this maybe be done for changed and removed too?
- Simpler to write with `if node.ty() == ...` rather than `match`
- Should it be `if node.ty() != NodeType::Primary` to match wlroots more precisely?

athul-krishna-kr:
> * Should this maybe be done for changed and removed too?

`let Some(device) = self.devices.get_mut(&node)`
this should return None in device_removed() and device_changed(). It'll be skipped in removed, and device_changed() calls device_added()

>     * Simpler to write with `if node.ty() == ...` rather than `match`
>
>     * Should it be `if node.ty() != NodeType::Primary` to match wlroots more precisely?
```
  if node.ty() != NodeType::Primary  {
      debug!("Not a primary node, skipping");
      return Ok(());
  }
```
is this okay?

YaLTeR:
> `let Some(device) = self.devices.get_mut(&node)`
> this should return None in device_removed() and device_changed(). It'll be skipped in removed, and device_changed() calls device_added()

Yes, but it also warns about unknown device, in this case unnecessarily.

> ```
>   if node.ty() != NodeType::Primary  {
>       debug!("Not a primary node, skipping");
>       return Ok(());
>   }
> ```
> is this okay?

Yes, just consistent casing with other messages: `"not a primary node, skipping"`

athul-krishna-kr:
@YaLTeR 

> > `let Some(device) = self.devices.get_mut(&node)`
> > this should return None in device_removed() and device_changed(). It'll be skipped in removed, and device_changed() calls device_added()
> 
> Yes, but it also warns about unknown device, in this case unnecessarily.

Added checks in device_changed() and device_removed()

YaLTeR:
Thanks

PR #2636, opened 2025-10-19
Remove unnecessary unsafe in `inhibit_power_key`

MarkusPettersson98:
Was reading through the code and saw `unsafe` being used to borrow an owned file descriptor. It seems like `zbus::zvariant::OwnedFd` does implement `AsFd` since [zbus 4.0](https://github.com/z-galaxy/zbus/issues/286), so this can be achieved without `unsafe` by simply removing the manual re-borrowing!

I've played around with Niri this past weekend, and it is *amazing*. Thanks for creating such a polished, approachable & hackable compositor! :yellow_heart:


YaLTeR:
Thanks

PR #2638, opened 2025-10-19
Create parents of screenshots directory

### LINKED ISSUE #2493: Screenshot tool does not create the screenshot folder
If niri tries to take a screenshot to a non-existent folder, it silently fails and doesn't save the screenshot anywhere. Ideally, it should try to create the folder with `mkdir -p`, or at least show a pop up error to the user.

### System Information

* niri version: niri 25.08 (Nixpkgs)

* Distro: NixOS Unstable

* GPU: Nvidia GeForce RTX 3060

* CPU: AMD Ryzen 7 3700x

miku4k:
Easy one-liner.

Fixes #2493

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -5665,7 +5665,7 @@ impl Niri {
                     if let Some(parent) = path.parent() {
                         // Relative paths with one component, i.e. "test.png", have Some("") parent.
                         if !parent.as_os_str().is_empty() {
-                            if let Err(err) = std::fs::create_dir(parent) {
+                            if let Err(err) = std::fs::create_dir_all(parent) {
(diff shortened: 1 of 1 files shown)


YaLTeR:
My only question here is that if the user misspells some component of the path, they may get a whole tree of unwanted directories created. Or you think that's fine?

miku4k:
I think that's fine. The user can already typo to create one unwanted directory, and it's not much different to delete a single directory than it is to delete a whole tree.

Vortriz:
The user would then find out about the wrongly created directories after one invocation of the command itself. They can go ahead and correct it.

As I see it, there is no harm here since existing directories stay intact.

YaLTeR:
Fair enough. Thanks.

PR #2644, opened 2025-10-20
feat: auto-center-when-space-available

baduhai:
This is my attempt at adding functionality to center columns whenever there is enough space to do so. This is more or less how karousel works, and how I expected niri to work, it's especially useful for ultrawide monitors.

In theory, this should center columns whenever sum_of_all_tile_widths < available_display_width.

In practice, it does so in almost every situation, but there is a race condition that I've been unable to track down. If I open two windows, and switch through one of the windows' widths using switch-preset-column-width quickly, the windows center just fine. If I do it slowly, it also centers just as well. But there is this middle of the road speed where I try to switch through column widths and niri fails to center the windows.

My best guess for this is that I'm trying to center the windows before the windows width change has been registered, or maybe it's between switching widths while the resize is hapenning, and the calculations are done with the wrong number.

I'm not 100% sure where I'm going wrong, I'd appreciate any help!

### DIFF
--- a/niri-config/src/layout.rs
+++ b/niri-config/src/layout.rs
@@ -19,6 +19,7 @@ pub struct Layout {
     pub preset_window_heights: Vec<PresetSize>,
     pub center_focused_column: CenterFocusedColumn,
     pub always_center_single_column: bool,
+    pub auto_center_when_space_available: bool,
     pub empty_workspace_above_first: bool,
     pub default_column_display: ColumnDisplay,
     pub gaps: f64,
@@ -42,6 +43,7 @@ impl Default for Layout {
             default_column_width: Some(PresetSize::Proportion(0.5)),
             center_focused_column: CenterFocusedColumn::Never,
             always_center_single_column: false,
+            auto_center_when_space_available: false,
             empty_workspace_above_first: false,
             default_column_display: ColumnDisplay::Normal,
             gaps: 16.,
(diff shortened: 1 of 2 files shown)


baduhai:
I've given up on getting this to work. I've been using the following script to acheive the same thing:

```bash
#!/usr/bin/env bash

while true; do
    ACTIVE_WORKSPACE=$(niri msg --json workspaces | jq -r '.[] | select(.is_active == true)')
    WORKSPACE_ID=$(echo "$ACTIVE_WORKSPACE" | jq -r '.id')
    OUTPUT_NAME=$(echo "$ACTIVE_WORKSPACE" | jq -r '.output')
    
    MONITOR_WIDTH=$(niri msg --json outputs | jq -r ".\"$OUTPUT_NAME\".logical.width")
    
    SUMMED_TILE_WIDTH=$(niri msg --json windows | jq --argjson wid "$WORKSPACE_ID" -r '
        [.[] | select(.workspace_id == $wid) | {col: .layout.pos_in_scrolling_layout[0], width: .layout.tile_size[0]}]
        | group_by(.col) | map(first.width) | add
    ')

    if awk "BEGIN {exit !($SUMMED_TILE_WIDTH < $MONITOR_WIDTH)}"; then
        niri msg action center-visible-columns
    fi

    sleep 0.1
done

```

PR #2681, opened 2025-10-26
Add building a Debian package to the release process using cargo deb

frankier:
I think it would be nice to build this on release. I think the package should be reasonably widely compatible as-is, but some tweaks/testing might be needed. I'm also not really sure whether I can test the workflow myself.

Once this is working, it could be added to https://github.com/wimpysworld/deb-get and this could be given as the instructions to get the latest upstream version for Debian/Ubuntu.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,10 @@ jobs:
           cargo vendor --locked > .cargo/config.toml
           tar cJf niri-${{ github.event.inputs.version }}-vendored-dependencies.tar.xz vendor/
 
+      - name: Build Debian package
+        run: |
+          cargo deb 
+
       - name: Build
         run: cargo build --all --frozen --release
 
@@ -59,5 +63,5 @@ jobs:
         with:
           draft: true
           tag_name: v${{ github.event.inputs.version }}
-          files: niri-${{ github.event.inputs.version }}-vendored-dependencies.tar.xz
+          files: niri-${{ github.event.inputs.version }}-vendored-dependencies.tar.xz target/debian/niri_${{ github.event.inputs.version }}_amd64.deb
           fail_on_unmatched_files: true


YaLTeR:
Are you sure it would be compatible outside Ubuntu 24.04? Niri does dynamically link a whole bunch of stuff after all.

Also, wouldn't this cause problems down the line when I want to switch the workflow to a different OS (say, newer Ubuntu), but then a bunch of users relying on this pre-built package would come complaining?

This kind of makes me reluctant to do this; ideally someone would just package niri for Ubuntu, just like all the other distros.

HigherOrderLogic:
Cant you just create a repo, with actions that checkout the source code, then build from there. That way, you can also freely employ a matrix to build the package on different Ubuntu version.

Sempyos:
I personally feel this should be done by packagers and not niri itself. We don't really cater to a specific distribution, similar to most other software. 

Doing this could also make other people want to add packages for other distros, which is just even more to maintain

frankier:
Yeah it all makes sense, but I was hoping there might be a way to keep up-to-date across a few machines/without building it myself. I've now found out about pacstall and that there's a package there. I've made a PR to highlight this which should hopefully help to steer others: https://github.com/YaLTeR/niri/pull/2699

OwenTrokeBillard:
I am installing niri on Ubuntu 24.04, which is a pretty popular operating system. However, I am no Linux expert. It has been a little confusing.

It looks like there is an apt repository set up for Ubuntu 25.10 but not 24.04, so I'll have to build from source. It's less convenient than I hoped, but I'm sure I'll figure it out.

If you want to drive adoption, it would _really_ help to make the installation as simple and braindead as possible. I'm offering the perspective of a typical user.

I understand it's less work to outsource the packaging, however, know that the installation is the first and one of the most important interactions every user will have with niri. If it's a headache to get it working, many people will just move on. 

Many folks, like me, are not too bright, and would greatly appreciate a couple terminal commands that "just work".

This is by no means a criticism—just perspective that might guide future efforts. I am excited to get niri working myself.

OwenTrokeBillard:
After reading about the challenges packaging presents maintainers, I take my comments back...

Hopefully someone in the community can step up and make some unofficial PPAs.

As an aside, I got niri working, and it is absolutely amazing.

YaLTeR:
They now exist. Ubuntu is mentioned here: https://yalter.github.io/niri/Getting-Started.html#quick-start, Debian here: https://build.opensuse.org/project/show/home:AvengeMedia:danklinux

vaskark:
It would be sweet if 26.04 was included. But it's great there's PPA repos for Ubuntu.

PR #2699, opened 2025-10-29
Update installation options in Getting-Started.md

frankier:
Added a link to a pacstall package for Debian-based distros.


HigherOrderLogic:
Why not update the repology table itself?

frankier:
> Why not update the repology table itself?

It **is** in the repology table -- I've added a note pointing to the package which is the not-entirely obviously named option for the popular Debian family of distros.

YaLTeR:
Thanks

PR #2707, opened 2025-10-30
fix: building with systemd on musl libc systems

valpackett:
This mostly affects postmarketOS as we're the ones pushing musl+systemd the most currently.

### DIFF
--- a/src/utils/spawning.rs
+++ b/src/utils/spawning.rs
@@ -211,7 +211,20 @@ mod systemd {
     use super::*;
 
     pub fn do_spawn(command: &OsStr, mut process: Command) -> Option<Child> {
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks. Is there a libc crate issue about having this function available where it's available?

valpackett:
No, it's not an issue with the libc crate at all: musl itself actually doesn't have this wrapper (as of right now) because [the maintainers have some interesting opinions](https://inbox.vuxu.org/musl/0fcddca6-2081-4dbf-8f7a-de149b51d6d8@gmail.com/T/) on it. :roll_eyes:

YaLTeR:
I see

PR #2712, opened 2025-10-30
Add option to prevent cursor warping when switching outputs

DerEchteJoghurt:
This PR adds a `disable-mouse-warp-to-focused-output` option which prevents warping the cursor to another monitor. See https://github.com/YaLTeR/niri/discussions/2704.

### DIFF
--- a/docs/wiki/Configuration:-Input.md
+++ b/docs/wiki/Configuration:-Input.md
@@ -103,6 +103,7 @@ input {
     // warp-mouse-to-focus
     // focus-follows-mouse max-scroll-amount="0%"
     // workspace-auto-back-and-forth
+    // disable-mouse-warps
 
     // mod-key "Super"
     // mod-key-nested "Alt"
@@ -363,6 +364,19 @@ input {
     workspace-auto-back-and-forth
 }
 ```
+
+#### `disable-mouse-warps`
+
+<sup>Since: 25.??</sup>
+
+Disables moving the cursor to a different monitor when switching monitors.
+
+```kdl
+input {
+    disable-mouse-warps
+}
+```
+
 
 #### `mod-key`, `mod-key-nested`
 
--- a/docs/wiki/Configuration:-Input.md
+++ b/docs/wiki/Configuration:-Input.md
@@ -367,7 +367,7 @@ input {
 
 #### `disable-mouse-warps`
 
-<sup>Since: 25.??</sup>
+<sup>Since: next release</sup>
 
 Disables moving the cursor to a different monitor when switching monitors.
 
@@ -376,7 +376,6 @@ input {
     disable-mouse-warps
 }
 ```
-
 
 #### `mod-key`, `mod-key-nested`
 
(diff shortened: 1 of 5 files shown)


docs/wiki/Configuration:-Input.md
```diff
@@ -364,6 +365,19 @@ input {
 }
 ```
 
+#### `disable-mouse-warps`
+
+<sup>Since: 25.??</sup>
```
flowerysong:
```suggestion
<sup>Since: next release</sup>
```
It's important to use the correct string here, both so it's clear to readers that this isn't yet in a released version and so that the release automation catches it and makes sure it's updated.

docs/wiki/Configuration:-Input.md
```diff
@@ -364,6 +365,19 @@ input {
 }
 ```
 
+#### `disable-mouse-warps`
```
flowerysong: I feel like this should be called `disable-mouse-warp-to-focused-output` or similar, to communicate that it doesn't affect all warps. But I haven't come up with a name I'm entirely happy with, partially because it gets wordy but mostly because it's difficult to align it with the existing `warp-mouse-to-focus`.

src/niri.rs
```diff
@@ -1042,6 +1042,10 @@ impl State {
     }
 
     pub fn move_cursor_to_output(&mut self, output: &Output) {
```
flowerysong, line 1044: Instead of modifying this function, you should add something like `maybe_warp_cursor_to_output()` that calls it. It requires touching more places, but it makes it clear to future people reading the callers that this function might decline to move the cursor.

flowerysong (changes requested):

docs/wiki/Configuration:-Input.md
DerEchteJoghurt: Ah cool, I didn't know this gets picked up automatically during releases. Changed it!

docs/wiki/Configuration:-Input.md
DerEchteJoghurt: I agree, renamed it to `disable-mouse-warp-to-focused-output` for now.

DerEchteJoghurt:
@YaLTeR Is this something you would consider merging? Do you have any better ideas for the configuration name?

davisoares454:
I would love this to be implemented!!

arqueon:
Another concrete use case for this: a taskbar widget on a panel (DMS unified taskbar) with a dual-monitor setup. Clicking a window that lives on the other output warps the cursor to that monitor's center even with `warp-mouse-to-focus` unset — activating a window on the same output correctly leaves the cursor alone. I traced it to `focus_window()` calling `move_cursor_to_output()` unconditionally when the active output changes, which is exactly the path this PR makes optional. Since Wayland clients can't reposition the pointer themselves, there's no way to work around this from the shell/plugin side — a config option like this is the only fix. Would love to see it merged.

PR #2718, opened 2025-10-31
tty: Fix a panic when updating ignored nodes on udev event

shaunren:
When called from the udev_dispatcher callback, the invocation udev_dispatcher.as_source_ref() will panic, since the dispatcher is being mutably borrowed.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -186,7 +186,7 @@ impl Backend {
 
     pub fn on_output_config_changed(&mut self, niri: &mut Niri) {
         match self {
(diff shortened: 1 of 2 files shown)


YaLTeR:
Sorry for taking long to review. I don't think adding an argument is a good solution here. Looking at how ignored nodes aren't even part of the output config (but rather the debug config), I think a better idea is to split out the "update ignored nodes" part into a separate function, and call it prior to update output config where it makes sense (and don't call it where it doesn't make sense).

YaLTeR:
Did that in 3eef19dc7dde989e27dc61f251a611d7f9d513de, which supersedes this PR.

Issue #2734, opened 2025-11-03
Non ASCII File Name Causes OSTree Deploy to Crash when Deploying Bootc Image with Niri Installed


noelmiller:
Distribution: Fedora 43
Niri Version: 25.08-1.fc43

`/usr/share/doc/niri/wiki/Layer‐Shell-Components.md` is reading as below:

`/usr/share/doc/niri/wiki/Layer%E2%80%90Shell-Components.md`

The %E2%80%90 part is a URL-encoded, 3-byte UTF-8 sequence for the character ‐ (U+2010, the "HYPHEN" character).

This is not the standard ASCII hyphen - (U+002D) that is on your keyboard.

When ostree (or any tool running in a C or POSIX locale) tries to read this filename, it sees the valid UTF-8 bytes E2 80 90 and fails because its "current locale" doesn't know how to represent that character.

The fix would be to make sure that Layer-Shell-Components.md is using a traditional ASCII hyphen for all of it's hyphens, just like all of the other docs files.

I managed to work around this problem by removing the offending file and then rebuilding my Bootc ISO: https://github.com/noelmiller/mordor/commit/8a687d4d19d00258cbf1c1c0d4f3ac4fe4678163

YaLTeR:
See https://github.com/YaLTeR/niri/pull/2447#issuecomment-3335890152

YaLTeR:
I think this will be fixed by https://github.com/ostreedev/ostree/pull/3559

PR #2735, opened 2025-11-03
fix (docs): Layer-Shell-Components renamed to have ASCII Hyphens

noelmiller:
Fixes: #2734

It's hard to see in the PR, but you can see it in the git log: 
image


LuckShiba:
Think it's a duplicated PR 
#2447 

Edit: Just saw you mentioned this and commented there, but looking at the changes... they look the exact same?

YaLTeR:
I think this will be fixed by https://github.com/ostreedev/ostree/pull/3559

PR #2746, opened 2025-11-04
feat: add new input-passthrough window rule

zitongcharliedeng:
Adds a new window rule that makes windows transparent to input events, allowing clicks to pass through to windows below.

Use cases: Gaming overlays, custom HUDs, desktop-wide status displays, or watching content on top without stealing input from applications underneath (Discord/OBS/video players over games or coding tools). Especially useful when coupled with floating and opacity rules to create custom overlays with ease. No other window manager has this feature surprisingly afaik.

## Implementation

- Added `input_passthrough` config property to window rules.
- Modified `Tile::hit()` to return None for passthrough windows.

Core logic is a simple check in `tile.rs`:
```rust
if self.window.rules().input_passthrough == Some(tr>
    return None; // Pass through all input
}
```

## Example usage:
```kdl
window-rule {
    match app-id=r#"^discord$"#
    opacity 0.5  
    //^ Optional - just to see content under if target window is 100% alpha
    input-passthrough true
}
```

### DIFF
--- a/niri-config/src/window_rule.rs
+++ b/niri-config/src/window_rule.rs
@@ -72,6 +72,8 @@ pub struct WindowRule {
     pub scroll_factor: Option<FloatOrInt<0, 100>>,
     #[knuffel(child, unwrap(argument))]
     pub tiled_state: Option<bool>,
+    #[knuffel(child, unwrap(argument))]
+    pub input_passthrough: Option<bool>,
 }
 
 #[derive(knuffel::Decode, Debug, Default, Clone, PartialEq)]
--- a/src/layout/tile.rs
+++ b/src/layout/tile.rs
@@ -860,6 +860,10 @@ impl<W: LayoutElement> Tile<W> {
         let offset = self.bob_offset();
         let point = point - offset;
 
+        if self.window.rules().input_passthrough == Some(true) {
+            return None;
+        }
+
         if self.is_in_input_region(point) {
             let win_pos = self.buf_loc() + offset;
             Some(HitType::Input { win_pos })
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -119,6 +119,7 @@ pub struct ResolvedWindowRules {
 
     /// Override whether to set the Tiled xdg-toplevel state on the window.
     pub tiled_state: Option<bool>,
+    pub input_passthrough: Option<bool>,
 }
 
 impl<'a> WindowRef<'a> {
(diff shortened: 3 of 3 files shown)


HigherOrderLogic:
Other compositor doesnt have this feature yet because it doesnt really make any sense.

Layer shell can achieve all things that you said above.

zitongcharliedeng:
As in I can convert existing windows into passthrough overlays using layer shell? Do you have any examples on how to do that? @HigherOrderLogic

HigherOrderLogic:
> As in I can convert existing windows into passthrough overlays using layer shell?

No, layer shell is a different thing than your normal windows. It is meant to be a layer that can be placed on your screen, above or below windows, and allow for things like input passthrough or focus grab.

zitongcharliedeng:
Yeah I think this is still a separate useful feature to be paired with floating window then, layer-shell sounds like I'd have to make overlays from scratch, rather than programmatically just allow any Niri window to turn into a passthrough HUD-like overlay whenever I want. To turn existing things like web based input overlays into live screen gaming keystroke tracking overlays using a single property

zitongcharliedeng:
We already have floating windows so I think it's fine for niri to also define passthrough here.

zitongcharliedeng:
@YaLTeR thoughts? We could add an IPC for it too, like toggle-input-passthrough.

This current small change was enough for it to work on a nested Niri env for me, and I got a nice floating desktop/gaming overlay working, showing me a keyboard input overlay using a simple webapp. Not sure if I need to contribute tests and stuff in this PR. 

And I'm pretty sure this fulfills a separate functionality than if we were to use layer-shells from what I looked into.

shaleh:
https://drewdevault.com/2018/07/29/Wayland-shells.html is a nice explainer for shells.

What I think I hear HigherOrderLogic saying is... what if you had a way to "promote" a window to particular layer? Toggle that layer as pass through or not. Then you can setup your HUD as "x, y, and z are in the overlay layer and pass through". Similar logic to what is in this PR, but implemented within the layer-shell framework.

ppenguin:
> And I'm pretty sure this fulfills a separate functionality than if we were to use layer-shells from what I looked into.

:+1:  :100: 

I just found this and I agree. Especially if you use e.g. a HUD/Flightsim dashboard with a touchscreen, it seems like pretty essential functionality and the best (uninvasive/simple) way to achieve it.  An example: you need to turn/slide a button on your HUD but it's interpreted as a drag event by the compositor, the passthrough functionality would instantly fix this.

This kind of behaviour belongs in window rules IMHO.

I'll be trying this PR as a patch on my daily driver and report back...

PR #2751, opened 2025-11-05
docs: add move-column-to-workspace focus parameter example

martinetd:
I was just opening a discussion when I found about it when looking at how hard it'd be to implement, and noticed it was already done recently but no doc was added in #1332 !

Not sure this is the best place to add this, but it's better than nothing -- happy to let you add it somewhere else and close it instead, will likely be faster than iterating on details.

-----------------

This feature was added a while ago but nothing in the doc mentions it anywhere. Add a basic usage example.

Fixes: #1332

### DIFF
--- a/docs/wiki/Configuration:-Key-Bindings.md
+++ b/docs/wiki/Configuration:-Key-Bindings.md
@@ -405,3 +405,15 @@ binds {
     Super+Alt+L allow-inhibiting=false { spawn "swaylock"; }
 }
 ```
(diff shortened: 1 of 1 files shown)


martinetd:
BTW is there any way to discover all valid parameters short of reading src/input/mod.rs ?
`niri msg action help` could perhaps list parameter names somehow?

(but I'll leave that as an exercise for the reader...)

PR #2757, opened 2025-11-06
screenshot: Add configurable compression, format, and resize options

szabolcsf:
This commit adds new configuration options to reduce screenshot file sizes, particularly beneficial for high DPI monitors where screenshots can exceed 15MB.

New config options under the `screenshot` block:
- clipboard-compression: PNG compression level (fast/default/best)
- clipboard-resize-width/height: Resize before copying to clipboard
- file-format: Choose PNG or JPEG for disk files
- file-quality: JPEG quality setting (1-100)
- resize-width/height: Resize before saving to disk

The implementation:
- Clipboard always uses PNG for compatibility (e.g., Firefox on Wayland)
- Resizing maintains aspect ratio when only one dimension is specified
- Only downscales, never upscales
- JPEG encoding automatically strips alpha channel
- All encoding happens in background threads

Example file size reductions on 4K displays:
- Clipboard PNG with resize + best compression: 15MB → 2-3MB
- Disk JPEG with resize: 15MB → 400KB

Adds jpeg-encoder 0.6.0 dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,6 +1993,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "jpeg-encoder"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b454d911ac55068f53495488d8ccd0646eaa540c033a28ee15b07838afafb01f"
+
+[[package]]
 name = "js-sys"
 version = "0.3.81"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2383,6 +2389,7 @@ dependencies = [
  "glam",
  "input",
  "insta",
+ "jpeg-encoder",
  "keyframe",
  "libc",
  "libdisplay-info",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,6 +84,7 @@ pango = { version = "0.20.12", features = ["v1_44"] }
 pangocairo = "0.20.10"
 pipewire = { git = "https://gitlab.freedesktop.org/pipewire/pipewire-rs.git", optional = true, features = ["v0_3_33"] }
 png = "0.18.0"
+jpeg-encoder = "0.6.0"
 portable-atomic = { version = "1.11.1", default-features = false, features = ["float"] }
 profiling = "1.0.17"
 sd-notify = "0.4.5"
--- a/docs/wiki/Configuration:-Miscellaneous.md
+++ b/docs/wiki/Configuration:-Miscellaneous.md
@@ -11,6 +11,14 @@ prefer-no-csd
 
 screenshot-path "~/Pictures/Screenshots/Screenshot from %Y-%m-%d %H-%M-%S.png"
 
+screenshot {
+    clipboard-compression "best"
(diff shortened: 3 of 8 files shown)


Sempyos:
This imo, really isn't a thing for niri itself to handle.

This should be much more up to the uh end user to decide if they need to compress and resize their screenshots with the proper tools

YaLTeR:
Yeah, sounds fairly out of scope tbh. You can have a script that watches the screenshot folder and recompresses.

szabolcsf:
Nope, it's mainly about reduction of screenshot size on the clipboard.

Try this: take a screenshot on a 4K display and paste it into telegram desktop. 18 freaking MB.

Sempyos:
...that's a you issue, not a niri one. All we do is provide you with the fully raw image, and it's up to you to decide how it should be post processed

Fireye04:
The extra dependency kinda sucks, but in all fairness, niri does implement a custom screenshot utility out of the box that many users would probably appreciate compression on (upload filesize limits etc...) and installing a second one to handle compression does seem a bit wasteful. Then again this could also probably be handled by a clipboard manager as well.

Frontear:
> ...that's a you issue, not a niri one. All we do is provide you with the fully raw image, and it's up to you to decide how it should be post processed

Sounds like a dumb take. If you're going to have a screenshot utility built into the compositor, then most people would expect it to have some simple features as a part of its implementation. You really think it's more acceptable to have a half-baked feature than completing it? Why have it at all then?

If it's about avoiding code churn in the compositor for a non-compositor specific feature, then move the screenshot utility to its own standalone program in a different repo, so that development can progress there instead.

YaLTeR:
> If you're going to have a screenshot utility built into the compositor, then most people would expect it to have some simple features as a part of its implementation. You really think it's more acceptable to have a half-baked feature than completing it? Why have it at all then?

If you don't like it then don't use it

Frontear:
> If you don't like it then don't use it

I think you're misunderstanding my point. I _do_ like it, and many others do as well. For a lot of us it's great to have an alternative to tools like `grimblast`, built right into the compositor itself.

My point is that it seems kind of silly to wave away any attempts to improve this functionality by saying it's "out-of-scope". Sure, it's not really compositor related, but since it's a feature that's built-in, it makes sense that the many people who rely on it would want improvements, no?

YaLTeR:
> wave away any attempts to improve this functionality by saying it's "out-of-scope"

No?.. Two last releases had screenshot UI improvements.

I'm not very interested in this either bad faith or misinformed arguing.

Frontear:
> No?.. Two last releases had screenshot UI improvements.
>
> I'm not very interested in this either bad faith or misinformed arguing.

I never once said that there has never been an improvement pushed to the screenshot functionality.

This entire chain started from the initial comment "...that's a you issue, not a niri one.", and everything I have said had been exclusively about that comment and this PR. In retrospect my first comment was worded a lot more aggressively than it should have been, so I apologize for that.

I still think the first comment was unnecessarily dismissive of a very obvious improvement. That's ultimately all I was really getting at. Take it as you will.

PR #2761, opened 2025-11-06
package shell completions in COPR

AldanTanneo:


YaLTeR:
What's an existing Fedora package I could look at to cross-check that this is correct?

kevinnkansah:
@YaLTeR https://docs.fedoraproject.org/en-US/packaging-guidelines/ShellCompletions/#_shell_completions_packaging

AldanTanneo:
https://src.fedoraproject.org/rpms/niri/blob/rawhide/f/niri.spec

I guess the niri spec in fedora rawhide might be a good reference :P I should probably move the completion generation to the 'build' section

YaLTeR:
Yeah, let's make it consistent with the official package spec

AldanTanneo:
I harmonized it with the official package spec, it works locally :)

I could also add a CI step to run `rpkg local` to test the spec, but maybe it would take too long (idk if it can be configured to build in debug mode?)

YaLTeR:
Thanks.

> I could also add a CI step to run `rpkg local` to test the spec, but maybe it would take too long (idk if it can be configured to build in debug mode?)

Nah, no need. niri-git builds every commit anyway, and sometimes something breaks there due to Fedora upgrades, don't want to break CI when that happens.

PR #2794, opened 2025-11-14
implement kde appmenu protocol

ilsubyeega:
- [x] Implement smithay and handler.
- [ ] Refactor smithay implementation; i believe this is not ideal implementation and could be more cleaner. (hope yalter can fix this)
- [x] Figure out where to save the state of client
- [ ] FIXME: make this protocol OPT-IN since most of client will hide the menu when this protocol exists in the session

Notes:
- used just raw xml protocol specifications but wayland-protocol-plasma exists by `smithay`, but not reexported
- compositor only need to save the 2 data(service_name and object_path) and then export somewhere, maybe realtime, if i understood correctly.
- for dbusmenu implementation just wip stuffs are here: https://github.com/ilsubyeega/appmenu

The appmenu works with (ignoring xwayland support):
- wayland compositor with `appmenu  protocol support`
- client that supports kde appmenu wayland protocol; when they do, they register new d-bus bus which contains `/MenuBar/...`, and then sends `org_kde_kwin_appmenu#send_address(service_name, object_path)` to compositor
- the application(shell) to listen the active/focused window and its `service_name/object_path`. probably niri ipc could do that

image

DO NOT MERGE YET: `todo!()` will panic your environment


resources/kde-appmenu.xml
HigherOrderLogic, line 1: I think you should use [wayland-protocols-plasma](https://docs.rs/wayland-protocols-plasma/latest/wayland_protocols_plasma/) instead of pulling the XML file into repo.

HigherOrderLogic (commented):
I dont think Yalter is going to accept this PR but anyway...

flake.nix
```diff
@@ -93,6 +94,7 @@
               libgbm
               pango
               wayland
+              gtk4
```
zimward: why do we need to pull gtk4 in?

ilsubyeega: `rust-analyzer` with vscode complained with that but i believe its fhs packaging issue with vscode tho, i could remove that but bumping the flake is essential since rust crate dependencies uses newer libraries

Sempyos:
Can you explain why we need to implement this protocol..?

ilsubyeega:
> Can you explain why we need to implement this protocol..?

@Sempyos In wayland area, there is no other way to support this feature at this time. See this 6-year old issue: https://github.com/Alexays/Waybar/issues/288

resources/kde-appmenu.xml
ilsubyeega: Done

flake.nix
ilsubyeega: Removed changes from `flake.nix`

ilsubyeega:
closing this as my code is worse than ai generated, but it has PoC so whatever someone can take this instead. o,o..

Issue #1024, opened 2025-01-20
Zoom/magnifier


YaLTeR:
Accessibility feature that zooms in the screen around the cursor.

This works https://github.com/negrel/wooz but is a bit fiddly to run.

Need to see how the magnifier works in other compositors, especially GNOME and KDE. I pretty much never used it so I'm not sure what the expected UX is.

Tentatively, I'm thinking discrete actions could look something like:
```
binds {
    Mod+Alt+WheelScrollUp   { set-zoom-level "+10%"; }
    Mod+Alt+WheelScrollDown { set-zoom-level "-10%"; }
}
```

But it should ideally be a gesture to use high-precision touchpad swipe input and high-precision scroll inputs (if the mouse supports it).

For other monitors (the ones where the cursor currently *isn't*) I guess we don't do any zooming?

When opening the screenshot UI, need to see what would feel good to use.

Implementation-wise, idk how cursed it would be to simply scale all render elements on an output? If it works, then it's probably fine for the first implementation. Though, it will do bilinear scaling rather than nearest.

If this doesn't work, or for nearest scaling, we would probably want to render into a texture, then render that texture, upscaled around the cursor with the nearest filter. Not sure how this will work with damage tracking though.

albassort:
This would be a very positive improvement and would very differentiate this project. To my knowledge, neither  Sway or Hyprland have any functioning zoom, and neither does other x11 compositors. This is quite important to accessibility, and for me personally, as a low-vision person, is mandatory.

YaLTeR:
Here a heavy screen magnifier user says KDE has a very good implementation: https://mastodon.social/@acidiclight/114065863939963061

They also gave detailed feedback to COSMIC: https://github.com/pop-os/cosmic-comp/issues/1237

Something to check out for reference.

SergioRibera:
For what it's worth, at the time I was making [this application](https://github.com/SergioRibera/zoomer) to have a magnifier in wayland inspired by MacOs, the repository has no readme because it was a very conceptual idea and I was not achieving my goal with winit (I had to make a fork).

https://github.com/user-attachments/assets/3b95a8c3-6ced-456f-8eba-c95c13459d94

systemd11:
@SergioRibera does this magnifier work in full screen too?

SergioRibera:
@systemd11 I should but I really don't think my application is ready for production

2gibbert:
[...]

PR #2797, opened 2025-11-14
shake-to-locate feature: enlarge the cursor when shaking

SergioRibera:
Adds an integrated implementation of the `shake-to-locate` functionality (enlarge the cursor when the user shakes it) and exposes it as a grouped setting under `cursor.shake`.

The details I implemented for the feature are:
- Detects rapid shaking using movement history and a threshold based on the distance-traveled/diagonal ratio of the area traveled (method inspired by the KDE solution).
- Smoothly animates the expansion and contraction of the cursor and allows the feature to be enabled/disabled.

Personally, I think it adds a little touch of “personality” and visual polish to Niri, improving the user experience without being intrusive. It also clearly improves accessibility because it helps locate the pointer on large screens or when using multiple monitors. In another particular case, combined with the zoom adjustment (https://github.com/YaLTeR/niri/issues/1024), it is very useful in presentations and courses (for example, when giving talks or Rust classes) because it allows you to draw the audience's attention visually and subtly.

This algorithm is based on the KDE implementation, so here are the references and sources of inspiration:
- KDE KWin: https://invent.kde.org/plasma/kwin/-/blob/96bc84d33d2a5913bfec17b96686b5f4bb4e41c8/src/plugins/shakecursor/shakedetector.cpp
- Hyprland plugin: https://github.com/VirtCode/hypr-dynamic-cursors/blob/9a5b0af83dd1e9de614db56ff26fd6e62e915fd5/src/other/Shake.cpp

https://github.com/user-attachments/assets/a319e709-6cc5-4831-aded-368c8a8052e5

### DIFF
--- a/docs/wiki/Configuration:-Animations.md
+++ b/docs/wiki/Configuration:-Animations.md
@@ -62,6 +62,16 @@ animations {
     recent-windows-close {
         spring damping-ratio=1.0 stiffness=800 epsilon=0.001
     }
+
+    cursor-expand {
+        duration-ms 200
+        curve "ease-out-cubic"
+    }
+
+    cursor-shrink {
+        duration-ms 300
+        curve "ease-out-cubic"
+    }
 }
 ```
 
@@ -436,6 +446,36 @@ The close fade-out animation of the recent windows switcher.
 animations {
     recent-windows-close {
         spring damping-ratio=1.0 stiffness=800 epsilon=0.001
+    }
+}
+```
+
+#### `cursor-expand`
+
+<sup>Since: 25.??</sup>
+
+Animation when the cursor grows due to shake detection.
+
+```kdl
+animations {
+    cursor-expand {
+        duration-ms 200
+        curve "ease-out-cubic"
+    }
+}
+```
+
+#### `cursor-shrink`
+
+<sup>Since: 25.??</sup>
+
+Animation when the cursor returns to normal size after shake.
+
+```kdl
+animations {
+    cursor-shrink {
+        duration-ms 300
+        curve "ease-out-cubic"
     }
 }
 ```
--- a/docs/wiki/Configuration:-Miscellaneous.md
+++ b/docs/wiki/Configuration:-Miscellaneous.md
@@ -222,6 +222,10 @@ cursor {
     shake {
         off
         max-multiplier 2.5
+        behavior "hold"
+        cooldown-ms 400
+        stopped-threshold-ms 50
+        shake-relax-ms 150
         post-expand-delay-ms 250
         expand-duration-ms 200
         decay-duration-ms 300
@@ -239,6 +243,48 @@ Maximum cursor scale multiplier applied when a shake is detected. For example, `
 ```kdl
 shake {
     max-multiplier 2.5
+}
+```
+
+#### `behavior`
+
+Selects the shake behaviour.
+  - `"hold"` (default): keep the cursor enlarged while the pointer is moving; when the pointer stops the shrink animation is scheduled.
+  - `"intensity"`: keep the cursor enlarged while shake intensity is high; start shrinking when intensity drops below `sensitivity`.
+
+```kdl
+shake {
+    behavior "hold"
+}
+```
+
+#### `cooldown-ms`
+
+After a shrink (decay) completes, ignore new shake-triggered expansions until this many milliseconds have passed. Prevents immediate re-expansions when the pointer is noisy.
+
+```kdl
+shake {
+    cooldown-ms 400
+}
+```
+
+#### `stopped-threshold-ms`
+
+Only for `behavior "hold"`. The pointer is considered "stopped" if no motion has been observed for this many milliseconds; when stopped the shrink is scheduled after `post-expand-delay-ms`.
+
+```kdl
+shake {
+    stopped-threshold-ms 50
+}
+```
+
+#### `shake-relax-ms`
+
+Only for `behavior "intensity"`. How long the shake factor must remain below `sensitivity` before decay is scheduled. Prevents transient dips from triggering shrink.
+
+```kdl
+shake {
+    shake-relax-ms 150
 }
 ```
 
--- a/docs/wiki/Configuration:-Miscellaneous.md
+++ b/docs/wiki/Configuration:-Miscellaneous.md
@@ -22,6 +22,16 @@ cursor {
 
     hide-when-typing
     hide-after-inactive-ms 1000
+    // shake {
+        // off
+        // max-multiplier 2.5
+        // inactivity-timeout-ms 250
+        // expand-duration-ms 200
+        // decay-duration-ms 300
+        // shake-interval-ms 400
+        // min-diagonal 100.0
+        // sensitivity 2.0
+    // }
 }
 
 overview {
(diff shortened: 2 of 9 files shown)


SergioRibera:
I have now implemented two different behaviors because I realized that I like both and think they can be useful in different situations:

**Intensity based**: this detects when the intensity of the shake decreases and only then begins to restore the size (similar behavior to MacOS).

https://github.com/user-attachments/assets/020e28d6-345c-4f63-89f1-dacd8cb2f57b

**Hold based**: this maintains the size of the cursor as long as the mouse is moving, regardless of the intensity once it has been activated (useful for presentations and such).

https://github.com/user-attachments/assets/86cdd4cf-f460-465e-a22f-e87c770bed56

SergioRibera:
@YaLTeR What do you think of this feature? Would you like to review it? Or would you prefer to discuss it? 😅

Uzlkav:
Hi, I tested this recently by applying a patch on top of [c814c](https://github.com/niri-wm/niri/commit/c814c656c53ea9d69f5afb45c88f4dc4d25338cd), but couldn't get it to work. This is the config I was using, though I also tried making the `sensitivity` and `shake-interval-ms` much larger to make sure I wasn't just not shaking hard enough.
```
cursor {
    shake {
        // off
        max-multiplier 4.5
        behavior "hold"
        shake-interval-ms 200
        min-diagonal 100.0
        sensitivity 3.0
    }
}
```
I have also previously tried pulling specifically this branch rather than patching on to main, which I recall did not work either. Any ideas about how I might troubleshoot this?

cwelsys:
> Hi, I tested this recently by applying a patch on top of [c814c](https://github.com/niri-wm/niri/commit/c814c656c53ea9d69f5afb45c88f4dc4d25338cd), but couldn't get it to work. This is the config I was using, though I also tried making the `sensitivity` and `shake-interval-ms` much larger to make sure I wasn't just not shaking hard enough.
> 
> ```
> cursor {
>     shake {
>         // off
>         max-multiplier 4.5
>         behavior "hold"
>         shake-interval-ms 200
>         min-diagonal 100.0
>         sensitivity 3.0
>     }
> }
> ```
> 
> I have also previously tried pulling specifically this branch rather than patching on to main, which I recall did not work either. Any ideas about how I might troubleshoot this?

you have off commented out, but off is the default so you need to explicitly say on.

SergioRibera:
I sincerely apologize for not responding sooner; I've been very busy. I'll try to rebase the main branch to bring it in line with the latest version of Niri and if anyone has comments or suggestions, I'm open to reviewing it

Uzlkav:
> you have off commented out, but off is the default so you need to explicitly say on.

Ah, I see. Will test this. Doesn't seem to be mentioned or indicated in the wiki commits, that's probably worth mentioning explicitly.

Uzlkav:
> I sincerely apologize for not responding sooner; I've been very busy. I'll try to rebase the main branch to bring it in line with the latest version of Niri and if anyone has comments or suggestions, I'm open to reviewing it

No worries! Thanks for your work on this, I'm excited about this feature. When I have the chance I'll test again and will be sure to share any feedback.

Uzlkav:
After initial testing, here's the things I've noticed:

- The detection works well, even on a trackpad
- Scaling looks nice and smooth
- On the default settings, scaling feels just a little too easy to activate, though this is mostly preference
- When using fractional scaling for the monitors, cursor scaling sometimes doesn't do anything:

It's possible this is due to the cursors I'm using not providing big enough sizes, as this didn't happen _always_ for _all_ the cursors I have installed. However I think it may be worth seeing if it's possible to get around this as with one of my cursors (graphite) at the size I like to use it at (24 with laptop monitor scaled to 1.25), the cursor isn't scaled at all.

I don't speak rust, so my analysis may be wrong, but it looks like you're pulling in an external crate to manage the actual size changes? My understanding of cursors on linux is that they come in a few standardized sizes; 12, 18, 24, 32, etc. Values in between are rounded up/down. My guess is that you're cycling through a few of these incrementally? Fractional scaling seems to mess this up in some way.

SergioRibera:
Hey @Uzlkav Thank you for the detailed testing and feedback!

I just pushed a change that might fix the scaling issue - the shake effect should now be consistently visible across all scaling configurations. Please give it a try and let me know if it works better for you!

The detection sensitivity and other parameters should work as expected now. Thanks again for catching this edge case! 🙌🏼

Uzlkav:
@SergioRibera Just tested your most recent changes. The issue with fractional scaling seems to be improved but not entirely fixed.
- Adwaita cursor works in all cases now, but the animation looks a little janky when using fractional scale. Sometimes it scales cleanly, sometimes it seems to just jump from big to small or vice versa without inbetweens.
- My [Wii](https://primmr.dev/projects/wii-pointer-cursors/) cursors work in most cases, but they don't grow with 1.1x scaling (which is wierd, since every other factor I tried did work with this cursor). They also have a similar jumpyness issue to the one that adwaita has at fractional scales.
- Graphite cursor still doesn't want to scale at all with fractional scaling.

Also, I thought to check whether there was any wierd behaviour when the cursor version changes (eg from normal to spinner) while it's scaled, didn't find anything, so that works well!

Uzlkav:
Update after testing latest changes:
The "jumpiness" of the cursor expansion seems to be gone entirely now, with all the cursors and scales I tested - Yay! My graphite cursor still doesn't scale all that much, but I'm pretty sure that's becuase it doesn't provide very many sizes.

When max-multiplier is set to something small, say `4.0`, it works as expected, clamping the cursor size to be only so much bigger than the default size. If you make it bigger than a certain number (seems to be about `5.0` for most of my cursors), the cursor doesn't actually get 100x bigger than it normally is, but is maxed out at a certain smaller scale. Additionally, setting `max-multipler` to be _really_ big seems to make the animation smoother. Video below

https://github.com/user-attachments/assets/c9195d02-5b3c-442c-909e-033575c9fac7

SergioRibera:
@Uzlkav Hey, I really appreciate you testing this so thoroughly, the video is really helpful. I’m thinking of making the transitions between fixed sizes smooth regardless of the scale 🤔 and maybe adding an option to ignore the maximum quality limit set by the cursor 🤔. It might look pixelated in some cases, but it could achieve the scaling effect. I’m going to try it out with some of the cursors you mentioned.

Uzlkav:
It would be fun (if not practical at all) to be able to have the cursor scale up so large it covers the whole screen like you can do with KDE :D 
I think smooth transitions in all cases is probably the way to go as well, I will continue to test that and general use as you push more commits. I'm very excited for this feature, so I'm glad I can help despite my lack coding prowess!

SergioRibera:
Hey @Uzlkav, I just pushed a change that should fix both things you flagged: the cap at higher max-multiplier values, and the jumpiness between cursor sizes. Now the cursor should grow smoothly all the way up, and once it goes past whatever max size your theme provides it just keeps scaling (it might get a bit soft/blurry past that point, but at least it won't plateau anymore — so the "cover the whole screen like KDE" idea should actually work now 😄).

Whenever you get a chance, give it a try with the cursors you tested before (especially Graphite) and let me know how it feels 😁

Uzlkav:
Sorry for the delay in response, scaling for all cursors seems to be smooth and consistent now, and you can indeed make the cursor grow as large as you want, which is _very_ fun lol.

druskus20:
I've been using this PR for a month or so with no noticeable issues :+1:  love it. Awesome work

PR #2804, opened 2025-11-15
feat: flash window on focus change

perstarkse:
Hi, first contribution to Niri.

The following commit contains changes enabling a opacity change on a newly focused window, similar to [hyprfocus](https://github.com/pyt0xic/hyprfocus) and discussed in [1380](https://github.com/YaLTeR/niri/discussions/1380). It's configurable with regards to on/off, the target opacity, duration and not flashing when a single window is displayed.

### DIFF
--- a/niri-config/src/appearance.rs
+++ b/niri-config/src/appearance.rs
@@ -680,6 +680,100 @@ pub struct TabIndicatorRule {
     pub urgent_gradient: Option<Gradient>,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct FocusOpacity {
+    pub enabled: bool,
+    pub flash_opacity: f32,
+    pub disable_on_solo: bool,
+}
+
+impl Default for FocusOpacity {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            flash_opacity: 0.8, // Valid range: [0.0, 1.0]
+            disable_on_solo: false,
+        }
+    }
+}
+
+impl MergeWith<FocusOpacityPart> for FocusOpacity {
+    fn merge_with(&mut self, part: &FocusOpacityPart) {
+        if part.off {
+            self.enabled = false;
+        } else if part.on {
+            self.enabled = true;
+        }
+
+        if let Some(flash_opacity) = part.flash_opacity {
+            // Clamp flash opacity to valid range [0.0, 1.0]
+            self.flash_opacity = (flash_opacity.0 as f32).clamp(0.0, 1.0);
+        }
+        if let Some(disable_on_solo) = part.disable_on_solo {
+            self.disable_on_solo = disable_on_solo.0;
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


breitnw:
+1 on this! I think it would be a useful feature, especially for folks who don't like the focus ring

nerdwave-nick:
Could maybe create an enhancement issue first, to see if this is even a feature niri should/wants to implement, and what the specification for that feature would be. 

Referencing the first point here https://github.com/YaLTeR/niri/blob/main/CONTRIBUTING.md#writing-pull-requests

Fireye04:
I don't necessarily see the problem with jumping straight into a PR as long as you're willing to rewrite or drop your specific implementation if it's not approved. Helps keep things moving forward :)

nerdwave-nick:
> I don't necessarily see the problem with jumping straight into a PR as long as you're willing to rewrite or drop your specific implementation if it's not approved. Helps keep things moving forward :)

Slightly off-topic now, so this'll be my last message on this, but I believe there could be value in straightening out the specs of what we actually want here, so I do see value in opening an issue up in advance. It not being a PR but an issue might make it easier (discoverability) for non-technical people to also share their mind on it. But ultimately yeah, fair enough 😄

perstarkse:
Interesting to hear your perspectives. Considering the discussion I linked to, there seemed to be some interest. Rejecting the PR is totally fine if it doesn’t align with the project. 

Re the specific code implementation I’m not particularly opinionated, this is somewhat outside the scope of what I typically do. So definitely interested in improving if there are suggestions

nerdwave-nick:
> Interesting to hear your perspectives. Considering the discussion... 

Fair enough. You saw the request and did the work.
I do not feel comfortable enough to review this change, as I am not nearly familiar enough with the code base. But I do want to endorse it. Sounds like a fun plan and creates the basic infra for this. Adding different animations shouldn't be too crazy after.

sashaduke:
So far the only thing I've not liked about niri (coming from hyprland) is how when changing focus between windows, the opacity snaps from translucent to opaque or vice versa. Would be keen to contribute by either extending this PR or making a new one to implement a smooth opacity transition as a configurable parameter

perstarkse:
> y thing I've not liked about niri (coming from hyprland) is how when changing focus between windows, the opacity snaps from translucent to opaque or vice versa. Would be keen to contribute by either extending this PR or making a new one to implement a smooth opacity transition as a configurable

I can take a look at that. For anyone potentially reviewing. I identified a couple of issues that needs resorting so changing status to draft. Hope that's fine

perstarkse:
> So far the only thing I've not liked about niri (coming from hyprland) is how when changing focus between windows, the opacity snaps from translucent to opaque or vice versa. Would be keen to contribute by either extending this PR or making a new one to implement a smooth opacity transition as a configurable parameter

I've looked at this and if I understand you correctly, I belive a solution for flash opacity would share a lot with your case. For testing, could you please share relevant aspects of your configuration for testing?

PR #2821, opened 2025-11-16
CI: free disk space before running nix flake check

Ind-E:
This prevents a `No space left on device` error as can be seen here: https://github.com/YaLTeR/niri/actions/runs/19403781415/job/55515251373#step:5:3659

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,6 +287,10 @@ jobs:
     steps:
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@v1.3.1
+        with:
+          dotnet: false
+          large-packages: false
+        
       - uses: actions/checkout@v4
         with:
           show-progress: false


Ind-E:
you can see a successful run here: https://github.com/Ind-E/niri/actions/runs/19412175230

YaLTeR:
Thanks. Though, maybe something can be done faster? Nix was already the slowest CI job, now it's even slower, that's a bit unfortunate:

image

HigherOrderLogic:
> Thanks. Though, maybe something can be done faster? 

The Nix flake check is just simply building in debug and run the tests I believe. Maybe we can simply just run `cargo check` or `cargo clippy` instead of running a full build?

Ind-E:
I experimented with using niri-flake's binary cache as a proof-of-concept, but that didn't even seem to speed things up very much. To be honest I'm not sure what exactly we can do. I'll try messing with some of the settings for the free-disk-space action

Ind-E:
ok, I changed some settings on the free-disk-space action and it now runs in ~1 minute instead of ~2 1/2 minutes. I think that's the best I can do for now

YaLTeR:
Alright, thanks for testing various options!

PR #2823, opened 2025-11-17
feat: add xwayland matcher for window rule

HigherOrderLogic:
Add xwayland matcher for window rules.

This requires user to use the built-in Xwayland-satellite integration.

~~TODO: add docs~~

### DIFF
--- a/src/utils/xwayland/satellite.rs
+++ b/src/utils/xwayland/satellite.rs
@@ -19,10 +19,18 @@ pub struct Satellite {
     abstract_token: Option<RegistrationToken>,
     unix_token: Option<RegistrationToken>,
     to_main: Sender<ToMain>,
+    pid: Option<u32>,
+}
+
+impl Satellite {
+    pub fn pid(&self) -> Option<u32> {
+        self.pid
+    }
 }
 
 enum ToMain {
     SetupWatch,
+    Pid(Option<u32>),
 }
 
 impl Satellite {
@@ -61,6 +69,11 @@ pub fn setup(state: &mut State) {
         .insert_source(rx, move |event, _, state| match event {
             calloop::channel::Event::Msg(msg) => match msg {
                 ToMain::SetupWatch => setup_watch(state),
+                ToMain::Pid(pid) => {
+                    if let Some(s) = state.niri.satellite.as_mut() {
+                        s.pid = pid;
+                    }
+                }
             },
             calloop::channel::Event::Closed => (),
         })
@@ -71,6 +84,7 @@ pub fn setup(state: &mut State) {
         abstract_token: None,
         unix_token: None,
         to_main,
+        pid: None,
     });
 
     setup_watch(state);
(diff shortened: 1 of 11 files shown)


src/window/mapped.rs
```diff
@@ -93,6 +93,9 @@ pub struct Mapped {
     /// Whether this window is floating.
     is_floating: bool,
 
+    /// Whether this windows in running under Xwayland.
```
YaLTeR: england

src/window/mod.rs
```diff
@@ -171,6 +171,13 @@ impl<'a> WindowRef<'a> {
             WindowRef::Mapped(mapped) => mapped.is_window_cast_target(),
         }
     }
+
+    pub fn is_xwayland(self) -> bool {
+        match self {
+            WindowRef::Unmapped(_) => false,
```
YaLTeR: Would be good for this to work. Which I guess means checking it already when creating the unmapped.

Actually, here's an idea: since the PID and is_xwayland is more of a Client property, it could be checked right upon a new Client connection and then stored somewhere in ClientState.

src/utils/xwayland/satellite.rs
```diff
@@ -307,6 +322,8 @@ fn spawn_and_wait(
         }
     };
 
+    let _ = to_main.send(ToMain::Pid(Some(child.id())));
```
YaLTeR, line 325: Another possible way to do it without PIDs is to create a separate Wayland socket for our xwayland-satellite, similarly to how we create a separate Wayland connection for sandboxed clients. Not sure how much effort that would be.

src/window/mod.rs
HigherOrderLogic: I tried but 1. you only get the client after you insert and 2. the data returned from `get_data` is not mutable so you cant edit `ClientState` after inserting. So I guess if you want this to work then it will need some very big refactoring.

src/utils/xwayland/satellite.rs
HigherOrderLogic: I afraid that'd be out of my current understanding of the Niri codebase.

src/window/mod.rs
YaLTeR: Well, `Cell<bool>` is always a thing

HigherOrderLogic: `Cell` doesnt implement `Sync`, which is required by `ClientData`.

YaLTeR: AtomicBool then, or ideally get credentials before inserting with whatever Smithay or [this](https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html#method.peer_cred) calls under the hood

YaLTeR:
Another question: what is this useful for?

HigherOrderLogic:
>what is this useful for?

Some Wine/Proton game (not necessarily Steam) doesnt open in full screen and cant be resized right away, which seems like a satellite bug. Thus I want this matcher to open it in full screen.

This is also useful to identify Xwayland windows without installing 3rd-party app like Xeyes.

YaLTeR:
> Some Wine/Proton game (not necessarily Steam) doesnt open in full screen and cant be resized right away, which seems like a satellite bug. Thus I want this matcher to open it in full screen.

But you don't want to blanket-match all xwayland windows for this?

HigherOrderLogic:
> But you don't want to blanket-match all xwayland windows for this?

Yeah, I'm also combining with other matchers too.

YaLTeR:
Why do you need the xwayland matcher for this specifically?

HigherOrderLogic:
Just to make sure it doesnt include normal Wayland windows.

Maybe a bit paranoid from my side.

PR #2825, opened 2025-11-17
Add a cap of 100% when changing audio volume

whiskeyPeak:
There is currently no cap on the volume which I unfortunately found out the hard way haha.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -374,7 +374,8 @@ binds {
     // Example volume keys mappings for PipeWire & WirePlumber.
     // The allow-when-locked=true property makes them work even when the session is locked.
     // Using spawn-sh allows to pass multiple arguments together with the command.
-    XF86AudioRaiseVolume allow-when-locked=true { spawn-sh "wpctl set-volume @DEFAULT_AUDIO_SINK@ 0.1+"; }
+    // "-l 1.0" limits the volume to 100%.
+    XF86AudioRaiseVolume allow-when-locked=true { spawn-sh "wpctl set-volume @DEFAULT_AUDIO_SINK@ 0.1+ -l 1.0"; }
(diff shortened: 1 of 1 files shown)


HigherOrderLogic:
No volume cap is intentional.

YaLTeR:
Is it supposed to be capped here in this command?

whiskeyPeak:
@YaLTeR I mean, you could add a policy to wireplumber but this way the cap is part of the config and can be customised.

@HigherOrderLogic Seems like an odd default. All DEs I have used have a cap.

HigherOrderLogic:
> Seems like an odd default. All DEs I have used have a cap.

Niri is not a DE. And the default config is just meant to be in an usable state, not trying to adapt to everyone's setup.

Also you should use a 3rd party program to cap this, it isnt really Niri's job here.

YaLTeR:
Do any other example configs like this cap it?

YaLTeR:
I think it's ok to add this, just not something I ever hit as a problem somehow, so I'm trying to learn more about it

Fireye04:
I think I'm largely with logic here, as most WMs leave this to the audio server afaik, but I dont hate the idea of modifying the default config to give newer users a more seamless experience.

whiskeyPeak:
Fair enough, I'll close this then.

neunato:
I actually think it makes sense to add this to the default config, as most users don't want to go above 100%.

> Is it supposed to be capped here in this command?

Yes it is.

> Do any other example configs like this cap it?

Here it is in [Hyprland example config](https://github.com/hyprwm/Hyprland/blob/b5a2ef77b7876798d33502f8de006f9c478c12db/example/hyprland.conf#L290).

YaLTeR:
Alright, makes sense

YaLTeR:
Thanks

PR #2826, opened 2025-11-17
Delay block-out release until content settles

### LINKED ISSUE #1348: Blocked out Window content visible on tabbing
Using a match rule, I'm blocking out firefox if it has a certain tab open. The blocking out works normally on switching to said tab, but when leaving it, the content is briefly visible.

https://github.com/user-attachments/assets/ba67de22-b358-4314-9108-d3f76fced2b2

### System Information

* niri version: niri stable v25.02 (commit b94a5db8790339cf9134873d8b490be69e02ac71)

* Distro: NixOS Unstable [(It's complicated, though)](https://forge.katzen.cafe/schrottkatze/nix-configs/src/commit/2de0639d54e3c24a813e432cfa25eb4d8302cd58/flake.nix#L5-L8)

* GPU: AMD RX 6700 XT

* CPU: AMD Ryzen 9 5950X

Acters:
In an attempt to reduce chances of leaked private details, the block-out-from has been hardened to intentionally delay the unblocking until subsurface draws reported by wayland compositor stops, or a hard cap timer is reached.

- Track block-out-from transition to unblock per window via the BlockOutHoldAction, so we know when a hold starts or clears.
 - Add per-window silence and force timers (30 ms debounce + 200 ms failsafe) to keep screencasts blocked until
 subsurface commits settle, resetting timers on each commit and cancelling them on unmap.
 - Extend the layout/Niri plumbing to locate windows by MappedId, manage timer lifecycles, and finish holds when
 timers fire.
 - Update compositor/xdg-shell handlers to hook into the new timer APIs and avoid borrowing conflicts while
 recomputing window rules.

Unfortunately, I am new to rust and had the help of AI to make these changes. 
Overall, it does correctly delay the unblocking of the window to allow the window to finish committing all the rendering changes.

fixes #1348


src/window/mapped.rs
```diff
@@ -1294,6 +1387,10 @@ impl LayoutElement for Mapped {
         &self.rules
     }
 
+    fn block_out_from(&self) -> Option<BlockOutFrom> {
+        self.effective_block_out_from()
```
EvansJahja: This pattern of just calling another method in `self` is confusing, it will raise confusion on the difference between `block_out_from` and `effective_block_out_from`, just consolidate them into one function

src/handlers/compositor.rs
```diff
@@ -257,14 +257,26 @@ impl CompositorHandler for State {
...
-                let window = mapped.window.clone();
-                let output = output.cloned();
-
-                let id = mapped.id();
-
+            let surface_was_mapped = is_mapped(surface);
+            if let Some((window, output, id, hold_active)) = self
+                .niri
+                .layout
+                .find_window_and_output_mut(surface)
```
EvansJahja: We're taking a mutable borrow just because we want to eventually return a boolean `hold_active`, I think this can be simplified by moving the call

src/handlers/xdg_shell.rs
```diff
@@ -1368,33 +1369,56 @@ impl State {
...
-                *rules = new_rules;
-            }
-        } else if let Some((mapped, output)) = self
-            .niri
-            .layout
-            .find_window_and_output_mut(toplevel.wl_surface())
+    pub fn update_window_rules(
+        &mut self,
+        toplevel: &ToplevelSurface,
+        delay_block_out_release: bool,
```
EvansJahja: This `delay_block_out_release` being a bool parameter makes it hard to read since it's not obvious what passing `true` means. Moreover, seems like all function call of this uses `true` anyway

src/window/mapped.rs
Acters: Hey, thank you for reviewing it. I "consolidated" this, so there is now a single `block_out_from()` path instead of a wrapper. Some of those weird AI things that I try to spot out, but I missed this one.

src/handlers/xdg_shell.rs
Acters: Yeah, that boolean parameter is another weird hiccup from gpt. I removed it from `update_window_rules()`
appreciate your review, haha

src/handlers/compositor.rs
Acters: I modified this from `find_window_and_output_mut()` to `find_window_and_output()`

Acters:
Closing as not planned.

phisch:
I'd have kept it open. I think it's a good idea still. It's just that the maintainer is busy with other things currently.

Acters:
I think it's a good idea but I wasn't really happy with it as it could probably been done better. I might get around to making a different PR that I like.

PR #2827, opened 2025-11-17
Global and output-specfied RTL-support

veighnsche:
https://github.com/YaLTeR/niri/issues/861

Status:
Done:
- when empty workspace. new tile starts at right wall.
- additional tiles spawn on the left side of active tile.
- some tests
- ipc
- when `switch-preset-column-width` then the left side of the tile should move.
- sanity tests ensuring **physical** semantics ("is left still left and right still right?" Read: Semantics A & B)

Doing:
- animation for `switch-preset-column-width`

Todo:
- correct animations

Semantics:
A) **Physical: Left Right** is the default semantics in development space.
There are 2 (3) main ways to define behavior when it comes to layout:
- **Physical**: Left, Right; (sane for development)
- **Logical**: Next, Previous, Before, After, Start, End; (intuitive for users) (more: Leading, Trailing)
- (**Array location**): N-spot in array in some _special_ locations.. (I'm trying to avoid this)

@YaLTeR said in [issue 861](https://github.com/YaLTeR/niri/issues/861) - near the end - that the logical semantics could be a new feature request. Out of simplicity - I have kept **physical** semantics for all of the RTL support. However. Logical semantics comes up naturally and requires effort to sanity test physicality.. (Example: Semantics B)
So I can make the argument that logical semantics has UX benefits.

B) **Leading edge**: There are some logical semantics that are better than using physical semantics. In animation the **Leading edge** points to the logical **"end"** of the workspace. And when pressing Mod+R it's always the Trailing edge that we have to "pin" (i know it's not _really pinning_ but it's some trick with the camera's view in RTL) And the Leading edge that has to move. 

Blockers:
- default direction sourced from language based on Gnome impl (I saw @YaLTeR still discussing this in the issue)
- ~~Home and End implementation (I think a native RTL should make the requirements for this, ~~I'm prone to overthinking~~, I'm starting to realize that `home == right` and `end == left` is the only thing that makes sense.)~~

### DIFF
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -2596,17 +2596,9 @@ screen_left_after={screen_left_after} screen_right_after={screen_right_after}",
         // Band origin: start of the column band in world space.
         let band_origin_x = match self.dir() {
             LayoutDirection::Ltr => {
-                // For LTR, when band nearly fills viewport, adjust origin so rightmost column
-                // right edge is at viewport width
-                let view_width = self.view_size.w;
-                // Only adjust when band is close to filling viewport (> 90%)
-                if _total_band_width < view_width && _total_band_width > view_width * 0.9 && data.len() >= 2 {
-                    // Rightmost column's right edge should be at view_width
-                    // band_origin_x + total_band_width = view_width
-                    view_width - _total_band_width
-                } else {
-                    0.0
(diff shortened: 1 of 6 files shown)


veighnsche:
Different approach in another branch

PR #2836, opened 2025-11-18
Add windows' fullscreen states to ipc

Sem1rose:
Solves [#discussion:1843](https://github.com/YaLTeR/niri/discussions/1843).

Adds a `fullscreen_state` to the Window struct, and a `WindowFullscreenStateChanged` event.


src/ipc/server.rs
```diff
@@ -504,6 +504,9 @@ fn make_ipc_window(
     workspace_id: Option<WorkspaceId>,
     layout: WindowLayout,
 ) -> niri_ipc::Window {
+    // Quick hack to fix the consistent halting if we tried to access the sizing mode
+    // inside the constructor
+    let fullscreen_state = mapped.sizing_mode().into();
```
nerdwave-nick: Will the "quick hack" stay? Usually a hack is not really a good thing for a project, although I don't have insight if this is avoidable or not

niri-ipc/src/lib.rs
```diff
@@ -1532,6 +1538,16 @@ pub enum Event {
...
+    /// A window's fullscreen state was changed changed.
+    WindowFullscreenStateChanged {
+        /// Id of the window.
+        id: u64,
+        /// Fullscreen state for the window
+        /// - `0`: not fullscreen (tiled or floating)
+        /// - `1`: maximized
+        /// - `2`: fullscreen
+        fullscreen_state: u8,
+    },
```
nerdwave-nick: Have you considered using `sizing_mode: SizingMode` instead? This could get rid of one translation (fullscreen_state -> sizing_mode), which can help with general reasoning about this code. Would also save on the comments above. Something like

```suggestion
    /// A window's sizing mode was changed changed.
    WindowSizingModeChanged {
        /// Id of the window.
        id: u64,
        sizing_mode: SizingMode,
    },
```
This would of course go with changes in the rest of the PR, but if there is no good reason to do the sizing_mode -> fullscreen_state translation, this would help with the mental indirection, and make things more obvious, which is always good in a large (or any) code base.

src/layout/mod.rs
```diff
@@ -526,6 +526,27 @@ impl SizingMode {
...
+impl From<SizingMode> for u8 {
+    fn from(value: SizingMode) -> Self {
+        match value {
+            SizingMode::Normal => 0,
+            SizingMode::Maximized => 1,
+            SizingMode::Fullscreen => 2,
+        }
+    }
+}
+
```
nerdwave-nick, line 556: Instead of this, the SizingMode could instead be a unit only enum, which saves on these functions, and goes hand in hand with the idea I proposed to change it to sizing_mode instead of fullscreen_state

nerdwave-nick (commented):
Posting this with the disclaimer that I barely know rust and this is my first look ever at the niri code base, but I think the simplification I have proposed could be a general improvement for clarity and simplicity. Unless, of course, there are good reasons to do it the way it is. 

Thank you for being my experiment/first attempt at reading the niri code 😄

src/ipc/server.rs
Sem1rose: well, it's not a hack per se. the `mapped.sizing_mode().into();` was supposed to go inside the constructor below, but i believe `mapped` gets locked inside the `with_toplevel_role` method, so when i try and access it inside the same funciton a deadlock occurs. i could be totally wrong here, i'm not sure if this is intended of `with_toplevel_role`, but saving the `fullscreen_state` beforehand avoids any halts.

niri-ipc/src/lib.rs
Sem1rose: Yeah, i actually considered this when creating the pr, but i instead decided to follow Hyprland's implementation. 
> this would help with the mental indirection, and make things more obvious

I agree, this would also eliminate some useless conversions between `u8` and `SizingMode` that only gets used once.

Sem1rose: Small issue is that `SizingMode` is defined inside the `niri` crate and is therefore inaccessable inside `niri_ipc`. it's possible to move the definition and implementations from `niri` to `niri_ipc`, but this would be a breaking change and it's just stupid to refactor the entire project for a single featture.

src/ipc/server.rs
nerdwave-nick: Hmmm... It's not necessary for any of the other values. That looks weird to me, but perhaps I'm just not knowledgeable enough regarding the code

niri-ipc/src/lib.rs
nerdwave-nick: Flexing my missing rust knowledge - could you serialzie the SizingMode as is? Who cares if it comes out as a lil string or number. But idk how rust is about this stuff

Viruzaum:
Does this also adds a match for window rules?

LucasOe:
> Does this also adds a match for window rules?

No. But I would be interested in adding support for `is-fullscreen` and `is-maximized` match rules based on this PR once it's merged.

Sem1rose:
I just added a window rules match for the sizing mode per @Viruzaum 's request, did some cleanup and removed the "quick hack". 

Also renamed the property and match to `sizing_mode` but i'm not sure if it's fitting. suggestions are welcome.

I'm also considering using an enum instead of an integer for the IPC and the window rules match, as per @nerdwave-nick 's suggestion, to improve UX. maybe in a future commit.
other than that i'm confident the pr is feature complete.

kanghengliu:
Would like to bump on this for [a DMS feature](https://github.com/AvengeMedia/DankMaterialShell/pull/2091). It's much easier to control the layer-shell layer's behavior with detailed window state metadata. 

In this case we'd like to determine if we want to hide the bar based on the current windows's fullscreen state: fullscreen(yes), maximized-fullscreen(no), default(no). 

Current implementation detects with Quickshell.wayland, the window is reported as fullscreen by wayland protocol when it's in maximized-fullscreen state. With this PR, we will be able to get such info via ipc.

YaLTeR:
I really don't think you should be hiding layer-shell based on a window state from IPC. It cannot work well. Let niri deal with it, fullscreen windows already render over the top layer, as is mentioned in the linked DMS PR.

kanghengliu:
@YaLTeR Good point. The reason for this feature stems from the window not occluding the top level bar when animating/transitioning to full screen. So when the fullscreen animation finishes, the bar will be immediately occluded to background without transitions.

YaLTeR:
The correct place to fix this is in niri and I will do it at some point, it's more complicated than it sounds

PR #2843, opened 2025-11-19
Maintain a single empty workspace at the edges

LoayGhreeb:
This PR ensures that workspaces maintain only one empty unnamed workspace at the end (or at both edges if `empty_workspace_above_first` is enabled).

A simple scenario: if there is only one workspace and a window is opened (which creates a new workspace at the end), then after closing that window, the last workspace should be removed. This makes the workspace state consistent: before opening a window and after closing it, the number of workspaces remains the same.

Similarly, when `empty_workspace_above_first` is enabled, the same behavior applies to the first workspace. So, if there is only one workspace, opening a window will create leading and trailing workspaces. After closing the window, it should return to having just one workspace, instead of three.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -1134,6 +1134,32 @@ impl<W: LayoutElement> Layout<W> {
                                 mon.workspaces.remove(1);
                                 mon.active_workspace_idx = 0;
                             }
+
+                            // If the last two workspaces are empty unnamed, remove the last one.
+                            let len = mon.workspaces.len();
+                            if len >= 2
+                                // FIXME: Should `!ws.has_windows_or_name()` be checked here
+                                // or it will be redundant?
+                                && !mon.workspaces[len - 1].has_windows_or_name()
+                                && !mon.workspaces[len - 2].has_windows_or_name()
+                                && mon.active_workspace_idx != len - 1
+                                && mon.workspace_switch.is_none()
+                            {
+                                mon.workspaces.remove(len - 1);
+                            }
+
+                            // If empty_workspace_above_first is set and the first two workspaces are
+                            // empty unnamed, remove the first one.
+                            if mon.options.layout.empty_workspace_above_first
+                                && mon.workspaces.len() >= 2
+                                && !mon.workspaces[0].has_windows_or_name()
+                                && !mon.workspaces[1].has_windows_or_name()
+                                && mon.active_workspace_idx != 0
+                                && mon.workspace_switch.is_none()
+                            {
+                                mon.workspaces.remove(0);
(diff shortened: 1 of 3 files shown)


ccjmne:
I can't quite make sense of this from your description.

- Isn't this automatic pruning of empty workspaces already done by niri?
- You state that opening a new window creates a new workspace: I haven't experienced that either.
Are you maybe using named workspaces? In which case, these are indeed meant to perdure even when empty.

Sempyos:
I feel the current behavior is intended, as we follow Gnome's dynamic workspaces, which will always have 2 workspaces loaded regardless of having a window in workspace 1

LoayGhreeb:
> Isn't this automatic pruning of empty workspaces already done by niri?

Yes, it is, and this PR handles a case while cleaning up workspaces. If you open a window in the last (empty) workspace and then close it, you'll notice that the last workspace is not removed.

> You state that opening a new window creates a new workspace: I haven't experienced that either.

Oh, wait this is the default behavior, which I don't think can be overridden. See: https://github.com/YaLTeR/niri/wiki/Workspaces#overview
> There's always one empty workspace at the end (at the bottom) of every monitor. When you open a window on this empty workspace, a new empty workspace will immediately appear further below it.

> Are you maybe using named workspaces? In which case, these are indeed meant to perdure even when empty.

I don't use any named workspaces.

For clarity, here's a comparison:
Main branch:
https://github.com/user-attachments/assets/b4afc74c-f1de-4bd2-a62c-2dc996db819b

PR:
https://github.com/user-attachments/assets/2fb74968-8acf-4486-98ef-e3ed956d8004

LoayGhreeb:
> I feel the current behavior is intended, as we follow Gnome's dynamic workspaces, which will always have 2 workspaces loaded regardless of having a window in workspace 1

I'm not sure whether this behavior is intended or not, but I've used the Cosmic compositor, and its dynamic workspaces behave the same way as in this PR.

ccjmne:
Ah, I misunderstood at first, it makes sense now.
Yeah, I suppose that niri never deletes a workspace that's currently active... I'm pretty sure it's very intentional, but it does look somewhat jumpy in cases like the one you're highlighting.

zgibberish:
Can you rebase this onto upstream main? I couldn't get the old version to start.

I tested manually and seems to work well.

Except for when moving window/column using keybinds it doesn't trigger workspace cleanup. But works fine when moving windows using the mouse.

Tests are failing, I keep getting this:

```text
thread 'layout::tests::random_operations_dont_panic' (225007) panicked at src/layout/monitor.rs:2212:13:
when empty_workspace_above_first is set, the first two workspaces cannot both be empty and unnamed
```

Issue #2393, opened 2025-09-15
scroll-factor does not affect on-button-down scrolling


ArijanJ:
Hi, I have a keyboard with a built-in trackball that I use to move the mouse and scroll. The scrolling is done `on-button-down`, but niri does not seem to care about the `scroll-factor` value when doing this.

This is what I currently have versus the sway equivalent:
```kdl
mouse {
    scroll-method "on-button-down"
    scroll-button 279
    natural-scroll
    scroll-factor 0.01
}
```
```kdl
input "43256:6196:Bastard_Keyboards_Charybdis_Mini_(3x6)_Splinky_Mouse" {
    scroll_factor 0.25
    natural_scroll on
    scroll_method on_button_down
    scroll_button BTN_TASK
    rotation_angle 0
}
```
In niri as opposed to sway, the scrolling is very fast no matter what I change the value to. I can't put `scroll-factor` in a 'trackball' block, niri does not allow it (but this specific device does not get picked up as a trackball either way). I also can't select the device by name, but that already has an open issue and probably doesn't matter here.

### System Information

* niri version: niri 25.08 (01be0e6)

* Distro: Arch

* GPU: N/A

* CPU: N/A

ArijanJ:
Oh, I also can't do Mod+middle drag to move the view with this method, let me know if I should open another issue for that

the-burrito-triangle:
I think I might have the same issue with a Logitech MX 2 Master mouse with "high resolution" scroll wheel. While I can configure my `scroll-factor` and it works for app windows, Niri layer-shell surfaces like fuzzel and waybar do not currently honor the `scroll-factor` value set under `input` and are difficult to control using the "high resolution" scroll wheel (too many events per detent making the selection move too fast).

YaLTeR:
> Niri layer-shell surfaces like fuzzel and waybar do not currently honor the `scroll-factor` value set under `input` and are difficult to control using the "high resolution" scroll wheel (too many events per detent making the selection move too fast).

Sounds like these applications are not handling scroll events correctly

PR #2844, opened 2025-11-19
Scroll factor configuration for `scroll-method: "on-button-down"`

### LINKED ISSUE #2841: `scroll-factor` does not apply to trackball devices
The `trackball` section of the input config does not accept `scroll-factor` like the `mouse` section does, thus preventing users of trackball mice from adjusting `on-button-down` scroll sensitivity.

In my case, I use a Ploopy Adept trackball mouse, and need to apply the following libinput settings for comfortable control:
```
sensitivity = -0.75;
accel_profile = "flat";
scroll_method = "on_button_down";
scroll_button = 277;
scroll_factor = 0.25
```

This works under hyprland, where `scroll_factor` can be applied from the top-level `input` section. But as niri does not offer a global or trackball-local setting, the result is scrolling being ~4x more sensitive than desired.

Hopefully this can be worked around at the system level with the appropriate libinput config files, but I'm still investigating that side of things.

Related: #2393

### System Information

* niri version: niri stable v25.08 (commit 01be0e65f4eb91a9cd624ac0b76aaeab765c7294)

* Distro: NixOS 25.05

* GPU: Radeon RX 6900XT

* CPU: Ryzen 9 5950X

ProfLander:
Implements `input.button-scroll-factor` using the existing `ScrollFactor` type, and applies it to events of type `AxisSource::Continuous`.

I've followed the hyprland pattern of having the `input` -level setting govern button-style scroll speed so as to avoid per-device clutter or collision with existing `scroll-factor` settings, and added a clarifying `button-` prefix to ensure users don't look at it and assume a global modifier, rather than an event-specific one.

Closes #2841, #2393

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2257,6 +2257,7 @@ dependencies = [
  "tracy-client",
  "wayland-backend",
  "wayland-client",
+ "wayland-protocols-plasma",
  "wayland-scanner",
  "wayland-server",
  "xcursor",
@@ -4218,6 +4219,7 @@ dependencies = [
  "wayland-client",
  "wayland-protocols",
  "wayland-scanner",
+ "wayland-server",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3476,7 +3476,7 @@ checksum = "b7c388c1b5e93756d0c740965c41e8822f866621d41acbdf6336a6a168f8840c"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/Smithay/smithay.git#f1d0d2c43e843233119af3dcf76d619e1afff4ee"
+source = "git+https://github.com/Smithay/smithay.git?rev=dce4d34e7421559b661af9c519904f4b24346148#dce4d34e7421559b661af9c519904f4b24346148"
 dependencies = [
  "aliasable",
  "appendlist",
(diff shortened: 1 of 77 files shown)


YaLTeR:
Maybe it should just follow the normal scroll-factor somehow? In sway there doesn't seem to be a separate button scroll factor. Besides, with specific input device config, there will be just a device-specific `scroll-factor` too.

ProfLander:
There is no 'normal' scroll factor as-is; `input.mouse.scroll-factor` modifies `AxisSource::Wheel` events, and `input.trackpad.scroll-factor` modifies `AxisSource::Finger` events. Trackball and trackpoint have no hardware-specific equivalent, so are assumed to emulate via `on-button-down`, which emits `AxisSource::Continuous`.

Since those are semantically distinct, and all types of pointer can use button scrolling, they need their own sensitivity modifiers to avoid coupling. For example: A mouse's line-based wheel sensitivity shouldn't be tied to its sensor-based button-scroll sensitivity, in the case where both are active.

Hyprland's solution is to treat button scrolling as a singleton like I have here, but that does miss the idea that button scrolling is DPI-relative; more exotic setups with multiple pointing devices could benefit from `button-scroll-sensitivity` being a device -level setting instead of a global one.

YaLTeR:
Hm well I don't see another way before some bigger refactoring so let's go with this. Do you mind adding it to the wiki page, not forgetting the Since: next release annotation?

PR #2852, opened 2025-11-20
Refer workspace by id

oisanjaya:
This PR enable workspace reference by ID for operations on `niri msg action`.

My idea was using `id:` prefix to refer workspace by its id instead of index, this is done in niri-ipc/src/lib.rs from_str trait implementation, that is checks user arguments; if it is integer number then assume user refer to index (to maintain backward compatibility of current operations); if it is prefixed by `id:` then assume user refering workspace's id; else assume name reference.

This PR also change `niri msg workspaces` command output to include workspace's id.


niri-ipc/src/lib.rs
```diff
@@ -1616,7 +1616,15 @@ impl FromStr for WorkspaceReferenceArg {
                 return Err("workspace index must be between 0 and 255");
             }
         } else {
-            Self::Name(s.to_string())
+            if s.starts_with("id:") {
```
helcoume: Using `s.strip_prefix("id:")` might be better here since it avoids slicing by index (s[3..])

Reduce nesting by splitting `Index` and `Id` cases into two separate `if` block could also be nice, but that’s probably a detail

oisanjaya: Agreed on strip_prefix, but keep nested `if` for now as it would result fewer checks, let's see how others opinion.

munich91:
so,If a workspace name is "id:1", will it also match by ID?

oisanjaya:
A workspace named 'id:1' will have its own id. That workspace will not be able to referred by its name but still can be referred by its id (which is not always '1')

PR #2857, opened 2025-11-20
Remove window comparison

### LINKED ISSUE #2856: Focus not triggered after closing overview
Focus is not being triggered for windows after the overview has been closed. 

https://github.com/user-attachments/assets/e725b56c-4505-4dab-a8c8-b73d29b07022

I have `focus-follows-mouse` on and expect the window the cursor is above to gain focus again after moving. This issue also occurs when the cursor is on top of an old window when a new one is launched.

The issue lies with `if !self.layout.is_overview_open() && current_focus.window.as_ref() != Some(window)` in niri.rs 6283. We queried the current `PointContents` with `self.contents_under` which ends up being the same as the `new_focus` var.

whiskeyPeak:
The issue lies with if !self.layout.is_overview_open() && current_focus.window.as_ref() != Some(window) in niri.rs 6283. We queried the current PointContents with self.contents_under which ends up being the same as the new_focus var.
Following video shows the fix.

https://github.com/user-attachments/assets/69c58daf-9b8c-4756-bd27-212e9bf0b782

Fixes #2856

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -6275,7 +6275,7 @@ impl Niri {
         }
 
         if let Some(window) = &new_focus.window {
-            if !self.layout.is_overview_open() && current_focus.window.as_ref() != Some(window) {
+            if !self.layout.is_overview_open() {
                 let (window, hit) = window;
 
                 // Don't trigger focus-follows-mouse over the tab indicator.


YaLTeR:
This check is there intentionally. Focus follows mouse should only trigger when a pointer motion moves into a new window, not when it's moving inside an unfocused window, which this PR would change. Regarding your issue, I'll comment there.

PR #2860, opened 2025-11-21
chore(deps): update various

Boosted-Bonobo:
Why wasn't this used instead? @Ind-E https://github.com/YaLTeR/niri/pull/2821
https://wimpysworld.com/posts/nothing-but-nix-github-actions
the space reclaimer which is used in nixpkgs through [nixpkgs-review-gha](https://github.com/Defelo/nixpkgs-review-gha/blob/main/.github/actions/setup-nix/action.yml#L19)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,9 +238,7 @@ jobs:
           show-progress: false
 
       - name: Install dependencies
-        run: |
-          sudo dnf5 update -y
-          sudo dnf5 install -y ${{ env.DEPS_DNF }} libadwaita-devel
+        run: sudo dnf5 install -y ${{ env.DEPS_DNF }} libadwaita-devel
 
       - uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
       - run: cargo build --all
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           show-progress: false
 
@@ -31,9 +31,9 @@ jobs:
           sudo apt-get update -y
           sudo apt-get install -y ${{ env.DEPS_APT }}
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@6d9817901c499d6b02debbb57edb38d33daa680b # stable
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
 
       - name: Build tests
         run: cargo test --no-run --all --exclude niri-visual-tests
@@ -49,7 +49,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           show-progress: false
 
@@ -58,9 +58,9 @@ jobs:
           sudo apt-get update -y
           sudo apt-get install -y ${{ env.DEPS_APT }}
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@6d9817901c499d6b02debbb57edb38d33daa680b # stable
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
 
       - name: Check (no default features)
         run: cargo check --no-default-features
@@ -91,14 +91,14 @@ jobs:
     runs-on: ubuntu-24.04
     container: alpine:3
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           show-progress: false
 
       - name: Install Deps
         run: apk add --no-cache ${{ env.DEPS_APK }}
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
 
       - name: Build
         run: cargo build --no-default-features --features dbus,xdp-gnome-screencast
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
```
YaLTeR: I don't like this GHA commit pinning, makes it a pain to update them

.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
     name: alpine musl
     runs-on: ubuntu-24.04
-    container: alpine:3
+    container: alpine:3.22.2@sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412 # https://hub.docker.com/layers/library/alpine/3.22.2/images/sha256-9eec16c5eada75150a82666ba0ad6df76b164a6f8582ba5cb964c0813fa56625
```
YaLTeR, line 92: Same here about pinning

.github/workflows/ci.yml
YaLTeR, line 1: This commit is not necessary I think

.github/workflows/ci.yml
```diff
@@ -238,9 +238,7 @@ jobs:
           show-progress: false
 
       - name: Install dependencies
-        run: |
-          sudo dnf5 update -y
-          sudo dnf5 install -y ${{ env.DEPS_DNF }} libadwaita-devel
+        run: sudo dnf5 install -y ${{ env.DEPS_DNF }} libadwaita-devel
```
YaLTeR, line 241: Can't this cause newer libadwaita-devel with stale other deps?

YaLTeR (changes requested):

.github/workflows/ci.yml
Boosted-Bonobo: When it installs the packages, it updates the repositories to do so.
image
Dnf5 should be smart enough to update everything it needs since not even `don't install weak dependencies` is turned on.

.github/workflows/ci.yml
Boosted-Bonobo: It shouldn't be a pain to update.
https://github.com/NixOS/nixpkgs/pull/458072/files
An automatic way to update them should be set, just like in nixpkgs, don't you think?

.github/workflows/ci.yml
Boosted-Bonobo: Without pinning, tags can change, especially until dockerhub adds a way to make tags/releases [immutable](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/immutable-releases) like github has and you won't have a way to know that the packages used haven't changed since the last CI run.

.github/workflows/ci.yml
Boosted-Bonobo: You are right, it is not necessary, but it's a good addition.
Even today you can use `yum install ...`, but at the same time it's a symlink to the new dnf.
Same with apt. So why not use the today's way instead of using symlinks for backwards compatibility?

YaLTeR: - `dnf` is the standard command everyone runs on Fedora. Nobody runs dnf5 by hand
- last time I checked `apt install` still had a scary warning that its CLI is not stable, and similarly everyone uses `apt-get` in CI. I think it's just safer to do the boring ones

.github/workflows/ci.yml
YaLTeR: I understand but it's not a problem

.github/workflows/ci.yml
YaLTeR: This repo is cursed and dependabot doesn't work on github actions, so we're stuck with updating by hand

HigherOrderLogic: > dependabot doesn't work on github actions

There is a bot called Renovate, which have been working just fine for me without any issues.

YaLTeR: I don't want to set up another bot

nerdwave-nick: > I don't like this GHA commit pinning, makes it a pain to update them

Definitely should though. While github is probably not going to be compromised, this shit is insanely important to avoid some supply chain attacks, and so a good idea for best practices

.github/workflows/ci.yml
Boosted-Bonobo: I know people that still use yum just because that's what get got used to.
Those that got used to use dnf, will continue to do so even if it's not there anymore.
That doesn't mean we shouldn't keep up with what has changed.
image

As for apt, there is a good comparison made by Amazon.
https://aws.amazon.com/compare/the-difference-between-apt-and-apt-get
With ubuntu 26.04, using apt instead of apt-get will be even more of an improvement because of the apt 3.0 release which is the default in debian 13.

.github/workflows/ci.yml
Boosted-Bonobo: Instead of saying cursed, it would be better to say misconfigured.
If it works for a much bigger repo such as nixpkgs, it should definitely work for this repo as well.
If reaching out to github support is needed as a last resort, then that is the plan.

But at the same time...
>I don't like this GHA commit pinning, makes it a pain to update them

https://github.com/YaLTeR/niri/blob/012700dd549a79c501bcab92839e5dcc46c372b5/.github/workflows/ci.yml#L303
Like come on... this was released in [2023](https://github.com/DeterminateSystems/nix-installer-action/releases/tag/v3).
With this kind of updates, there should be no difference and no pain in updating because they weren't done before either.

YaLTeR: > Instead of saying cursed, it would be better to say misconfigured.
> If it works for a much bigger repo such as nixpkgs, it should definitely work for this repo as well.
> If reaching out to github support is needed as a last resort, then that is the plan.

I am saying cursed because that's what it is. https://github.com/dependabot/dependabot-core/issues/11201

YaLTeR: > With this kind of updates, there should be no difference and no pain in updating because they weren't done before either.

Increasing 1 digit is way simpler than going to look up the commit hash, should I really explain this?

nerdwave-nick: > Increasing 1 digit is way simpler than going to look up the commit hash, should I really explain this?

But 1 digit also does nothing against supply chain attacks. Tags are mutable. Commit SHAs are not. Malicious actions could be used to inject significant vulnerabilities or even backdoors into niri releases, which would not be visible in the code. That's a significant attack vector, which nobody would notice until the harm has been done and a bunch of people's systems have been compromised. Unlikely with github's actions, but not too crazy with the third party actions being used in various steps. Especially if niri grows and becomes a more attractive target.

Updating is a lil more annoying, but one could grab release SHAs within like 30s and find+replace them in another 30s across the whole workflow folder, which keeps updates quite manageable. 

This PR doesn't fully address this issue though, but this is a good change from a security standpoint.

**TL;DR**: Find+Replace or sed take care of the issue with updating them, and it's a pretty good change from a security standpoint and should be done in principle and dogmatically.

PR #2879, opened 2025-11-23
feat: refactor the logic for removing columns

lonerOrz:
I don't want to have a large blank area on the screen after the delete window, I think the window should automatically fill in.


Sempyos:
I will note, you are able to already to do this with the IPC,

lonerOrz:
Can we possibly support it directly?

veighnsche:
I think that there needs to be a public IPC script that does exactly this - rather then putting it in the codebase... Mainly because I think that the config for this behavior needs to be refined... and it's best to find all the config requirements during IPC script development... rather than ad-hoc in the niri codebase.

PR #2888, opened 2025-11-25
RTL-support | And also up-to-down and down-to-up support | 2D workspace support

veighnsche:
Hi @YaLTeR,

I hope everything is well. 

This PR is informational. I figured that this PR is too much of a sink to give it any review. However I am committed to support RTL. So assuming that time constraint would prevent you from checking a large PR. I first aimed to keep the RTL-support as small as possible. But since I found out that was not possible - I decided to swing the pendulum the other way and do what I know best.

I have made a snapshot generator (all the 300+ files are generated snapshots). that makes a snapshot of ALL the LTR behaviors and animations and essentially freeze the behaviors into place. (which could be an addition to the codebase if you want, but I'm still capturing ALL of them). Those LTR snapshots goes into a LTR to RTL generator. That basically "mirrors" the snapshot. And I develop based on the RTL requirements.

Well I understand that you won't accept this PR because it is huge and not what anybody is asking for. and it's going to be a complete overhaul of scrolling.rs and workspace.rs. And I really don't want to be disrespectful of your code. In the contrary even. Studying scrolling.rs and workspace.rs was so mind-opening that I am imagining so many features that I want to experiment with. For me - it is a very good and educative codebase.

The reason that I made this PR is to not leave you in the dark while I maintain an experimental fork with features that I want to add, like 2D support (added UTD and DTU snapshots) or gutter-bars (like a waybar in between tiles and columns).

Just wanted to keep you in the loop.

Thanks again for maintaining an amazing project.

Vince

### DIFF
--- a/src/layout/tests/test_configs.rs
+++ b/src/layout/tests/test_configs.rs
@@ -5,7 +5,7 @@
 //! that can be converted to both `Options` (for Rust tests) and `.kdl` files
 //! (for manual testing).
 //!
-//! To regenerate `.config/*.kdl` files from these presets:
+//! To regenerate `manual_tests/.config/*.kdl` files from these presets:
 //! ```
 //! cargo xtask generate-test-configs
 //! ```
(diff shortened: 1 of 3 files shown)


Fireye04:
This has a lot of interesting and useful features that would definitely benefit the upstream but as you acknowledge, the changes are huge and frankly almost certainly un-reviewable in their current state. I'd love to see them included upstream at some point though; is there any chance that we see this stuff chopped up a bit into smaller, more reviewable PRs so they can be added little by little?

veighnsche:
Hi @Fireye04 ,
 
 https://github.com/veighnsche/niri/tree/2d-canvas
 That is the latest branch for reference.
 
I am currently moving very fast with breaking changes... But there are definitely some things that I've discovered that I want to share upstream. I already made a golden-snapshot branch (https://github.com/veighnsche/niri/tree/golden-snapshots) with untouched behaviors and only the snapshot testing logic that I want to send upstream first. This is good to freeze all the current behaviors into place. good for ci etc.

I plan to make my first small PR in the coming week.

PR #2900, opened 2025-11-26
Fix MRU tracking when overview is open

### LINKED ISSUE #2834: Inconsistent recents behavior in overview
In the overview the most recent window changes depending on whether recents is opened on an empty workspace or not. I hope the video illustrates the issue.

https://github.com/user-attachments/assets/3ec467cc-eb14-4c7f-bace-e2d539d3642c

(tbh I find the intended behavior of the recents in overview confusing, to me it would make more sense if would behave like it does outside of overview where the window receives focus)

### System Information

* niri version: niri unstable 2025-11-17 (commit 7f19d268b3563e3683eaca02d9665fed66ce862e)

* Distro: nixos unstable

* GPU: Whatever is in the AMD Ryzen 7 PRO 5850U

* CPU: AMD Ryzen 7 PRO 5850U

### LINKED ISSUE #2839: Small issue with Alt+Tab
Alt tabbing works like expected as a 'return to last window' if pressed quickly, if held shows preview, good, but if the 'overview' is used it does not return to the last window as expected and selects the same window again instead.

### System Information

* niri version: 
niri 25.08 (commit 6ca21e7bfdd559263dbb54fb2bef51c97ba791fc)

* Distro: 
Linux nixos 6.12.58 1-NixOS SMP PREEMPT_DYNAMIC Thu Nov 13 20:34:41 UTC 2025 x86_64 GNU/Linux

* GPU: 
RTX 2080 Ti (Driver Version: 580.105.08)

* CPU:
AMD Ryzen 7 7745HX

LoayGhreeb:
This PR fixes MRU focus window tracking when the overview is open. The issue was that Alt+Tab would work once, but subsequent presses wouldn't switch back to the previous window.

The root cause is that MRU timestamps were updated based on keyboard focus changes, but when the overview is open, keyboard focus stays on the overview even though layout focus changes (via arrow keys, MRU advances, or opening new windows).

The fix adds a `previous_layout_focus_id` field to track layout focus separately from keyboard focus. On each refresh cycle, if the layout focus changed, the MRU timestamp is updated for the newly focused window.

Fixes: #2834
Fixes: #2839

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -387,6 +387,7 @@ pub struct Niri {
 
     pub window_mru_ui: WindowMruUi,
     pub pending_mru_commit: Option<PendingMruCommit>,
+    pub previous_layout_focus_id: Option<MappedId>,
 
     pub pick_window: Option<async_channel::Sender<Option<MappedId>>>,
     pub pick_color: Option<async_channel::Sender<Option<niri_ipc::PickedColor>>>,
@@ -767,6 +768,8 @@ impl State {
         self.niri.popups.cleanup();
         self.refresh_popup_grab();
         self.update_keyboard_focus();
+
+        self.update_focus_mru_timestamp();
 
         // Should be called before refresh_layout() because that one will refresh other window
         // states and then send a pending configure.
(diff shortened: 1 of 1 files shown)


YaLTeR:
I'm still somewhat undecided what's the expected/correct way for this to work, if it's keyboard focus or layout focus. One potential argument for keyboard focus is supporting layer surfaces in the future, though also not sure if they should be present in the same menu.

LoayGhreeb:
> One potential argument for keyboard focus is supporting layer surfaces in the future

I'm wondering why this would be an argument in this case?

As far as I understand, layer surfaces could still be tracked separately using keyboard focus, regardless of whether we use keyboard focus or layout focus to detect focus changes for windows. So if layer surfaces need to be tracked in the future, they could use keyboard focus if that's the correct detection method for them.

LoayGhreeb:
@YaLTeR, sorry for the ping, but are there any updates on this PR?

YaLTeR:
Evidently there aren't. I'm not ready to page in the reasoning to answer this PR right now sorry

PR #2904, opened 2025-11-27
dbus: emit when window config changed

HigherOrderLogic:
Emit the signal through Gnome Introspect DBus channel when windows config changed.

### DIFF
--- a/src/dbus/gnome_shell_introspect.rs
+++ b/src/dbus/gnome_shell_introspect.rs
@@ -51,8 +51,6 @@ impl Introspect {
         }
     }
 
-    // FIXME: call this upon window changes, once more of the infrastructure is there (will be
-    // needed for the event stream IPC anyway).
     #[zbus(signal)]
     pub async fn windows_changed(ctxt: &SignalEmitter<'_>) -> zbus::Result<()>;
 }
--- a/src/ipc/server.rs
+++ b/src/ipc/server.rs
@@ -5,7 +5,7 @@ use std::os::unix::net::{UnixListener, UnixStream};
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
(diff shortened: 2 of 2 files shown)


Sempyos:
If I may ask, why do we need this..?

HigherOrderLogic:
For completeness according to the spec. Also to help clients not needing to fetch window config everytime.

YaLTeR:
This impl is kinda cursed:
- It emits WindowsChanged overeagerly, even when the D-Bus properties (title and app_id) don't actually change, which is just extra traffic. We don't want this.
- I think it's racy because we don't have any good correlation with IPC event stream changes and the D-Bus interface currently.

Doing it properly is complicated because we don't have Arc-Mutex access to the event stream state. The DisplayConfig D-Bus works around the problem due to event stream not having any output events yet, so it uses its own separate ipc_outputs list. When event stream gets output events, it will have a similar problem.

I'm not sure off the top of my head how to refactor this to work well but I'd rather do it properly.

HigherOrderLogic:
> * It emits WindowsChanged overeagerly, even when the D-Bus properties (title and app_id) don't actually change, which is just extra traffic. We don't want this.

In the spec, there're 2 events:

```
    <!--
        RunningApplicationsChanged:
        @short_description: Notifies when the running applications changes
    -->
    <signal name="RunningApplicationsChanged" />
```

and 

```
    <!--
        WindowsChanged:
        @short_description: Notifies when any window opens or closes
    -->
    <signal name="WindowsChanged" />
```

So yeah, we should split this. The current implementation is wrong i guess.

YaLTeR:
I think WindowsChanged is intended to signal any changes to the output of GetWindows, since there's no other mechanism. So it should be emitted when (and only when) something would change about that output (so windows opening, closing, or those two exposed properties changed). You can also check how consumers like xdp-gnome use it.

PR #2906, opened 2025-11-27
screenshot_ui: add ctrl+a command to select entire output

lineargraph:
<details>
<summary>images</summary>

image
image

</details>

adds a simple <kbd>ctrl+a</kbd> keybind to the screenshot ui to select the entire output

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1636,7 +1636,11 @@ impl State {
                 self.niri.queue_redraw_all();
             }
             Action::MaximizeColumn => {
-                self.niri.layout.toggle_full_width();
+                if self.niri.screenshot_ui.is_open() {
+                    self.niri.screenshot_ui.select_entire_output();
+                } else {
+                    self.niri.layout.toggle_full_width();
+                }
             }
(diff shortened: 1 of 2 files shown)


Fireye04:
This is already a bind I believe; https://yalter.github.io/niri/Configuration%3A-Key-Bindings.html#screenshot-screenshot-screen-screenshot-window

Though alternatively a little heads up reminding users that the alternative is an option if they're selecting the full screen in the section UI might be helpful, as I actually didn't know that this was a thing before looking it up just now, and was doing it manually haha.

lineargraph:
i so far like the workflow of having one screenshotting tool to rule them all. but given that this already exists ill probably be fine just adding some extra keybindings for the entire screen screenshot.

YaLTeR:
Maybe it should (also?) be what the fullscreen-window action does in the screenshot UI? This would be consistent with how the screenshot UI handles other actions.

lineargraph:
that would be relatively easy to implement. ill check that out after work.

lineargraph:
i changed it to reuse the existing binds. hopefully this is alright :)

PR #2926, opened 2025-11-29
ipc: add output toggle option

markstos:
Output toggling is part of the Sway IPC.

In a multi-monitor setup, it can be useful to bind a key
to toggle a secondary monitor and off to temporary increase focus or save energy.

## Context about the change

Unlike the `on` and `off` commands, the `toggle` option has to first read the state, so a bit of refactoring was needed. This required for some code to be re-indented, which makes the diff look larger than it is.

## QA Log

- Tested that docs appear in CLI
- Tested manually toggling a monitor on and off.

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -993,6 +993,8 @@ pub enum OutputAction {
     Off,
     /// Turn on the output.
     On,
+    /// Toggle the output on or off.
+    Toggle,
     /// Set the output mode.
     Mode {
         /// Mode to set, or "auto" for automatic selection.
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -1824,38 +1824,31 @@ impl State {
     }
 
     pub fn apply_transient_output_config(&mut self, name: &str, action: niri_ipc::OutputAction) {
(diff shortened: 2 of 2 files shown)


averageluadev:
would this allow me to bind a key to turn off my laptop screen ? cant seem to find another way to do this as lid switch does not work

markstos:
@averageluadev I believe you can already bind a key to turn off your laptop screen and bind a second key to turn it back on. This patch would allow the same keybind to toggle it:

```
Mod+Shift+m { spawn-sh "niri msg output 'eDP-1' toggle" }
```

One IPC design question I have though is why does `niri msg output` exist at all, because all it contains are actions, which is inconsistent with putting all the other actions under `niri msg action`? 

I can guess the reason why is that perhaps it was expected that people would not want to bind keys to change their outputs, as that's normally down by something like Kanshi at startup and not changed at run time.

That's generally true, I think perhaps except perhaps for a toggling a display on and off.

It's true that even if this patch were accepted, it seems like a binding a key to toggle the display would still require a `spawn` call to `niri msg` because the function is not considered an "action" under `niri msg action`.

PR #2962, opened 2025-12-03
Add serialize/deserialize to EventStreamState in niri-ipc

Perodactyl:


HigherOrderLogic:
Why tho? This struct is never (de)serialized.

Perodactyl:
When users of niri (like me) are writing ipc scripts, it's nice to be able to directly serialize states as JSON so you could, for example, pass them to eww or waybar.

HigherOrderLogic:
My point is, this struct is never serialized, so adding this trait there is useless nonetheless.

Feel free to correct me tho.

markstos:
@Perodactyl Could you provide an example script to show how this is useful?

PR #2964, opened 2025-12-03
Optimize thumbnail creation and testing strategies

XansiVA:
Refactor thumbnail creation for performance optimizations and flexibility. Litteraly RAW SPEED DAWG....
I changed notably the create_thumbnail function
arbitrary_thumbnail(), you (whoever wrote this) were creating a thumbnail and then mutating it, I just wrote a wrapper instead.

### DIFF
--- a/src/ui/mru/tests.rs
+++ b/src/ui/mru/tests.rs
@@ -1,59 +1,113 @@
 use proptest::prelude::*;
 use proptest_derive::Arbitrary;
-
 use super::*;
 
-fn create_thumbnail() -> Thumbnail {
-    Thumbnail {
-        id: MappedId::next(),
-        timestamp: None,
-        on_current_output: false,
-        on_current_workspace: false,
-        app_id: None,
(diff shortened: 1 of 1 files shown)


YaLTeR:
But why? There's no problem with current code, and this is making it much more complex..

markstos:
@XansiVA Since this is about performance, do you have before/after benchmarks that demonstrate the performance benefit?

Xansidev:
> @XansiVA Since this is about performance, do you have before/after benchmarks that demonstrate the performance benefit?

I am okay with doing bench marks I am just a simple guy lol, right now I am going to test it to see how much it does change I just figured it'd be helpful and I didn't think this is significantly more complex?

Xansidev:
Edit; I was pretty bored, and basically have nothing else to do, I wanted to try out optimization, but it doesn't give anything on any charts i can see through system monitor, i tried, if you want to reject this then its fine.
Thank you for looking tho.

PR #2997, opened 2025-12-07
Hidden workspaces

argosnothing:
# Hidden Workspaces

### Justification
[this thread](https://github.com/YaLTeR/niri/discussions/329#discussioncomment-15181848) proposes several implementations of a niri-scratchpad system. These implementations tend to rely on a declared `stash` workspace. Currently these workspaces are visible and treated by the compositor like any other workspace. My proposol is to introduce a flag to workspaces through the Workspace { hide true } option that effectively hides that workspace in the overview, drawn windows inside the overview, and socket. 

In addition, there are workflows that will benefit from being able to dynamically "hide" a workspace in this fashion, consider a workflow in which you have a workspace you dedicate to social media, or any category that you have no reason to navigate to or see information from, you could add windows to that workspace, "hide" it, and then bring it back later when you actually need to use it in the same session. ( https://github.com/YaLTeR/niri/pull/2997#issuecomment-3632460562 )

### Implementation
Here is what I did:  
* Added: `hidden` option to workspace config
* Added: `ToggleWorkspaceVisibility` to binds
* On workspaces i've added both `hidden` and `original_idx fields`
* Changes, mainly in monitor.rs for updates on how to handle idx with hidden-aware behavior. 

### Justification for Implementation
My goal is to try and touch as little of the workspace logic as possible, with minimal additions to the code. Having a nullable idx, for example would require substantial changes to everything that works with workspaces to handle variants. 

Update in workspace logic. 
* If a workspace is hidden, we do not want it to be shown on bars or anything consuming niri workspace state, they should be effectively hidden from tools consuming that state unless explicitly requested, ( through `niri msg workspaces-with-hidden` ). I have omitted hidden workspaces from the event stream, but keep them in the state for `workspaces-with-hidden`. 
* If a workspace is hidden, we do not want it to present any gaps in workspaces that are visible, therefor we need to maintain that all hidden workspaces are at idx past any visible workspace. This means i've had to adjust several spots that deal with adding workspaces to bottom, and cleaning up workspace code. 
* A workspace, on hide should maintain it's original idx before being hidden, so it can "return" to said idx on unhide. 

- [x] Hidden option for workspace in config. 
- [x] Hidden workspaces are not viewable in overview
- [x] Hidden workspaces are not navigable to through up and down bindings
- [x] Hidden workspaces do not display in ext/workspaces
- [x] Hidden workspaces do not show through IPC ( partial, needs rewrite to maintain compatibility with code using ipc_crate ) 
- [x] Hidden workspaces id management. When hidden a hidden workspace should not cause *skips* in nonhidden workspace ids. 
- [x] Hidden workspaces *remember* their last position when unhidden. 
- [x] Hidden workspaces can be toggled through a binding. partial https://github.com/YaLTeR/niri/pull/2997/commits/d5f4772c9fad73ae4772f44410177037c1d0a436#diff-608918babe5a4eb5db908e910f16f33a173c6821eb12ef6e1077bf277160da76
- [x] Hidden workspace windows do not display in recent windows, including `All`
- [x] Clean up dead code added when experimenting with alternative IPC
- [ ] Edge case: monitor with only empty workspace, but a hidden stash workspace, shows a empty workspace below when it shouldn't. 
- [x] Crash on config change that adds hidden true to a named workspace. 
- [x] Navigating to a workspace by Idx still allows you to reach, and then interact with a hidden workspace ( even though it's windows are still hidden ). Behavior should be the same as navigating to a workspace that's higher than the last visible workspace instead, meaning if it's past the last visible workspace, it should instead snap you to the last visible workspace.
- [ ] Handle Unset workspace name, and workspaces that have names. On unset workspace name, the workspace needs to lose any hidden it has. 
- [x] Binds that change focus to a workspace through that workspaces name should unhide the workspace while in their view. 
- [x] Workspaces revealed through go to workspace [name] should rehide on leaving the workspace. 
- [ ] Monitor disconnect/reconnect, hidden workspaces becoming visible/undefined behavior
- [x] https://github.com/YaLTeR/niri/pull/2997#issuecomment-3787416682 
## Config add

```kdl
workspace "stash" {
    hidden true
}
```

### Other
If anything I would just like to use this to start a dialogue about how we could go about implementing hidden workspaces. I believe there are many different ways to go about a scratchpad implementation, and it's understandable that it will take some time. In the meantime I think having more flexibility for how the user regards workspaces can only be a good thing and allow third party implementations to act as a temporary stopgap. 

### Dailing this PR:
I try and keep this PR itself up to date with master so any potential merge will be easier, but for dailing I am matching the latest release as a separate branch from my repository. I have come back to niri, and due to some issues with recent smithay versions i'd had to make a new version off the release commit. 

[Latest Release PR](https://github.com/argosnothing/niri/tree/v25.11-hidden-workspaces) . I recommend you use this if you want hidden workspaces, unless you really want to run off main. :)


Fireye04:
Glad to see some progress here! Looking ahead at indexing implementation, I think negative numbers would make sense in this context? Especially as these hidden workspaces would be un-switchable.

argosnothing:
> Glad to see some progress here! Looking ahead at indexing implementation, I think negative numbers would make sense in this context? Especially as these hidden workspaces would be un-switchable.

that was my thought, but i think WorkspaceId wraps a u64

Fireye04:
Hmm yeah if that's a rigid implementation then that seems like an annoying workaround. In theory, you could wrap around and index backwards from 0xFFFFFFFFFFFFFFFF, then find a way to discount hidden workspaces from being detected & interpolated between. 

I'd say a swap to i64 would be theoretically cleaner but that's definitely something for YaLTeR to comment on.

argosnothing:
> Hmm yeah if that's a rigid implementation then that seems like an annoying workaround. In theory, you could wrap around and index backwards from 0xFFFFFFFFFFFFFFFF, then find a way to discount hidden workspaces from being detected & interpolated between.
> 
> I'd say a swap to i64 would be theoretically cleaner but that's definitely something for YaLTeR to comment on.

That's not a bad method. If we force the hidden workspaces at the end of u64 we shouldn't need to worry about interpolation. But ill have to play with that. 

As for larger changes like changing types on workspace id. Yeah, my hope is to step on as few as toes as possible with this, haha.

markstos:
Every workspace is that's not the active one is already hidden (not visible) outside of the overview, while the overview is useful for reviewing what can't normally be seen. What's the benefit completely hiding workspaces even from the overview? I don't quite see the use over a normal workspace named "stash" that you perhaps treat differently from others using different keybindings for it.

argosnothing:
> Every workspace is that's not the active one is already hidden (not visible) outside of the overview, while the overview is useful for reviewing what can't normally be seen. What's the benefit completely hiding workspaces even from the overview? I don't quite see the use over a normal workspace named "stash" that you perhaps treat differently from others using different keybindings for it.

It declutters the interface for those using those stash workspaces. It's a concession every implementation of scratchpad has to make because niri has no hidden workspace. Workspaces that are "hidden" by virtue of not being in the current view can still be erroneously navigated to and also display through different protocols. Those using the workspace in this way will never need to see the workspace their scratchpads go to.

danneu:
> Every workspace is that's not the active one is already hidden (not visible) outside of the overview, while the overview is useful for reviewing what can't normally be seen. What's the benefit completely hiding workspaces even from the overview? I don't quite see the use over a normal workspace named "stash" that you perhaps treat differently from others using different keybindings for it.

The scratchpad use-case might be too incidental to be the main selling point of this feature.

But I have a direct use-case. I work on a large project made up of subprojects that I open at the start of my work day.

- I open a workspace for each subproject I'm working on (each one with its own editor, terminal, browser, etc), but in any 20min block, I only need a subset of workspaces visible.
- I can't close the ones I don't need because they're stateful from when I used them 20min ago, I'll need them again in the near future, and they may be running necessary processes (like a dev server) in the terminal foreground.
- Having all eight workspaces open is mentally taxing and has me constantly navigating past workspaces I don't need at the moment. 
  - The niri/workspaces waybar widget gets so long that it crowds out the rest of my waybar
  - The workspaces also kind of look the same in the overview (full width editor next to a browser) so I get lost.

Another potential solution would be some sort of virtual output that holds workspaces offscreen.

argosnothing:
> > Every workspace is that's not the active one is already hidden (not visible) outside of the overview, while the overview is useful for reviewing what can't normally be seen. What's the benefit completely hiding workspaces even from the overview? I don't quite see the use over a normal workspace named "stash" that you perhaps treat differently from others using different keybindings for it.
> 
> The scratchpad use-case might be too incidental to be the main selling point of this feature.
> 
> But I have a direct use-case. I work on a large project made up of subprojects that I open at the start of my work day.
> 
>     * I open a workspace for each subproject I'm working on (each one with its own editor, terminal, browser, etc), but in any 20min block, I only need a subset of workspaces visible.
> 
>     * I can't close the ones I don't need because they're stateful from when I used them 20min ago, I'll need them again in the near future, and they may be running necessary processes (like a dev server) in the terminal foreground.
> 
>     * Having all eight workspaces open is mentally taxing and has me constantly navigating past workspaces I don't need at the moment.
>       
>       * The niri/workspaces waybar widget gets so long that it crowds out the rest of my waybar
>       * The workspaces also kind of look the same in the overview (full width editor next to a browser) so I get lost.
> 
> 
> Another potential solution would be some sort of virtual output that holds workspaces offscreen.

I like this idea. I was also worried that this story would be pointless without a third party tie that sends things to hidden. So would this entail keybinds that toggle hidden on workspaces by their id, name, etc? 

EDIT: i've gone ahead and taken a shot at this req.   
https://github.com/YaLTeR/niri/pull/2997/commits/d5f4772c9fad73ae4772f44410177037c1d0a436#diff-608918babe5a4eb5db908e910f16f33a173c6821eb12ef6e1077bf277160da76

src/layout/workspace.rs
```diff
@@ -522,7 +540,9 @@ impl<W: LayoutElement> Workspace<W> {
     }
 
     pub fn update_output_size(&mut self) {
-        let output = self.output.as_ref().unwrap();
+        let Some(output) = self.output.as_ref() else {
+            return;
+        };
```
argosnothing, line 545: On the case of updating the output size of the ws a hidden workspace will not be rendered so we need to exit early if there is none. At least that seems to be the need, as otherwise i'd get core dump on the initial hide toggle.

src/layout/workspace.rs
```diff
@@ -110,6 +110,13 @@ pub struct Workspace<W: LayoutElement> {
 
     /// Unique ID of this workspace.
     id: WorkspaceId,
+
+    /// Hidden workspaces need to track their original idx
+    ///
+    pub original_idx: Option<usize>,
```
argosnothing: Ideally, I'd like a hidden workspace to try and keep it's idx so it can be at an intuitive place when you unhide it, similar to workspace movement on monitor connect/disconnect.

argosnothing:
tmp QsDgnBxr3O

Bit of progress on this.

argosnothing:
Currently, for my usecase this is functional for me, and I encourage anyone willing to experiment to try this PR and let me know if there are issues with it, or edgecases I didn't test. I have also added a separate branch on my [scratchpad project](https://github.com/argosnothing/niri-scratchpad-rs/tree/hidden-workspaces) that makes use of this PR's new niri_ipc WorkspacesWithHidden functionality to summon and stash windows bound to hidden workspaces.

argosnothing:
An alternative approach I've been thinking about:  
We maintain a separate vec of workspaces, remove the original_idx field, and simply use the workspace idx for the hidden workspaces vec to map onto the original workspace instead. I think this would be cleaner, and potentially lets us avoid unnecessary data on structs like hidden and original_idx. We will know if a workspace is hidden simply by it being in the hidden_workspaces vec, and we can determine the idx it wants to go by what its idx is, and simply insert by that idx when we unhide it. 

So, we hide a workspace. We remove the workspace from the monitors workspace vec ( but retain its output. ). Then insert that workspace into a hidden_workspaces vec on that monitor. Because it's in a different vec we shouldn't need extra logic for managing idxs of the visible workspaces. When we unhide that workspace, we simply insert it by its idx into the main workspace vec at its desired position. 

My hope is this method will require far fewer changes, less extra struct fields, less complexity for the same functionality.

https://github.com/YaLTeR/niri/compare/main...argosnothing:niri:hidden-workspaces-2 an attempt was made. I'm not really convinced this is a better route after all. It seems to be even more code add than simply keeping hidden workspaces in the same vec and adjusting indexes to keep out of the visible idx range.

src/ipc/client.rs
```diff
@@ -338,6 +339,54 @@ pub fn handle_msg(mut msg: Msg, json: bool) -> anyhow::Result<()> {
                 println!("The change will apply when it is connected.");
             }
         }
+        Msg::WorkspacesWithHidden => {
```
argosnothing, line 343: Initially I had added options to Workspaces for different filtering options with Hidden workspaces, but this method seems better since it won't break projects that use the rust_ipc crate as it's opt in.

src/ipc/server.rs
```diff
@@ -278,6 +278,18 @@ async fn process(ctx: &ClientCtx, request: Request) -> Reply {
             Response::Outputs(outputs.collect())
         }
         Request::Workspaces => {
+            let state = ctx.event_stream_state.borrow();
```
argosnothing, line 281: We do not include hidden workspaces in normal workspace requests. If consumers want to access hidden workspaces, they'll need to use the new endpoint.

src/ipc/server.rs
```diff
@@ -670,6 +716,26 @@ impl State {
             state.apply(event.clone());
             server.send_event(event);
         }
+
```
argosnothing, line 732: I played around with different methods to keep hidden workspaces out of event stream while still keeping it for the state, this just ended up being the only one I could get to work without horrible things happening.

src/layout/monitor.rs
```diff
@@ -690,6 +761,21 @@ impl<W: LayoutElement> Monitor<W> {
...
+            return;
+        }
+        if self.options.layout.empty_workspace_above_first && idx == 0 {
+            self.add_workspace_top();
+        }
+
+        self.workspaces[idx].hidden = true;
+        let mut ws = self.remove_workspace_by_idx(idx);
+        ws.original_idx = Some(idx);
+        self.workspaces.insert(self.workspaces.len(), ws);
```
argosnothing: This is the main "trick" of how im doing these hidden workspaces. If it's hidden, we make sure it's at the end of the vec, and that empty workspace is shown right before, creating the illusion that the empty workspace is the last workspace, and the hidden ones arn't behind it.

alexjp:
Hi,

I don't want to add complexity, but may I ask if this method of hiding workspaces supports also hidding named workspaces?

I have a couple of usecases that would be great to have named workspaces (which have specific layouts) be hidden!

Thanks!

argosnothing:
> Hi,
> 
> I don't want to add complexity, but may I ask if this method of hiding workspaces supports also hidding named workspaces?
> 
> I have a couple of usecases that would be great to have named workspaces (which have specific layouts) be hidden!
> 
> Thanks!

Yes, in fact it is designed to only work on named workspaces. 

Currently I have a workspace called stash that i send scratchpad windows to. This is usually hidden, but i can toggle it with a keybind. I have another workspace called "work" that Is unhidden but I toggle to hidden when im not working.

alexjp:
> Yes, in fact it is designed to only work on named workspaces.

Right, sorry, dumb question!

I tried your branch `hidden-workspaces`, and it works, I managed to hide some workspaces at startup and toggle them hidden/visible.

Some things I missed, and wondering if it is in scope of features:
- If a workspace is hidden, but I have a shortcut to focus it, if it could be made visible automatically (and not focus a dynamically created workspace)
- If the workspace is hidden, visible when selected, and then left (switch to another workspace), without any window open, if it could be hidden again. (maybe a setting "hide-when-empty")
- If the bind `toggle-workspace-visibility` could work with current workspace, something like `toggle-current-workspace-visibility`

I understand if these are not wanted features, since I think I am trying to use this in a different usecase. Instead of one hidden "stash" workspace, I am trying to have several workspaces that I not always use, configured.
For example, my usecase is something like:
- Workspace named "MultiMedia"
  - Has special layout settings
  - My `wlr-which-key` script with check the name of the workspace, and if it exists, it will show the specific menu (like using `wlr-which-key -k "a m"`)
- Workspace named "Chat"
  - Has special layout settings 
  - My `wlr-which-key` script with check the name of the workspace, and if it exists, it will show the specific menu (like using `wlr-which-key -k "a c"`)


Mentioning again, totally understand if this usecase it outside of scope!

Thanks!

argosnothing:
> > Yes, in fact it is designed to only work on named workspaces.
> 
> Right, sorry, dumb question!
> 
> I tried your branch `hidden-workspaces`, and it works, I managed to hide some workspaces at startup and toggle them hidden/visible.
> 
> Some things I missed, and wondering if it is in scope of features:
> 
>     * If a workspace is hidden, but I have a shortcut to focus it, if it could be made visible automatically (and not focus a dynamically created workspace)
> 
>     * If the workspace is hidden, visible when selected, and then left (switch to another workspace), without any window open, if it could be hidden again. (maybe a setting "hide-when-empty")
> 
>     * If the bind `toggle-workspace-visibility` could work with current workspace, something like `toggle-current-workspace-visibility`
> 
> 
> I understand if these are not wanted features, since I think I am trying to use this in a different usecase. Instead of one hidden "stash" workspace, I am trying to have several workspaces that I not always use, configured. For example, my usecase is something like:
> 
>     * Workspace named "MultiMedia"
>       
>       * Has special layout settings
>       * My `wlr-which-key` script with check the name of the workspace, and if it exists, it will show the specific menu (like using `wlr-which-key -k "a m"`)
> 
>     * Workspace named "Chat"
>       
>       * Has special layout settings
>       * My `wlr-which-key` script with check the name of the workspace, and if it exists, it will show the specific menu (like using `wlr-which-key -k "a c"`)
> 
> 
> Mentioning again, totally understand if this usecase it outside of scope!
> 
> Thanks!

1 and 2 seem like good ideas and would be fun to implement. 

The issue with the third, and why I only have it setup for named workspaces ( although i believe i still need to properly represent this in the config ) 
1. You are on idx 3. so the third workspace in a monitor. there is a workspace on 4 as well. 
2. You remove workspace on idx 3 ( because you are currently on that workspace and hide it)
3. What happens is workspace at idx 3 becomes hidden, and its idx actually ends up being 5 ( or 6 when an empty workspace is there )
4. The workspace that was idx 4 now becomes idx 3, taking that hidden workspaces place. 
5. You don't really have any method to access that workspace again because to the user workspaces are really regarded by that idx. 

Your usecase makes sense though, and for the first two i'll add that to my todo on this PR. 

EDIT: I will need to support dynamically assigned and unassigned workspace as well, so this might suit your first requirement after all once implemented.

alexjp:
> 1 and 2 seem like good ideas and would be fun to implement.
> 
> The issue with the third, and why I only have it setup for named workspaces ( although i believe i still need to properly represent this in the config )

Great to hear!

I, in my own use case, don't see much issue in the third not being implemented if the first 2 are. Because if it "auto hides" based on having windows or not, I don't need much of "manual hidding".

For the manual hidding (and really hide windows), I can set a proper static scratchpad workspace, like the use case of this pull request.

Thanks!

argosnothing:
> > 1 and 2 seem like good ideas and would be fun to implement.
> > The issue with the third, and why I only have it setup for named workspaces ( although i believe i still need to properly represent this in the config )
> 
> Great to hear!
> 
> I, in my own use case, don't see much issue in the third not being implemented if the first 2 are. Because if it "auto hides" based on having windows or not, I don't need much of "manual hidding".
> 
> For the manual hidding (and really hide windows), I can set a proper static scratchpad workspace, like the use case of this pull request.
> 
> Thanks!

hopefully this is easy to follow, but this should be an example of both 1 and 2? It's not through a config option, the idea is that if you have focus-workspace to a named workspace that is hidden, it will reveal itself to you until you switch off from it, in which case it rehides. This isn't really ready to be merged into the main PR yet. ( the idx finding probably needs work and as you can see, when leaving the needs hidden workspace it tends to skip past 1 ) 
example

alexjp:
Just tested.

Yeah, as you said exactly :)

It is interesting, maybe for temporary apps/windows, that I don't mind losing when closing Niri.
The issue with them being hidden as soon as I leave, is that then I need to look for open apps in every workspace.

I am testing with Dank shell, which allows to give special icons for named desktops in Niri. (one issue with Dank shell, is if I click on the taskbar, the icon of a window on a hidden workspace, it does not change to it)

Would be great to have an option to leave them visible if they have windows, but, this way is also interesting! (already have a use case in mind... to use llama-serve, which I want running but it would be handy to be hidden when not on it, and I would not care to close it before exiting Niri).

Thanks, looking good! :)

argosnothing:
> Just tested.
> 
> Yeah, as you said exactly :)
> 
> It is interesting, maybe for temporary apps/windows, that I don't mind losing when closing Niri. The issue with them being hidden as soon as I leave, is that then I need to look for open apps in every workspace.
> 
> I am testing with Dank shell, which allows to give special icons for named desktops in Niri. (one issue with Dank shell, is if I click on the taskbar, the icon of a window on a hidden workspace, it does not change to it)
> 
> Would be great to have an option to leave them visible if they have windows, but, this way is also interesting! (already have a use case in mind... to use llama-serve, which I want running but it would be handy to be hidden when not on it, and I would not care to close it before exiting Niri).
> 
> Thanks, looking good! :)

Next time i do a rebuild i'll bring in these changes myself. When testing niri it actually crashed my scratchpad daemon so i need to retag my windows in my stash workspace to pick them back up after i restart the daemon. Would save me a keypress as I never need the workspace to stay unhidden when im done with it.  

As for the other stuff i'll probably hold off until I get some additional feedback on what things other people would want/expect for a PR like this. Your suggestion of auto unhiding when explicitly focused makes intuitive sense to me though.

fnzr:
I'm trying out this branch, and things have been pretty smooth so far. I'm also using https://github.com/argosnothing/niri-scratchpad-rs/tree/hidden-workspaces and it's working as expected.

I just did some basic sanity tests, but that's a feature I'm really looking forward to, I'll keep using it to see if something comes up. Good job!

Something I did notice, if you create a hidden workspace while niri is running, niri crashes. 

I mean, adding

```
workspace "stash" {
   hidden true
}
```
and saving to hot reload, it causes a crash.

Creating a `hidden false` workspace does not crash, and creating a hidden false workspace and changing it to true also does not crash. Running niri again after the crash works normally

argosnothing:
> I'm trying out this branch, and things have been pretty smooth so far. I'm also using https://github.com/argosnothing/niri-scratchpad-rs/tree/hidden-workspaces and it's working as expected.
> 
> I just did some basic sanity tests, but that's a feature I'm really looking forward to, I'll keep using it to see if something comes up. Good job!
> 
> Something I did notice, if you create a hidden workspace while niri is running, niri crashes.
> 
> I mean, adding
> 
> ```
> workspace "stash" {
>    hidden true
> }
> ```
> 
> and saving to hot reload, it causes a crash.
> 
> Creating a `hidden false` workspace does not crash, and creating a hidden false workspace and changing it to true also does not crash. Running niri again after the crash works normally

Ah i forgot to add that bug to my todos! I did run into this. I'll add it to the list. I try to get a few things knocked out on this PR a week and do my best to keep it up to date with master. I appreciate the feedback!

argosnothing:
I've added the fix to `Crash on config change that adds hidden true to a named workspace.`. 

One note is that hidden workspaces "want" to be sent to where they were previously when you hide and unhide. I use the idx for this. On a newly added hidden workspace that idx will be where that workspace would have been sent to if it were not hidden. 

So lets say you have non hidden workspaces like 
|0|1|2|

on hidden workspace add it's original_idx ( where it wants to go when unhidden ) will be set to 0, pushing the idx of other workspaces up, it's actual position on add will be at |3| (at the end of the vec, and hidden, inaccessible), once it gets unhid it will go to it's 0 position.

GoodbyeNJN:
After using this feature daily for nearly a month, I find it to be sufficiently reliable and very practical. Are there any plans to convert this PR from a draft to a formal PR?

argosnothing:
> After using this feature daily for nearly a month, I find it to be sufficiently reliable and very practical. Are there any plans to convert this PR from a draft to a formal PR?

I would like some input from contributors and yalter first. There are currently many Issues in the pipeline so I'd rather get input before I add to the pile. There's also several edge cases I know exist like travelling hidden workspaces that I need to brainstorm workarounds for.


Prs to this are welcome as well, my plan is to keep this updated with niri major releases.

MilitaoLucas:
Hello, great work! I am using this, and I am having a problem:
```niri
window-rule {
    match app-id="qalculate-gtk"
    open-floating true
    open-on-workspace "stash"
    default-column-width { proportion 0.75; }
    default-window-height { proportion 0.75; }
    default-floating-position x=0 y=100 relative-to="bottom"
}
workspace "stash" {
   hidden true
}
```
as soon as I open the application I get the workspace ID of the current workspace even though it was supposed to open in the stash workspace:
```
Window ID 13:
  Title: "Qalculate!"
  App ID: "qalculate-gtk"
  Is floating: yes
  PID: 107243
  Workspace ID: 1
  Layout:
    Tile size: 1200 x 642
    Workspace-view position: 200, 158
    Window size: 1200 x 642
    Window offset in tile: 0 x 0
```
It should show workspace ID 3:
```zsh
❯ niri msg workspaces-with-hidden
Output "DP-1":
 * 1
   2

Output "HDMI-A-2":
 * 1
   2
   3 "stash"
```
I don't know if this is related to having 2 monitors. But I also can't see the window but know it is running because of `niri msg windows`.

MilitaoLucas:
> Hello, great work! I am using this, and I am having a problem:
> 
> ```
> window-rule {
>     match app-id="qalculate-gtk"
>     open-floating true
>     open-on-workspace "stash"
>     default-column-width { proportion 0.75; }
>     default-window-height { proportion 0.75; }
>     default-floating-position x=0 y=100 relative-to="bottom"
> }
> workspace "stash" {
>    hidden true
> }
> ```
> 
> as soon as I open the application I get the workspace ID of the current workspace even though it was supposed to open in the stash workspace:
> 
> ```
> Window ID 13:
>   Title: "Qalculate!"
>   App ID: "qalculate-gtk"
>   Is floating: yes
>   PID: 107243
>   Workspace ID: 1
>   Layout:
>     Tile size: 1200 x 642
>     Workspace-view position: 200, 158
>     Window size: 1200 x 642
>     Window offset in tile: 0 x 0
> ```
> 
> It should show workspace ID 3:
> 
> ```shell
> ❯ niri msg workspaces-with-hidden
> Output "DP-1":
>  * 1
>    2
> 
> Output "HDMI-A-2":
>  * 1
>    2
>    3 "stash"
> ```
> 
> I don't know if this is related to having 2 monitors. But I also can't see the window but know it is running because of `niri msg windows`.

As a reference, I solved this by keeping track of when the windows is hidden by using a file in the tmp dir.

argosnothing:
> Hello, great work! I am using this, and I am having a problem:
> 
> ```
> window-rule {
>     match app-id="qalculate-gtk"
>     open-floating true
>     open-on-workspace "stash"
>     default-column-width { proportion 0.75; }
>     default-window-height { proportion 0.75; }
>     default-floating-position x=0 y=100 relative-to="bottom"
> }
> workspace "stash" {
>    hidden true
> }
> ```
> 
> as soon as I open the application I get the workspace ID of the current workspace even though it was supposed to open in the stash workspace:
> 
> ```
> Window ID 13:
>   Title: "Qalculate!"
>   App ID: "qalculate-gtk"
>   Is floating: yes
>   PID: 107243
>   Workspace ID: 1
>   Layout:
>     Tile size: 1200 x 642
>     Workspace-view position: 200, 158
>     Window size: 1200 x 642
>     Window offset in tile: 0 x 0
> ```
> 
> It should show workspace ID 3:
> 
> ```shell
> ❯ niri msg workspaces-with-hidden
> Output "DP-1":
>  * 1
>    2
> 
> Output "HDMI-A-2":
>  * 1
>    2
>    3 "stash"
> ```
> 
> I don't know if this is related to having 2 monitors. But I also can't see the window but know it is running because of `niri msg windows`.

Interesting. I haven't tested with against window rules. I'll add that to my todos. Thanks for letting me know.

MilitaoLucas:
@argosnothing Something that is more of a QOL tip, would it be possible to just show some symbol or something in the normal workspace listing command to say that the workspace is hidden? Maybe it is something like: 

3 "~~stash~~"

MilitaoLucas:
> @argosnothing Something that is more of a QOL tip, would it be possible to just show some symbol or something in the normal workspace listing command to say that the workspace is hidden? Maybe it is something like: 
> 
> 3 "~~stash~~" 

I think that most scripts use json, so it shouldn't break a lot of stuff to unify the commands and add a flag. But, maybe you just created the command to make it more clear while testing the PR.

I will let you know if I find anything else. I plan to keep using it. I will also soon put the scripts I use to make the complete scratchpad setup work even with that issue.

Thank you for the awesome work! I don't have the time to help with code as I am not very proficient in Rust, but I hope I can help with reproducible bug reports.

argosnothing:
> > @argosnothing Something that is more of a QOL tip, would it be possible to just show some symbol or something in the normal workspace listing command to say that the workspace is hidden? Maybe it is something like:
> > 3 "~stash~"
> 
> I think that most scripts use json, so it shouldn't break a lot of stuff to unify the commands and add a flag. But, maybe you just created the command to make it more clear while testing the PR.
> 
> I will let you know if I find anything else. I plan to keep using it. I will also soon put the scripts I use to make the complete scratchpad setup work even with that issue.
> 
> Thank you for the awesome work! I don't have the time to help with code as I am not very proficient in Rust, but I hope I can help with reproducible bug reports.

Yes, the initial intention was to never touch previous IPC code. If i remember correctly, I initially put the hidden in normal workspace list calls, with an additional hidden flag, but the issue is shells like noctalia do not know about this flag, and will simply include workspaces by default. 

I initially added a flag to that endpoint, and this does work. The issue is mainly on applications that hook into the niri_ipc rust crate would also need to be updated for that version. In the end I think it does make more sense to pass this along any custom ipcs niri exposes, but since I also use this daily and make use of noctalia ( and also don't want to modify noctalias own dependencies to match ), I went ahead and just made a special endpoint to grab hidden.

argosnothing:
> > Hello, great work! I am using this, and I am having a problem:
> > ```
> > window-rule {
> >     match app-id="qalculate-gtk"
> >     open-floating true
> >     open-on-workspace "stash"
> >     default-column-width { proportion 0.75; }
> >     default-window-height { proportion 0.75; }
> >     default-floating-position x=0 y=100 relative-to="bottom"
> > }
> > workspace "stash" {
> >    hidden true
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as soon as I open the application I get the workspace ID of the current workspace even though it was supposed to open in the stash workspace:
> > ```
> > Window ID 13:
> >   Title: "Qalculate!"
> >   App ID: "qalculate-gtk"
> >   Is floating: yes
> >   PID: 107243
> >   Workspace ID: 1
> >   Layout:
> >     Tile size: 1200 x 642
> >     Workspace-view position: 200, 158
> >     Window size: 1200 x 642
> >     Window offset in tile: 0 x 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It should show workspace ID 3:
> > ```shell
> > ❯ niri msg workspaces-with-hidden
> > Output "DP-1":
> >  * 1
> >    2
> > 
> > Output "HDMI-A-2":
> >  * 1
> >    2
> >    3 "stash"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't know if this is related to having 2 monitors. But I also can't see the window but know it is running because of `niri msg windows`.
> 
> As a reference, I solved this by keeping track of when the windows is hidden by using a file in the tmp dir.

This should be on the branch now. Develop also has the merge of the current latest yalter commits in it. Will work on moving those over to this PR later.

AmmoniumX:
Thank you for adding this feature! I'd love this to get out of draft status soon and get merged, this is one of the biggest missing features for niri to become my ideal daily use DE (besides allowing floating windows to render on top of fullscreen windows). I use dynamic/unnamed workspaces and having a "hidden" workspace would allow me to clean up a lot of stuff, such as a) hiding the tiny blank window that seems to be created by proton every time I launch a game using it, and b) have some window applications running in the "background" if I need them to and they don't have a built-in "windowless" option.

I noticed a minor issue using the default config (with dynamic workspaces) and adding this:
```
window-rule {
    match app-id="com.github.PintaProject.Pinta"
    open-on-workspace "stash"
}

workspace "stash" {
  hidden true
}
```

When opening an application and a window-rule is set to open on the "stash" workspace by default, the dynamic workspace order is altered and messes up with navigation.

Example:
1. Start on default workspace 1:
```
❯ ./target/debug/niri msg workspaces                                            
Output "eDP-1":
 * 1
   2
```
To avoid confusion, let's call this current workspace one the "main" workspace, but note that it's not actually named like that, it's just "workspace 1"

2. Open any app with default window-rule opening it on stash, in this example "pinta"
3. Check new workspace order:
```
❯ ./target/debug/niri msg workspaces
Output "eDP-1":
   1 "stash"
 * 2
   3
```
Now, pressing Mod+1 to focus workspace 1 sends me to a different workspace than my previos workspace 1. Now my "main" workspace was assigned to workspace 2. 

If I go and focus the stash workspace and then go back to my main workspace, then it seems to fix itself:
```
❯ ./target/debug/niri msg workspaces
Output "eDP-1":
 * 1
   2
```

This happens when opening pretty much any window set to open on stash via windowrule.

argosnothing:
> Thank you for adding this feature! I'd love this to get out of draft status soon and get merged, this is one of the biggest missing features for niri to become my ideal daily use DE (besides allowing floating windows to render on top of fullscreen windows). I use dynamic/unnamed workspaces and having a "hidden" workspace would allow me to clean up a lot of stuff, such as a) hiding the tiny blank window that seems to be created by proton every time I launch a game using it, and b) have some window applications running in the "background" if I need them to and they don't have a built-in "windowless" option.
> 
> I noticed a minor issue using the default config (with dynamic workspaces) and adding this:
> 
> ```
> window-rule {
>     match app-id="com.github.PintaProject.Pinta"
>     open-on-workspace "stash"
> }
> 
> workspace "stash" {
>   hidden true
> }
> ```
> 
> When opening an application and a window-rule is set to open on the "stash" workspace by default, the dynamic workspace order is altered and messes up with navigation.
> 
> Example:
> 
>     1. Start on default workspace 1:
> 
> 
> ```
> ❯ ./target/debug/niri msg workspaces                                            
> Output "eDP-1":
>  * 1
>    2
> ```
> 
> To avoid confusion, let's call this current workspace one the "main" workspace, but note that it's not actually named like that, it's just "workspace 1"
> 
>     2. Open any app with default window-rule opening it on stash, in this example "pinta"
> 
>     3. Check new workspace order:
> 
> 
> ```
> ❯ ./target/debug/niri msg workspaces
> Output "eDP-1":
>    1 "stash"
>  * 2
>    3
> ```
> 
> Now, pressing Mod+1 to focus workspace 1 sends me to a different workspace than my previos workspace 1. Now my "main" workspace was assigned to workspace 2.
> 
> If I go and focus the stash workspace and then go back to my main workspace, then it seems to fix itself:
> 
> ```
> ❯ ./target/debug/niri msg workspaces
> Output "eDP-1":
>  * 1
>    2
> ```
> 
> This happens when opening pretty much any window set to open on stash via windowrule.

The intended behavior I had was that on opening on a hidden workspace, we unhide that workspace, and I believe switch to it. I'm not certain on the default configuration for window rule opening so let me know what the expected behavior is there.

If open on stash workspace toggles visibliity of the workspace, and does not switch your view to the workspace, that wasn't my intention. My idea was 

```  
*1
 2  
```
open application with window rule for stash : 
```
 * stash
 2
 3
 ```
 on moving off of the hidden workspace we rehide it. I do not use these window rules so just let me know what the more intuitive flow would be

**Comment by AmmoniumX** [comment]
> The intended behavior I had was that on opening on a hidden workspace, we unhide that workspace, and I believe switch to it. I'm not certain on the default configuration for window rule opening so let me know what the expected behavior is there.

> If open on stash workspace toggles visibliity of the workspace, and does not switch your view to the workspace, that wasn't my intention.

> on moving off of the hidden workspace we rehide it. I do not use these window rules so just let me know what the more intuitive flow would be

Hmm, so you're saying that it's intentional behavior? I'm not sure I like it. From my tests, opening something on stash does make stash visible when it wasn't before, and as a result all other workspaces get "shifted" down by one. My expectation would be that my workspace 1 remains workspace 1 no matter if I move something to stash or not. Stash should remain "hidden" by default and not change unless explicitly asked to, even if something is opened there. But as of right now, it becomes visible automatically

**Comment by argosnothing** [comment]
> > The intended behavior I had was that on opening on a hidden workspace, we unhide that workspace, and I believe switch to it. I'm not certain on the default configuration for window rule opening so let me know what the expected behavior is there.
> 
> > If open on stash workspace toggles visibliity of the workspace, and does not switch your view to the workspace, that wasn't my intention.
> 
> > on moving off of the hidden workspace we rehide it. I do not use these window rules so just let me know what the more intuitive flow would be
> 
> Hmm, so you're saying that it's intentional behavior? I'm not sure I like it. From my tests, opening something on stash does make stash visible when it wasn't before, and as a result all other workspaces get "shifted" down by one. My expectation would be that my workspace 1 remains workspace 1 no matter if I move something to stash or not. Stash should remain "hidden" by default and not change unless explicitly asked to, even if something is opened there. But as of right now, it becomes visible automatically

I will need to check the default behavior. I though opening a window with a workspace rule would have you follow *to* that workspace, but I guess in the cash of a hidden workspace we dont want this.

**Comment by argosnothing** [comment]
alright, I think you were right. It seems more intuitive to simply move the window to the hidden workspace if it's currently hidden. Incoming changes

Edit: Both this PR and the stable branch off the previous version are updated. Let me know how it works!

**Comment by AmmoniumX** [comment]
> alright, I think you were right. It seems more intuitive to simply move the window to the hidden workspace if it's currently hidden. Incoming changes
> 
> Edit: Both this PR and the stable branch off the previous version are updated. Let me know how it works!

Works great, thank you for the change!

**Comment by MilitaoLucas** [comment]
I never had any issue with this PR. I am running for a while now. It just randomly switches sometimes to a hidden workspace but I can't reproduce it. Do you have a roadmap to make it a normal PR? It is fine anyway, I will keep building it in my nix setup, but this is something more people should have as it is very nice!

**Comment by argosnothing** [comment]
> I never had any issue with this PR. I am running for a while now. It just randomly switches sometimes to a hidden workspace but I can't reproduce it. 

Did this just start occurring recently or has this been happening for awhile now? 


---
>  Do you have a roadmap to make it a normal PR? It is fine anyway, I will keep building it in my nix setup, but this is something more people should have as it is very nice!

I think I've been a bit cagey on undrafting this, so I'll just explain where I'm at here: 

Many parts of it are largely untested, and automated tests still fail. Things like how this interacts with moveable workspaces, removal/adding of a named workspace through running config changes, monitor connect/disconnect, etc are still untested. 

#### Hide/Unhide weirdness
There is also behavior I don't like when it comes to multiple named workspaces toggling before and after adding a bunch of new workspaces. On hide a workspace uses it's idx at time of hide and inserts itself into that idx on unhide. 

The problem is that what if you have

(all unhidden)
stash1
stash2
3
4

then hide stash 1. Now stash1 on rehide will always insert at idx 1. But after you hide stash1 and then hide stash2, well because stash2 is now at idx 1 it uses that same idx on its own unhide meaning those temporarily hidden workspaces will randomly be at either idx 1 or 2, depending on hide and unhide order.

Similar strange behavior happens if you hide a workspace between visible workspaces. Hidden workspaces stick to the end of the vec, so you will never get idx gaps, but on rehide hidden workspaces will try and go to it's `original_idx`. 

One thought I had was on stash2 hide, we first check to see if any currently hidden workspaces have their `original_idx` set to where stash2 is at that moment. in the above example stash1 is hidden with it's `original_idx` set to 1, the current idx of stash2. So we know we should not set the original_idx to 1, and we repeat so forth for every taken idx so no two hidden workspaces have the same original_idx (although this name would need to be changed to reflect what it's actually for now). 

The issue is that this *can still* produce undefined behavior as workspaces are entirely dynamic, so trying to reinsert workspaces in a specific spot that makes sense by the time you reinsert *seems* impossible to me, espcially because you can arbitrarily remove, add, hide, workspaces at any point. 

It almost seems like the better behavior would be to simply insert the workspace at the end of the vec ( but before where we have hidden workspaces ) and unhide.

---

Testing is appreciated, as well as ideas for how to manage the different edgecases here.

**Comment by MilitaoLucas** [comment]
> Did this just start occurring recently or has this been happening for awhile now?

Always happened. I believe this is mainly an issue with my 2 monitor of different resolutions setup. When I change monitors I notice that I can navigate to the hidden workspace for a while for the first time I do it.

The other issues are probably something to do with the architecture of this change. Unfortunately I don't have the time at the moment to work on helping in this. I know it must be annoying to just use it and not contribute with nothing but some vague bug reports. I am sorry. 

The order problem seems like something that could be solved by limiting the amount of un-hidden workspaces to one per monitor. I don't see why you would need to see more than 2 hidden workspaces at a time. I also thinking putting the workspace as the last element in the vector is more consistent. 

The above suggestion could maybe let you switch between the hidden workspaces with a bind. Or maybe even a switch, where you use a binding to turn all visible workspaces hidden and all hidden visible could work. I don't know how feasible this is to implement

**Comment by argosnothing** [comment]
> > Did this just start occurring recently or has this been happening for awhile now?
> 
> Always happened. I believe this is mainly an issue with my 2 monitor of different resolutions setup. When I change monitors I notice that I can navigate to the hidden workspace for a while for the first time I do it.
> 
> The other issues are probably something to do with the architecture of this change. Unfortunately I don't have the time at the moment to work on helping in this. I know it must be annoying to just use it and not contribute with nothing but some vague bug reports. I am sorry.
> 
> The order problem seems like something that could be solved by limiting the amount of un-hidden workspaces to one per monitor. I don't see why you would need to see more than 2 hidden workspaces at a time. I also thinking putting the workspace as the last element in the vector is more consistent.
> 
> The above suggestion could maybe let you switch between the hidden workspaces with a bind. Or maybe even a switch, where you use a binding to turn all visible workspaces hidden and all hidden visible could work. I don't know how feasible this is to implement

Knowing that bug has been awhile narrows down a lot so I appreciate it! If its been an issue for several months than it would have been some earlier adds :)

I will likely move it to end of vec only, this will also let me do away with `original_idx` field on workspace which i didn't like anyways. As for multiple hidden workspaces, I've actually been keeping two personally. One that is permahidden called "stash", and another that I keep for non work related stuff so it's visually cleaner and less distracting.

**Comment by AmmoniumX** [comment]
@argosnothing 
> #### Hide/Unhide weirdness
> 
> There is also behavior I don't like when it comes to multiple named workspaces toggling before and after adding a bunch of new workspaces. On hide a workspace uses it's idx at time of hide and inserts itself into that idx on unhide.
> 
> The problem is that what if you have
> 
> (all unhidden) stash1 stash2 3 4
> 
> then hide stash 1. Now stash1 on rehide will always insert at idx 1. But after you hide stash1 and then hide stash2, well because stash2 is now at idx 1 it uses that same idx on its own unhide meaning those temporarily hidden workspaces will randomly be at either idx 1 or 2, depending on hide and unhide order.
> 
> Similar strange behavior happens if you hide a workspace between visible workspaces. Hidden workspaces stick to the end of the vec, so you will never get idx gaps, but on rehide hidden workspaces will try and go to it's `original_idx`.

What about basing the insert position off of *context* instead of their index? This is similar to how git applies patches and merge conflicts for example, instead of just "reinsert at X position" it looks for what was *around* itself. Something like this:

```
struct HiddenWorkspace {
    id: usize,
    prev_neighbor_id: Option<usize>,
    next_neighbor_id: Option<usize>,
}

fn hide(visible: &mut Vec<usize>, hidden: &mut Vec<HiddenWorkspace>, id: usize) {
    let idx = visible.find(id);

    // remember left and right neighbors
    hidden.push(HiddenWorkspace {
        id,
        prev_neighbor_id: visible.get(idx - 1),
        next_neighbor_id: visible.get(idx + 1),
    });

    visible.remove(idx);
}

fn unhide(visible: &mut Vec<usize>, hidden: &mut Vec<HiddenWorkspace>, id: usize) {
    let entry = hidden.remove(id);

    // Priority 1: place after our old left neighbor, if it's still visible
    //             (transitively walk the chain if it's also hidden)
    if let Some(prev_id) = find_visible_neighbor(entry.prev_neighbor_id, going_left) {
        visible.insert(after: prev_id, entry.id);
        return;
    }

    // Priority 2: place before our old right neighbor, if it's still visible
    if let Some(next_id) = find_visible_neighbor(entry.next_neighbor_id, going_right) {
        visible.insert(before: next_id, entry.id);
        return;
    }

    // Fallback: both neighbors are gone (e.g deleted dynamic workspace), just append
    visible.push(entry.id);
}

// Walks the hidden chain until it finds a neighbor that is currently visible
fn find_visible_neighbor(start_id: Option<usize>, direction: Direction) -> Option<usize> {
    let mut current = start_id;

    while let Some(id) = current {
        if visible.contains(id) {
            return Some(id);  // found a visible anchor
        }
        // This neighbor is also hidden — follow its neighbor in the same direction
        current = hidden.get(id).neighbor_in(direction);
    }

    None  // no visible anchor found
}
```

This is guaranteed to "just work" by following the hidden chain as long as workspaces aren't removed/deleted, which isn't really the case for dynamic workspaces, but if it can't find their neighbouring context I think appending would be a reasonable fallback.

**Comment by argosnothing** [comment]
> @argosnothing
> 
> > #### Hide/Unhide weirdness
> > There is also behavior I don't like when it comes to multiple named workspaces toggling before and after adding a bunch of new workspaces. On hide a workspace uses it's idx at time of hide and inserts itself into that idx on unhide.
> > The problem is that what if you have
> > (all unhidden) stash1 stash2 3 4
> > then hide stash 1. Now stash1 on rehide will always insert at idx 1. But after you hide stash1 and then hide stash2, well because stash2 is now at idx 1 it uses that same idx on its own unhide meaning those temporarily hidden workspaces will randomly be at either idx 1 or 2, depending on hide and unhide order.
> > Similar strange behavior happens if you hide a workspace between visible workspaces. Hidden workspaces stick to the end of the vec, so you will never get idx gaps, but on rehide hidden workspaces will try and go to it's `original_idx`.
> 
> What about basing the insert position off of _context_ instead of their index? This is similar to how git applies patches and merge conflicts for example, instead of just "reinsert at X position" it looks for what was _around_ itself. Something like this:
> 
> ```
> struct HiddenWorkspace {
>     id: usize,
>     prev_neighbor_id: Option<usize>,
>     next_neighbor_id: Option<usize>,
> }
> 
> fn hide(visible: &mut Vec<usize>, hidden: &mut Vec<HiddenWorkspace>, id: usize) {
>     let idx = visible.find(id);
> 
>     // remember left and right neighbors
>     hidden.push(HiddenWorkspace {
>         id,
>         prev_neighbor_id: visible.get(idx - 1),
>         next_neighbor_id: visible.get(idx + 1),
>     });
> 
>     visible.remove(idx);
> }
> 
> fn unhide(visible: &mut Vec<usize>, hidden: &mut Vec<HiddenWorkspace>, id: usize) {
>     let entry = hidden.remove(id);
> 
>     // Priority 1: place after our old left neighbor, if it's still visible
>     //             (transitively walk the chain if it's also hidden)
>     if let Some(prev_id) = find_visible_neighbor(entry.prev_neighbor_id, going_left) {
>         visible.insert(after: prev_id, entry.id);
>         return;
>     }
> 
>     // Priority 2: place before our old right neighbor, if it's still visible
>     if let Some(next_id) = find_visible_neighbor(entry.next_neighbor_id, going_right) {
>         visible.insert(before: next_id, entry.id);
>         return;
>     }
> 
>     // Fallback: both neighbors are gone (e.g deleted dynamic workspace), just append
>     visible.push(entry.id);
> }
> 
> // Walks the hidden chain until it finds a neighbor that is currently visible
> fn find_visible_neighbor(start_id: Option<usize>, direction: Direction) -> Option<usize> {
>     let mut current = start_id;
> 
>     while let Some(id) = current {
>         if visible.contains(id) {
>             return Some(id);  // found a visible anchor
>         }
>         // This neighbor is also hidden — follow its neighbor in the same direction
>         current = hidden.get(id).neighbor_in(direction);
>     }
> 
>     None  // no visible anchor found
> }
> ```
> 
> This is guaranteed to "just work" by following the hidden chain as long as workspaces aren't removed/deleted, which isn't really the case for dynamic workspaces, but if it can't find their neighbouring context I think appending would be a reasonable fallback.

I actually like this more than the current system of previous id of the hidden workspace itself, because it plays nicely with the dynamic nature of niri workspaces. Currently a workspace is considered hidden through a flag on the workspace struct itself, and this would entail adding adding another field to the workspace struct. I was considering just removing the previous_id and always simply pushing to the end of the vec, meaning the field add would only be a bool representing of a workspace is currently hidden. 

I'm open to an alternative method, but my goal for this PR is for as few structural changes to workspaces as possible so there is minimal line changes.

**Comment by AmmoniumX** [comment]
@argosnothing

> I actually like this more than the current system of previous id of the hidden workspace itself, because it plays nicely with the dynamic nature of niri workspaces. Currently a workspace is considered hidden through a flag on the workspace struct itself, and this would entail adding adding another field to the workspace struct. I was considering just removing the previous_id and always simply pushing to the end of the vec, meaning the field add would only be a bool representing of a workspace is currently hidden.
> 
> I'm open to an alternative method, but my goal for this PR is for as few structural changes to workspaces as possible so there is minimal line changes.

So, would you rather the method stay as it currently is (storing previous_id), change it to be context-dependant, or just simply always append? I can volunteer to implement whichever you prefer if it would help to get the PR out of draft

**Comment by argosnothing** [comment]
> @argosnothing
> 
> > I actually like this more than the current system of previous id of the hidden workspace itself, because it plays nicely with the dynamic nature of niri workspaces. Currently a workspace is considered hidden through a flag on the workspace struct itself, and this would entail adding adding another field to the workspace struct. I was considering just removing the previous_id and always simply pushing to the end of the vec, meaning the field add would only be a bool representing of a workspace is currently hidden.
> > I'm open to an alternative method, but my goal for this PR is for as few structural changes to workspaces as possible so there is minimal line changes.
> 
> So, would you rather the method stay as it currently is (storing previous_id), change it to be context-dependant, or just simply always append? I can volunteer to implement whichever you prefer if it would help to get the PR out of draft

In my opinion always append would be the least intrusive, I'm not on niri currently ( or linux for that matter due to work issues ), so feel free to PR any changes and I can review the code to double check. 

Edit: Windows has finally made it unuseable for work vs niri and linux so i return!

**Comment by argosnothing** [comment]
this PR should now be caught up with the blur change.

**Comment by Roshan-R** [comment]
Great work @argosnothing! Just wondering: once hidden workspaces are merged, do you think window swallowing would become relatively straightforward to implement in niri?

ref: https://github.com/niri-wm/niri/discussions/762

**Comment by argosnothing** [comment]
> Great work @argosnothing! Just wondering: once hidden workspaces are merged, do you think window swallowing would become relatively straightforward to implement in niri?
> 
> ref: #762

Appreciate the ref, hidden workspaces don't interact in any way with this behavior. They simply represent a workspace that is not shown with some additional functionality to work with windows inside a hidden workspace.

**Comment by AmmoniumX** [comment]
I'm experiencing some weirdness/glitchy behavior with named workspaces, specifically when switching between them and on the overview mode, when I switch to them they appear empty for a second and then the windows flash into existance, and on overview mode they don't show up at all:


https://github.com/user-attachments/assets/d8ded082-b6c4-4cbd-8117-15dc7a41bec7

This is using the default config with the only changes being switching to named workspaces, the issue doesn't occur with the default config nor on upstream

Config used: https://gist.github.com/AmmoniumX/b4aa6a72dbd47720447dc9d3c21341e6

**Comment by barrulus** [comment]
I've been running this feature in my fork ([barrulus/biri](https://github.com/barrulus/biri), picked up via AmmoniumX/niri-qol) and found a bunch of bugs while hardening it against the rest of the layout code. All of these are verified against the current branch here (3508510). I have fixes and regression tests for all of them on my side, so ask if you want any of that.

The invariants that most of these break:

1. hidden workspaces are contiguous at the end of `Monitor::workspaces`
2. an empty unnamed workspace sits directly before the hidden block (it doubles as the normal trailing empty workspace)
3. hidden workspaces are always named. unhide/toggle look up by name, so an unnamed hidden workspace is unreachable forever

Crashes:

1. `move_workspace_to_idx` isn't hidden-aware. `old_idx` can point at a hidden workspace (reachable from IPC), and `new_idx = 0` puts it at the front. The next `clean_up_workspaces` then does `last_hidden_idx - 1` with `last_hidden_idx == 0`, usize underflow, panic. Even without the panic, moving a visible workspace to `len - 1` lands it after the hidden block.

2. `original_idx` is set at hide time and never updated while the visible region churns (cleanup removing empties, workspaces moving to other monitors). On unhide, `insert_workspace(ws, original_idx, false)` only handles `idx == len`. Hide at index 5, let the region shrink to 3, unhide: `Vec::insert` panics. If it's merely stale instead of out of range, the workspace gets inserted inside the hidden block.

3. with `empty-workspace-above-first`, the `len == 2` special case in `clean_up_workspaces` asserts both workspaces are empty and unnamed. `[empty-top, hidden-named]` is a legal state (hide the only occupied workspace) and trips the assert.

Wrong or lost workspaces:

4. `hide_workspace_by_idx(0)` with `empty-workspace-above-first`: `add_workspace_top()` runs first and shifts the target to index 1, but the function goes on to hide index 0, which is now the fresh empty unnamed workspace. So the workspace you asked to hide stays visible, and you get an unnamed hidden workspace that can never be unhidden (invariant 3).

5. unnaming a hidden workspace while no outputs are connected: the `MonitorSet::NoOutputs` arm of `unname_workspace_by_id` calls `ws.unname()` without clearing `hidden`/`needs_hidden`. When an output comes back it goes through `insert_hidden_workspace` as hidden and unnamed, unreachable forever. The Monitor path is fine because `unname_workspace` unhides first.

6. `move_workspace_to_output_by_id` can resolve `old_idx` to a hidden workspace and then `target.insert_workspace(ws, active + 1, ...)` puts a still-hidden workspace in the middle of the target's visible region. The "hidden workspaces keep their output because they're re-added to the same monitor" assumption in `remove_workspace_by_idx` doesn't hold here either, and `original_idx` still refers to the old monitor.

Invariant corruption (no immediate crash, corrupts over time):

7. `append_workspaces` (called on output removal) does `let empty = workspaces.remove(len - 1)` assuming the last workspace is the trailing empty. With a hidden block on the primary that's a hidden named workspace, which gets pushed back at the end as if it were the empty, and the incoming workspaces get spliced after the remaining hidden ones. The removed monitor's workspaces can also contain hidden ones themselves, which then land mid-vec. My fix splices incoming visible workspaces before the empty guard and re-appends incoming hidden ones at the end.

8. `move_workspace_up`/`move_workspace_down` swap blindly, so moving the last visible workspace down swaps it with the first hidden one.

9. operator precedence in `workspaces_with_render_geo_mut`:

```rust
.filter(move |(ws, geo)| !ws.hidden && !cull || geo.intersection(output_geo).is_some())
```

`&&` binds tighter than `||`, so this is `(!ws.hidden && !cull) || intersects` and any hidden workspace whose geometry intersects the output passes. The other two filters do it right, this one wants `!ws.hidden && (!cull || intersects)`.

10. `unhide_workspace_by_id` doesn't check that the workspace is actually hidden, and `unname_workspace` calls it unconditionally. So unnaming a normal visible named workspace does a full remove + reinsert with all the side effects (extra empty workspaces, active-index adjustment, workspace_switch cancelled). Unnaming the focused workspace can shift focus to the one above. An early return when `!ws.hidden` fixes it.

11. `set_workspace_name`'s "was this the last workspace" check uses `workspaces.last()`, which is now a hidden workspace, so naming the empty guard before the hidden block doesn't add a new empty and invariant 2 breaks.

Most of these weren't found by review. I made the layout tests' `verify_invariants` hidden-aware (the three invariants above) and added hide/unhide to the existing randomized op tests, and proptest found 8 and the index-0 case of 1 within seconds. Worth doing here too, I can port the invariant checks and regression tests over if that helps.

argosnothing:
> I've been running this feature in my fork ([barrulus/biri](https://github.com/barrulus/biri), picked up via AmmoniumX/niri-qol) and found a bunch of bugs while hardening it against the rest of the layout code. All of these are verified against the current branch here ([3508510](https://github.com/niri-wm/niri/commit/3508510fae37d82096e95fcd6b6c8d9fb4ff8548)). I have fixes and regression tests for all of them on my side, so ask if you want any of that.
> 
> The invariants that most of these break:
> 
> 1. hidden workspaces are contiguous at the end of `Monitor::workspaces`
> 2. an empty unnamed workspace sits directly before the hidden block (it doubles as the normal trailing empty workspace)
> 3. hidden workspaces are always named. unhide/toggle look up by name, so an unnamed hidden workspace is unreachable forever
> 
> Crashes:
> 
> 1. `move_workspace_to_idx` isn't hidden-aware. `old_idx` can point at a hidden workspace (reachable from IPC), and `new_idx = 0` puts it at the front. The next `clean_up_workspaces` then does `last_hidden_idx - 1` with `last_hidden_idx == 0`, usize underflow, panic. Even without the panic, moving a visible workspace to `len - 1` lands it after the hidden block.
> 2. `original_idx` is set at hide time and never updated while the visible region churns (cleanup removing empties, workspaces moving to other monitors). On unhide, `insert_workspace(ws, original_idx, false)` only handles `idx == len`. Hide at index 5, let the region shrink to 3, unhide: `Vec::insert` panics. If it's merely stale instead of out of range, the workspace gets inserted inside the hidden block.
> 3. with `empty-workspace-above-first`, the `len == 2` special case in `clean_up_workspaces` asserts both workspaces are empty and unnamed. `[empty-top, hidden-named]` is a legal state (hide the only occupied workspace) and trips the assert.
> 
> Wrong or lost workspaces:
> 
> 4. `hide_workspace_by_idx(0)` with `empty-workspace-above-first`: `add_workspace_top()` runs first and shifts the target to index 1, but the function goes on to hide index 0, which is now the fresh empty unnamed workspace. So the workspace you asked to hide stays visible, and you get an unnamed hidden workspace that can never be unhidden (invariant 3).
> 5. unnaming a hidden workspace while no outputs are connected: the `MonitorSet::NoOutputs` arm of `unname_workspace_by_id` calls `ws.unname()` without clearing `hidden`/`needs_hidden`. When an output comes back it goes through `insert_hidden_workspace` as hidden and unnamed, unreachable forever. The Monitor path is fine because `unname_workspace` unhides first.
> 6. `move_workspace_to_output_by_id` can resolve `old_idx` to a hidden workspace and then `target.insert_workspace(ws, active + 1, ...)` puts a still-hidden workspace in the middle of the target's visible region. The "hidden workspaces keep their output because they're re-added to the same monitor" assumption in `remove_workspace_by_idx` doesn't hold here either, and `original_idx` still refers to the old monitor.
> 
> Invariant corruption (no immediate crash, corrupts over time):
> 
> 7. `append_workspaces` (called on output removal) does `let empty = workspaces.remove(len - 1)` assuming the last workspace is the trailing empty. With a hidden block on the primary that's a hidden named workspace, which gets pushed back at the end as if it were the empty, and the incoming workspaces get spliced after the remaining hidden ones. The removed monitor's workspaces can also contain hidden ones themselves, which then land mid-vec. My fix splices incoming visible workspaces before the empty guard and re-appends incoming hidden ones at the end.
> 8. `move_workspace_up`/`move_workspace_down` swap blindly, so moving the last visible workspace down swaps it with the first hidden one.
> 9. operator precedence in `workspaces_with_render_geo_mut`:
> 
> ```rust
> .filter(move |(ws, geo)| !ws.hidden && !cull || geo.intersection(output_geo).is_some())
> ```
> 
> `&&` binds tighter than `||`, so this is `(!ws.hidden && !cull) || intersects` and any hidden workspace whose geometry intersects the output passes. The other two filters do it right, this one wants `!ws.hidden && (!cull || intersects)`.
> 
> 10. `unhide_workspace_by_id` doesn't check that the workspace is actually hidden, and `unname_workspace` calls it unconditionally. So unnaming a normal visible named workspace does a full remove + reinsert with all the side effects (extra empty workspaces, active-index adjustment, workspace_switch cancelled). Unnaming the focused workspace can shift focus to the one above. An early return when `!ws.hidden` fixes it.
> 11. `set_workspace_name`'s "was this the last workspace" check uses `workspaces.last()`, which is now a hidden workspace, so naming the empty guard before the hidden block doesn't add a new empty and invariant 2 breaks.
> 
> Most of these weren't found by review. I made the layout tests' `verify_invariants` hidden-aware (the three invariants above) and added hide/unhide to the existing randomized op tests, and proptest found 8 and the index-0 case of 1 within seconds. Worth doing here too, I can port the invariant checks and regression tests over if that helps.

This is terrific work! There have been a few bugs I were tracking and just haven't found any time in my schedule to address them. feel free to make a PR to the branch for this. Some behavior that I feel should be changed on hidden workspaces is to just completely remove the last idx and logic around that. If a hidden workspace is shown it's ultimately just simpler to put it at the end of the vec rather than try and track a relative idx.

barrulus:
> free to make a PR to the branch for this.

Sure. I will do this in the next day or so

ymolists:
Hi. 

I've been quietly monkey-guarding this PR to see where it goes. 😄

I'm coming from i3, where there is a programmatic way for external tools to determine/control what workspaces are presented through the i3bar workspace protocol. https://i3wm.org/docs/i3bar-workspace-protocol.html

It got me wondering whether a similar model could work for niri: rather than baking all of the workspace visibility logic into niri itself, could niri expose a protocol that lets an external shell script or binary determine which workspaces should be presented?

The big advantage, IMO, would be that workspace visibility could then be driven by arbitrary user-defined logic. Things like scratchpads, project-based workspaces, application categories, dynamically generated workspaces, etc. could all be implemented externally without requiring niri to grow a configuration option for every possible workflow.

I'm curious what the niri maintainers think about the general idea, particularly the trade-offs around shelling out to an external process when niri needs to determine the current list of visible workspaces. Obviously there are concerns around latency, process spawning, failure modes, and keeping the compositor responsive, so perhaps a persistent IPC mechanism would make more sense than literally spawning a process each time.

Would love to hear any and all arguments for or against this approach, or whether there's a better architectural direction I'm missing.

barrulus:
> Would love to hear any and all arguments for or against this approach, or whether there's a better architectural direction I'm missing.

I am not a niri maintainer, I run a fork that has incorporated this hidden workspace logic successfully and I have a lot of other thiings on the go around this.

Having a "protocol based" approach is just anothe rsurface that requires maintenance, no different to building it all internally really. Sure it might make it easier for some to integrate certain things, but spending the time to build a robust, reliable, protocol for compositor external interactions, is likely harder to do in the short/medium term...

argosnothing:
> love to

It's an interesting idea, if there is a standard way used by other wm's to show or hide things, this PR could be wired to use the same protocol so tools that look for it work.

argosnothing:
> > Would love to hear any and all arguments for or against this approach, or whether there's a better architectural direction I'm missing.
> 
> I am not a niri maintainer, I run a fork that has incorporated this hidden workspace logic successfully and I have a lot of other thiings on the go around this.
> 
> Having a "protocol based" approach is just anothe rsurface that requires maintenance, no different to building it all internally really. Sure it might make it easier for some to integrate certain things, but spending the time to build a robust, reliable, protocol for compositor external interactions, is likely harder to do in the short/medium term...

Yeah if said protocol isn't already a standard, this isn't going to be the place for it to be one i agree.

mikenrafter:
> > Would love to hear any and all arguments for or against this approach, or whether there's a better architectural direction I'm missing.
> 
> I am not a niri maintainer, I run a fork that has incorporated this hidden workspace logic successfully and I have a lot of other thiings on the go around this.
> 
> Having a "protocol based" approach is just anothe rsurface that requires maintenance, no different to building it all internally really. Sure it might make it easier for some to integrate certain things, but spending the time to build a robust, reliable, protocol for compositor external interactions, is likely harder to do in the short/medium term...
> 
> 
> 

Niri already exposes IPC for naming workspaces. Why wouldn't it be as simple as `niri msg action set-workspace-visibility hidden` and `niri msg action set-workspace-visibility visible`? Then `niri msg workspaces` filters for visible, keeping indices intact? (i.e. 1:visible 2:visible 3:hidden 4:visible would still show 1,2,4 for indices)

barrulus:
@mikenrafter agreed, and exactly why I push back against the i3bar model where niri asks an external process which workspaces to show. That's an outbound protocol niri would have to define, spawn or hold open, time out, and fall back from when the helper is slow or dead. That's the maintenance surface I meant. 
Its better if niri owns the flag and tools set it over the IPC as that's already there. No new surface, and it's what this PR already does: `niri msg workspaces` filters hidden ones, `is_hidden` is on the IPC struct, and `WorkspacesWithHidden` returns the full list.

The one gap is that `toggle-workspace-visibility` is only a bind action here, not a niri msg action. I added the IPC variants in my fork (barrulus/biri#26: toggle, plus idempotent hide-workspace / unhide-workspace with optional --focus). Small change if @argosnothing wants to fold it in.

On sparse indices (1,2,4): I'd keep them compacted as the PR does. `idx` is positional in niri, focus-workspace 3 means the third workspace and bars lay out by it, so a gap either breaks navigation or changes what idx means. Hidden workspaces stay reachable by name.

PR #2998, opened 2025-12-07
layout: add anchor [left|right]

markstos:
Controls whether windows are laid out left to right, the default or
right to left.

anchor "[left|right]" is a new config option with layout {},
or may be overridden by a per-output layout{} section.

Fixes: https://github.com/YaLTeR/niri/discussions/1642
Fixes: https://github.com/YaLTeR/niri/discussions/658

## Naming

There are other names which might well for this. Suggestions welcome if you have a better idea!
## Using

As noted in the commit, with this patch you an add

`anchor "right"` to the top level `layout {}` block
or have per-output overrides within an output's `layout` block.

`anchor "left"` is also supported, but is the same as not adding this
at all.

## Status

Works for basic cases. In draft state for feedback.

### Shrinking a window

If a window is resized to be smaller on an output with anchor=right, the right edge of the window should stay fixed (anchored!) and the left edge should move towards it.

### Growing the width of the output

- If you resize an output to be wider, a window that was right-anchored will lose the anchor. This is visually different behavior than left anchored. This is because the window is being widened by adding space on the right.
Not sure if we could widen the output on the left side in this case, or if we'd run into a problem with negative numbers.

### Combining with `always-center-single-column`

If using `always-center-single-column` in combining with anchor=right,
when adding a second window, the second window appear on the left as expected, and both windows will move to become right-aligned. This is nice, but differs from the anchor=left behavior. With the default anchor=left behavior, the first window centered, and add the second window to the right of it.

To me, the anchor=left behavior looks more like a bug-- as soon as there is more than one window, I would expect left or right anchoring to apply.

### Restoring state when unfullscreening.

If a right-anchored window is fullscreened and then unfullscreened, it returns to be left-anchored instead of right anchored. 

## Scope

Excluded from scope is the concept of anchoring windows
to the center. That is, a first window might
appear in the center, then when a second window appears
they move so that both are centered.

First, there really *two* center anchoring options people
might like, one where the second window appears on the right
and a second where the window appears on the left.

There are also already actions for "center-column", "center-window"
and "center-visible-columns", so if you want to you center columns,
there are already ways to do that.

### DIFF
--- a/niri-config/src/layout.rs
+++ b/niri-config/src/layout.rs
@@ -24,6 +24,7 @@ pub struct Layout {
     pub gaps: f64,
     pub struts: Struts,
     pub background_color: Color,
+    pub anchor: Anchor,
 }
 
 impl Default for Layout {
@@ -52,6 +53,7 @@ impl Default for Layout {
                 PresetSize::Proportion(2. / 3.),
             ],
             background_color: DEFAULT_BACKGROUND_COLOR,
+            anchor: Anchor::default(),
         }
     }
 }
@@ -78,6 +80,7 @@ impl MergeWith<LayoutPart> for Layout {
             default_column_display,
             struts,
             background_color,
+            anchor,
         );
 
         if let Some(x) = part.default_column_width {
@@ -126,6 +129,8 @@ pub struct LayoutPart {
     pub struts: Option<Struts>,
     #[knuffel(child)]
     pub background_color: Option<Color>,
+    #[knuffel(child, unwrap(argument))]
+    pub anchor: Option<Anchor>,
 }
 
 #[derive(knuffel::Decode, Debug, Clone, Copy, PartialEq)]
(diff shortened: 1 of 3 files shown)


yngvildr-the-voracious:
Hello, just to ask : how easy is it to make your work (should it work in the coming weeks or months, depending on your availability or that or other devs) capable of being set by the user in a single config.kdl line ? 
Is there any way I can help as someone with very bare bones coding abilities (mostly in nodejs api handling and basic bash, but I would love to learn)

markstos:
@yngvildr-the-voracious I've switched to [scroll](https://github.com/dawsers/scroll) am not developing this further. Scroll doesn't have an option to reverse the layout direction either, but it does have an option to align any window to the left or right side of an output, so I use that. 

> Is there any way I can help as someone with very bare bones coding abilities (mostly in nodejs api handling and basic bash, but I would love to learn)

Yes, you can try with a vibe coding tool.

markstos:
A reflection I have after trying this approach is that the idea that the first entry in an array being the left-most window is deeply engrained in Niri. An ideal solution would probably be designed differently than what I have here and might be considerably more work: It might go back to the fundamental question of how you design data structures and operations if you were designing the system to support both LTR and RTL in a fundamentally equal way. Then rework whatever code needs to change to adapt to that design.

One idea might be not start window numbers at 0 but at some very high number, like 1000 or more, so you could still naturally add windows to the "left" of this array index (with lower numbers), before you hit zero and run out of indexes. 
But in that case you are decrementing an index when adding a window on the left and incrementing to add a window to the right.  

@YaLTeR would have a much better idea about the ideal structure for this then I think I would.

PR #3004, opened 2025-12-08
feat: allow customising the color of the fullscreen backdrop

Rolv-Apneseth:
Small PR to allow customising the colour of the fullscreen backdrop, instead of just having a hard-coded black colour. Figured this could be a potential solution for people (like me) who want transparency for fullscreen windows (see #1399), while also giving a bit more customisation to others too. Unfortunately, using a transparent colour lets you see that layer shells like `waybar` don't get hidden, but I believe handling that should be a separate PR.

Totally understand if you don't want this as a feature for niri, no hard feelings. If you are open to it though, just let me know if you want any changes made.

**Edit**: Forgot to briefly describe how this is actually achieved:

Added a `fullscreen-backdrop-color` option under `layout`, which can be set to any colour (including a transparent one):

```kdl
layout {
    fullscreen-backdrop-color "transparent"
}
```

The backdrop colour also updates when the config is saved, as would be expected.


saygo-png:
Tested the patch and it works well. Thanks!

src/layout/tile.rs
```diff
@@ -225,6 +227,11 @@ impl<W: LayoutElement> Tile<W> {
         if self.options.layout.preset_window_heights != options.layout.preset_window_heights {
             self.floating_preset_height_idx = None;
         }
+        // Update backdrop colour
+        let new_color = Color32F::from(options.layout.fullscreen_backdrop_color);
+        if new_color != self.fullscreen_backdrop.color() {
+            self.fullscreen_backdrop.set_color(new_color);
+        }
```
helcoume: This can be simplified to:
```rust
self.fullscreen_backdrop
    .set_color(options.layout.fullscreen_backdrop_color);
```

A no-op guard is already implemented in `SolidColorBuffer::set_color()`, and `Color` implements `Into<Color32F>`.
This is consistent with how `background_color` updates are handled in `Workspace::update_config()`.
As a result, `use smithay::backend::renderer::Color32F` would no longer be needed in this file.

Rolv-Apneseth: Ah, nice, thank you

Rolv-Apneseth:
Rebased and merged commits into just 1. Let me know if anything else is required for this patch, happy to make changes.

workibear:
Screenshot from 2026-07-19 18-53-48
Screenshot from 2026-07-19 18-53-55
I also encountered the same situation, but I found that in addition to full screen, a black background is also injected when using the launcher or opening the configuration. Is it the same problem?  If yes, how to alleviate it?  Any guidance would be greatly appreciated.

Rolv-Apneseth:
> I also encountered the same situation, but I found that in addition to full screen, a black background is also injected when using the launcher or opening the configuration. Is it the same problem? If yes, how to alleviate it? Any guidance would be greatly appreciated.

I'm not entirely sure but maybe? I'm actually waiting for this to be merged to give Niri another shot, since this and blur were some of my bigger issues last time.

As for how to alleviate it: give this PR a shot (if you know how). I'm sure more people confirming it works as intended would help with reviewing it.

workibear:
> > I also encountered the same situation, but I found that in addition to full screen, a black background is also injected when using the launcher or opening the configuration. Is it the same problem? If yes, how to alleviate it? Any guidance would be greatly appreciated.
> 
> I'm not entirely sure but maybe? I'm actually waiting for this to be merged to give Niri another shot, since this and blur were some of my bigger issues last time.
> 
> As for how to alleviate it: give this PR a shot (if you know how). I'm sure more people confirming it works as intended would help with reviewing it.

Thank you for your reply. I tried your repair branch. It is indeed a good idea to customize the background color. However, there is a problem that my status bar is still displayed after using the transparent branch. I know that it can be solved by configuring smart hiding of the status bar, but I think adding a mask to cover it is indeed a compromise solution.

In addition, the black background in my terminal is another application behavior. The principle should be the same as the black background inserted in the full screen, but it can be modified through configuration. I have modified it to an effect that satisfies me, like this
image
image

Reference link https://docs.noctalia.dev/v5/compositor-settings/niri/?section=blur#blur

PR #3005, opened 2025-12-08
Update issue template to put niri config into `<details>` block

DerRockWolf:
Since using Niri and reading through quite a few issues, I've found that I often scroll entirely past the config. From a user's point of view, I'm usually not interested in the reporter's entire config.

This makes issues much more readable and prevents readers from needing to scroll all the way past the config.


YaLTeR:
Makes sense, thanks

PR #3008, opened 2025-12-09
wiki/include: add path resolution section

imnotpoz:
I couldn't find anything related to this in the code so I tested the behaviour manually

tested on commit ba29735, latest at the time of writing

### DIFF
--- a/docs/wiki/Configuration:-Include.md
+++ b/docs/wiki/Configuration:-Include.md
@@ -28,6 +28,37 @@ layout {
 }
 ```
 
+### Path resolution
+
+Paths of included files are relative to the directory of the current
+config file, unless they start with `/` - then they're absolute:
+
+```kdl,must-fail
+// /tmp/config.kdl
+
+include "foo.kdl" // `/tmp/foo.kdl`
+
+include "bar/foo.kdl" // `/tmp/bar/foo.kdl`
+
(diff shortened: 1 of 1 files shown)


docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,31 @@ layout {
 }
 ```
 
+### Path resolution
+
+Paths of included files are relative to the directory
+your main config file is in, unless they start with `/`:
```
HigherOrderLogic: I think this should match [`PathBuf::push` method's doc](https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html#method.push), which is used to resolve the path.

YaLTeR: But this is wrong? They are relative to the current config's directory...

imnotpoz: I added a link to that function and explain the behaviour

imnotpoz: thank you, fixed

YaLTeR: no why would you link technical rust documentation from the niri wiki in this case

imnotpoz: I changed it to the initial version but now correctly describing that it's relative to the current file's directory

PR #3016, opened 2025-12-10
[WIP] Add Hot Corner Pressure

### LINKED ISSUE #1473: Hot corner with barriers
Currently, hot corner is a simple implementation which triggers when the mouse lands on the pixel at x=1 y=1 of the monitor. It would be better to instead use a barrier implementation similar to what GNOME Shell does, where you need to move the mouse a bit "against" the corner to make it trigger. This will help to avoid some misfires.

https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/6d86bd0dfa62ff2b0b1d21d4be90aa29803e3fd0/js/ui/layout.js?page=2#L1164

Fireye04:
fixes #1473 

This PR adds pressure to hot corners, requiring users to both mouse over the hot corner and also give a certain amount of pressure to cause it to trigger, reducing accidental activation. 

Reviews, test feedback, and comments are, as always, more than welcome.

## Current Design

In summary, `PointerMotionEvent` deltas are being read and added to a `hot_corner_pressure` variable, which is then used to measure total pressure against the hot corner. 

There are a few values that you can tweak to get the pressure feeling right, listed here:

- dividend: This is the big one, basically it divides the combined x and y deltas to get the absolute pressure. You can get more extreme values by decreasing this number, or less extreme ones by increasing it.
- pressure_cap: This should change as a function of the dividend. I've found that I want the pressure cap to be reachable if you take a big swipe at the hot corner, but only really in that scenario. This allows you to trigger the hot corner with a large swipe, which is something that I find I do regularly, but was something the original design made significantly more difficult. (Sidenote, but this behavior can be removed if it's found to be generally undesirable)
- target: This is the threshold that the total pressure needs to reach in order to trigger the hot corner. Eg: If your dividend goes down, this will trip faster.

I don't think I've got the balance all the way down at the moment, but it's close, as tested on my laptop's track pad. I'd like to get a more complete idea of how it interacts with different hardware though, so please feel free to test on whatever you happen to have and let me know how things feel and what settings you prefer.

## Disclaimers

A couple of things to take care of before i'd consider this thing usable, as well as some things of note:

First off, this must be tested off the TTY. I'm currently running off `nix develop` and using the following command to obtain program logs `cargo run -- --session 2>&1 | tee logs_tty.txt`. (this is how you can tell if you're hitting your pressure cap)

Currently it's only implemented for the bottom left hot-corner (I wanted to tune the pressure settings before handling the boilerplate). Theoretically I could swap to an absolute value implementation, but that might be a bit odd for multiple monitor setups so I think I'm leaning towards a more rigid implementation here.

Further, the pointer is not currently being captured on entering a hot corner, which again might create problems for multi-monitor setups where the hot corner is placed between monitors. I'm not sure if this is a desirable behavior, but I reckon it's something good to include unless otherwise determined.

Finally, we're smack dab in the middle of my uni's finals season, so please forgive me if I end up taking a second to finish polishing this thing up. 

## TODO
- [x] add event stream
- [ ] capture & release pointer
- [ ] implement all four corners (whoaw!)
- [ ] finalize default tuning
- [ ] Implement config
- [ ] Add docs


Fireye04:
Alright did some digging, and studied some of gnome-shell's [homework](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/6d86bd0dfa62ff2b0b1d21d4be90aa29803e3fd0/js/ui/layout.js?page=2#L1164); gonna document my findings here, so a) I don't forget and b) people can voice opinions on the implementation.

### gnome-shell implementation
- Two pressure barriers are monitored: vertical pressure and horizontal pressure.
- If either is triggered, the hot corner triggers, regardless of the state of the other.
- the horizontal barrier discards an input if the vertical pressure is greater than the horizontal, and vice versa for the vertical barrier.
- A pressure event stream is maintained. Events beyond the timeout are culled, and no longer contribute pressure to the threshold.
- Total pressure per-event is capped at a constant, but is allowed to be lower (this is currently implemented). If the running total after the current frame's cull and pressure increment is greater than the threshold, the pressure barrier triggers.

As far as I can tell, the dual barrier approach only really serves to take the greater axis of input, rather than combining the two into a single vector, though please correct me if I'm wrong.

Fireye04:
Alright, still very much WIP, but we now have an event stream in, and mimic the vertical/horizontal pressure barriers from gnome shell. Feels quite good at the moment, but still probably needs a bit of tuning.

Issue #3012, opened 2025-12-09
`niri validate -c $CONFIG_INCLUDING_FILE_WHICH_ENOENT_NOW` fails


johnrichardrinehart:
**First, I'm unsure if this is a _bug_ or a feature request. So, sorry if I'm misclassifying.**

If I have a configuration file, `/etc/niri/config.kdl` like
```kdl
include "/tmp/enoent_so_far.kdl"
```
which describes an inclusion of a file which does not yet exist, but will at various points during `niri`'s runtime (as I hack on a config) then I'd like `niri validate -c /etc/niri/config.kdl` to not fail around [here](https://github.com/YaLTeR/niri/blob/83a733e08512d9a6ff5332ef53acc0e8e635c2ea/niri-config/src/lib.rs#L371-L376). It [looks like the](https://github.com/YaLTeR/niri/blob/83a733e08512d9a6ff5332ef53acc0e8e635c2ea/src/utils/watcher.rs#L76) `watcher` logic already supports gracefully handling missing configuration files. But, I'm not exactly sure. That's why I'm filing this as a potential "bug". But, it depends on maintainers' expected/desired behavior.

I think a _warning_ emitted from `niri validate` makes sense. I don't think an _error_ does, though, especially if both: 1) people block `niri` changes on valid configuration (exit code 0 from `niri validate`) and 2) `watcher` is currently satisfied if configuration files in the `include` tree don't exist.

The context for this is that I'm using NixOS in a way that makes my default `niri` configuration immutable at runtime. So, I want to add a line at the bottom like
```
include "/tmp/niri.kdl"
```
which allows for creating and modifying a file at runtime to play around with different configuration. And, I also want to block system configuration changes based on valid configuration. So, I'm kind of stuck unless I add a `tmpfile` rule/activation for that path. But, I think it might make more sense to tweak `niri`'s configuration validation logic to warn instead of failing in this case. I.e. remove a currently broken code path where it might ought proceed (changing behavior so that things don't fail when they ought not to sounds like "bug fixing").

Alternatively, if users want to be guaranteed at `niri validate`-time that their configuration file hierarchy is valid/exists, then maybe we could support a variant like `?include` or `maybe_include` which won't fail if something is missing, while `include` would be a "strict" include. But, that increases complexity so I'm a priori opposed to that change.

Sempyos:
[...]

PR #3022, opened 2025-12-10
feat: allow optional includes

johnrichardrinehart:
Cf. #3012 

This seems to have the desired behavior, regarding optional configuration. It may be desirable to drop the `tracing::warn!`. But, I figured it doesn't hurt to give users the feedback if they explicitly run `niri validate`. No toast/notification appears on log-in which would otherwise be a bit scary.

This seems to work in my system configuration https://github.com/johnrichardrinehart/JohnOS/commit/b75129a2cfad7a7dcd53cbbfc7d4083077e91300. I also tested this ex situ (first). `optional=false` fails if the file doesn't exist. `optional=true` succeeds. Omitting is the same as `optional=false`.

Also, no changes needed to be made to the watcher logic, apparently. My "included" configuration changes were picked up within 500ms of `rm /tmp/niri.kdl && mv /tmp/niri.kdl.bkp /tmp/niri.kdl` and removed within 500ms of `mv /tmp/niri.kdl /tmp/niri.kdl.bkp`. So, all should be good.

Just let me know about dropping that `warn!` commit.

# Justification/Rationale

1. Many users describe their system configuration in a way that makes my default `niri` configuration immutable at runtime. So, if they want to do things like
    ```
    include "/tmp/niri.kdl"
    ```
    they can't unless `/tmp/niri.kdl` already exists. And, this file may be used only for development purposes so it doesn't make sense to ensure that it always exists.

1. This is a small patch to the grammar and behavior of `niri`, available only under opt-in conditions. So, it should affect a relatively small subset of users (low support burden)

1. This has already been requested [independently](https://github.com/YaLTeR/niri/issues/3012) in a [few places](https://github.com/YaLTeR/niri/pull/2482#issuecomment-3352515217). And, this PR (description) has only positive responses, so far. So, let's give the people what they want!

### DIFF
--- a/docs/wiki/Configuration:-Include.md
+++ b/docs/wiki/Configuration:-Include.md
@@ -118,8 +118,8 @@ window-rule {
 
 <sup>Since: next release</sup>
 
-By default, including a nonexistent file will cause an error. You can allow
-nonexistent includes by setting `optional=true`:
+By default, including a nonexistent file will cause an error.
+You can allow nonexistent includes by setting `optional=true`:
 
 ```kdl,must-fail
 // Won't fail if this file doesn't exist.
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -294,7 +294,10 @@ where
                     // Parse the path argument
                     let mut iter_args = node.arguments.iter();
                     let path_val = iter_args.next().ok_or_else(|| {
-                        DecodeError::missing(node, "additional argument for include path is required")
+                        DecodeError::missing(
+                            node,
+                            "additional argument for include path is required",
+                        )
                     })?;
                     let path: PathBuf = knuffel::traits::DecodeScalar::decode(path_val, ctx)?;
 


kchibisov:
IIRC some projects that have 'config include`s, basically ignore the error if imported path doesn't exist, so the extra config syntax is probably not needed?

YaLTeR:
I don't want to ignore include errors by default since they mean a broken config (especially in the middle between editing, etc.); the syntax in this PR is my suggestion.

kchibisov:
I mean, the fact that config exists or not, not the error inside the config.

YaLTeR:
Yeah, I know. But for things like moving configuration around or renaming files where you first rename (leaving you with an include pointing at a missing file) and then edit the include, etc.

kchibisov:
if you move the file with `optional` keyword you'll have the same problem.

YaLTeR:
Yes but in this case you specifically asked for it so you know it can happen

johnrichardrinehart:
There's also a warning emitted to stderr in the case it doesn't exist. So, it's not treated the same as silently ignoring that it doesn't exist. I think requiring users to opt in to something that may relatively-silently succeed is good.

@YaLTeR what would you like to see before this is merged?

YaLTeR:
Idk, some more justification on why this is needed in niri, rather than solving it elsewhere I guess

kchibisov:
globs can kind of solve it as well without changing much, but I don't know if niri has them already for something to re-use logic.

YaLTeR:
Nah we definitely are not doing globs, that's hell for file watching

kchibisov:
Kind of, yeah, though, you generally watch dir and not file, so it's not that bad.

johnrichardrinehart:
> Idk, some more justification on why this is needed in niri, rather than solving it elsewhere I guess

There are lots of ways to skin a cat, so this isn't _needed_. But, it's low-cost (to express originally and to maintain later) and people seem to want it and other solutions are awkward. So, I added a section in the PR description that motivates acceptance. Please let me know if you need/want anything else.

niri-config/src/lib.rs
```diff
@@ -291,7 +291,21 @@ where
...
+                    })?;
+                    let path: PathBuf = knuffel::traits::DecodeScalar::decode(path_val, ctx)?;
+
+                    // Parse the optional property
+                    let mut optional = false;
+                    for (name, val) in &node.properties {
+                        if &***name == "optional" {
+                            optional = knuffel::traits::DecodeScalar::decode(val, ctx)?;
+                        }
+                    }
```
YaLTeR, line 328: Needs to fail on unrecognized arguments and properties

niri-config/src/lib.rs
```diff
@@ -369,10 +383,16 @@ where
                             }
                         }
                         Err(err) => {
-                            ctx.emit_error(DecodeError::missing(
-                                node,
-                                format!("failed to read included config from {path:?}: {err}"),
-                            ));
+                            if optional && err.kind() == std::io::ErrorKind::NotFound {
+                                // Warn about missing optional includes
+                                warn!("optional include not found: {path:?}");
```
YaLTeR, line 418: If you have a missing optional include (e.g. the "private config file in public dots" case) will this print on every config reload? Seems like in this case it should only print when a particular optional include first becomes missing. So e.g. continued reloads of the same config due to changes elsewhere don't warn repeatedly, but reloading the config adding a new missing optional include should still warn once for it.

YaLTeR (changes requested):
Also please document this on the wiki

niri-config/src/lib.rs
johnrichardrinehart: > If you have a missing optional include (e.g. the "private config file in public dots" case) will this print on every config reload? 

It'll print in the logs repeatedly, yeah. That's what I expected, but I also tested with:
```
$ date; niri msg action load-config-file; date; niri msg action load-config-file
Thu Dec 18 09:33:58 AM EST 2025
Thu Dec 18 09:33:58 AM EST 2025
```
in one shell while 
```
$ journalctl --user -u niri --follow 
<...>
Dec 18 09:33:58 framie niri[130340]: 2025-12-18T14:33:58.557094Z  WARN niri_config: optional include not found: "/tmp/niri.kdl"
Dec 18 09:33:58 framie niri[130340]: 2025-12-18T14:33:58.557129Z DEBUG niri_config: loaded config from "/etc/niri/config.kdl"
Dec 18 09:33:58 framie niri[130340]: 2025-12-18T14:33:58.579590Z  WARN niri_config: optional include not found: "/tmp/niri.kdl"
Dec 18 09:33:58 framie niri[130340]: 2025-12-18T14:33:58.579631Z DEBUG niri_config: loaded config from "/etc/niri/config.kdl"
```

> Seems like in this case it should only print when a particular optional include first becomes missing. So e.g. continued reloads of the same config due to changes elsewhere don't warn repeatedly, but reloading the config adding a new missing optional include should still warn once for it.

Hm, I disagree. But, I'm happy to implement this if you insist. The reason I disagree is that, as a user, I want reloading to be a sort of pure behavior - I don't want configuration loading behavior to depend on `niri` state. I'd rather have more information than less and if my `niri` session has been running for days (assumedly) generating enough logs that the warning is out of view I may want to be reminded of it.
 
I'd prefer to limit noise in the logs only after people complain that they're too verbose. But, if you _do_ insist on this, then I'd plan to implement this using an in-memory `Vec<std::fs::File>` (so, users would see this warning across `niri` process restarts and we wouldn't be trying to optimize lookup using something like a `HashMap` because I only expect <dozen layered configuration files in practice).

niri-config/src/lib.rs
johnrichardrinehart: I tried to implement this in b43236af1882df9ff7be81906a580823c6b83fd8. Please let me know if you'd like a different change set to implement this.

niri-config/src/lib.rs
YaLTeR: Hm, well, alright, let's keep it warning every time for now.

niri-config/src/lib.rs
YaLTeR: Seems ok

niri-config/src/lib.rs
johnrichardrinehart: Let's goooooooooo. Thanks for working with me on this!

johnrichardrinehart:
> Also please document this on the wiki

This should happen after the merge, no, since it's out of VCS? Sorry, I don't have much/any experience with GitHub Wikis.

YaLTeR:
It is in fact all in the VCS: https://github.com/YaLTeR/niri/blob/main/docs/wiki/Configuration%3A-Include.md

johnrichardrinehart:
> It is in fact all in the VCS: https://github.com/YaLTeR/niri/blob/main/docs/wiki/Configuration%3A-Include.md

Ah, okay. My misunderstanding, thanks. I've made a Wiki change in 08a87d576f5a81229512c977a5e80f0b64815e5c. Please review when you get a minute.

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
 }
 ```
 
+### Optional includes
+
+<sup>Since: [#3022](https://github.com/YaLTeR/niri/pull/3022)</sup>
```
YaLTeR:
```suggestion
<sup>Since: next release</sup>
```

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
 }
 ```
 
+### Optional includes
```
YaLTeR: I think maybe put this section after Positionality and before Merging

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
...
+
+<sup>Since: [#3022](https://github.com/YaLTeR/niri/pull/3022)</sup>
+
+You can mark includes as optional to prevent errors when the included file doesn't exist:
```
YaLTeR:
```suggestion
By default, including a nonexistent file will cause an error.
You can allow nonexistent includes by setting `optional=true`:
```

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
...
+
+```kdl,must-fail
+// Won't fail if this file doesn't exist
```
YaLTeR:
```suggestion
// Won't fail if this file doesn't exist.
```

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
...
+include optional=true "optional-config.kdl"
+
+// Regular include - will fail if file doesn't exist
```
YaLTeR:
```suggestion
// Regular include, will fail if the file doesn't exist.
```

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
...
+include "required-config.kdl"
+```
+
+When an optional include file is missing, niri will emit a warning in the logs on every config reload.
+This allows you to be notified that the file is missing while still allowing the config to load successfully.
```
YaLTeR:
```suggestion
This reminds you that the file is missing while still loading the config successfully.
```

docs/wiki/Configuration:-Include.md
```diff
@@ -28,6 +28,28 @@ layout {
...
+
+The optional file is still watched for changes, so if you create it later, the config will automatically reload and apply the new settings.
+
+Note that `optional` only affects whether a missing file causes an error.
+If the file exists but contains invalid syntax or other errors, those errors will still be reported normally.
```
YaLTeR:
```suggestion
If the file exists but contains invalid syntax or other errors, those errors will still cause a parsing failure.
```

docs/wiki/Configuration:-Include.md
johnrichardrinehart: No problem.

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Changed in 25c449a38c005326af6bd49204a4491d859bc6af

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Moved in 25c449a38c005326af6bd49204a4491d859bc6af

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Included in 25c449a38c005326af6bd49204a4491d859bc6af

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Added a period in 25c449a38c005326af6bd49204a4491d859bc6af

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Copy-pasted in 25c449a38c005326af6bd49204a4491d859bc6af

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Changed in 25c449a38c005326af6bd49204a4491d859bc6af

docs/wiki/Configuration:-Include.md
johnrichardrinehart: Cf. 25c449a38c005326af6bd49204a4491d859bc6af

johnrichardrinehart:
@YaLTeR Any outstanding concerns?

YaLTeR:
Thanks

PR #3045, opened 2025-12-15
Use Grab&Grabbing cursors for interactive move

kchibisov:
There was no real indication that something can be dragged and thus it's generally harder to discover for someone not familiar with Mod+LMB to start dragging window around.

--

I was surprised that `Move` is usually linked to `Default`. The `Grab` might be missing, however `Grabbing` is rather common. Not feeling strong about this one, mostly a preference.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2859,12 +2859,20 @@ impl State {
                         let start_data = PointerOrTouchStartData::Pointer(start_data);
                         let icon = CursorIcon::Grabbing;
                         if let Some(grab) =
-                            MoveGrab::new(self, start_data, window.clone(), false, icon)
+                            MoveGrab::new(self, start_data, window.clone(), false, Some(icon))
                         {
                             pointer.set_grab(self, grab, serial, Focus::Clear);
-                            self.niri
-                                .cursor_manager
-                                .set_cursor_image(CursorImageStatus::Named(icon));
+
+                            // Set the cursor to Grabbing right away for Mod+LMB since it doesn't
+                            // do any other gesture.
+                            //
+                            // In the overview, we click to activate window and close the overview,
+                            // in this case setting the cursor right away would be distracting.
+                            if !is_overview_open {
+                                self.niri
+                                    .cursor_manager
+                                    .set_cursor_image(CursorImageStatus::Named(icon));
+                            }
                         }
                     }
                 }
@@ -4112,13 +4120,9 @@ impl State {
                         location: pos,
                     };
                     let start_data = PointerOrTouchStartData::Touch(start_data);
-                    let icon = CursorIcon::Grabbing;
-                    if let Some(grab) = MoveGrab::new(self, start_data, window.clone(), true, icon)
+                    if let Some(grab) = MoveGrab::new(self, start_data, window.clone(), true, None)
                     {
                         handle.set_grab(self, grab, serial);
-                        self.niri
-                            .cursor_manager
-                            .set_cursor_image(CursorImageStatus::Named(icon));
                     }
                 }
 
--- a/src/input/move_grab.rs
+++ b/src/input/move_grab.rs
@@ -43,7 +43,7 @@ impl MoveGrab {
         start_data: PointerOrTouchStartData<State>,
         window: Window,
         enable_view_offset: bool,
-        move_icon: CursorIcon,
+        move_icon: Option<CursorIcon>,
     ) -> Option<Self> {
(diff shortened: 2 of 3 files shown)


YaLTeR:
What do other compositors do?

In Mutter, if I'm not mistaken, moving windows by titlebars uses the default cursor, while moving with Mod+LMB uses the (nonexistent) move cursor: https://gitlab.gnome.org/GNOME/mutter/-/blob/3a41f04f19c1fcd4d4c93ff49b87b7515a9b5f54/src/compositor/meta-window-drag.c#L349-355

kchibisov:
Sway uses `Grab` for both start of grab and actual moving, which looks rather strange for moving. The most intuitive looking is grab(before actual move)->grabbing(holding the thing you move). Especially given that niri has resistance before actually starting grabbing window.

It seems like macOS uses the default cursor, so that probably explains why `Move` is aliased as `Default`?

kchibisov:
Hyprland seems to use `Grabbing`, but it doesn't have that resistance before starting moving. Just generally using `Grabbing` would be also fine, imho.

YaLTeR:
> It seems like macOS uses the default cursor, so that probably explains why `Move` is aliased as `Default`?

Seems to be because `Move` is intended to mean "DnD and we will move the file to the destination" (as opposed to copy the file) but everyone is misusing it as grabbing. See links here: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/commit/aec748ad276aa6ca84fafddcd091387c7954822a

`Grabbing` makes sense I guess, at least for Mod+LMB (and maybe in the overview). Not sure about dragging by the window titlebar since I think the default cursor is generally used here.

`Grab` seems to be more of a hover cursor: "drag-and-drop: something can be grabbed" (from [cursor-shape](https://wayland.app/protocols/cursor-shape-v1#wp_cursor_shape_device_v1:enum:shape)).

kchibisov:
>Grab seems to be more of a hover cursor: "drag-and-drop: something can be grabbed" (from [cursor-shape](https://wayland.app/protocols/cursor-shape-v1#wp_cursor_shape_device_v1:enum:shape)).

I guess it's just grabbing straight away then and for that patch I'd need to just put `grabbing` on `Mod+LMB` start and drop the code from `begin_move`?

YaLTeR:
Yeah, let's try that. On Mod+LMB and on LMB in the overview (might be the same place in the code).

src/input/mod.rs
```diff
@@ -4106,8 +4112,13 @@ impl State {
...
                     };
                     let start_data = PointerOrTouchStartData::Touch(start_data);
-                    if let Some(grab) = MoveGrab::new(self, start_data, window.clone(), true) {
+                    let icon = CursorIcon::Grabbing;
+                    if let Some(grab) = MoveGrab::new(self, start_data, window.clone(), true, icon)
+                    {
                         handle.set_grab(self, grab, serial);
+                        self.niri
+                            .cursor_manager
+                            .set_cursor_image(CursorImageStatus::Named(icon));
```
kchibisov: Don't quite get what this place is, since overview is handled above.

src/input/move_grab.rs
```diff
@@ -27,6 +27,7 @@ pub struct MoveGrab {
     window: Window,
     gesture: GestureState,
     enable_view_offset: bool,
+    move_icon: CursorIcon,
```
kchibisov, line 30: This change is rather conservative. I'd assume that you generally can not change cursor shape while grabbing, etc, so it could be possible to do it on the caller side, like in other similar places.

src/input/mod.rs
```diff
@@ -2857,8 +2857,14 @@ impl State {
...
                         let start_data = PointerOrTouchStartData::Pointer(start_data);
-                        if let Some(grab) = MoveGrab::new(self, start_data, window.clone(), false) {
+                        let icon = CursorIcon::Grabbing;
+                        if let Some(grab) =
+                            MoveGrab::new(self, start_data, window.clone(), false, icon)
+                        {
                             pointer.set_grab(self, grab, serial, Focus::Clear);
+                            self.niri
+                                .cursor_manager
+                                .set_cursor_image(CursorImageStatus::Named(icon));
```
kchibisov: This is for instant feedback.

YaLTeR:
Did a few changes:
- Removed setting cursor for touch move since it's hidden there anyway
- Made instant feedback only on Mod+LMB and not in the overview because there you click to activate and you don't want cursor change on that

YaLTeR:
Thanks

PR #3054, opened 2025-12-16
feat: improve primary gpu selection

yuxqiu:
This pull request ports and enhances [Smithay's primary GPU selection logic](https://docs.rs/smithay/latest/src/smithay/backend/udev.rs.html#237-265) to better suit niri's requirements.

It addresses issues on setups such as Asahi Linux with DisplayLink docks, where no `boot_vga` attribute exists. Previously, on my setup, `udev::primary_gpu` would always fall back to the `all_gpus` path, which in turn always selected `card0`, a device without an associated render node. This forced me to manually specify `render-drm-device` in the configuration. (I'm unsure whether this change will help users with a touch bar though.)

Nevertheless, this PR explicitly documents the priority order of the criteria used to select the primary GPU, which may also be useful as documentation.

My main concern is whether this change belongs upstream in Smithay or should remain in niri.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -4,7 +4,7 @@ use std::fmt::Write;
 use std::iter::zip;
 use std::num::NonZeroU64;
 use std::os::fd::{AsFd, OwnedFd};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks. Hmm, I think this would be useful in Smithay itself. Do you mind opening a PR there too?

yuxqiu:
> Thanks. Hmm, I think this would be useful in Smithay itself. Do you mind opening a PR there too?

Sure

YaLTeR:
Included in 7ca46b44b2910d49b612fd06859560a4db2c1528

yuxqiu:
> Included in [7ca46b4](https://github.com/YaLTeR/niri/commit/7ca46b44b2910d49b612fd06859560a4db2c1528)

Thanks!

PR #3076, opened 2025-12-20
config: add hide-pointer-by-default screenshot option

EmmetZ:
Add a global option `hide-pointer-by-default` to set whether the pointer is displayed when taking a screenshot. But you can still override this in keybindings or IPC.

This pr breaks the original `screenshot-path` config and moves it to `screenshot { path }`, I'm not sure if this is appropriate.

### DIFF
--- a/docs/wiki/Configuration:-Miscellaneous.md
+++ b/docs/wiki/Configuration:-Miscellaneous.md
@@ -9,7 +9,9 @@ spawn-sh-at-startup "qs -c ~/source/qs/MyAwesomeShell"
 
 prefer-no-csd
 
-screenshot-path "~/Pictures/Screenshots/Screenshot from %Y-%m-%d %H-%M-%S.png"
+screenshot {
+    path "~/Pictures/Screenshots/Screenshot from %Y-%m-%d %H-%M-%S.png"
+}
 
 environment {
     QT_QPA_PLATFORM "wayland"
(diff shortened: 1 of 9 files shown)


ToxicMushroom:
Breaking configs is not appropriate imo.

HigherOrderLogic:
> Breaking configs is not appropriate imo.

It is, in fact, against the [breaking change policy](https://yalter.github.io/niri/Configuration%3A-Introduction.html#breaking-change-policy).

Fireye04:
This looks like a great little proposal! 

Though, as a rule, we want to avoid breaking config changes if at all possible, and if unavoidable, we make sure old configs can still build on newer versions. 

What that would mean here, is probably continuing support for `screenshot-path`, while also adding an override for it in your new `screenshot { path }`, if that makes sense.

ToxicMushroom:
Also how does this this new option help, you can already [configure](https://yalter.github.io/niri/Configuration%3A-Key-Bindings.html#screenshot-screenshot-screen-screenshot-window) this in the bind and the IPC for screenshots itself no ?

Fireye04:
Ooh I didn't know about that, very nifty!

Yeah in that case it appears this is already implemented, unless there's some sort of usecase not covered by the bind. (I can't think of any though)

EmmetZ:
You're right, that's against the rules. My fault. I'll be sure to read the guidelines more closely next time.

PR #3105, opened 2025-12-23
Implement hyprland-lock-notify-v1

### LINKED ISSUE #2573: Hypridle reporting that Niri does not implement hyprland-lock-notify-v1
hypridle is logging:
```
[WARN] Compositor is missing hyprland-lock-notify-v1!
```

This seems to prevent hypridle from knowing when the lock command has been executed and as a result when suspending the machine it doesn't get to lock the screen before suspending.

This then means that when the machine is awoke it is briefly not locked and any sensitive information on the screen is visible for a short time period before the lock screen starts.

The documentation on this protocol can be found [here](https://wayland.app/protocols/hyprland-lock-notify-v1)

### System Information

* niri version: `niri unstable 2025-10-02 (commit ba6e5e082a79901dc89b0d49c5da1b769d652aec)`

* Distro: NixOS

* GPU: AMD Radeon 6900XT

* CPU: AMD Ryzen 9 7950X

jbms:
Fixes https://github.com/YaLTeR/niri/issues/2573

### DIFF
--- a/resources/hyprland-lock-notify-v1.xml
+++ b/resources/hyprland-lock-notify-v1.xml
@@ -1,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="hyprland_lock_notify_v1">
+  <copyright>
+    Copyright © 2025 Maximilian Seidler
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions are met:
+
+    1. Redistributions of source code must retain the above copyright notice, this
+       list of conditions and the following disclaimer.
+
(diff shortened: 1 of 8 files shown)


YaLTeR:
Thanks but I don't think we need this, things already work fine without it

jbms:
The hypridle on_unlock_cmd option doesn't work without this protocol.  With this protocol it works (confirmed through testing).  Other features of hypridle do work without it, though.

jbms:
> Thanks but I don't think we need this, things already work fine without it

Is there an alternative protocol that hypridle could use to receive lock and unlock events instead?

YaLTeR:
We notify the locked hint over the logind D-Bus

PR #3110, opened 2025-12-24
input: resync keyboard focus when backend state desyncs

### LINKED ISSUE #2314: wtype makes input unusable
`wtype` (a binary that inputs keys through the virtual-keyboard protocol) works as it should but once it exits (and closes the virtual keyboard), the selected app stops receiving input events from your keyboard.

### Reproduction

- run ` wtype -k p` in a shell
- observe that `p` got written (as it should)
- observe that your keyboard does not input anymore until you unfocus the app (niri keybinds still work)

### System Information

* niri version: niri 25.05.1 (Nixpkgs)

* Distro: nixos

* GPU: /

* CPU: /

luochen1990:
When a virtual input device (like wtype) is destroyed, Smithay might clear the keyboard focus. However, Niri's internal state (self.niri.keyboard_focus) remains unchanged. This causes a desynchronization where Niri thinks a window is focused, but the backend doesn't send key events to it.

This patch adds a check in update_keyboard_focus: if Niri's focus target hasn't changed but the underlying backend focus is different, we force a set_focus call to restore synchronization.

Fixes #2314

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -1212,6 +1212,7 @@ impl State {
         };
 
         let keyboard = self.niri.seat.get_keyboard().unwrap();
+        let current_focus = keyboard.current_focus();
         if self.niri.keyboard_focus != focus {
             trace!(
                 "keyboard focus changed from {:?} to {:?}",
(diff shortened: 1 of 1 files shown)


YaLTeR:
I don't think this change is correct. Keyboard grabs can change the actual focus, it doesn't need to be "synchronized" to niri.

PR #3118, opened 2025-12-25
Allow trackball scrolling in overview

### LINKED ISSUE #2544: Scroll-method "on-botton-down" does not work in overview screen
Scrolling with trackball and `scroll-method "on-button-down"` does not work in overview screen, although it works just fine within applications.

```kdl
    trackball {
        natural-scroll
        accel-speed 0.8
        scroll-method "on-button-down"
        scroll-button 275
        left-handed
    }
```

### System Information

* niri version: niri 25.08 (01be0e6)

* Distro: Arch 

* GPU: Advanced Micro Devices, Inc. [AMD/ATI] Picasso/Raven 2 [Radeon Vega Series / Radeon Vega Mobile Series] (rev d8)

* CPU: quad core AMD Ryzen 5 PRO 3400G with Radeon Vega Graphics

ArijanJ:
Simply allows scrolling in the overview with a trackball (on-button-down). Tested on an Arch PC and a Fedora laptop, everything works as expected

Closes #2544

I wanted to also handle `mods_with_finger_scroll_binds` (mod+scroll in main view), but that seems pretty coupled with `horizontal_finger_scroll_tracker` so I'm leaving it for now.

Not sure which sensitivity is used here, but when #2844 is merged we can probably use that value

Recording 2025-12-25 at 22 20 30

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -3255,8 +3255,8 @@ impl State {
         let horizontal_amount = event.amount(Axis::Horizontal);
         let vertical_amount = event.amount(Axis::Vertical);
 
-        // Handle touchpad scroll bindings.
-        if source == AxisSource::Finger {
+        // Handle touchpad and continuous scroll bindings.
+        if source == AxisSource::Finger || source == AxisSource::Continuous {
             let mods = self.niri.seat.get_keyboard().unwrap().modifier_state();
             let modifiers = modifiers_from_state(mods);
 


ArijanJ:
> I wanted to also handle [...] mod+scroll in main view, [...] I'm leaving it for now.

For anyone finding this in the future, this is implemented in the following branch of my fork: https://github.com/ArijanJ/niri/tree/mod-trackball-scroll (commit https://github.com/YaLTeR/niri/commit/f2172609b04984b25116273b26a4e5b8b6293ff5): mod + on-button-down scroll will move the view up/down as a touchpad does.

I vibecoded it so I won't be opening a PR, as the API is not optimal and doesn't quite fit very well with the rest of the architecture in main. Perhaps a more experienced contributor can adapt it into an acceptable state :)

**edit:** The easiest way to implement non-overview scrolling would be to just add another condition to this if statement: 
```rust
if (should_handle_in_overview && modifiers.is_empty()) || (source == AxisSource::Continuous && modifiers.contains(mod_key.to_modifiers())) {
```
But then it seems we don't have have that direction-deciding buffer that comes with SpatialMovementGrab and the other normal APIs like the 3f scroll tracker - I'm not sure how to implement those correctly with a continuous axis source.

jfrconley:
@Sempyos Is there anything blocking this from being merged? Accepting continuous scroll would be a big usability improvement for trackballs, even without the additional changes to support mod scroll.

Separately, I do wonder if button scroll should be effectively treated the same as touchpad swiping everywhere since it has many of the same physical characteristics. Though I do imagine you would still want different acceleration.

Sempyos:
That is not my place to say, I just add tags and occasionally comment

YaLTeR:
Thanks, tested on a regular mouse with on-button-down and it works.

PR #3133, opened 2025-12-29
Improve OpenRC/non-systemd compatibility

coleleavitt:
- Add runtime check for systemctl/dinitctl before running
- Use POSIX 'command -v' instead of bash 'hash'
- Warn when running on TTY without --session flag

This eliminates spurious warnings on OpenRC systems where systemctl doesn't exist, while still running dbus-update-activation-environment.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -69,6 +69,21 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let cli = Cli::parse();
 
+    // Warn if running as a TTY compositor without --session, as this likely means
+    // D-Bus services (ScreenCast, etc.) won't be registered.
+    if !cli.session
+        && env::var_os("WAYLAND_DISPLAY").is_none()
+        && env::var_os("WAYLAND_SOCKET").is_none()
+        && env::var_os("DISPLAY").is_none()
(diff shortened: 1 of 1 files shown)


coleleavitt:
the script I am using on gentoo, I don't know if this could be baked in if that's best practice or if having a script such as this is the right move:

```bash
#!/bin/sh
set -e

XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-/run/user/$(id -u)}"
PW_SOCKET="$XDG_RUNTIME_DIR/pipewire-0"

pkill -9 pipewire 2>/dev/null || true
pkill -9 wireplumber 2>/dev/null || true
rm -f "$XDG_RUNTIME_DIR"/pipewire*.lock "$XDG_RUNTIME_DIR/pulse/pid" 2>/dev/null || true
sleep 0.1

pipewire &
i=0; while [ ! -S "$PW_SOCKET" ] && [ $i -lt 50 ]; do sleep 0.1; i=$((i+1)); done

wireplumber &
pipewire-pulse &
i=0; while ! wpctl status >/dev/null 2>&1 && [ $i -lt 50 ]; do sleep 0.1; i=$((i+1)); done

exec /home/cole/RustProjects/active/niri/target/release/niri --session
```

PR #3155, opened 2025-12-31
Implement the `open-play-animation` window rule

vars1ty:
Fixes my request at #3153.
# 
Implements the `open-play-animation` `bool` window rule, which allows the user to enable/disable the opening animation of a window.


YaLTeR:
We're not gonna do it this way, we'll likely have animations block in window rule

vars1ty:
> We're not gonna do it this way, we'll likely have animations block in window rule

Sounds like a better idea, I'll close the PR due to that and assuming the final animations window rule gets put into a PR; I'll be happy to test it :)

Issue #2248, opened 2025-08-16
Add window stacking order to IPC


YaLTeR:
#1265 added window positions and sizes, but notably absent is window stacking order, important for floating windows.

The main problem is it's not entirely clear what's the best way to expose it that is both easy to work with and not too spammy.

> One idea I had is to have the stacking order on a workspace as a `Vec<WindowId>`, then have some kind of `WorkspaceWindowsRestacked { ws_id: WorkspaceId, new_order: Vec<WindowId> }`. Any windows missing from that should be considered below, or something.

Also, however we expose it should be forward-compatible. For example, when we add always-on-top windows, they will likely be on top of all other floating windows, yet they will have a stacking order among themselves. This can add new IPC but should not require breaking changes to any existing IPC.

my4ng:
What about treating changing the stack order as a permutation of the `WindowId`? That way any change can be communicated in the cycle notation with unchanged order as the omitted fixed points, i.e. `Vec<Vec<WindowId>>`. For example, `0 1 2 3 4 -> 0 3 1 2 4` can be efficient encoded as `(1 3 2)` and `0 1 2 3 4 -> 0 2 1 4 3` as `(1 2) (3 4)`. The cycles can be computed and applied efficiently too.

my4ng:
I guess it also depends on the assumption of what a single stack change could entail; if only a single window can be moved forward or backward in an event, e.g. ` 0 1 2 3 4` -> `0 2 3 1 4` then a circular shift with the index of the window id in question and a single offset suffices, `(1, +2)`. However, this could lead to breaking changes if it is later not general enough.

YaLTeR:
Interesting idea, though yeah, need to think about future changes. Also it sounds like it may pose some trouble for downstream consumers (non obvious implementation)

my4ng:
Using the full cycle notation should be general enough for arbitrary changes, and the implementation should be fairly trivial (correction: should be `Vec<Vec<usize>>` rather than  `Vec<Vec<WindowId>>`):

```rust
order_new = order.clone();
for c in cycles:
  for i in 0..c.count():
    order_new[c[i]] = order[c[i + 1 % c.count()]];
```

Using `A, B, C, D...` for `WindowId` for clarity:

```
   A B C D E F          (1 2 3 4)(0 5)
-> A C C D E F
-> A C D D E F
-> A C D E E F
-> A C D E B F
-> F C D E B F
-> F C D E B A
```

I can cook up a draft PR and see how it goes from there.

PR #3156, opened 2025-12-31
Add window stacking order to IPC

my4ng:
This implements #2248.

The stacking order is defined by the totally ordered `StackingOrder` struct which contains both a `StackingOrderGroup` and an index, which indicates position within the group. This can be later extended to support always-on-top windows by adding it as a new group. Per convention, a greater index implies a window is more on top. 

When a window's stacking order goes from `None -> Some(...)` or vice versa, the change is observed in the existing `WindowOpenedOrChanged` event. Otherwise, `Some(...) -> Some(...)` changes are batched in the `WindowStackingOrdersChanged` event.

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1307,6 +1307,30 @@ pub struct Window {
     ///
     /// The timestamp comes from the monotonic clock.
     pub focus_timestamp: Option<Timestamp>,
+    /// Stacking order of the window, if applicable.
+    pub stacking_order: Option<StackingOrder>,
+}
+
+/// Stacking order group of a [`Window`].
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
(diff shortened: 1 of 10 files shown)


YaLTeR:
> When a window's stacking order goes from None -> Some(...) or vice versa, the change is observed in the existing WindowOpenedOrChanged event. Otherwise, Some(...) -> Some(...) changes are batched in the WindowStackingOrdersChanged event.

If a window goes between floating and tiling are there internally inconsistent stacking orders in the middle between any messages?

my4ng:
> If a window goes between floating and tiling are there internally inconsistent stacking orders in the middle between any messages?

I am not quite sure I understand the question. 😅 When a window goes from having a stacking order of `None` when tiling to floating `Some((Floating, idx)`, the `WindowOpenedOrChanged` event is emitted first, while any `WindowStackingOrdersChanged` event is batched and always emitted at the end. It would mean that after the first event and before the second event, two windows might potentially share the same stacking order, if that's what you mean.

The obvious fix is to include this change in the second event, and ignore it in the first. Going from floating to tiling is of course not affected by this issue.

YaLTeR:
I'm thinking from the perspective of implementing an IPC client, how error prone is it

Issue #1533, opened 2025-05-06
Add support for monitor bit depth configuration


Gwenodai:
Currently you can set resolution, refresh rate, vrr, scale, transform, etc for each output. However, there's no way to control the bit depth of a display through Niri (Such as running a display in 8 or 10 bit mode).

This seems like a strange thing to not have access to considering all of the other output options available. Maybe it was just forgotten?

YaLTeR:
Some reasons:
1. I don't have a high bit depth or an HDR monitor.
2. I'm not sure if high bit depth is useful without the Wayland color management stuff that we don't have yet.
3. Sometimes the output defaults to 10 bit which can result in higher bandwidth, meaning less monitors can be enabled at the same time. So I made niri force 8 bit for now to increase the chances that a given monitor configuration will work.

Gwenodai:
Thanks for the response, that's all totally understandable. It was just something I noticed, not exactly a needed feature imo.

I actually didn't realise point 2 was still going to hold that back anyway, and point 3 is very true as well. Though I could see just defaulting all 10 bit displays to 8 bit (unless otherwise specfied by the user) being the 'fix' to that issue.

Quackdoc:
> I'm not sure if high bit depth is useful without the Wayland color management stuff that we don't have yet.

high bitdepth is indeed useful even without the protocol. At the very least Cosmic, Sway and KDE both have a notable improvement when running with a high bitdepth. on high bit displays it can be verified using  https://jursonovicst.github.io/gradient/test_sequences/3840x2160/gradient_3840-2160_0-50_yuv444p10_x265.mp4 IF you have a 10bit capable player mpv is one

> Sometimes the output defaults to 10 bit which can result in higher bandwidth, meaning less monitors can be enabled at the same time. So I made niri force 8 bit for now to increase the chances that a given monitor configuration will work.

There are other issues too, kmsgrab on ffmpeg usually breaks.

quadans:
It feels so bad to use a 1000€ OLED monitor to use 8 bit .... why cant I choose this in niri myself?

Gwenodai:
> It feels so bad to use a 1000€ OLED monitor to use 8 bit .... why cant I choose this in niri myself?

~~There is a debug flag that prevents Niri from limiting the your monitors bit depth.~~

~~https://yalter.github.io/niri/Configuration%3A-Debug-Options.html#keep-max-bpc-unchanged~~

[...]

PR #3158, opened 2025-12-31
Add per output max-bpc config and ipc action

my4ng:
This adds per output maximum bits per channel (max-bpc) config and ipc action as requested in #1533. 

```
output "DP-1" {
    ...
    max-bpc 10
}
```

Supported values are 6, 8, 10, 12, 14, 16. If not explicitly set, niri will not change the existing `max bpc` connector property.

### DIFF
--- a/docs/wiki/Configuration:-Outputs.md
+++ b/docs/wiki/Configuration:-Outputs.md
@@ -15,6 +15,7 @@ output "eDP-1" {
     variable-refresh-rate // on-demand=true
     focus-at-startup
     backdrop-color "#001100"
+    bpc 10
 
     hot-corners {
         // off
@@ -276,6 +277,21 @@ The alpha channel for this color will be ignored.
 ```kdl
 output "HDMI-A-1" {
     backdrop-color "#001100"
+}
+```
+
+### `bpc`
+
+<sup>Since: next release</sup>
+
+Set the maximum bits per channel (BPC) for this output.
+This is not needed for most outputs unless encountering bandwidth or lower-than-expected colour depth issues.
+
+Valid values are `8`, `10`.
+
+```kdl
+output "HDMI-A-1" {
+    bpc 10
 }
 ```
 
--- a/docs/wiki/Configuration:-Outputs.md
+++ b/docs/wiki/Configuration:-Outputs.md
@@ -15,6 +15,7 @@ output "eDP-1" {
     variable-refresh-rate // on-demand=true
     focus-at-startup
     backdrop-color "#001100"
+    max-bpc 8
 
     hot-corners {
         // off
@@ -276,6 +277,23 @@ The alpha channel for this color will be ignored.
 ```kdl
 output "HDMI-A-1" {
     backdrop-color "#001100"
+}
+```
+
+### `max-bpc`
+
+<sup>Since: next release</sup>
+
+Set the maximum bits per channel (BPC) for this output.
+You *do not* need to set this option unless you hit bandwidth issues (can't set a monitor configuration that works on other compositor) or lower-than-expected color depth.
+Particularly, 10-bit color output may work even if max-bpc is 8.
+
+Valid values are `6`, `8`, `10`, `12`, `14`, `16`.
+
+```kdl
+// Set 8 max-bpc on HDMI-A-1 to lower the bandwidth.
+output "HDMI-A-1" {
+    max-bpc 8
 }
 ```
 
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1139,7 +1139,9 @@ mod tests {
                                 y: 20,
                             },
                         ),
-                        bpc: _10,
+                        bpc: Some(
+                            _10,
+                        ),
                         mode: Some(
                             Mode {
                                 custom: false,
@@ -1185,7 +1187,7 @@ mod tests {
                         scale: None,
                         transform: Normal,
                         position: None,
-                        bpc: _8,
+                        bpc: None,
                         mode: Some(
                             Mode {
                                 custom: true,
@@ -1212,7 +1214,7 @@ mod tests {
                         scale: None,
                         transform: Normal,
                         position: None,
-                        bpc: _8,
+                        bpc: None,
                         mode: None,
                         modeline: Some(
                             Modeline {
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -725,6 +725,7 @@ mod tests {
                 transform "flipped-90"
                 position x=10 y=20
                 mode "1920x1080@144"
+                bpc 10
                 variable-refresh-rate on-demand=true
                 background-color "rgba(25, 25, 102, 1.0)"
                 hot-corners {
@@ -837,7 +838,7 @@ mod tests {
                 window-open { off; }
 
                 window-close {
-                    curve "cubic-bezier" 0.05 0.7 0.1 1  
+                    curve "cubic-bezier" 0.05 0.7 0.1 1
                 }
 
                 recent-windows-close {
@@ -1138,6 +1139,7 @@ mod tests {
                                 y: 20,
                             },
                         ),
+                        bpc: _10,
                         mode: Some(
                             Mode {
                                 custom: false,
@@ -1183,6 +1185,7 @@ mod tests {
                         scale: None,
                         transform: Normal,
                         position: None,
+                        bpc: _8,
                         mode: Some(
                             Mode {
                                 custom: true,
@@ -1209,6 +1212,7 @@ mod tests {
                         scale: None,
                         transform: Normal,
                         position: None,
+                        bpc: _8,
                         mode: None,
                         modeline: Some(
                             Modeline {
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -745,6 +745,7 @@ mod tests {
                 transform "flipped-90"
                 position x=10 y=20
                 mode "1920x1080@144"
+                max-bpc 10
                 variable-refresh-rate on-demand=true
                 background-color "rgba(25, 25, 102, 1.0)"
                 hot-corners {
@@ -857,7 +858,7 @@ mod tests {
                 window-open { off; }
 
                 window-close {
-                    curve "cubic-bezier" 0.05 0.7 0.1 1  
+                    curve "cubic-bezier" 0.05 0.7 0.1 1
                 }
 
                 recent-windows-close {
@@ -1160,6 +1161,11 @@ mod tests {
                                 y: 20,
                             },
                         ),
+                        max_bpc: Some(
+                            MaxBpc(
+                                _10,
+                            ),
+                        ),
                         mode: Some(
                             Mode {
                                 custom: false,
@@ -1205,6 +1211,7 @@ mod tests {
                         scale: None,
                         transform: Normal,
                         position: None,
+                        max_bpc: None,
                         mode: Some(
                             Mode {
                                 custom: true,
@@ -1231,6 +1238,7 @@ mod tests {
                         scale: None,
                         transform: Normal,
                         position: None,
+                        max_bpc: None,
                         mode: None,
                         modeline: Some(
                             Modeline {
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1228,6 +1228,10 @@ pub struct Output {
     ///
     /// `None` if the output is not mapped to any logical output (for example, if it is disabled).
     pub logical: Option<LogicalOutput>,
+    /// Maximum bit per color  (bit depth), if known.
+    pub max_bpc: Option<u8>,
+    /// Pixel format for framebuffers, if known.
+    pub format: Option<String>,
 }
 
 /// Output mode.
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1097,6 +1097,12 @@ pub enum OutputAction {
         #[cfg_attr(feature = "clap", command(flatten))]
         vrr: VrrToSet,
     },
+    /// Set the maximum bits per channel (bit depth).
+    MaxBpc {
+        /// Maximum bits per channel to set.
+        #[cfg_attr(feature = "clap", arg())]
+        max_bpc: MaxBpc,
+    },
 }
 
 /// Output mode to set.
@@ -1228,6 +1234,8 @@ pub struct Output {
     ///
     /// `None` if the output is not mapped to any logical output (for example, if it is disabled).
     pub logical: Option<LogicalOutput>,
+    /// Maximum bits per channel (bit depth), if known.
+    pub max_bpc: Option<u8>,
 }
 
 /// Output mode.
@@ -1289,6 +1297,32 @@ pub enum Transform {
     /// Rotated by 270° and flipped horizontally.
     #[cfg_attr(feature = "clap", value(name("flipped-270")))]
     Flipped270,
+}
+
+/// Output maximum bits per channel.
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub enum MaxBpc {
+    /// 6-bit.
+    #[serde(rename = "6")]
+    _6 = 6,
+    /// 8-bit.
+    #[default]
+    #[serde(rename = "8")]
+    _8 = 8,
+    /// 10-bit.
+    #[serde(rename = "10")]
+    _10 = 10,
+    /// 12-bit.
+    #[serde(rename = "12")]
+    _12 = 12,
+    /// 14-bit.
+    #[serde(rename = "14")]
+    _14 = 14,
+    /// 16-bit.
+    #[serde(rename = "16")]
+    _16 = 16,
 }
 
 /// Toplevel window.
@@ -1865,6 +1899,30 @@ impl FromStr for Transform {
                 r#""flipped", "flipped-90", "flipped-180" or "flipped-270""#
             )),
         }
+    }
+}
+
+impl TryFrom<u8> for MaxBpc {
+    type Error = &'static str;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            6 => Ok(MaxBpc::_6),
+            8 => Ok(MaxBpc::_8),
+            10 => Ok(MaxBpc::_10),
+            12 => Ok(MaxBpc::_12),
+            14 => Ok(MaxBpc::_14),
+            16 => Ok(MaxBpc::_16),
+            _ => Err("invalid max-bpc, can be 6, 8, 10, 12, 14, 16"),
+        }
+    }
+}
+
+impl FromStr for MaxBpc {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::try_from(s.parse::<u8>().unwrap_or_default())
     }
 }
 
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -14,7 +14,7 @@ use anyhow::{anyhow, bail, ensure, Context};
 use bytemuck::cast_slice_mut;
 use drm_ffi::drm_mode_modeinfo;
 use libc::dev_t;
-use niri_config::output::Modeline;
+use niri_config::output::{Bpc, Modeline};
 use niri_config::{Config, OutputName};
 use niri_ipc::{HSyncPolarity, VSyncPolarity};
 use smithay::backend::allocator::dmabuf::Dmabuf;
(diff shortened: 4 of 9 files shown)


YaLTeR:
Do we want to do this? Over just listing all formats always?

my4ng:
I think this will mitigate the bandwidth issue you mentioned by limiting the max bpc (for some monitors), and also any potential driver/display issue where they advertise more than what they support. One change I will make is to set the default based on the "max dpc" that is reported from DRM. Nevertheless, I don't see any disadvantage for finer-grain control as done in the case of KWin and Mutter.

YaLTeR:
I'm not sure if there *is* a bandwidth issue, just hypothesizing. If possible I'm for no config.

> One change I will make is to set the default based on the "max dpc" that is reported from DRM.

max bpc is something the compositor can change though?

> as done in the case of KWin and Mutter.

What do they do?

my4ng:
> max bpc is something the compositor can change though?

Sorry I meant from EDID. Though it might be easier to use the DRM set default if no bpc is give.

From [Mutter](https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/2412):

> Useful for lowering bandwidth requirements on systems that can't maintain a stable signal. Or for increasing the displayed colour depth on kernel drivers that default to too low (allegedly amdgpu has done this in the past).

Though I think we also need to also set the `max bpc` property too along side the color format changes.

my4ng:
A few changes:

1. ~~`COLOR_FORMATS_10` now used by default (unless bpc 8 explicitly set)~~ Unified `COLOR_FORMATS` with 10 bpc then 8 bpc formats used.
2. Readd `set_max_bpc` called on device init/session resume only when bpc explicitly set

EDIT: It means that by default, without any config, niri will use ~~10 bpc when possible~~ default driver set  (or potentially other compositor set) `max bpc`.

my4ng:
@YaLTeR should be ready for review :+1:

GregoryBai:
Thanks, man! 🙏

One step closer to total goodness

GregoryBai:
Очень жду одобрения этого PR'a)

Gwenodai:
> Nevertheless, I don't see any disadvantage for finer-grain control as done in the case of KWin and Mutter.

Why not consider adding a debug option to allow for setting arbitrary values? That way if someone does want to enable say 12bpc on an expensive display, they can do so easily. 

Although I'd say the use case for bpc values outside of 8 and 10 are so minor it probably doesn't matter at this point in time. But I figured I'd propose it before this pr is merged, unless it'd require more than a simple change to implement.

my4ng:
> Why not consider adding a debug option to allow for setting arbitrary values? That way if someone does want to enable say 12bpc on an expensive display, they can do so easily.

While there is no _technical_ issue against it, since for example the amdgpu driver supports a range from 8 to 16 on my system, IMO it isn't helpful. Users might fall for the 'bigger is better' fallacy, and choose the largest value which might lead to subtle bugs due to (many documented) driver/display/bandwidth issues.

In addition to the rarity of such high-end professional displays you have mentioned, they really require a proper color management implementation to work properly. This is more or less a stopgap solution before that is realised.

src/backend/tty.rs
```diff
@@ -3212,35 +3272,71 @@ impl<'a> ConnectorProperties<'a> {
...
-            .context("error setting property")?;
+    fn set_max_bpc(&mut self, bpc: Bpc) -> anyhow::Result<u64> {
+        let (info, value) = self.find(c"max bpc")?;
+
+        let property::ValueType::UnsignedRange(min, max) = info.value_type() else {
+            bail!("wrong property type")
+        };
+
+        let bpc = bpc as u64;
+        let bpc = bpc.clamp(min, max);
```
quinnr:
```suggestion
        let bpc = bpc.clamp(*min,*max);
```
I have been using this patch and when merged with upstream it now kicks back a compile error about a variable cast on my system. The compiler suggests this change and it appears to work at least to compile again.

quinnr (commented):
This patch has been working great! I ran into a compile error when merging with upstream and have a suggested fix. I'm no expert, just what the compiler suggested seems to work fine.

src/backend/tty.rs
my4ng: `drm::control::property::Info::value_type` now returns `&ValueType` instead of `ValueType` by not cloning, and not bumping pre-1.0 minor version despite the breaking change...

YaLTeR: You may want to open an issue there to bump the version

my4ng: @YaLTeR done, 0.14.2 yanked and 0.15.0 released by the drm people.

YaLTeR: Awesome, thanks (it's the same people as Smithay)

docs/wiki/Configuration:-Outputs.md
```diff
@@ -279,6 +280,21 @@ output "HDMI-A-1" {
 }
 ```
 
+### `bpc`
```
YaLTeR: Perhaps we should call this `max-bpc` matching the DRM property? Since this is more of a technical property than something that should be used to configure bit depth (I imagine we might even have a separate setting for that, but idk yet.)

docs/wiki/Configuration:-Outputs.md
```diff
@@ -279,6 +280,21 @@ output "HDMI-A-1" {
...
+
+<sup>Since: next release</sup>
+
+Set the maximum bits per channel (BPC) for this output.
+This is not needed for most outputs unless encountering bandwidth or lower-than-expected colour depth issues.
+
+Valid values are `8`, `10`.
```
YaLTeR: I think I agree with the comment that let's allow this to be an int from 8 to 16, since once again it's a technical setting. My monitors currently default to (unneeded) 16, and it'd be good to have this available for testing.

docs/wiki/Configuration:-Outputs.md
```diff
@@ -279,6 +280,21 @@ output "HDMI-A-1" {
 }
 ```
 
+### `bpc`
+
+
+Set the maximum bits per channel (BPC) for this output.
+This is not needed for most outputs unless encountering bandwidth or lower-than-expected colour depth issues.
```
YaLTeR:
```suggestion
You *do not* need to set this option unless you hit bandwidth issues (can't set a monitor configuration that works on other compositor) or lower-than-expected color depth.
```

niri-ipc/src/lib.rs
```diff
@@ -1228,6 +1228,10 @@ pub struct Output {
     ///
     /// `None` if the output is not mapped to any logical output (for example, if it is disabled).
     pub logical: Option<LogicalOutput>,
+    /// Maximum bit per color  (bit depth), if known.
```
```suggestion
    /// Maximum bits per channel (bit depth), if known.
```

niri-ipc/src/lib.rs
```diff
@@ -1228,6 +1228,10 @@ pub struct Output {
...
+    pub max_bpc: Option<u8>,
+    /// Pixel format for framebuffers, if known.
```
```suggestion
    /// Framebuffer pixel format, if known.
```

niri-ipc/src/lib.rs
```diff
@@ -1228,6 +1228,10 @@ pub struct Output {
...
+    pub format: Option<String>,
```
YaLTeR: Do we want this here? Especially as a String? I think ppl who want can check drm_info.

src/backend/tty.rs
```diff
@@ -1285,18 +1306,27 @@ impl Tty {
...
             match get_panel_orientation(&props) {
                 Ok(x) => orientation = Some(x),
                 Err(err) => {
                     trace!("couldn't get panel orientation: {err:?}");
                 }
             }
+
+            if let Err(err) = props.commit() {
+                warn!("failed to atomically commit properties: {err}");
+            }
```
YaLTeR: I think this should go before get_panel_orientation(), just after both "set"s

src/backend/tty.rs
```diff
@@ -67,10 +67,14 @@ use crate::frame_clock::FrameClock;
...
-use crate::render_helpers::{resources, shaders, RenderCtx, RenderTarget};
+use crate::render_helpers::{resources, shaders, RenderTarget};
 use crate::utils::{get_monotonic_time, is_laptop_panel, logical_output, PanelOrientation};
 
-const SUPPORTED_COLOR_FORMATS: [Fourcc; 4] = [
+const COLOR_FORMATS: [Fourcc; 8] = [
+    Fourcc::Xrgb2101010,
+    Fourcc::Xbgr2101010,
+    Fourcc::Argb2101010,
+    Fourcc::Abgr2101010,
```
YaLTeR: Unrelated change? Did we figure out if this even works?

src/backend/tty.rs
```diff
@@ -2402,6 +2441,21 @@ impl Tty {
...
+                    {
+                        if let Err(err) = props.set_max_bpc(bpc) {
+                            warn!("failed to get `max bpc` property: {err}");
+                        };
+                        if let Err(err) = props.commit() {
+                            warn!("failed to atomically commit properties: {err}");
+                        }
+                    } else {
+                        warn!("failed to get connector properties");
+                    }
```
YaLTeR: Seems like with these warnings and all, repeated in several places of the code, this could be extracted in a helper function

src/backend/tty.rs
```diff
@@ -3242,35 +3298,71 @@ impl<'a> ConnectorProperties<'a> {
...
+                property::Value::UnsignedRange(bpc).into(),
+            );
+            self.has_change = true;
+        }
+
+        Ok(bpc)
     }
 
-    Ok(())
+    fn commit(self) -> anyhow::Result<()> {
```
YaLTeR: Good call making this into an atomic commit sequence I think

niri-config/src/lib.rs
```diff
@@ -729,6 +725,7 @@ mod tests {
                 transform "flipped-90"
                 position x=10 y=20
                 mode "1920x1080@144"
+                bpc 10
```
YaLTeR: I think we might also want to add the corresponding `niri msg output ...` setter

YaLTeR (changes requested):
For splitting into commits:
- Please avoid "fix" commits. Squash fixes into the relevant commits, we want a clean history.
- My suggestion for commit splits: 1. remove two spaces from test, 2. add max-bpc config, 3. add max-bpc into ipc

the-burrito-triangle:
So I wanted to add some data points from my own setup.

My Intel iGPU using the xe kernel driver allows for a range of `6` to `12` for the `max bpc` property, while my AMD dGPU using amdgpu kernel driver allows for a range of `8` to `16` `max bpc`.

**Currently, this value is always set to the maximum in both Niri 26.04 and GNOME 50**, so 12 for Intel and 16 for AMD. The only compositor that I know of that actually changes the `max bpc` value is Sway, which forces 8 for some reason.

Setting the `max bpc` property only looks to be useful for displays without DSC or with a flaky DSC implementation, where limiting the GPU to 8 bpc may help. Otherwise, setting this, to say 10 bpc, doesn't guarantee a 10 bpc link between the display and GPU: the handshake between the display and GPU could still select 8 bpc.

So I would rename this setting from `bpc` (which gives false confidence in what it does) to `max bpc` to better explain what it actually is (only a limit).

Using Niri 26.04 **without this patch**  shows I already have a 10 bpc link to the display (as shown by the display's built-in info panel--we just need 10 bpc framebuffer formats to get the full 10bpc pipeline):

```
./drm_info | grep -B2 -A6 -E "object framebuffer = [1-9]"

    │   └───Properties
    │       ├───"type" (immutable): enum {Overlay, Primary, Cursor} = Primary
    │       ├───"FB_ID" (atomic): object framebuffer = 426
    │       │   ├───Object ID: 426
    │       │   ├───Size: 3840×2160
    │       │   ├───Format: XRGB8888 (0x34325258)
    │       │   ├───Modifier: I915_FORMAT_MOD_Y_TILED (0x0100000000000002)
    │       │   └───Planes:
    │       │       └───Plane 0: offset = 0, pitch = 15360 bytes
--
    │   └───Properties
    │       ├───"type" (immutable): enum {Overlay, Primary, Cursor} = Cursor
    │       ├───"FB_ID" (atomic): object framebuffer = 427
    │       │   ├───Object ID: 427
    │       │   ├───Size: 64×64
    │       │   ├───Format: ARGB8888 (0x34325241)
    │       │   ├───Modifier: DRM_FORMAT_MOD_LINEAR (0x0000000000000000)
    │       │   └───Planes:
    │       │       └───Plane 0: offset = 0, pitch = 256 bytes
```

```
./drm_info | grep -B2 -A12 -E "MPG 274"

│   │       ├───"EDID" (immutable): blob = 421
│   │       │   ├───Manufacturer: Microstep
│   │       │   ├───Model: MPG 274URF QD
│   │       │   └───Serial: 0x00000001
│   │       ├───"DPMS": enum {On, Standby, Suspend, Off} = On
│   │       ├───"link-status": enum {Good, Bad} = Good
│   │       ├───"non-desktop" (immutable): range [0, 1] = 0
│   │       ├───"TILE" (immutable): blob = 0
│   │       ├───"CRTC_ID" (atomic): object CRTC = 150
│   │       ├───"subconnector" (immutable): enum {Unknown, VGA, DVI-D, HDMI, DP, Wireless, Native} = Native
│   │       ├───"audio": enum {force-dvi, off, auto, on} = auto
│   │       ├───"Broadcast RGB": enum {Automatic, Full, Limited 16:235} = Automatic
│   │       ├───"max bpc": range [6, 12] = 12
│   │       ├───"Colorspace": enum {Default, BT709_YCC, XVYCC_601, XVYCC_709, SYCC_601, opYCC_601, opRGB, BT2020_CYCC, BT2020_RGB, BT2020_YCC, DCI-P3_RGB_D65, RGB_WIDE_FIXED, RGB_WIDE_FLOAT, BT601_YCC} = Default
│   │       ├───"HDR_OUTPUT_METADATA": blob = 0
```

docs/wiki/Configuration:-Outputs.md
my4ng: In KDE implementation, the valid color depths are multiples of 2, i.e. `6`, `8`, `10`, `12`, `14`, `16`. Would that be better than the full range [6, 16]? I don't mind either way.

niri-ipc/src/lib.rs
my4ng: Yeah I agree, it was mostly for convenience and probably very rarely needed. I will remove it then.

src/backend/tty.rs
my4ng: This is confirmed to work in the related issue where it is required for true 10-bit color depth (mpv banding). The driver picks the first format in the list that works.

docs/wiki/Configuration:-Outputs.md
YaLTeR: Let's do multiples

src/backend/tty.rs
YaLTeR: Perhaps separate commit on top then?

niri-config/src/lib.rs
my4ng: Should we also support 'unsetting' the max-bpc, i.e. if the config specifies `max-bpc 10`, and `niri msg output max-bpc none`, then on resuming session max-bpc isn't set anymore? If so, should it be `none` or something else?

YaLTeR: It seems that we don't in say `output transform` so let's not

my4ng:
@YaLTeR I've addressed all the issues short of splitting them into three separate commits since they are too tangled together. I have however put the supported color format change in a separate commit.

docs/wiki/Configuration:-Outputs.md
```diff
@@ -73,7 +74,7 @@ If the refresh rate is omitted, niri will pick the highest refresh rate for the
 If the mode is omitted altogether or doesn't work, niri will try to pick one automatically.
 
 Run `niri msg outputs` while inside a niri instance to list all outputs and their modes.
-The refresh rate that you set here must match *exactly*, down to the three decimal digits, to what you see in `niri msg outputs`.
+The refresh rate that you set here must match _exactly_, down to the three decimal digits, to what you see in `niri msg outputs`.
```
YaLTeR: I have a feeling that you ran a .md formatter and it reformatted this. We use `*` for italic in the rest of the docs. (same about `_do not_` below)

```suggestion
The refresh rate that you set here must match *exactly*, down to the three decimal digits, to what you see in `niri msg outputs`.
```

niri-ipc/src/lib.rs
```diff
@@ -1291,6 +1299,32 @@ pub enum Transform {
...
 }
 
+/// Output maximum bits per channel.
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub enum MaxBpc {
+    /// 6-bit.
+    #[serde(rename = "6")]
+    _6 = 6,
```
YaLTeR, line 1309: Does our code even support 6? We only have the 8-bit color format. If it doesn't work let's omit 6-bit

src/backend/tty.rs
```diff
@@ -3262,60 +3269,123 @@ impl<'a> ConnectorProperties<'a> {
...
+    max_bpc: Option<MaxBpc>,
+    reset_hdr: bool,
+) {
+    if let Some(bpc) = max_bpc {
+        if let Ok(mut props) = ConnectorProperties::try_new(drm, connector) {
+            if let Err(err) = props.set_max_bpc(bpc) {
+                warn!("failed to get `max bpc` property: {err}");
+            };
+
+            if reset_hdr {
```
YaLTeR: Reset HDR shouldn't be gated behind max_bpc being available

niri-ipc/src/lib.rs
```diff
@@ -1097,6 +1097,12 @@ pub enum OutputAction {
         #[cfg_attr(feature = "clap", command(flatten))]
         vrr: VrrToSet,
     },
+    /// Set the maximum bit per color (bit depth).
+    MaxBpc {
+        /// Maximum bit per color to set.
```
```suggestion
    /// Set the maximum bits per channel (bit depth).
    MaxBpc {
        /// Maximum bits per channel to set.
```

niri-ipc/src/lib.rs
```diff
@@ -1868,6 +1902,33 @@ impl FromStr for Transform {
...
+            6 => Ok(MaxBpc::_6),
+            8 => Ok(MaxBpc::_8),
+            10 => Ok(MaxBpc::_10),
+            12 => Ok(MaxBpc::_12),
+            14 => Ok(MaxBpc::_14),
+            16 => Ok(MaxBpc::_16),
+            _ => Err(concat!(
+                r#"invalid max-bpc, can be "6", "8", "10", "12", "#,
+                r#""14", "16""#
+            )),
```
```suggestion
            _ => Err("invalid max-bpc, can be 6, 8, 10, 12, 14, 16"),
```
since this is not a string in the config.

src/backend/tty.rs
```diff
@@ -1301,21 +1304,15 @@ impl Tty {
...
-            }
-
-            match get_panel_orientation(&props) {
-                Ok(x) => orientation = Some(x),
-                Err(err) => {
-                    trace!("couldn't get panel orientation: {err:?}");
-                }
+        set_connector_properties(&device.drm, connector.handle(), config.max_bpc, true);
+        let orientation = match ConnectorProperties::try_new(&device.drm, connector.handle())
+            .and_then(|c| c.get_panel_orientation())
```
YaLTeR: I'd prefer to reuse ConnectorProperties like the code used to instead of fetching them twice because it's kinda inefficient.

src/backend/tty.rs
```diff
@@ -3262,60 +3273,123 @@ impl<'a> ConnectorProperties<'a> {
...
-            .context("error setting property")?;
+    fn set_max_bpc(&mut self, bpc: MaxBpc) -> anyhow::Result<u64> {
+        let (info, value) = self.find(c"max bpc")?;
+
+        let property::ValueType::UnsignedRange(min, max) = info.value_type() else {
+            bail!("wrong property type")
+        };
+
+        let bpc = bpc.0 as u64;
+        let bpc = bpc.clamp(min, max);
```
YaLTeR: Do we want to clamp here or bail with a warning? So that the user knows when they try to set max-bpc to outside what their monitor supports.

YaLTeR (changes requested):

niri-ipc/src/lib.rs
my4ng: I believe it does? I would imagine it simply discards the least significant two bits. AMDGPU doesn't support 6 bits so I can't test it myself, but according to @the-burrito-triangle intel iGPU does.

src/backend/tty.rs
my4ng: I am thinking clamp *and* with a warning, like this?

```rs
        if !(min..=max).contains(&max_bpc) {
            warn!("max-bpc outside valid range of [{min}, {max}]");
        }
        let max_bpc = max_bpc.clamp(min, max);
```

src/backend/tty.rs
my4ng: Yeah I am not sure what's the best way to combine it with set_connector_properties that's not ugly and clunky.

src/backend/tty.rs
YaLTeR: There's one problem, this doesn't provide any context (if the user has several connectors, which one has the issue?)

YaLTeR: Also I think I'd prefer to not change the property in this case rather than warn and clamp, but not sure

niri-ipc/src/lib.rs
YaLTeR: If so then alright

src/backend/tty.rs
my4ng: that's true I agree. I will have it bail then.

YaLTeR:
Pushed some fixes and tested on my devices. Seems to work, though I noticed that apparently 10-bit output works fine even when max-bpc is set to 8 on my devices. Like, the plane format stays the exact same and the mpv test works, despite DSC turning off at max-bpc = 8.

YaLTeR:
Thanks

PR #3168, opened 2026-01-01
protocols: implement `ext-foreign-toplevel-list-v1`

HigherOrderLogic:
Closes #2044, closes #1450.

Basically Sodiboo's work but rebased upon latest master and fix merge conflict, and also apply outstanding change requests by Yalter.

Testing video will follow soon.

### DIFF
--- a/src/protocols/foreign_toplevel.rs
+++ b/src/protocols/foreign_toplevel.rs
@@ -1,10 +1,16 @@
 use std::collections::hash_map::Entry;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
 use arrayvec::ArrayVec;
 use smithay::output::Output;
+use smithay::reexports::wayland_protocols::ext::foreign_toplevel_list::v1::server::{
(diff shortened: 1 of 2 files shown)


YaLTeR:
But this is unaddressed?

> My remaining concern is the naming one that I mentioned above (`EXT_VERSION`-`WLR_VERSION`)

HigherOrderLogic:
> But this is unaddressed?
> 
> > My remaining concern is the naming one that I mentioned above (`EXT_VERSION`-`WLR_VERSION`)

Addressed. Must have missed that since I was only looking at the reviews, not comments.

YaLTeR:
What clients can I use to test this, especially continuous listening to events (rather than start, print a list of windows, then exit)?

HigherOrderLogic:
I have written this (rather simple) [test client](https://github.com/HighOrderLogic/foreign-toplevel-list-test) to test the protocol. It just logs out the received event and updated the internal state to reflect the change.

The implementation seems to be correct from my local testing with the aforementioned client.

YaLTeR:
Thanks, looks good

PR #1856, opened 2025-06-19
Per-device touch and tablet config

stefanboca:
Allows configuring `touch` and `tablet` by device name, which can be found with  `libinput list-devices` or `libinput debug-events`:

```kdl
input {
    tablet { map-to-output "eDP-1"; }
    tablet "ELAN9009:00 04F3:2F2A Stylus" { map-to-output "DP-1"; }
    touch { map-to-output "eDP-1"; }
    touch "ELAN9009:00 04F3:2F2A" { map-to-output "DP-1"; }
}
```

My Asus Zenbook Duo has dual touchscreens which are also tables, so I have locally tested this change.

Related to #371, Closes #1231

Issue #372, opened 2024-05-21
Gesture bindings


YaLTeR:
Add some way to customize gesture bindings. I don't have a concrete design yet, but:
1. Users should be able to customize how many fingers to use for the swipe gestures (supersede https://github.com/YaLTeR/niri/pull/315, cc @MagneFire)
2. I should be able to add new gestures without breaking users' configs
3. Users should be able to bind gestures only where it makes sense, i.e. swipe actions to swipe gestures and not to pinch gestures for example

I'm thinking something like a `binds` section, but for gestures. To satisfy 1. and 2., maybe encode number of fingers explicitly into the "key"?

```
gestures {
    touchpad-swipe-3-horizontal horizontal-view-movement
    touchpad-swipe-3-vertical workspace-switch
    Mod+Mouse3-horizontal horizontal-view-movement
    Mod+touch-swipe-3-horizontal horizontal-view-movement
}
```

This way, I can add new defaults when this section is missing from the config, and when it is present in the config the user will just need to add new gestures manually.

I don't entirely like this though, looks kinda awkward.

Also, I can see a problem in the future where there may be a 2D gesture, and so you will need to be able to bind *either* `touchpad-swipe-N` to a 2D gesture, or separate `touchpad-swipe-N-horizontal/vertical` to 1D gestures. But also maybe that's not a problem and can just be verified during parsing.

Also, this "gestures" section that I have in mind seems to be mainly about continuous gestures (swipe and pinch) and not about discrete gestures like double-resize-click (these seem more fit for the regular binds section).

YaLTeR:
How about this:

```
gestures {
    touchpad-swipe-3 horizontal="view-movement" vertical="workspace-switch"
    Mod+Mouse3 horizontal="view-movement" vertical="workspace-switch"
    Mod+Mouse2 "resize-window"
}
```

You can only set either an argument, or the properties. All three have only those variants that make sense for that property (i.e. no vertical-only gestures in the horizontal property).

I'm still not sure about encoding the finger count into the "key" name.

> in the future where there may be a 2D gesture

Examples: currently existing interactive window resize; future interactive window move. For the resize it even makes sense to be able to bind just the horizontal part (since we're a scrolling WM).

IvanTurgenev:
Yeah even mouses nowadays have gesture buttons ![image](https://github.com/user-attachments/assets/413db647-0e91-444f-8379-b458da391f0b)

nakibrayan3:
how about some thing like this:

[...]

PR #3180, opened 2026-01-02
feat: Add touch gesture support (mostly hardcoded)

Atan-D-RP4:
- Brings the touchscreen gestures to parity with touchpad/trackpad gestures, as best I can tell
- Matches behavior of touchpad gestures closely
- Updated docs with `Since` annotation
- Added natural-scroll config option for touchscreen gestures

**Testing**:
- 3-finger horizontal swipe correctly adjusts view offsets.
- 3-finger vertical swipe correctly switches workspaces
- 2-finger start into 3rd mid-gesture -> works
- During 3-finger gesture, when one finger is removed mid-gesture, the
  transition is held in place by the two fingers and finishes if a 2nd is
  removed and correctly snaps into place.
- Gestures during:
    - window animations -> works (tested with spawning windows with
      animations enabled)
    - window dragging -> works (though very slow when dragging windows in

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -4076,7 +4076,7 @@ impl State {
 
         // Check if we're tracking a multi-finger gesture (2+ fingers).
         // If so, we should not forward events to clients.
-        let tracking_gesture = self.niri.touch_gesture_points.len() >= 2;
+        let tracking_gesture = self.niri.touch_gesture_points.len() > 2;
 
         let serial = SERIAL_COUNTER.next_serial();
 
@@ -4210,7 +4210,7 @@ impl State {
         let slot = evt.slot();
 
         // Check if we're tracking a multi-finger gesture before removing this touch point.
-        let tracking_gesture = self.niri.touch_gesture_points.len() >= 2;
+        let tracking_gesture = self.niri.touch_gesture_points.len() > 2;
 
         // Remove touch point from gesture tracking.
         self.niri.touch_gesture_points.remove(&Some(slot));
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -4063,6 +4063,15 @@ impl State {
         };
         let slot = evt.slot();
 
+        // Track touch point for multi-finger gesture detection.
+        let was_single = self.niri.touch_gesture_points.len() == 1;
+        self.niri.touch_gesture_points.insert(Some(slot), pos);
+
+        // When second finger arrives, start gesture recognition.
+        if was_single && self.niri.touch_gesture_points.len() == 2 {
+            self.niri.touch_gesture_cumulative = Some((0., 0.));
+        }
+
         let serial = SERIAL_COUNTER.next_serial();
 
         let under = self.niri.contents_under(pos);
@@ -4191,6 +4200,23 @@ impl State {
         };
         let slot = evt.slot();
 
+        // Remove touch point from gesture tracking.
+        self.niri.touch_gesture_points.remove(&Some(slot));
+
+        // End gesture when fewer than 2 fingers remain.
+        if self.niri.touch_gesture_points.len() < 2 {
+            self.niri.touch_gesture_cumulative = None;
+
+            // End any ongoing gesture animations.
+            if let Some(output) = self.niri.layout.workspace_switch_gesture_end(Some(true)) {
+                self.niri.queue_redraw(&output);
+            }
+            if let Some(output) = self.niri.layout.view_offset_gesture_end(Some(true)) {
+                self.niri.queue_redraw(&output);
+            }
+            self.niri.layout.overview_gesture_end();
+        }
+
         if let Some(capture) = self.niri.screenshot_ui.pointer_up(Some(slot)) {
             if capture {
                 self.confirm_screenshot(true);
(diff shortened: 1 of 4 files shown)


Sempyos:
I'm just going to say my opinions on this, 

I am not terribly a uh fan of having hardcoded gesture bindings, as this could Interfere with some apps that utilize 3 fingers, additionally this will break quite a few people's configurations with lisgd and similar programs. This should also probably, uh be customizable and would require time to design

Atan-D-RP4:
> I'm just going to say my opinions on this,
> 
I appreciate you taking the time do so. Thank youc

> I am not terribly a uh fan of having hardcoded gesture bindings, as this could Interfere with some apps that utilize 3 fingers, additionally this will break quite a few people's configurations with lisgd and similar programs. This should also probably, uh be customizable and would require time to design

Yeah, fair criticism. I'd rather be they be more customizable too but for now I only matched current touch pad gestures.

This and basically a bunch of other changes to gestures stuff is waiting on the author to settle on the design for the gestures interface.

Thanks for letting me know about lisgd! Seems great and I'm gonna play with it, and test this there too see how things break.

Do you think the current hardcoded touchpad gestures also cause the same problems you described here?

Sempyos:
I don't think I have run into any app that has needed me to use more than 2 fingers..? Usually the most I need is 2 for zooming and such

Atan-D-RP4:
Rebased on current main. Tracking issue https://github.com/YaLTeR/niri/issues/372.

Has there any update on the design for gesture bindings?

Sempyos:
Last I checked it still needs design

Additionally, feel free to ping me if you wants me to test this PR

x1unix:
@Atan-D-RP4 If you're looking for volunteers to test -  I've a spare raspberry PI 4 device with touchscreen

Atan-D-RP4:
> @Atan-D-RP4 If you're looking for volunteers to test - I've a spare raspberry PI 4 device with touchscreen

Please do! I actually developed this for my own 2-in-1 laptop which has a touchscreen, so testing hasn't been an issue. It works pretty well, enough that I'm daily driving it on my system.

But any further work is blocked by lack finalization for the total gesture and gesture bindings interface in Niri, which is being discussed in the tracking issue of this PR.

Going forward, I'll clean up the code, polish the implementation so that when the design stabilizes, the touch gestures can be configured directly off of this PR.

yuitayuun:
i'll comment since i hang around artists quite a bit, and that's the group of people that would be really disappointed in hardcoded gestures. it's common for drawing programs to have 3 finger taps and such to ctrl+z and do other stuff. that said i can't wait for some touch gestures to be added, lisgd is not treating me well. wishing you luck!

jarmuszz:
@Atan-D-RP4 Some notes from testing on Framework 12
* The touch sensitivity is too high for me. A horizontal swipe that is enough to switch a workspace in GNOME/KDE makes me travel through 3-4 full-screen windows. It's kinda hard to just go to the next window. Maybe this should be configurable?
* Similar thing for vertical swipes. I often go 2 workspaces up/down. Maybe it would be better if a vertical swipe would just switch to the next/previous workspace? It's how it works in other compositors I tried.
* Gestures are being "leaked" to apps. I think the only app I have that understands ≥3 finger gestures is [foot](https://codeberg.org/dnkl/foot). Using the gestures makes the terminal scroll which is realllly annoying.

Feel free to ping if you'd need more tests.

Tellinq:
An additional note I’d like to point out is that the 3 finger gestures will focus on the workspace that the cursor is focused on, instead of the device that I’ve mapped to a specific display. For instance, if I move my cursor to an external monitor, and then do a 3 finger gesture. I’ll be moving windows or switching workspaces on that external monitor. Once I move the cursor back to my main laptop screen, the gestures then apply to that display instead.

Currently testing on:
- Samsung Galaxy Book3 Pro 360
- Lenovo Yoga Book 9i (however it combines this PR and #1856 to allow me to individually map each display out)

I did set up a bit of a workaround on my Yoga Book 9i to instead only target the display a touch device is mapped to as this is important for a dual-screen device such as this one. However, I have not tested my changes with the Book3 Pro 360 without mapping my device to a specific output. Once I do some further testing, I’ll look into sending a PR to this branch/PR.

If anyone else does have a touchscreen device alongside with an external display they could test out, I’d like to know if anyone else is also experiencing the same thing.

Aside from that, jarmuszz did cover the other main three issues I’ve noticed. Another example to try to replicate gestures being “leaked” is doing a workspace switch and on a chromium based browser web page. I’ve noticed that also invokes a pinch to Zoom behavior at the same time.

Feel free to also ping me if you need any further testing. I’ll see if I can look into addressing the other issues, but my knowledge with Rust is very little, so no guarantees.

Atan-D-RP4:
Thank you all for your feedback!
I'm not at my desk right now and won't be until tomorrow. I also won't have access to my main touch machine until Tuesday for personal reasons so I'll get to updating and reviewing this PR's work then.

But for now of the top of my head and the PR diff:
Yeah I definitely need to make it less sensitive which I have not yet figured out how since this is for absolute pointer motion devices. Shouldn't be too conplicated.
The leak is very interesting. I don't know where ot might be coming from. Maybe I missed consuming the gesture events? Bound to be fun and educational.

To @Tellinq, I don't get it. 
Do you mean 3F touch screen gestures focus on the workspace under the cursor? If so, then I'm pretty sure that's intended behavior since a gesture is still a touch  event which can focus the workspace or output by putting the cursor there.

Or wait do you mean the opposite? That the used gesture is not focusing where it was used(display/workspace) but instead applies on or from the one under the cursor which is elsewhere?

On a re-read of your comment, I understood what you both meant by "leak". Yes, I've had the browser pinch-zoom triggering for me too. It's definitely due to incomplete implementation here from me.

Need to do some more reading and research on how wayland/Niri consumes or forwards to clients the touch input events.

Tellinq:
Hi @Atan-D-RP4, I apologize for the lack of clarification. What I meant was the 3F gestures focus on the workspace where the cursor is at. If my cursor is at my main laptop's screen (where the touch screen is at), the 3F gestures will apply there. If I then move that cursor to an external display, the gestures will focus on that external display instead.

This seems to apply regardless if map-to-output is defined.

I've provided a video on what I mean. On the right is my Samsung Galaxy Book3 Pro 360 laptop, on the left is my Minisforum V3 being treated as an external display.

https://github.com/user-attachments/assets/cc8f2f20-4aa5-4fc0-9979-48f022d515ad

I have touch configured as such:
```kdl
touch {
    natural-scroll
    map-to-output "eDP-1"
}
```
My main expectation with this configuration is that if I define map-to-output to a touch device, I would expect the gestures to specifically apply to that display. In this scenario, my laptop screen. 

It is worth noting that I did use 1-finger gestures to swipe between windows in the Overview in this video. That is correctly applying only to the map-to-output display.

Perhaps the current behavior might be acceptable if map-to-output isn't defined? I don't think this should be the behavior if map-to-output is defined.

Does this help better clarify?

Atan-D-RP4:
Ah, ok. So I did get it right on the second read-through. Thank you for the confirmation and clarification.
The current behavior without map-to-output _seems_ fine to me but I've never used multi-monitor setups so I'm no authority on that.

Atan-D-RP4:
Updated the branch.
I prevented forwarding of touch events when gestures are in progress and am using the output at the touch event pos instead of the cursor pos, as I had done previously.

src/input/mod.rs
```diff
@@ -4244,6 +4285,117 @@ impl State {
...
+
+                // Check if we're still in recognition phase.
+                if let Some((cx, cy)) = &mut self.niri.touch_gesture_cumulative {
+                    *cx += delta_x;
+                    *cy += delta_y;
+
+                    // Check if gesture moved far enough to decide direction.
+                    // Threshold matches touchpad gestures (16px from GNOME Shell).
+                    let (cx, cy) = (*cx, *cy);
+                    if cx * cx + cy * cy >= 16. * 16. {
```
Aleksanaa, line 4285: Perhaps we can make this configurable? 16 is too sensitive for me.

src/input/mod.rs
```diff
@@ -4093,11 +4093,17 @@ impl State {
...
+        // Actual gestures (workspace/view/scroll) require 3+ fingers and are processed
+        // in on_touch_motion once the third finger arrives and moves.
         if was_single && self.niri.touch_gesture_points.len() == 2 {
             self.niri.touch_gesture_cumulative = Some((0., 0.));
         }
 
+        // Check if we're tracking a multi-finger gesture (2+ fingers).
+        // If so, we should not forward events to clients.
+        let tracking_gesture = self.niri.touch_gesture_points.len() >= 2;
+
```
jarmuszz, line 4080: I think this also blocks clients from detecting **2**-finger gestures (at least it does in my setup). Changing `>=` to `>` does the thing for me.

src/input/mod.rs
```diff
@@ -4226,6 +4235,9 @@ impl State {
         };
         let slot = evt.slot();
 
+        // Check if we're tracking a multi-finger gesture before removing this touch point.
+        let tracking_gesture = self.niri.touch_gesture_points.len() >= 2;
+
```
jarmuszz, line 4214: ditto

src/input/mod.rs
```diff
@@ -4218,6 +4259,117 @@ impl State {
...
+                    .is_some()
+                {
+                    self.niri.queue_redraw_all();
+                }
+
+                // Swipe right → scroll left (prev window), swipe left → scroll right (next window).
+                if self
+                    .niri
+                    .layout
+                    .view_offset_gesture_update(gesture_delta_x, timestamp, true)
```
Ficik, line 4353: Is the is_touchpad=true in gesture update and begin intentional? If it's false it skips the touchpad to screen multiplier and required swipe distance feels about right (framework 12, I don't have different size touchscreen to test)

src/input/mod.rs
horriblename: I'd suggest using physical units (touch position * monitor physical size in mm) for threshold detection. This is what android does (except they use dip not mm). It's still desirable to make the threshold configurable, but it's a decent default that works well-enough on almost any device

in [hyprgrass](https://github.com/horriblename/hyprgrass) we use a default threshold of 100px, which works well on a phone, but you'd need to turn the sensitivity up to 4.0 on a tablet (=> 25px threshold), and this value varies widely depending on your monitor physical size and pixel density

on the other hand, if the threshold defaults to 2mm (just an example value), it works fairly well everywhere. As per android's suggestions, it's still desirable to configure it per-device

One problem with physical units is, some outputs (e.g. projectors) don't have a physical unit, so some kind of fallback might be needed, and the default will almost certainly not work well for the user

well ultimately this is just for a better default, if a pixel based approach is used, it just means the users won't get out-of-the-box gestures, it will still work well after configured properly (assuming the threshold is made configurable)

YaLTeR: I haven't looked at this PR but most of that code should be operating in logical (scaled) pixels, not physical

PR #3187, opened 2026-01-03
Stop hardcoding the niri path in niri.service

Axlefublr:
From `man systemd.service`:
> Searched directories include /usr/local/bin/, /usr/bin/, and their sbin/ counterparts (only on systems using split bin/ and sbin/).

i.e. `/usr/bin/niri` is redundant: `ExecStart` can already “see” `niri`

The actual benefit of this pr, though, is the idea that you might have a personal niri fork, the binary of which you'll place in `/usr/local/bin/niri`.
Then you can easily use the infrastructure that the system package provides to you, and only worry about your changes towards the niri binary. Very convenient! \
When you do this, though, placing the resulting binary into `/usr/bin/niri`, overriding the package's one, is a bit annoying, because on the next update your patched binary will get **replaced**. \
You might rely on some changes in your fork for niri to start well, and so you may run into an annoyance if it's replaced beneath your feet.

But with this pr, you just blammo it into `/usr/local/bin/niri`, and continue to never worry again 😌 \
Since it comes *before* /usr/bin in both $PATH and in `ExecStart`, it'll correctly override the /usr/bin/niri binary as expected.

### DIFF
--- a/resources/niri.service
+++ b/resources/niri.service
@@ -11,4 +11,4 @@ Before=xdg-desktop-autostart.target
 [Service]
 Slice=session.slice
 Type=notify
-ExecStart=/usr/bin/niri --session
+ExecStart=niri --session


Axlefublr:
as an irrelevant but maybe fun / interesting aside: thanks to this feature of ExecStart, you can do something like `ExecStart=env some-command` to effectively make systemd search for a command in your $PATH \
`env` is in /usr/bin/env, and `env` is what we often use in shebangs to not hardcode paths. similar idea here, but without even needing to hardcode the path to `env`, as it's visible to ExecStart

YaLTeR:
Do other compositors do this?

Axlefublr:
I haven't checked. But this feels to me like an easy mistake to make: why bother searching up the behavior of ExecStart, if you can just hardcode the path and not worry about it.

Axlefublr:
sway doesn't provide a .service in its source code, hyprland only provides an “example” that I assume isn't even meant to be provided directly by hyprland; probably banking on packagers to actually confirm / do the systemd part.

Axlefublr:
like, that example forces a swaybg dependency, for instance. so, not an example to follow as a good practice.

YaLTeR:
Alr thanks

PR #3190, opened 2026-01-03
 nix: fix path replace in SystemD service 

LuckShiba:
Fix a build error in the Nix flake due to the path change in the just-merged #3187

~~I also changed the Nix workflow to _not_ continue on check errors, so we won't have a false-positive success check in the future like it just happened with this other PR.~~

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,7 +64,7 @@
           postPatch = ''
             patchShebangs resources/niri-session
             substituteInPlace resources/niri.service \
-              --replace-fail '/usr/bin' "$out/bin"
+              --replace-fail 'ExecStart=niri' "ExecStart=$out/bin/niri"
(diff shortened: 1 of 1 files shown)


YaLTeR:
The continue on errors is intended. I don't know nix and can't fix it when it breaks, so I don't want it to break our CI. Unfortunately, GitHub doesn't seem to offer a way to see that a job has failed without failing the entire workflow...

LuckShiba:
> The continue on errors is intended. I don't know nix and can't fix it when it breaks, so I don't want it to break our CI. Unfortunately, GitHub doesn't seem to offer a way to see that a job has failed without failing the entire workflow...

I see. I removed that commit.

YaLTeR:
Ty

PR #3191, opened 2026-01-03
use real error type

bend-n:
i think this should work

this makes the result type easier to use with anyhow, as it impls Error now.


YaLTeR:
sorry I'm not adding many lines of macro code for questionable benefit

bend-n:
i mean the macro is inlinable in which case it becomes many lines of similar code, is that better?

YaLTeR:
No, not really

bend-n:
could be as simple as `ReplyError(Cow<'static, str>)`

PR #3192, opened 2026-01-03
Feat/per output power control

Atan-D-RP4:
Built on-top #3077. Rebases cleanly.
Tested it with wlopm and it works. 
But I don't actually have multi-monitor setup so I can't test that part. Can some one who does please take a look at this?
Definitely needs more tests too.


r-h-7:
I just tested it with wlopm on an AMD GPU (not sure if that parts important) with multiple monitors and it works great. Thank you for implementing this and let me know if there's any info I can provide to help.

PR #3193, opened 2026-01-04
Improve dinit service files

markK24:
1. In the dinit section of the `niri-session` script, I made the script wait for the service to terminate. Otherwise, the script ends right there and some login managers immediately terminate the session because of that. Also I made some changes to the script for consistency with the systemd section.
2. Since the latest dinit version, variable substitution if now supported in all paths, thus I changed the dinit service files to have a better model: `niri-session` starts the `niri.target` service, which has dependencies on user (`$HOME/.config/dinit.d/niri.d`) and system (`/etc/dinit.d/user/niri.d`), and the `niri` service inself.
   Now, we can have niri start services like `swaybg` by creating a link/file with name `swaybg` in one of the abovementioned folders. I also mentioned this in the discussion davmac314/dinit#496.

Also I fixed documentation to reflect the changes.


YaLTeR:
Thanks

PR #3194, opened 2026-01-04
feat: add dynamic overview zoom controls

barrulus:
## Summary

- Add `zoom-presets` config option for overview
- Add `overview-zoom-cycle`, `overview-zoom-in`, `overview-zoom-out` actions
- Add `overview-zoom` animation config
- Zoom changes are animated smoothly
- Zoom persists across overview close/reopen; close animates from the current zoom
- `overview-zoom-out` and `overview-zoom-cycle` auto-open the overview if closed
- `overview-zoom-in` auto-closes the overview when already at the highest preset
- Only two binds (zoom-in/zoom-out) needed to fully control overview + zoom

This allows users to dynamically adjust the overview zoom level to see more or fewer workspaces at once. The zoom actions double as overview open/close controls, so users can bind zoom-in and zoom-out to scroll wheel and seamlessly enter/exit the overview by scrolling.

## Design alignment

This implementation follows niri's design principles:

- **Animated transitions**: Zoom changes animate smoothly using a configurable spring animation, consistent with other niri animations.
- **No performance impact when unused**: If `zoom-presets` is not configured, zoom actions are no-ops. The runtime zoom state is minimal (one f64 per monitor).
- **Mindful of invisible state**: The zoom level resets to the config default when closing the overview, so reopening it always starts from a predictable state. Users won't be surprised by a forgotten zoom level from a previous overview session.
- **Natural auto-open/close**: Zooming out from a closed overview opens it at the highest (least zoomed-out) preset, providing a smooth entry. Zooming in past the highest preset closes the overview, providing a natural exit. The zoom level is preserved during auto-close so the closing animation looks correct.

## Example config

```kdl
overview {
    zoom 0.5
    zoom-presets 0.5 0.25 0.1
}

animations {
    overview-zoom {
        spring damping-ratio=1.0 stiffness=800 epsilon=0.0001
    }
}

binds {
    // Scroll to zoom in/out — also opens/closes the overview
    Mod+WheelScrollUp   cooldown-ms=150 { overview-zoom-in; }
    Mod+WheelScrollDown cooldown-ms=150 { overview-zoom-out; }

    // Or cycle through presets with a key
    Mod+Z { overview-zoom-cycle; }
}
```

## Test plan

- [x] Enter overview → should use config `zoom` value
- [x] Press zoom cycle key → should animate through presets
- [x] Ctrl+Scroll → should animate to next/prev preset
- [x] Close overview → zoom should reset to config default
- [x] Re-enter overview → should start at config default again
- [x] No presets configured → zoom actions should no-op
- [x] Zoom-out when overview closed → should auto-open at highest preset
- [x] Zoom-cycle when overview closed → should auto-open at first preset
- [x] Zoom-in at highest preset → should auto-close overview
- [x] Auto-close preserves zoom level for closing animation

Closes: discussion #3047


youngcoder45:
Nice i really loved your idea zoomable overview hope this gets accepted

kumawatdarshan:
can we reuse the code here to have a generic zoomer(that follows mouse)? edit: nvm, there #3246

saygo-png:
I'm testing this patch right now. One thing that would be an improvement is making easy transitions between the overview mode and normal mode. In practice:

1. Using `overview-zoom-out` in normal mode should bring you to the zoomed out overview mode based on the preset
2. When in overview mode, zooming in to 1 (the default zoom of normal mode) should put you back into normal mode on whatever workspace you are looking at.

Doing this would allow the user to unbind `toggle-overview` and only use the zoom cycles. You no longer would have to toggle overview and then zoom, only zoom. I think this would be a nicer user experience, making the modes feel more cohesive.

I don't know Rust nor this codebase so I don't know the technical implications/requirements.
Thanks for creating this!

barrulus:
> allow the user to unbind `toggle-overview` and only use the zoom cycles. You no longer would have to toggle overview and then zoom, only zoom. I think this would be a nicer user experience, making the modes feel more cohesive.
> 

At the time I specifically chose not to do that as niri's contributors guideline specifies to keep things as simple as possible.

How you have phrased this actually makes it more simple. 

Right now I have 3 binds PLUS the overview bind. Doing this would halve that down to 2!

I will 4have a look into it, I don't think it will be too difficult. Thanks for testing!

barrulus:
Quick status update:

- Rebased onto current `main` (v26.04). The only conflict was a one-liner in `Options::from_config` where upstream added the new `blur` field while this PR's `Overview` struct gained `zoom_presets: Option<Vec<f64>>` (so `Overview` is no longer `Copy` and needs `.clone()`). Resolved.
- Re: @saygo-png's suggestion about auto-transitioning between overview and normal mode via zoom — that's already implemented on this branch:
  - `dcce19b7` feat: auto-open/close overview on zoom in/out — `overview-zoom-out` and `overview-zoom-cycle` auto-open the overview if closed; `overview-zoom-in` auto-closes when already at the highest preset.
  - `836e734f` fix: preserve zoom level when auto-closing overview on zoom-in — smooths the close animation so it doesn't snap to config default.

This means users can unbind `toggle-overview` and drive the whole flow with just the zoom binds, as suggested. Pushing the rebased branch now.

jtaala:
Fantastic work @barrulus!

The only feedback I can give (nitpick?) is that the animation seems to "jump" to the default zoom setting, and then smoothly animate in, when doing the standard niri exit-overview actions (e.g. clicking a window, or hitting "enter") when in overview.

That is, when I zoom out to "0.4" preset, then click a window - you can see a rapid jump to 0.8 before smoothly animating in.

For reference, my config is:
image

You can see this happening better if you do something like:
image

What's happening is when you click (or hit "enter") in overview, you'll see it "jump out" to 0.4 (the default zoom level) and then smoothly animate in.

Otherwise, great work!

barrulus:
Thanks. 
I guess I haven't looked that thoroughly at it because I Ctrl+mouse wheel zoom in in small steps right through.

jtaala:
> Thanks. I guess I haven't looked that thoroughly at it because I Ctrl+mouse wheel zoom in in small steps right through.

Hey @barrulus, just sent in a PR to your fork which should address the issue I mentioned (it essentially takes the `overview_zoom_presets` into account when returning the standard (default) zoom level that is used in standard niri overview actions and animations).

barrulus:
Ooh thanks I will check it out later

barrulus:
> Hey @barrulus, just sent in a PR to your fork which should address the issue I mentioned (it essentially takes the `overview_zoom_presets` into account when returning the standard (default) zoom level that is used in standard niri overview actions and animations).

I like what you have done. I will merge this to my fork and update this PR with those changes - I don't think this PR will ever see the lioght of day but its good work worth preserving :)

jtaala:
> > Hey @barrulus, just sent in a PR to your fork which should address the issue I mentioned (it essentially takes the `overview_zoom_presets` into account when returning the standard (default) zoom level that is used in standard niri overview actions and animations).
> 
> I like what you have done. I will merge this to my fork and update this PR with those changes - I don't think this PR will ever see the lioght of day but its good work worth preserving :)

Fingers crossed @YaLTeR can consider this one!  For me this functionality is very useful (I find I use the different zoom/overview levels often for different purposes - e.g. just peeking what I have around me vs zooming out to see muliple workspaces and windows etc.).

We could even combine it with the current zoom setting since it naturally fits in there, e.g.

```
overview {
    zoom 0.8 0.6 0.4
}
```
which is a natural extension of the current "zoom" setting.  I.e. just set 1 (like it is now), or set more if wanted.  I think it does extend and fit nicely in niri's functionality.

In any case, nice work @barrulus!

barrulus:
Thank you!

I have been messing with exposing every output as a carousel item Cover Flow style into the zoomed overview. 

Have paused it in Barrulus-custom for a bit as I needed to reset my tooling

barrulus:
Merged @jtaala's fix (barrulus/biri#2) into this branch — it's the `d122995e` commit above. Closing the overview with a click or enter no longer snaps to the default zoom before the close animation runs; it animates out from wherever you were zoomed. Thanks Jay.

One behaviour note for reviewers: the zoom level now persists across close/reopen instead of resetting to the config default. I prefer it this way — the overview comes back where you left it — but it does make the "zoom resets to config default when overview closes" bullet in the description outdated. I'll update the description to match.

jtaala:
> One behaviour note for reviewers: the zoom level now persists across close/reopen instead of resetting to the config default. I prefer it this way

I prefer this as well.  But I did think that maybe others might not.  This does seem to make the overview "zoom" config setting pretty much redundant.  Which got me thinking about just having a single "zoom" setting to be like presets, e.g.

image

Having one value would be as niri's current behaviour, and adding more feels quite natural (from a config pov).

jtaala:
> One behaviour note for reviewers: the zoom level now persists across close/reopen instead of resetting to the config default. I prefer it this way — the overview comes back where you left it — but it does make the "zoom resets to config default when overview closes" bullet in the description outdated. I'll update the description to match.

Hey @barrulus, I've just submitted a PR (https://github.com/barrulus/biri/pull/3) for you to have a look at.  It addresses the above by adding option `zoom-use-last-preset` to control this behaviour.  Setting to true allows standard niri overview actions (e.g. activate by hot-corner or keybind) to use the last active zoom preset (which some of us prefer as mentioned).  Setting to false (the default) causes overview to use niri's zoom setting (which is 0.5 by default) - which others may prefer.

image

Issue #932, opened 2025-01-05
Pinned/sticky/show on all workspaces floating windows


YaLTeR:
Add some way to show a floating window on all workspaces. For things like picture-in-picture, webcam, etc.

I'm not entirely sure what would be the best way to structure the code.

1. UX-wise, it would make sense to have an extra floating space for pinned windows on a Monitor, that renders above all workspaces. This would require somehow integrating keyboard navigation between workspace-local and pinned floating spaces, so e.g. `focus-window-up` considers windows in both.
1. Other compositors seem to duplicate pinned windows in all workspaces. So they look like they sit inside each workspace's own floating space, but on all workspaces at once. This makes keyboard navigation obvious, but our Layout code really doesn't expect one window to live in several places at once.

I think option 1 should be easier to implement and result in clearer UX, but as usual, need to try to see what kind of roadblocks I bump into.

Of course both of these should have actions to toggle/set pinned status, a window rule to open-pinned, probably a window rule matcher for is-pinned.

sindrenm:
Perhaps a separate FR, but similar sticky functionality for bars (like Waybar) would be great! It feels a little awkward to have it move together with the rest of the windows when switching workspaces.

luleyleo:
@sindrenm I'm using Waybar and mine is sticky. Maybe you haven't set `"layer": "top"` in your config?

sindrenm:
Oh thanks, setting that absolutely did the trick! I wasn't aware of how layer shells worked, that's pretty nifty!

bugeats:
> This would require somehow integrating keyboard navigation

If the "sticky" windows are always fixed to the head or the tail of the row, then no changes to keyboard nav are required. So on your monitor they would appear pinned to the left or right edge of the screen.

Something like this:

```
        --------------------Monitor Frame-----------------------       
        |                                                      |       
        ----------------                       -----------------       
 -------|              |-----------------------|               |-------
 |      |      W1      |  |        |        |  |       W6      |      |
 |  W2  |              |  |   W3   |   W4   |  |               |  W5  |
 |      | sticky: left |  |        |        |  | sticky: right |      |
 -------|              |-----------------------|               |-------
        ----------------                       -----------------       
        |                                                      |       
        --------------------------------------------------------       
```
      
I would behave pretty much like the CSS property `position: sticky`.

I suppose the position of the normal windows might have to account for the sticky boxes when doing things like centering.

My use case is having always-visible windows for stuff like chat and video calls.

[...]

PR #3205, opened 2026-01-05
Add pinned window support

my4ng:
This attempts to implement #932 based on https://github.com/YaLTeR/niri/issues/932#issuecomment-3634047242, by pinning windows to a monitor across its workspaces. More concretely, this means all pinned windows are moved to the new workspace on switching, without introducing a new `PinnedSpace` or duplicating windows across workspaces. 

**EDIT**: Windows are first moved to a temporary space of the monitor, and fully moved to the new workspace only after animation is done (to appear stationary).

Since pinned windows are just normal windows, navigation between windows remains the same as before. The disadvantage is that pinned windows only appear on the focused workspace in overview, ~~and pinned windows are animated alongside the new workspace on workspace switching~~.

~~Currently, as long as the pinned windows are on the active workspace *and* floating, regardless of their status when pinned, they will be moved to the newly switched workspace, otherwise they will be unpinned.~~

**EDIT**: Pinned windows will keep their latent pinned status even if they are not floating on workspace switch, though they won't be carried to the new workspace. As soon as they become floating, they will remain fixed to the monitor on workspace switch.

Windows can be pinned via the `ToggleWindowPinned` action, e.g.

```
binds {
    Mod+P { toggle-window-pinned; }
    ...
}
```

New window rules `open-pinned` to open a window pinned (orthogonal to `open-floating`) and `is-pinned` to match:

```
window-rule {
    ...

    open-pinned true
}

window-rule {
    match is-pinned=true

    ...
}

```

New IPC action to toggle window pinned status: 

```
niri msg action toggle-window-pinned
niri msg action toggle-window-pinned --id <ID>
```

**NOTE**: This is only *a* solution to pinned windows, but IMO much, much simpler than alternative designs. Please check it out and provide feedback on the design/pinning behaviour logic if you are interested.

### DIFF
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -541,7 +541,7 @@ window-rule {
 }
 ```
 
-Only *floating* pinned windows will be actively pinned to the monitor on workspace switch, so you likely want `open-floating` too.
+Only *floating* pinned windows will remain fixed to the monitor on workspace switch, so you likely want `open-floating` too.
 
 ```kdl
 // Open the Firefox picture-in-picture window as floating and pinned.
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -36,6 +36,7 @@ window-rule {
     match is-floating=true
     match is-window-cast-target=true
     match is-urgent=true
+    match is-pinned=true
     match at-startup=true
 
     // Properties that apply once upon window opening.
@@ -48,6 +49,7 @@ window-rule {
     open-fullscreen true
     open-floating true
     open-focused false
+    open-pinned true
 
     // Properties that apply continuously.
     draw-border-with-background false
@@ -177,7 +179,7 @@ You can find the title and the app ID of a window by running `niri msg pick-wind
 
 > [!TIP]
 > Another way to find the window title and app ID is to configure the `wlr/taskbar` module in [Waybar](https://github.com/Alexays/Waybar) to include them in the tooltip:
-> 
+>
 > ```json
 > "wlr/taskbar": {
 >     "tooltip-format": "{title} | {app_id}",
@@ -298,6 +300,19 @@ Matches windows that request the user's attention.
 ```kdl
 window-rule {
     match is-urgent=true
+}
+```
+
+#### `is-pinned`
+
+<sup>Since: next release</sup>
+
+Can be `true` or `false`.
+Matches windows that are pinned.
+
+```kdl
+window-rule {
+    match is-pinned=true
 }
 ```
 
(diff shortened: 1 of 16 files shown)


my4ng:
Updated to remove workspace switching animation and add window rules. Please try it out and see if you like the current design!

Sempyos:
I will probably play around with this later today and see how it is

YaLTeR:
I haven't taken a look at this, but just to set expectations: I will definitely want to explore adding a separate pinned floating space first. My idea is to try to have pinned windows retain their position in the Overview, kinda like top and overlay layer-shell surfaces. As I understand, this is what you want for many usecases.

my4ng:
@YaLTeR I should have updated the comment about the overview: the pinned windows appear in overview, but *only* on the active workspace. This is by design, since, in terms of implementation, the pinned windows can be easily moved to the temporary space on overview open just like on workspace switch.

My rationale is quite simple. If I have a pinned PnP and a pinned terminal that take half the space of the monitor, I would like to see what they would be covering on other workspaces in the overview. It doesn't make sense to display them on all workspaces that simply obstruct the overview.

YaLTeR:
Half screen wide terminal doesn't seem like a good use for the pinned that I have in mind?

my4ng:
I meant two or three pinned windows that may sum up to a substantial part of the screen space. Placing them across the overview feels obstructive to me, but that is just an opinion. If people do want that, it can be easily changed.

YaLTeR:
Covering too much space with pinned windows isn't really among the primary usecases I've been considering for designing niri built-in pinned windows. I don't think this usecase should be blocking 

Anyway feel free to leave this PR, idk when I'll get to pinned windows anyway

my4ng:
@Sempyos Any feedback? Would really appreciate it

Sempyos:
I don't have anything to provide

chek1337:
Thanks for this feature. I find it really necessary, especially when the workstation has only one monitor (for example, a laptop). 

Here are a few problems that I found. 
First, when switching between workspaces, the logic of selecting the current window is not entirely clear. I'm seeing an effect similar to a quick window focus and then unfocus.

https://github.com/user-attachments/assets/71defe35-2898-46ef-815e-1092a00b58d7

Also, when moving windows, the pinned window "detaches" until I return back to the workspace.

https://github.com/user-attachments/assets/1793b48e-dd2e-4680-910c-5f33348ffb69

Hikiru:
> I haven't taken a look at this, but just to set expectations: I will definitely want to explore adding a separate pinned floating space first. My idea is to try to have pinned windows retain their position in the Overview, kinda like top and overlay layer-shell surfaces. As I understand, this is what you want for many usecases.

this feels like the best UX. it should also mean that pinned windows stay above fullscreen windows, unlike normal floating windows, which i'm not sure if this PR does. i like how umbriel handles pinned windows.

PR #3235, opened 2026-01-11
Set the NIRI_BUILD_COMMIT in flake.nix

CoffeeImpliesCode:
When I have `github:YaLTeR/niri/main` (or any other rev) as a flake input to my NixOS config + its overlay, `niri --version` yields
``` niri 25.11 (unknown commit) ```
meaning that all other ways to get the commit info have failed at https://github.com/YaLTeR/niri/blob/45fc76328145e0bf91b13edb7a43ac1f5770cb48/src/utils/mod.rs#L98-L99
So this PR provides an additional fallback which works (at least for my case).

One thing I am unsure about is whether providing an envvar results in dependencies maybe not being able to be properly cached? (if they are cached at all)
I don't know enough about the nix/rust build process to know whether that might be an issue and I haven't come up with an alternative.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,39 +4,38 @@ version = 4
 
 [[package]]
 name = "accesskit"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf203f9d3bd8f29f98833d1fbef628df18f759248a547e7e01cfbf63cda36a99"
+checksum = "3eca13c82f9a5cd813120b2e9b6a5d10532c6e4cd140c295cebd1f770095c8a5"
 
 [[package]]
 name = "accesskit_atspi_common"
-version = "0.14.2"
+version = "0.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "890d241cf51fc784f0ac5ac34dfc847421f8d39da6c7c91a0fcc987db62a8267"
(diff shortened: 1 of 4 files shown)


YaLTeR:
Why? Just use NIRI_BUILD_COMMIT. This is what it's for

CoffeeImpliesCode:
I didn't want to break existing behavior. You write in the docs that for packaging niri maintainers should set NIRI_BUILD_COMMIT, but if I override that in the env section that would break.

YaLTeR:
flake.nix is a moral equivalent of a package isn't it

CoffeeImpliesCode:
Just to be sure: are you suggesting I change that I change NIRI_SHORTENV to NIRI_BUILD_COMMIT
or are you suggesting that I should make a personal overlay that fixes this for me, as this should not be fixed upstream?

YaLTeR:
I suggest using NIRI_BUILD_COMMIT in flake.nix. Let's also ask other nix ppl like @LuckShiba

CoffeeImpliesCode:
Okay I changed it.

LuckShiba:
> I suggest using NIRI_BUILD_COMMIT in flake.nix. Let's also ask other nix ppl like @LuckShiba

It works (first command)
image

An alternative would be setting the full version string too, like the second screenshot (niri-flake), but I think this is already good enough.

LuckShiba (approved):

YaLTeR:
Thanks

CoffeeImpliesCode:
Thx!

PR #3284, opened 2026-01-20
docs: remove unnecessary clipboard dependencies from Arch installation

Xarth-Mai:
Removes `wl-clipboard` and `cliphist` from the Arch Linux installation instructions.

**Reason:**
`dms` has recently updated to its own zero-dependency clipboard management implementation. Therefore, external clipboard utilities are no longer necessary, and removing them helps keep the installation command minimal.

**More info:**
https://danklinux.com/blog/v1-2-release#native-clipboard-and-history
[https://github.com/AvengeMedia/DankMaterialShell/commit/ec8ab47
](https://github.com/AvengeMedia/DankMaterialShell/commit/ec8ab4746228f53e3885f9319c2a3121d60cc4fe)


YaLTeR:
Thanks

PR #3299, opened 2026-01-22
layout/monitor: round workspace render geo to physical pixels

### LINKED ISSUE #3298: Top-edge pointer events missed on non-first workspace (scale ≠ 1)
When using a fractional output scale (scale ≠ 1), pointer events at the very top edge of the screen do not reach maximized windows unless the window is on workspace 1.

Steps to reproduce with Chrome (tab bar):
- Switch to workspace 2 (or any workspace other than 1).
- Open and maximize Chrome.
- Move the cursor to the top edge.

Pointer events does not reach the window:

Image

skrmc:
Fixes #3298.

The workspace render rectangles could end up at fractional physical positions, causing a mismatch near the edge.

This PR rounds the workspace origin to physical pixels using existing helpers:
`to_physical_precise_round(scale).to_logical(scale)`.

### DIFF
--- a/src/layout/tests.rs
+++ b/src/layout/tests.rs
@@ -3691,6 +3691,41 @@ fn expel_pending_left_from_fullscreen_tabbed_column() {
     check_ops(ops);
 }
 
+#[test]
+fn workspace_render_geo_at_fractional_scale() {
+    let ops = [
+        Op::AddScaledOutput {
+            id: 1,
+            scale: 1.1,
+            layout_config: None,
+        },
+        Op::AddWindow {
+            params: TestWindowParams::new(1),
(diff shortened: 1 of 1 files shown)


YaLTeR:
Good catch but the problem is actually deeper. All components of the position are already rounded to physical pixels, but the final coordinate ends up very very slightly off due to floating-point rounding. I'm thinking what's the best fix here is and what other places could be affected.

YaLTeR:
Specifically, in this case, the cursor y is 0 and the workspace y is 0.0000000000002

YaLTeR:
Thanks. I added a clarifying comment and a test

PR #3304, opened 2026-01-23
Add tablet option map-to-active-output

qqwa:
Makes it more convenient to use a tablet as a mouse replacement.

Still needs a separate way to change the active output, like using touch/mouse to move the cursor over the edge or with a key binding for example running the action `focus-monitor-next`.

### DIFF
--- a/docs/wiki/Configuration:-Input.md
+++ b/docs/wiki/Configuration:-Input.md
@@ -89,6 +89,7 @@ input {
     tablet {
         // off
         map-to-output "eDP-1"
+        // map-to-focused-output
         // left-handed
         // calibration-matrix 1.0 0.0 0.0 0.0 1.0 0.0
     }
@@ -280,6 +281,10 @@ input {
 Valid output names are the same as the ones used for output configuration.
 
 <sup>Since: 0.1.7</sup> When a tablet is not mapped to any output, it will map to the union of all connected outputs, without aspect ratio correction.
+
+Setting specific to `tablet`:
+
+- `map-to-focused-output`: <sup>Since: next release</sup> will map the tablet to the focused output, takes precedence over `map-to-output`.
 
 ### General Settings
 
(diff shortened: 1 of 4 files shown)


docs/wiki/Configuration:-Input.md
```diff
@@ -281,6 +282,10 @@ Valid output names are the same as the ones used for output configuration.
 
 <sup>Since: 0.1.7</sup> When a tablet is not mapped to any output, it will map to the union of all connected outputs, without aspect ratio correction.
 
+Setting specific to `tablet`:
+
+- `map-to-active-output`: will map the tablet to the active output, takes precedence over `map-to-output`.
```
LuckShiba: Missing the `<sup>Since: next release</sup>` prefix here

LuckShiba (commented):
Seems to be fine, worked with my tablet and I think it makes sense for it to get precedence. Very nice-to-have feature!

docs/wiki/Configuration:-Input.md
qqwa: Thanks for the review, I added the release prefix

YaLTeR:
Does sway or others have this kind of setting? If so then what do they call it?

LuckShiba:
> Does sway or others have this kind of setting? If so then what do they call it?

Sway doesn't seem to have this. In Hyprland it is `output = current` (may be a good design!),

YaLTeR:
1. Let's rename to `map-to-focused-output`, it's consistent with our terminology elsewhere (`niri msg focused-output`, `focus-at-startup`, etc.)
2. While `map-to-output "focused"` would naturally prevent the ambiguity, in this case I think it's worse because `map-to-output` accepts any string as an output name, so misspellings such as `map-to-output "focuesd"` won't get automatically caught. We could add custom parsing to validate it but that may be too much effort so let's keep it as is I guess.

@qqwa do you want to change this or should I do it?

qqwa:
I changed it, thank you

YaLTeR:
Thanks, works well

PR #3305, opened 2026-01-23
Window geometries relative to the output

Szybet:
I needed a way to know what is seen now on the screen in what parts, exactly where (As current available information (niri msg windows) didn't show anything if for example the window is centered, or offset in any way). I came up with this, I followed https://github.com/YaLTeR/niri/pull/1265 for suggestion on how to implement this. Just a sketch, suggestions are welcome


jbms:
This is a feature I'm also interested in.

A few comments: 
- the WindowGeometry is partially redundant with the existing WindowLayout, and I believe it is fully redundant for floating windows.  To make it non-redundant, you could eliminate the new WindowGeometry type and instead add a field to Workspace that specifies what the current viewport is and how it maps to the output coordinate space. 
- that would also provide a natural way to integrate with the event steam without producing an excessive number of messages as you move the viewport.
- there is the question of how overview mode should be handled.  Maybe there should be a separate Option field added to Workspace indicating the workspace to output coordinate map for the overview.
- there is the question of whether to expose the geometry of windows within the recent window view.  I think that could be skipped unless there is a use case for it.

Szybet:
Good ideas, let's wait for YaLTeR for his thoughts, then I can apply the changes

ycf-anon:
@YaLTeR,

sorry for the ping, but we really need this

YaLTeR:
i'm sorry do we have a support contract?

binarin:
I'm also experimenting with different options to make information about which windows are (partially) visible, and their spatial relationship - I want to have this in `emacs`, first to make `frame-visible-p` to really work.

First thing I did was is to go against https://github.com/niri-wm/niri/pull/1265#issuecomment-2812309501 (the last paragraph) and add this information [to scrolling layout `tiles_with_ipc_layout`](https://github.com/binarin/niri/blob/8ed0da44d974c32c6877d2f4630c314da0717ecb/src/layout/scrolling.rs#L2429).

(Note: I don't know why, but for me github strips the comment anchor in the link above when clicked directly, but I can go to the exact comment using "open in the new tab" on it)

This was the simplest change that kinda worked, but there are way more things that need to be resolved to make it universally useful:
- When application itself wants to be aware about its own windows state within niri, it's not straightforward to map app windows to niri window ids (there are `pid`/`title`, but it falls short when titles are not unique). For now I've temporary injecting random string into title to establish the mapping.
- Outputs configuration is not present in event stream, so some workaround was needed - for now I'm using a set of unique output names in `WorkspacesChanged` to detect change and requests outputs config separately via rpc - but it's just a proxy (it works quite fine for the most common case of external display(s) being (dis)connected).
- Windows in a tabbed column can be within the view, but still not visible - this information is also not exposed.

It looks like this PR is adding to much complexity (new `tiles_with_geometry` method, etc.) for too little gain. 

In the merged PR @YaLTeR [proposed that all this geometry stuff should be calculated mostly client side](https://github.com/niri-wm/niri/pull/1265#discussion_r1995523193). And given that everyone has different needs, maybe it'd be better to just expose raw information in niri-ipc event-stream, to make every sort of calculation possible: viewport information for active workspaces, column widths (and tile heights within columns), information about window order in tabbed columns, output configuration changes, etc...

so1ve:
I think the existing render-position API can simplify this now. `Workspace::tiles_with_render_positions()` already provides render-order positions and tab visibility, and `Tile::animated_tile_size()` provides the matching size. The geometry stays in logical `f64` coordinates so fractional scaling does not lose precision:

https://gist.githubusercontent.com/so1ve/6e19e5da1c437cea301b9c71abc940b5/raw/pr-3305-use-render-positions.patch

PR #3307, opened 2026-01-24
wiki: Rename Layer-Shell-Components.md

SED4906:
The name contained a U+2010, which looks like a '-' but isn't.
See also: https://github.com/ostreedev/ostree/issues/3431

### DIFF
--- a/docs/wiki/_Sidebar.md
+++ b/docs/wiki/_Sidebar.md
@@ -7,7 +7,7 @@
 * [Tabs](./Tabs.md)
 * [Overview](./Overview.md)
 * [Screencasting](./Screencasting.md)
-* [Layer‐Shell Components](./Layer%E2%80%90Shell-Components.md)
+* [Layer-Shell Components](./Layer-Shell-Components.md)
 * [IPC, `niri msg`](./IPC.md)
 * [Application-Specific Issues](./Application-Issues.md)
 * [Nvidia](./Nvidia.md)


Sempyos:
See https://github.com/YaLTeR/niri/pull/2447

YaLTeR:
I am still waiting for someone to definitively show me that this won't break the GitHub Wiki heading.

SED4906:
It'll change how the heading looks on the GitHub Wiki, I'm not sure there's a way to avoid that. I've seen both Layer-Shell and Layer Shell used interchangeably, so it's probably not that important whether the hyphen is visible.

YaLTeR:
What about https://github.com/YaLTeR/niri/pull/2447#issuecomment-3479017120

SED4906:
It used to be possible to define the page title within the markdown, it hasn't been possible for many years.

YaLTeR:
Do I understand right that this PR will break existing links to this wiki page?

SED4906:
Yes, the links are from the filename. This particular page doesn't seem to have many links from elsewhere that would need to be updated.

YaLTeR:
How necessary is this PR anyway? I'm successfully using niri on Silverblue, and there are bootc images with niri (Zirconium). Plus, the underlying issue seems to be on its way to be fixed.

SED4906:
It seems to only affect installer ISOs, `bootc switch` is unaffected. Removing `/usr/share/doc` seems to be a common step in many bootc images, and that works. I assume that the underlying problem in ostree will be fixed before a new release of niri, anyways.

PR #3318, opened 2026-01-26
avoid clobbering XDG_CURRENT_DESKTOP

cmm:
Don't set `XDG_CURRENT_DESKTOP` to "niri" if it's already set to whatever.

This (plus wrapping the session script to set `XDG_CURRENT_DESKTOP` to "GNOME", specifically) makes a niri session work properly on my corporate laptop (Ubuntu 22.04 plus every piece of enterprise spyware imaginable plus some custom monstrosities courtesy of the IT security dept).  And I really did try doing it Properly (editing `OnlyShowIn` entries in autostart desktop files, monitoring DBus, etc.) -- nothing worked.  But saying "it's totally a GNOME session sir I swear!" just did, so.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -89,8 +89,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             }
         }
 
-        // Set the current desktop for xdg-desktop-portal.
-        env::set_var("XDG_CURRENT_DESKTOP", "niri");
+        if env::var_os("XDG_CURRENT_DESKTOP").is_none() {
+            // Set the current desktop for xdg-desktop-portal.
+            env::set_var("XDG_CURRENT_DESKTOP", "niri");
+        }
         // Ensure the session type is set to Wayland for xdg-autostart and Qt apps.
         env::set_var("XDG_SESSION_TYPE", "wayland");
     }


YaLTeR:
What didn't work? Also does it work if you set to something like `niri:gnome`?

cmm:
It's been a while (a couple of months, but I'm getting a bit senile) so I don't quite remember _everything_ that didn't work.  But one specific and critical thing was launching the corporate VPN client from browser -- every tool I tried was telling me that all the portals are working fine and my niri session is working as good as it's ever going to, but only giving up and setting `XDG_CURRENT_SESSION` to "GNOME" actually fixed that.

Maybe the base system GNOME/XDG plumbing is just too old (I mean, almost 4 years), maybe I missed something, who knows.

(If I set `XDG_CURRENT_SESSION` to "niri:gnome", won't (unpatched) niri just reset it to "niri"? I probably don't understand the question)

YaLTeR:
> (If I set `XDG_CURRENT_SESSION` to "niri:gnome", won't (unpatched) niri just reset it to "niri"? I probably don't understand the question)

I mean with your patch, try `niri:gnome` instead of just `GNOME`. I wonder if it will work.

Also btw you can override this env variable for specific apps that want them. Even set it in the niri config `environment {}` and/or manually import into systemd global environment after starting niri.

cmm:
> Also btw you can override this env variable for specific apps that want them. Even set it in the niri config environment {} and/or manually import into systemd global environment after starting niri.

None of those helped when I tried it back then iirc, but I'll retry at the nearest opportunity.

cmm:
So:

- The PR is indeed not needed, as this addition to `config.kdl` does the trick (not sure why this or a moral equivalent didn't work for me earlier.  Oh well, skill issues everywhere):
```
environment {
    "XDG_CURRENT_DESKTOP" "GNOME:niri"
}
spawn-at-startup "dbus-update-activation-environment" "--systemd" "XDG_CURRENT_DESKTOP"
```

- To your other question: anything beginning with "GNOME" satisfies whatever piece of software that apparently checks the value of `XDG_CURRENT_DESKTOP`.  So "GNOME" is fine, "GNOME:blah" is fine, "GNOME:niri" is fine, but "niri:gnome" isn't.

cmm:
(forgot to close)

YaLTeR:
Interesting. FWIW this kind of setup won't help with services autostarted quickly enough by systemd, so if that becomes a problem we can think about something like this PR.

cmm:
maybe niri can postpone notifying systemd that it's ready until after all `spawn-at-startup`s are done (except how would you define "done" here properly without in effect specifying/implementing half of systemd yourself).

the robust thing to do is letting the user control the value of `XDG_CURRENT_DESKTOP` niri sets in the first place, like this PR does, except maybe in a more explicit (less accident-prone) way, like with a command-line argument?

cmm:
> Interesting. FWIW this kind of setup won't help with services autostarted quickly enough by systemd, so if that becomes a problem we can think about something like this PR.

[ Narrator: the IT changed something, and this indeed has now become a problem :frowning_face:.  but #3322 works great! ]

PR #3320, opened 2026-01-26
Docs: Update Electron info

Sempyos:
There were changes made that remove the env variable: https://github.com/electron/electron/issues/48001

### DIFF
--- a/docs/wiki/Application-Issues.md
+++ b/docs/wiki/Application-Issues.md
@@ -2,7 +2,7 @@
 
 Electron-based applications can run directly on Wayland, but it's not the default.
 
-For Electron > 39, you can use the command-line flag if the app does not default to Wayland:
+For Electron ≥ 39, you can use the command-line flag if the app does not default to Wayland:
 ```
 --ozone-platform=wayland
 ```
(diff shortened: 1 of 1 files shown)


Sempyos:
I am unsure if it would be a better idea to remove the bottom section, and only have these two instead, with rewording of course. For now I feel this works

LuckShiba:
Shouldn't the app always default to Wayland in newer versions?

Sempyos:
It should, but if it does not, it would be nice to provide a solution in the docs

docs/wiki/Application-Issues.md
YaLTeR, line 1: One of these comparisons should be inclusive. Either ≥ 39 or ≤ 39. Which one is it?

YaLTeR: Also, this would be for electron ≤ 28 then?

Sempyos: Correct, I added that to the PR for more clarity

Sempyos: Added ≥ 39

YaLTeR:
Thanks, also resolve comments once you're done with them

Sempyos:
Alright

PR #3327, opened 2026-01-28
backend/winit: DMA-BUF setup for Nvidia support

Fingel:
Hello! Thank you for your amazing work on Niri, I moved from PaperWM and haven't looked back.

I wanted to hack on Niri a bit (There's a feature that's been bouncing around my head, I wanted to see if it was feasible first) but when I went to run it using the winit backend (for the quick iteration!) I was met with errors when trying to start any application:

```
2026-01-28T04:26:33.451301Z  WARN niri::render_helpers::surface: failed to import surface: Error accessing the buffer (BufferAccessError::NotManaged)
2026-01-28T04:26:33.452208Z  WARN niri::layout::tile: error rendering window opening animation: error rendering to offscreen buffer

Caused by:
    Failed to bind Framebuffer
2026-01-28T04:26:33.452314Z  WARN niri::layout::scrolling: error creating a closing window animation: error rendering contents

Caused by:
    0: error rendering to texture
    1: error binding texture
    2: Failed to bind Framebuffer
```
To be clear, this **only happens** on the winit backend, the tty backend has been working perfectly for me. I did some looking around and this appears to be due to the lack of a global DMA buffer, NVIDIA will just not work with wl_drm alone.

I tried to match the existing structure of the TTY module as much as possible. The rest of the code is basically [copied directly from Smithay/Anvil](https://github.com/Smithay/smithay/blob/f9d4d7ffd82d743fb89737b3ffaed57505b477d1/anvil/src/winit.rs#L148). I'm not super knowledgeable about this stuff so I tried to play follow the leaders here. 

This fixes the issue on my machine. I also tested in a VM where it follows the fallback code path and works there as well.

Thanks for taking a look! 

Original commit message:
--
Creates a dmabuf global in the same manner as the tty backend. This fixes applications failing to launch on Nvidia when using the winit backend.

This code is adapted from Smithay's Anvil compositor. See smithay/anvil/src/winit.rs

### DIFF
--- a/src/backend/winit.rs
+++ b/src/backend/winit.rs
@@ -6,6 +6,7 @@ use std::sync::{Arc, Mutex};
 
 use niri_config::{Config, OutputName};
 use smithay::backend::allocator::dmabuf::Dmabuf;
+use smithay::backend::egl::EGLDevice;
 use smithay::backend::renderer::damage::OutputDamageTracker;
 use smithay::backend::renderer::gles::GlesRenderer;
 use smithay::backend::renderer::{DebugFlags, ImportDma, ImportEgl, Renderer};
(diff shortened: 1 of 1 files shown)


Ramblurr:
I can report that this works great for me with an Nvidia 4090 (open source drivers). 

I've been blocked from hacking on niri, and this has totally unlocked it. Thanks @Fingel

YaLTeR:
Thanks!

PR #3348, opened 2026-01-30
doc: Add note about border and included configs

NicolaiRuckel:
If you split up the config first and then want to enable borders it is super easy to miss that you have to turn them on explicitly. Hence, I think there should be a note in the border documentation.


YaLTeR:
Perhaps better to adjust the example snippet to have an "on" and a comment saying that "on" is required in included files?

NicolaiRuckel:
I like that it would be more difficult to miss that way. The only downside I can see is that we can't have a link to the border special case that way, right?

YaLTeR:
Could have both a note and a mention in the comment in the example perhaps

NicolaiRuckel:
Sounds good to me. I updated my patch accordingly.

PR #3366, opened 2026-02-01
feat: keycode resolution, keycode triggers

tyrypyrking:
Solves the issues with multiple layouts and exotic xkb keymaps behaving undesirably. By allowing for binds to physical keys and/or resolving all keybinds against default xkb layout. This allows niri to match behavior of majority of applications in regards with keysym resolution and mapping.

**Related discussions:** https://github.com/YaLTeR/niri/discussions/3255 (solves), https://github.com/YaLTeR/niri/discussions/2134 (useful proposal, but stale), https://github.com/YaLTeR/niri/discussions/1919 (solves), 
**Related issues**: https://github.com/YaLTeR/niri/issues/283 (solves)
**Similar PR:** https://github.com/YaLTeR/niri/pull/1920 (draft)

**Note on variants:** Pull request contains two variants for the same functionality provided in two commits. Cleaner (var2) and "Safer" (var1). Var1 is not much safer because avoiding xkb mutex lock and unwrap doesn't yeild any significant performance, nor will it remove other unwraps used by the libraries handling xkb connection.

PR #3387, opened 2026-02-03
Docs: Added AZERTY input info

Denperidge:
Hi! I love using Niri and apparently also love using a keyboard layout with a low usershare.

The default keybinds do not work for Azerty, and requires some digging using [wev](https://git.sr.ht/~sircmpwn/wev) (or a similar tool) to find the equivalent keynames.

So here's a workaround! A docs addition that documents the correct keynames, and has a copy-paste azerty version of the default keybinds

### DIFF
--- a/docs/wiki/Configuration:-Azerty.md
+++ b/docs/wiki/Configuration:-Azerty.md
@@ -1,0 +1,417 @@
+### AZERTY
+When using Niri with an AZERTY keyboard layout, the default workspace keybinds (or any keybinds using numbers) will not work as intended.
+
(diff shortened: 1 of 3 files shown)


ToxicMushroom:
I think you should also specify which variants of azerty this works for (There are a couple)
Otherwise I like the docs on this.

tyrypyrking:
Hi, I'm from the same indigenous people like you, just another tribe. Please checkout my [PR](https://github.com/YaLTeR/niri/pull/3366) it should be of help. It resolves keybinds to physical keys instead of dynamic names. You can still go back to how it was and use keycode:{num} instead of keysym for keybind assignment. Ex. Mod+Shift+keycode:41

Denperidge:
@ToxicMushroom Variants should be all accounted for! Required putting it on a different page though

PR #3395, opened 2026-02-04
ipc: allow load-config to relocate the path of the config

zimward:
After long hours at the nix caffeine-aided enrichment center to break config hot-reloading despite your best efforts from preventing that, i implemented a solution to increase the difficulty for the next round of "nix entities break hot-reloading somehow". 

With this one can relocate the config path at runtime, which should be useful for scripting purposes too and not just for us nix entities that want to use it for a systemd reload hook.

### DIFF
--- a/src/ipc/server.rs
+++ b/src/ipc/server.rs
@@ -475,12 +475,10 @@ fn validate_action(action: &Action) -> Result<(), String> {
         }
     }
 
-    if let Action::LoadConfigFile { path } = action {
-        if let Some(path) = path {
-            let p = Path::new(path);
-            if !p.exists() || !p.is_file() {
-                return Err(format!("file {path} does not exist"));
-            }
+    if let Action::LoadConfigFile { path: Some(path) } = action {
+        let p = Path::new(path);
+        if !p.is_file() {
+            return Err(format!("path does not point to a file: {path}"));
         }
     }
 


YaLTeR:
I feel like this one might be useful to mention somewhere on the wiki too

niri-ipc/src/lib.rs
```diff
@@ -936,7 +937,11 @@ pub enum Action {
     ///
     /// Can be useful for scripts changing the config file, to avoid waiting the small duration for
     /// niri's config file watcher to notice the changes.
-    LoadConfigFile {},
+    LoadConfigFile {
+        /// Path of the config file to load.
+        #[cfg_attr(feature = "clap", arg(long))]
+        path: Option<PathBuf>,
```
YaLTeR: In the Screenshot action I went with a String for whatever, probably good, reason, so let's do the same here

YaLTeR: Should have the same is_absolute validation and handling as the Screenshot path. See https://github.com/YaLTeR/niri/blob/189917c93329c86ac2ddd89f459c26a028d590ba/src/ipc/server.rs#L469-L473 and the cli counterpart.

docs/wiki/Integrating-niri.md
```diff
@@ -3,14 +3,17 @@ First, for creating a niri package, see the [Packaging](./Packaging-niri.md) pag
 
 ### Configuration
 
-Niri will load configuration from `$XDG_CONFIG_HOME/niri/config.kdl` or `~/.config/niri/config.kdl`, falling back to `/etc/niri/config.kdl`.
+Niri will by default load configuration from `$XDG_CONFIG_HOME/niri/config.kdl` or `~/.config/niri/config.kdl`, falling back to `/etc/niri/config.kdl`.
```
YaLTeR:
```suggestion
Niri will load configuration from `$XDG_CONFIG_HOME/niri/config.kdl` or `~/.config/niri/config.kdl`, falling back to `/etc/niri/config.kdl`.
```

I guess this is not really needed here, let's not overload the sentence

docs/wiki/Integrating-niri.md
```diff
@@ -3,14 +3,17 @@ First, for creating a niri package, see the [Packaging](./Packaging-niri.md) pag
...
 If both of these files are missing, niri will create `$XDG_CONFIG_HOME/niri/config.kdl` with the contents of [the default configuration file](https://github.com/YaLTeR/niri/blob/main/resources/default-config.kdl), which are embedded into the niri binary at build time.
 
 This means that you can customize your distribution defaults by creating `/etc/niri/config.kdl`.
 When this file is present, niri *will not* automatically create a config at `~/.config/niri/`, so you'll need to direct your users how to do it themselves.
 
 Keep in mind that we update the default config in new releases, so if you have a custom `/etc/niri/config.kdl`, you likely want to inspect and apply the relevant changes too.
 
+The default configuration locations can be overridden with the `NIRI_CONFIG` environment variable.
+Changing the configuration path during runtime is also possible via the niri IPC or using the command `niri msg action load-config --path <path-to-niri.kdl>`.
```
YaLTeR:
```suggestion
You can also change the configuration path at runtime via the niri IPC or using the command `niri msg action load-config --path <path-to-config.kdl>`.
```

YaLTeR (changes requested):

niri-ipc/src/lib.rs
zimward: the way its done now it also works with relative paths like `./config.kdl`, which makes it a bit nicer to use imo.

zimward: ah nevermind. i just had the file in my home dir, which happens to be the working dir of niri

docs/wiki/Integrating-niri.md
```diff
@@ -11,6 +11,9 @@ When this file is present, niri *will not* automatically create a config at `~/.
 
 Keep in mind that we update the default config in new releases, so if you have a custom `/etc/niri/config.kdl`, you likely want to inspect and apply the relevant changes too.
 
+The default configuration locations can be overridden with the `NIRI_CONFIG` environment variable.
+You can also change the configuration path at runtime via the niri IPC or using the command `niri msg action load-config --path <path-to-config.kdl>`.
```
YaLTeR:
```suggestion
You can also change the configuration path at runtime via the niri IPC or using the command `niri msg action load-config-file --path <path-to-config.kdl>`.
```

niri-ipc/src/lib.rs
```diff
@@ -936,7 +936,11 @@ pub enum Action {
     ///
     /// Can be useful for scripts changing the config file, to avoid waiting the small duration for
     /// niri's config file watcher to notice the changes.
-    LoadConfigFile {},
+    LoadConfigFile {
+        /// Path of the config file to load.
```
```suggestion
        /// Path of a new config file to load.
        ///
        /// If unset, reloads the current config file.
```

src/ipc/server.rs
```diff
@@ -474,6 +475,15 @@ fn validate_action(action: &Action) -> Result<(), String> {
...
 
+    if let Action::LoadConfigFile { path } = action {
+        if let Some(path) = path {
+            let p = Path::new(path);
+            if !p.exists() || !p.is_file() {
+                return Err(format!("file {path} does not exist"));
+            }
+        }
+    }
+
```
YaLTeR:
```suggestion
    if let Action::LoadConfigFile { path: Some(path) } = action {
        let p = Path::new(path);
        if !p.is_file() {
            return Err(format!("path does not point to a file: {path}"));
        }
    }

```

YaLTeR:
Thanks

PR #3398, opened 2026-02-05
Add Ctrl+Space to copy screenshot path to clipboard

domenkozar:
I'd often want to paste the path to screenshot created rather than the image itself, I've added this under `Ctrl+Space`. Also adds `Ctrl-C` to the help text.

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -114,6 +114,7 @@ pub enum Action {
     #[knuffel(skip)]
     ConfirmScreenshot {
         write_to_disk: bool,
+        copy_path: bool,
     },
     #[knuffel(skip)]
     CancelScreenshot,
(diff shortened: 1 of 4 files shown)


abcfy2:
Ctrl+Space is the default system-wide shortcut for toggling input methods across virtually all operating systems. Please don't override it.

domenkozar:
> Ctrl+Space is the default system-wide shortcut for toggling input methods across virtually all operating systems. Please don't override it.

Ctrl+C used here is also used to copy thing but taking a screenshot is kind of a modal operation, I wouln't expect having to change input method while doing a screenshot is really something one would ever do.

Either way, I don't mind what the shortcut is, as long as I can get a path into my clipboard

domenkozar:
Any suggestion what to use?

Issue #913, opened 2025-01-01
Long-running custom shaders


YaLTeR:
Currently you can set a custom shader for short animations (window open, close, resize) but not for something long-running. For example, you can't put a custom shader on a window or an output in general.

This can be used for fun effects, but also for things like grayscale mode, or desaturating unfocused windows and such.

The reason I haven't added it yet is that it requires some damage tracking design. While short animations can sidestep the problem and always draw with full damage, anything long-running should report correct damage whenever possible since this massively affects power consumption and background system load.

I have two ideas on how to handle this.

One is to have a property that you must set when using a custom shader that tells niri how it affects damage. The two obvious values are: does not affect damage (i.e. the shader changes each pixel independently), requires full damage (the shader can do whatever it wants, but same input texture always produces the same output texture), and requires full damage and is animated (the shader can do whatever it wants and accepts current time as input for animations).

Second idea is to also offer different shader interfaces for these values. Like, for shaders that do not affect damage, offer a `vec4 process(vec4 color)` interface that processes each pixel individually. I suspect this could be unnecessarily limiting, and well, if you're writing a custom shader and set the damage property wrong, then the worst thing that can happen is you get some weird redrawing on screen.

kosayoda:
Wanted to chime in that this feature would allow plugging in a blue-light filter shader (like the one in https://github.com/loqusion/hyprshade) for a simple replacement of gammastep, wlsunset etc.

alicealysia:
I prefer your second idea as it *seems* more performant.

partisani:
> I suspect this could be unnecessarily limiting

I don't know much about shader implementation, but what is the problem in adding something like `vec4 process(vec4 uv, vec4 color)`? Does it suffer from damage tracking too?

YaLTeR:
I haven't given this too much thought. Either way both options should be simple enough code wise

keksc:
I switched to niri today because Hyprland has pretty poor shader support and niri seems to have projects for long-running animation shaders, w/ hyprland i couldnt force the damage even if I had the time uniform and didnt want to use a custom build because I already have NVIDIA shit which takes ages to build when i updgrade my nixos system. I greatly enjoy the other niri features but would greatly appreciate if you implemented long-running custom shaders because It's what I've been searching for for a long time.

YaLTeR:
Curious, what is your use case?

Vescrity:
[...]

PR #3400, opened 2026-02-06
Add saturation support to layer and window rules

repentsinner:
image

## Summary                                                                                                              

  - Adds a `saturation` property to window rules and layer rules, allowing per-window/layer desaturation (`0.0` = grayscale, `1.0` = normal color) while waiting on long-running shader support described in #913.
  - Implements a `SaturatedSurfaceRenderElement` wrapper with a custom fragment shader, following the same pattern as `ClippedSurfaceRenderElement`
  - Integrates saturation into the existing `clipped_surface` shader so windows with corner rounding and desaturation don't need separate passes
  - Adds wiki documentation for both window rules and layer rules

  ## Motivation

  A common request is to visually distinguish inactive/unfocused windows (see [Discussion
  #2642](https://github.com/YaLTeR/niri/discussions/2642)). The existing `opacity` rule works but reduces readability by bleeding through the background. Desaturation is a subtler alternative that preserves contrast and legibility while clearly indicating focus.

  ```kdl
  window-rule {
      match is-active=false
      saturation 0.0
  }
```

## Test plan

  - Set saturation 0.0 on inactive windows, verify they turn grayscale and return to color on focus
  - Set saturation 0.5 for partial desaturation
  - Test layer rules with saturation on a layer surface (e.g. waybar)
  - Verify windows with corner rounding + saturation render correctly (single shader pass)
  - Verify cargo test -p niri-config passes (snapshot + wiki KDL parsing)
  - Verify cargo clippy --all --all-targets is clean

### DIFF
--- a/niri-config/src/layer_rule.rs
+++ b/niri-config/src/layer_rule.rs
@@ -13,6 +13,8 @@ pub struct LayerRule {
     #[knuffel(child, unwrap(argument))]
     pub saturation: Option<f32>,
     #[knuffel(child, unwrap(argument))]
+    pub color_filter: Option<String>,
+    #[knuffel(child, unwrap(argument))]
     pub block_out_from: Option<BlockOutFrom>,
     #[knuffel(child, default)]
     pub shadow: ShadowRule,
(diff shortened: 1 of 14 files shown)


HigherOrderLogic:
This should be covered by long-running shader (#913).

bluelinden:
> This should be covered by long-running shader (#913).

this is also likely superseded by yalter's work in the wip/branch branch, which already has this feature.

repentsinner:
closing until https://github.com/YaLTeR/niri/tree/wip/branch gets merged to main. If needed at that point, I'll have another crack at generic color shaders.

PR #3409, opened 2026-02-07
fix(backend/tty): add delayed rescan for connectors missing EDID on hotplug

coleleavitt:
## Problem

USB-C docks with DP MST / alt-mode (e.g. Lenovo ThinkPad USB-C Dock Gen 2) can report connectors as `Connected` to the kernel DRM subsystem **before** EDID data has been read. When this happens:

1. `connector.modes()` returns an empty list
2. `pick_mode()` returns `None`
3. `connector_connected()` logs `"no mode"` and skips activation

Smithay's `ConnectorScanner` treats `(Connected, Connected)` as a no-op — it does not re-emit events for already-connected connectors. This means the output gets stuck in a permanent "connected but never activated" dead state.

Whether activation succeeds depends entirely on timing: if a second `UdevEvent::Changed` fires after EDID completes, the connector recovers. This makes the bug **intermittent** — monitors sometimes come up and sometimes don't on the same hardware.

## Root Cause

The EDID race is in the kernel/dock firmware timing, but niri had no retry path for connectors that were connected but could not be activated due to missing modes.

## Fix

Two complementary mechanisms:

### 1. Handle `DrmScanEvent::Changed` (new in smithay PR #1923)

When a connector's mode list changes while it stays connected (e.g. EDID arrives after the initial probe returned empty/fallback modes), smithay now emits a `DrmScanEvent::Changed` event. We handle this by registering the crtc in `known_crtcs` (if no surface exists yet) so `on_output_config_changed()` can connect it. If a surface already exists, `on_output_config_changed()` will re-evaluate mode selection automatically.

**Note**: This requires bumping smithay to at least rev `9219bf8a9` (includes the `Changed` variant). The `rev` pin in `Cargo.toml` can be removed once niri updates smithay past this point.

### 2. Bounded rescan timer (defense-in-depth)

The `Changed` event only fires when `scan_connectors()` is called. It doesn't trigger rescans on its own — the kernel must fire a udev Changed event (or the timer must trigger a rescan). Since the kernel doesn't always fire a second udev event after EDID completes, we keep a bounded retry timer:

- After `device_changed()` processes connectors, `schedule_rescan_if_needed()` checks for connected connectors that have no matching surface (not yet activated) and are not non-desktop connectors
- If found, schedules a `calloop::Timer` (2 s delay) that re-invokes `device_changed()`, giving the kernel time to complete EDID reads
- Retries are capped at `MAX_RESCAN_RETRIES` (3) to prevent infinite rescheduling
- The timer self-clears when all connectors are successfully activated
- Existing timers are cancelled before scheduling new ones (cancel-and-reschedule)
- Timers are cleaned up on device removal

### New fields on `OutputDevice`

- `rescan_timer_token: Option<RegistrationToken>` — handle to cancel pending timers
- `rescan_retry_count: u8` — bounded counter, reset on successful activation

## Testing

- **Hardware**: ThinkPad P16 Gen 3, NVIDIA RTX PRO 4000 (nvidia-drm), Intel iGPU (i915), Lenovo USB-C Dock Gen 2, two LEN S27q-10 monitors (HDMI + DP MST)
- Builds cleanly (`cargo check`, `cargo clippy`, `cargo build --release`)
- Follows existing calloop timer patterns used elsewhere in `tty.rs` (e.g. VRR timers, redraw timers)

## Alternatives Considered

- **Polling loop**: Rejected in favor of bounded async timer to avoid blocking the event loop
- **Infinite retries**: Rejected — capped at 3 to avoid pathological cases
- **Timer only (no Changed event)**: Works but less responsive. The `Changed` event gives immediate feedback when modes actually arrive, while the timer serves as a fallback.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,14 +567,14 @@ dependencies = [
 
 [[package]]
 name = "calloop"
-version = "0.14.4"
+version = "0.14.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4dbf9978365bac10f54d1d4b04f7ce4427e51f71d61f2fe15e3fed5166474df7"
+checksum = "cb9f6e1368bd4621d2c86baa7e37de77a938adf5221e5dd3d6133340101b309e"
 dependencies = [
  "async-task",
  "bitflags 2.10.0",
  "futures-io",
- "nix 0.31.1",
+ "nix",
  "polling",
  "rustix 1.1.3",
  "slab",
@@ -599,7 +599,7 @@ version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "138efcf0940a02ebf0cc8d1eff41a1682a46b431630f4c52450d6265876021fa"
 dependencies = [
- "calloop 0.14.4",
+ "calloop 0.14.3",
  "rustix 1.1.3",
  "wayland-backend",
  "wayland-client",
@@ -940,7 +940,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1068,7 +1068,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1416,7 +1416,7 @@ dependencies = [
  "gobject-sys",
  "libc",
  "system-deps",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1735,7 +1735,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi 0.5.2",
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


coleleavitt:
## Upstream Root Cause & Fix

After deeper analysis, the underlying issue is in smithay's `ConnectorScanner` in `smithay-drm-extras`. The `(Connected, Connected)` arm is a no-op, which means mode-list changes on already-connected connectors are silently ignored. When a USB-C dock connector reports as `Connected` before EDID is ready (empty mode list), and a later rescan finds modes populated, no `Connected` event is re-emitted.

I've filed an issue and PR upstream:

- **Smithay issue**: Smithay/smithay#1922 — ConnectorScanner does not re-emit Connected event when modes change on already-connected connector
- **Smithay PR**: Smithay/smithay#1923 — Fix: compare mode lists on `(Connected, Connected)` and re-emit when they differ
- **niri issue**: #3410 — Tracking issue for this bug with cross-compositor analysis

### Why this PR is still valuable even with the smithay fix

The smithay fix ensures that when a rescan happens and modes have appeared, the compositor gets notified. However, this niri-level rescan timer is still needed because:

1. **No guaranteed second udev event**: The kernel may only fire one `Changed` event before EDID is ready — without a timer-based rescan, there's nothing to trigger a second `ConnectorScanner::scan()`
2. **Defense in depth**: The timer provides a bounded retry window (3 × 2s = 6s) regardless of udev event timing
3. **Complementary**: The smithay fix makes each rescan more effective (mode changes are detected), while this PR ensures rescans actually happen

YaLTeR:
I suppose this needs updating to use the new Changed event instead?

coleleavitt:
> I suppose this needs updating to use the new Changed event instead?

yes I'll update it today if I get time; thanks

YaLTeR:
I updated Smithay on main and added logging for `DrmScanEvent::Changed` in a217ad6424e58198519e2776bc3949bf92311960.

As far as I can see, no extra logic is actually needed. The delayed rescan is unnecessary because now we have `Changed`, and your commit with extra logic on `Changed` didn't actually do anything because the connector must've always been in `known_crtcs`, so the `if` condition never triggers. (It can trigger only if the connector failed to get a CRTC initially, and then got one before the `Changed` event, but I'm not sure this happens in practice. In any case we the drm scanner doesn't report this kind of later CRTC map event so it would be broken regardless.)

Please test niri-git and tell me if it works now.

PR #3414, opened 2026-02-07
Add support for window labels

metlos:
Hey!

First of all thanks for an utterly awesome piece of software. Niri makes my time using a GUI actually enjoyable :)

Recently, I wanted to mark certain windows such that they would render slightly differently (in my case with a red border, to remind me I should be careful working with them. Looking around the config gave me no idea how to do it but I came across several discussions on the topic (https://github.com/YaLTeR/niri/discussions/884, https://github.com/YaLTeR/niri/discussions/3120) and possibly others.

This lead me to believe I'm not alone needing this and so I rolled up my sleeves and tried to make my first contribution to your wonderful creation.

This PR makes it possible to programmatically tag or label windows where label is a pair of a key and an optional value. There can be multiple labels present on a window and many windows can have the same label. The PR also adds a match rule for these labels, making it possible to change the behavior/look of the windows based on the presence/absence of labels.

I didn't call the "tags", that the above mentioned discussions mention, tags but rather "labels" (taking inspiration from Kubernetes) because https://github.com/YaLTeR/niri/discussions/884 mentioned "tag" conflicts with the name of a wayland protocol.

This PR I think doesn't make https://github.com/YaLTeR/niri/discussions/884 fully possible, because it doesn't add "tag" and "remove-tag" (or rather "label" and "remove-label") as window rules but I think something similar could be achieved through scripting. Having "tag" both as a window rule and a match (so that it can be reacted upon) is IMHO dangerous because it could lead to infinite loops if used carelessly.

I think https://github.com/YaLTeR/niri/discussions/3120 would be fully possible with this PR.


bluelinden (approved):
LGTM, implementation seems intuitive and well thought-out.

src/window/mapped.rs
```diff
@@ -583,6 +588,36 @@ impl Mapped {
     pub fn is_urgent(&self) -> bool {
         self.is_urgent
     }
+
+    pub fn set_label(&mut self, name: String, value: Option<String>) {
+        let mut labels = self.labels.take().unwrap_or_else(HashMap::new);
```
HigherOrderLogic: You should be able to use an `.as_mut()` here instead of taking the value and do an extra allocation.

src/window/mapped.rs
```diff
@@ -583,6 +588,36 @@ impl Mapped {
...
+        let old = labels.insert(name, value.clone());
+
+        let changed = old.map(|v| v != value).unwrap_or(true);
```
HigherOrderLogic:
```suggestion
        let changed = old.is_some_and(|v| v != value);
```

src/window/mod.rs
```diff
@@ -171,6 +172,13 @@ impl<'a> WindowRef<'a> {
             WindowRef::Mapped(mapped) => mapped.is_window_cast_target(),
         }
     }
+
+    pub fn labels(self) -> Option<&'a HashMap<String, Option<String>>> {
+        match self {
+            WindowRef::Unmapped(_) => None,
```
HigherOrderLogic: Shouldnt windows be able to carry over their labels when they go from mapped to unmapped and vice versa?

metlos: Sorry it took me so long to get back to this. The reason unmapped windows weren't handled is purely my ignorance and not finding issues with it during manual testing :)

Hopefully the labels should be propagated now.

winged:
I'm currently daily-driving a patched Niri version with this PR applied. It feels stable and very useful: Especially with the matching rules, I can now mark critical windows (like where I'm SSHing into production servers). 

Very much looking forward to having this in the official release in one form or other!

metlos:
@bluelinden, @HigherOrderLogic is there anything more you want addressed in this PR? any design or UX concerns?

I'm willing to spend time on this, if you think more work is needed, because I think this is genuinely a useful feature (as evidenced by comments and emojis on the PR).

ycna07 (approved):

metlos:
@YaLTeR, sorry to ping directly, but what are your plans with this PR?

YaLTeR:
Hey, I don't have plans for this PR yet, I haven't had a chance to look at it, sorry

metlos:
> Hey, I don't have plans for this PR yet, I haven't had a chance to look at it, sorry

No worries. Looking forward to discussing it when you get to it.

PR #3424, opened 2026-02-10
Replace yalter/niri and vortriz/awesome-niri with niri-wm/niri and niri-wm/awesome-niri in README

bluelinden:
Figured I would quickly write this to save you the trouble - this just updates README links to point to the correct locations. Media still works as far as I can tell.


YaLTeR:
Ty

PR #3428, opened 2026-02-11
bump smithay: fix zxdg_exported crash

### LINKED ISSUE #3421: Crash with misbehaving client (KMail)
KMail seems a bit ill-behaved. It has caused problems in e.g. Sway as well. This crash happens while interacting with KMail, and is hard to reproduce on purpose. KMail itself does not crash.

I did not have full debug info installed at the time of the crash (running out of disk space on this laptop) so the log output only has this:

```
thread 'main' (2036) panicked at vendor/wayland-backend-0.3.11/src/sys/server_impl/mod.rs:1576:17:
Callback creating object zxdg_exported_v2@286 did not provide any object data.
```

After installing debug info, I was able to retrieve this from the coredump (in case it’s useful): [GDB backtrace](https://github.com/user-attachments/files/25196665/gdb.txt)

### System Information

* niri version: https://github.com/Naxdy/niri/commit/c8db88d9f15bcbcfb2dee89a62db1b82195caf48 — I have run into exactly the same issue with the upstream 25.11 release before, though, so I chose to report it here.
* Distro: openSUSE Tumbleweed 20260206
* GPU: Intel Haswell-ULT Integrated
* CPU: Intel Core i5-4210U

cmeissl:
this bumps smithay exactly one commit to https://github.com/Smithay/smithay/commit/ae14fa12f6ee3fc4dd4c766ffb21848f991a6a18 which includes the fix to prevent a crash when a client
requests to export an non xdg toplevel.

fixes #3421

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3390,7 +3390,7 @@ checksum = "b7c388c1b5e93756d0c740965c41e8822f866621d41acbdf6336a6a168f8840c"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/Smithay/smithay.git#2928e4f34541d957b7b3c3b3e13b2539cd44990f"
+source = "git+https://github.com/Smithay/smithay.git#ae14fa12f6ee3fc4dd4c766ffb21848f991a6a18"
 dependencies = [
  "aliasable",
  "appendlist",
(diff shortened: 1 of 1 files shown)


YaLTeR:
Thanks

PR #3439, opened 2026-02-13
bump calloop to resolve timer buildup

cmeissl:
~this tracks https://github.com/Smithay/calloop/pull/238~ (merged and released as 0.14.4)

resolves an issue where cancelling
timers can result in performance degradation
over time and memory buildup

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -567,14 +567,14 @@ dependencies = [
 
 [[package]]
 name = "calloop"
-version = "0.14.3"
+version = "0.14.4"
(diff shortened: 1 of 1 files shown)


hackeryarn:
Before this patch, I saw CPU climb to 50-90% over the course of the day. Ran this all day today and did not see it creep up like before.

CPU usage will get a small spike with high frequency input device, but seems to then settle down. Everything seems to work as expected.

YaLTeR:
Thanks

PR #3452, opened 2026-02-15
openbsd: implement close_range workaround

tobhe:
OpenBSD doesn't have `close_range()` so we have to close `< raw` manually in a loop and use `close_from` for everything `> raw`.

This depends on `close_from()` being available in the libc crate. I upstreamed support in https://github.com/rust-lang/libc/commit/2aabbf03958074908a4de9fa8bade1a2e8a43165 which should be available since 2.171. Niri is currently on 2.180 so that should work.


YaLTeR:
Thanks

PR #3453, opened 2026-02-15
fix: prioritize activation region hits over CSD overflow in scrolling hit testing

odtgit:
## what this fixes

focus-follows-mouse edge panning doesn't work reliably with CSD windows. with struts enabled and two 0.5-proportion columns side by side, moving the mouse to the screen edge should focus the adjacent peeking column - but CSD windows like firefox and alacritty block it most of the time, while SSD windows work fine.

## root cause

CSD windows have buffers and input regions that extend beyond their `geometry()` bounds (shadows, resize handles etc). in `ScrollingSpace::window_under()`, tiles are checked in render order and the first tile whose `hit()` returns `Some` wins. since `hit()` checks `is_in_input_region()` first - which delegates to the wayland client's input region without clipping to tile bounds - the CSD shadow bleeds into the strut zone, the active column steals the hit, focus-follows-mouse sees the same window already focused, and nothing happens.

SSD windows don't have this issue because their `geometry().loc` is `(0,0)` and their buffer doesn't extend beyond the geometry.

worth noting that `clip-to-geometry` only affects rendering, not input - so it can't help here either.

## the fix

instead of unconditionally clipping all pointer input to the activation region (which would break legitimate CSD surfaces extending beyond tile bounds, like GTK 3 subsurface popups or non-grabbing popups), this uses a two-pass approach in `ScrollingSpace::window_under()`:

1. **first pass** - only accept hits within the tile's activation region (tile bounds). this means the peeking column wins over the active column's CSD shadow in strut zones.
2. **second pass** - fall back to the full `hit()` including CSD overflow regions. this preserves normal pointer input on legitimate surfaces extending beyond tile bounds.

the first pass almost always matches (pointer is within some tile's bounds), so the second pass only runs when the pointer is in a CSD overflow zone outside all tile bounds - negligible performance impact.

adds `Tile::hit_within_activation_region()` which checks activation region first, then input region within it. the original `Tile::hit()` is unchanged.

also updated `resize_edges_under()` in `workspace.rs` to use `hit_within_activation_region()` for consistency - it had a comment explicitly noting it should match `window_under()` behavior, and resize edge calculations are relative to tile geometry so CSD overflow hits would produce meaningless edges anyway.

## what about other hit testing paths

- **floating windows** - use single-pass `hit_tile()`. CSD overlap between floating windows is a z-order concern, not a strut/peeking issue. left as-is.
- **interactive move** - uses single-pass `hit_tile()` on the single dragged tile. no overlap with itself, no change needed.
- **overview mode** - converts all hits to `Activate` and disables resize. unaffected.
- **tab indicators** - checked before tiles in the first pass so they win correctly. not rechecked in the second pass, but tab indicators don't extend into other columns' CSD zones in practice.

## tested with

same tight layout - 4px gaps, 2px struts, 2px focus ring border, `clip-to-geometry true`, `prefer-no-csd` on:

```kdl
layout {
    gaps 4
    struts {
        left 2
        right 2
    }
    border {
        width 2
    }
}
```

- edge panning via focus-follows-mouse works consistently on CSD windows (firefox, alacritty, chromium, discord, spotify)
- normal pointer input on CSD regions extending beyond tile bounds (shadows, resize handles) still works
- mod+right-click resize behaves correctly at tile edges
- SSD windows unaffected (hit() and hit_within_activation_region() are equivalent for them)
- also tested with different gap/strut/border geometries - all working

tried to test for regressions with the second pass (GTK 3 subsurface popups, non-grabbing popups) but couldn't find a good client that reliably produces CSD surfaces extending beyond tile bounds in a way that would exercise the fallback path. if anyone has a test case for that I'm happy to verify.

would definitely benefit from further testing, but since this enables a purely mouse-driven horizontal scrolling use-case that was broken before, I think it's worth checking out.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -594,6 +594,17 @@ impl HitType {
             .map(|hit| (tile.window(), hit.offset_win_pos(tile_pos)))
     }
 
+    /// Like `hit_tile()`, but clipped to the tile's activation region.
+    pub fn hit_tile_within_activation_region<W: LayoutElement>(
+        tile: &Tile<W>,
+        tile_pos: Point<f64, Logical>,
+        point: Point<f64, Logical>,
+    ) -> Option<(&W, Self)> {
+        let pos_within_tile = point - tile_pos;
+        tile.hit_within_activation_region(pos_within_tile)
+            .map(|hit| (tile.window(), hit.offset_win_pos(tile_pos)))
+    }
+
     pub fn to_activate(self) -> Self {
         match self {
             HitType::Input { .. } => HitType::Activate {
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -2998,6 +2998,31 @@ impl<W: LayoutElement> ScrollingSpace<W> {
                 // Round to physical pixels.
                 let tile_pos = tile_pos.to_physical_precise_round(scale).to_logical(scale);
 
+                // Prefer hits within the activation region so CSD shadows can't steal
+                // hits from adjacent tiles (e.g. peeking columns in strut zones).
+                if let Some(rv) =
+                    HitType::hit_tile_within_activation_region(tile, tile_pos, pos)
+                {
+                    return Some(rv);
+                }
+            }
+        }
+
+        // Fall back to CSD input regions extending beyond tile bounds (shadows, resize
+        // handles, GTK 3 subsurface popups) so normal pointer input still reaches them.
+        for (col, col_x) in self.columns_in_render_order() {
+            let col_off = Point::from((col_x, 0.));
+            let col_render_off = col.render_offset();
+
+            for (tile, tile_off, visible) in col.tiles_in_render_order() {
+                if !visible {
+                    continue;
+                }
+
+                let tile_pos =
+                    view_off + col_off + col_render_off + tile_off + tile.render_offset();
+                let tile_pos = tile_pos.to_physical_precise_round(scale).to_logical(scale);
+
                 if let Some(rv) = HitType::hit_tile(tile, tile_pos, pos) {
                     return Some(rv);
                 }
--- a/src/layout/tile.rs
+++ b/src/layout/tile.rs
@@ -867,6 +867,27 @@ impl<W: LayoutElement> Tile<W> {
         activation_region.contains(point)
     }
 
+    /// Like `hit()`, but only returns a result if the point is within the tile's activation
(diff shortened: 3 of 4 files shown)


Sempyos:
Have you tried using negative struts instead..?

odtgit:
Yes, that's how the investigation started. It was never working reliably
regardless of the strut size, even when hovering over the window (clicking
would work though). With the fix it's working for me regardless of the
strut size, as long as there is one, the focus will shift properly.

Tested - alacritty, firefox, chromium, discord, spotify -> all work, unless
maximized, ie. as intended.

YaLTeR:
The problem is that there can be a legit part of a surface that sticks out. Most popups should be fine as they grab pointer input, but what about nongrabbing popups? (I'm not sure if there's a good test client, but I think if you unconditionally return from the xdg toplevel `grab()` handler then GTK popups will stick around without a grab.) Some clients use subsurfaces for popups, most notably GTK 3. You can tell by them getting clipped with clip-to-geometry. This isn't good Wayland behavior, but it's behavior that exists and works, and it would be weird if those parts of the surface just didn't accept input.

Perhaps restricting focus-follows-mouse specifically (and not normal pointer input) from triggering outside the activation region is better?

odtgit:
Rebased on latest main. Regarding the concern about non-grabbing popups and GTK 3 subsurface popups:

The current implementation already handles this — it's a two-pass approach:

1. **First pass**: prefer hits within the activation region (tile bounds), so CSD shadows can't steal hits from adjacent tiles
2. **Fallback pass**: if no activation-region hit is found, fall back to full CSD input regions (shadows, resize handles, subsurface popups)

So surfaces extending beyond tile bounds still receive normal pointer input — they just can't win over a tile that the point is actually inside of. This prevents the edge-panning/focus-follows-mouse issue without breaking input delivery to legitimate CSD regions.

I did attempt a narrower approach that only restricted focus-follows-mouse (separate `window_under_for_focus` path), but it introduced subtle bugs due to mismatches between the FFM hit testing and the regular `contents_under` comparison, and actually had a larger blast radius in practice. The two-pass priority approach is simpler and more robust.

larsch:
I'm just trying this fix, because firefox being activated constantly without mouse actually going into the tile area is very annoying.

However, with this PR, it seems to still activate a firefox window which is completely off-screen, if you move the cursor the edge of the screen, so this PR does not solve it fully for me.

I really wish it would not activate an off-screen window.

PR #3465, opened 2026-02-16
Add active-window-* as options for floating window relative-to rule

contagnas:
This adds `"active-window-*"` options to the `RelativeTo` enum for `default-floating-position`. These position the new floating window relative to the currently active window. My use case to emulating window parenting for applications` that don't support it, but I could imagine other uses as well.

Demo:

https://github.com/user-attachments/assets/7cba3091-c6e9-4534-94eb-2d4c2190e7c6

If there is no active window the new window goes to the center of the screen.

Considered some alternative approaches, because IMO this enum is a little bloated after this change:

* new option in window-rule like `default-floating-position-relative-to-active-window` that would accept the same `relative-to` options
* hook to an external program that would decide coordinates

but just adding these to the enum seems more straightforward and understandable.


contagnas:
somewhat related: https://github.com/niri-wm/niri/issues/978

PR #3481, opened 2026-02-20
Feat: layer-animations

Atan-D-RP4:
Implements layer animations.
There is surprisingly no open issue for this. Just a couple discussions.

EDIT: UPDATED DEMO
https://github.com/user-attachments/assets/01c7f300-fbfe-4c24-ab27-e03667e8195a

TODOs:
- [x] Add matchers for existings heuristics 
     - layer - top, bottom, overlay, backgroud (6bcaaf9d on main)
     - [x] anchor - top, bottom, left, right (#3862)
     - [x] has_exclusive_zone (#3862)
     - [x] Maybe some others (#3862)
- [x] animations block in layer-rules (with proper dedup for shaders)

~~So layer-*-open/close use a heuristic based on layer anchors and stuff to set the animation configs for them.~~
~~If we just expose some of them as matchers in layer-rules, we can get the exact same behavior as it currently has through layer-rules instead of having it be hardcoded.~~

So we have this now:
``` kdl
layer-rule {
   match namespace="launcher"
   match anchor-sides=3
   match anchor="top,left" // both top AND left
   match anchor="right" // OR right
   match exclusive-zone="exclusive" // or "neutral" or "dontcare"
   match keyboard-interactivity="on-demand" //  or "none" or "exclusive"
   animations {
      layer-open {}
      layer-close {}   
   }
}
```

EDIT:
This now properly depends on #3862 and #4098, and is stacked on top of them. I learned a bunch of new git-fu getting that right!


src/backend/tty.rs
```diff
@@ -1790,6 +1796,21 @@ impl Tty {
...
+                .next()
+                .map(|surface| (render_node, surface.compositor.format()))
+        }) {
+            let mut renderer = self
+                .gpu_manager
+                .renderer(&self.primary_render_node, &render_node, format)
+                .ok()?;
+            return Some(f(renderer.as_gles_renderer()));
+        }
+
```
YaLTeR: What is this change?

src/handlers/layer_shell.rs
```diff
@@ -115,11 +147,15 @@ impl State {
...
             let was_unmapped = self.niri.unmapped_layer_surfaces.remove(surface);
+            let config = self.niri.config.borrow();
 
-            // Resolve rules for newly mapped layer surfaces.
-            if was_unmapped {
-                let config = self.niri.config.borrow();
-
+            if let Some((_, mut mapped)) = self.niri.closing_layers.remove(layer) {
+                mapped.clear_close_animation();
+                mapped.start_open_animation(&config.animations);
```
YaLTeR: I think it would make more sense for it to work like closing windows work where they are not tied in any way to a Mapped. It would remove this sort of juggling.

src/layer/closing_layer.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::niri_render_elements;
+use crate::render_helpers::primary_gpu_texture::PrimaryGpuTextureRenderElement;
+use crate::render_helpers::shader_element::ShaderRenderElement;
+use crate::render_helpers::shaders::{mat3_uniform, ProgramType, Shaders};
+use crate::render_helpers::snapshot::RenderSnapshot;
+use crate::render_helpers::texture::{TextureBuffer, TextureRenderElement};
+use crate::render_helpers::{render_to_encompassing_texture, RenderTarget};
+
+#[derive(Debug)]
+pub struct ClosingLayer {
```
YaLTeR, line 22: This is pretty much exactly the same as ClosingWindow, maybe they could be somehow combined together?

src/layer/mapped.rs
```diff
@@ -39,13 +43,37 @@ pub struct MappedLayer {
...
+    /// Whether the open animation has been started at least once.
+    open_animation_started: bool,
+
+    last_geometry: Option<Rectangle<i32, Logical>>,
+
+    /// Closing layer animation.
+    pub closing_layer: Option<ClosingLayer>,
+
+    /// Cached render elements for close animation (captured on last commit before close).
+    cached_close_elements: Option<CachedCloseElements>,
```
YaLTeR: This is called "unmap snapshot" for windows, let's keep the terminology

src/layer/mapped.rs
```diff
@@ -179,10 +340,9 @@ impl MappedLayer {
             );
             push(elem.into());
         } else {
-            // Layer surfaces don't have extra geometry like windows.
             let buf_pos = location;
-
             let surface = self.surface.wl_surface();
+
```
YaLTeR: Unrelated changes

src/layer/opening_layer.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+niri_render_elements! {
+    OpeningLayerRenderElement => {
+        Offscreen = RelocateRenderElement<RescaleRenderElement<OffscreenRenderElement>>,
+        Shader = ShaderRenderElement,
+    }
+}
+
+impl OpeningAnimation {
```
YaLTeR: Also if we can merge this with the opening animation from the Tile that would be good as they are also exactly the same

src/layer/mapped.rs
```diff
@@ -233,4 +393,78 @@ impl MappedLayer {
             );
         }
     }
+
+    /// Render with animation if an open or close animation is ongoing.
+    ///
+    /// Returns true if animation was rendered, false to indicate fallback to normal rendering
+    /// needed.
+    pub fn render_with_animation<R: NiriRenderer>(
```
YaLTeR: Why not follow the pattern from Tile where there's render() that deals with open anim and underlying render_inner()

src/handlers/layer_shell.rs
```diff
@@ -165,11 +252,73 @@ impl State {
...
+                    let mut blocked_out_contents = Vec::new();
+                    push_elements_from_surface_tree(
+                        renderer,
+                        surface,
+                        Point::from((0., 0.)).to_physical_precise_round(scale),
+                        scale,
+                        alpha,
+                        Kind::ScanoutCandidate,
+                        &mut |elem| blocked_out_contents.push(elem),
+                    );
```
YaLTeR: This stuff duplicated in 2 places likely needs to go into a separate function called store_unmap_snapshot() (see how window Mapped does it)

src/layer/mapped.rs
```diff
@@ -233,4 +393,78 @@ impl MappedLayer {
...
+        geo_size: Size<f64, Logical>,
+        push: &mut dyn FnMut(LayerSurfaceRenderElement<R>),
+    ) -> bool {
+        let alpha = self.rules.opacity.unwrap_or(1.).clamp(0., 1.);
+        let location = location + self.bob_offset();
+
+        // Blocked out layers fall back to normal rendering
+        if target.should_block_out(self.rules.block_out_from) {
+            return false;
+        }
```
YaLTeR: Why? Blocked-out tiles get rendered with anims, layers should too

src/handlers/layer_shell.rs
```diff
@@ -142,6 +178,57 @@ impl State {
...
+                        push_elements_from_surface_tree(
+                            renderer,
+                            surface,
+                            Point::from((0., 0.)).to_physical_precise_round(scale),
+                            scale,
+                            alpha,
+                            Kind::ScanoutCandidate,
+                            &mut |elem| contents.push(elem),
+                        );
+
```
YaLTeR: So the logic here is to store a snapshot of contents at first commit to use if the surface is destroyed without unmapping? That's kinda weird tbh, those contents are way out of date by then. And anyhow this shouldn't be a problem I think if you store unmap snapshot in CompositorHandler::destroyed() just like regular windows do

src/handlers/layer_shell.rs
```diff
@@ -165,11 +252,73 @@ impl State {
                 self.niri.layer_shell_on_demand_focus = Some(layer.clone());
             }
         } else {
-            // The surface is unmapped.
-            if self.niri.mapped_layer_surfaces.remove(layer).is_some() {
-                // A mapped surface got unmapped via a null commit. Now it needs to do a new
-                // initial commit again.
-                self.niri.unmapped_layer_surfaces.insert(surface.clone());
+            // The surface is unmapped. Capture elements now (while surface still has content)
+            // and add to closing_layers for animation rendering.
```
YaLTeR: And this needs to be in pre-commit hook I believe (just like normal windows unmap snapshot)

niri-config/src/animations.rs
```diff
@@ -193,6 +203,48 @@ impl Default for WindowCloseAnim {
...
+pub struct LayerOpenAnim {
+    pub anim: Animation,
+    pub custom_shader: Option<String>,
+}
+
+impl Default for LayerOpenAnim {
+    fn default() -> Self {
+        Self {
+            anim: Animation {
+                off: false,
```
YaLTeR, line 273: I'd like to have a design discussion about when and which animations should be enabled, especially by default. Definitely not all animations on all layers.

For one thing, bars need a slide-in animation while centered launchers need something closer to the default window-open. Fullscreen surfaces should probably not have any animation by default. And now new shells (DMS, etc.) do their own anims which we probably also don't want to interfere with by default.

The bar slide-in, and generally animating anything with exclusive zones, I'd rather handle separately at a later time, when I figure out how to do window transactions for these. So no anim by default if layer has exclusive zone.

Let's say we play it safe and have no layer anims by default. I think it would make sense then if the user can simply write something like `layer-open { on; }` and get an animation matching their configured window-open (and same for window-close), at least for "centered launcher" kind of layer, because those are very similar to floating windows.

As I'm writing this I start to feel that layer animations in general should be added only after window-rule animations because layer surfaces are quite different, and you likely want different anims for launchers (window-open) vs. background (fade-in) vs. bars (slide-in)...

YaLTeR (changes requested):
Haven't looked through in detail

niri-config/src/animations.rs
Atan-D-RP4: > I'd like to have a design discussion about when and which animations should be enabled, especially by default. Definitely not all animations on all layers.
> 
> For one thing, bars need a slide-in animation while centered launchers need something closer to the default window-open. Fullscreen surfaces should probably not have any animation by default. And now new shells (DMS, etc.) do their own anims which we probably also don't want to interfere with by default.
> 
> The bar slide-in, and generally animating anything with exclusive zones, I'd rather handle separately at a later time, when I figure out how to do window transactions for these. So no anim by default if layer has exclusive zone.
> 
> Let's say we play it safe and have no layer anims by default. I think it would make sense then if the user can simply write something like `layer-open { on; }` and get an animation matching their configured window-open (and same for window-close), at least for "centered launcher" kind of layer, because those are very similar to floating windows.

I've been having similar thoughts, but honestly just played it safe with the simple config options that were the same as for windows.

Changing the defaults here is fine, but again, it's best if we can configure these from layer rules.

> 
> As I'm writing this I start to feel that layer animations in general should be added only after window-rule animations because layer surfaces are quite different, and you likely want different anims for launchers (window-open) vs. background (fade-in) vs. bars (slide-in)...

You've seem to come to similar conclusion too.

src/layer/mapped.rs
Atan-D-RP4: I think I got the wrong idea from render_popups(). I also don't use screensharing+block-out-from much so I didn't notice.

niri-config/src/animations.rs
YaLTeR: While layer-rule anim customization should definitely be a thing, I wonder if we need extra configuration on top. Like (I don't like this particular version, but to give the idea) `layer-open-bar {}`, `layer-open-launcher {}`, `layer-open-wallpaper {}`. Or how could this work?

The goal is to have sensible defaults for broad existing types of layer-shell client, without forcing the user to hardcode layer-rules with namespaces.

src/backend/tty.rs
Atan-D-RP4: Debug artefact. Will cleanup.

src/layer/closing_layer.rs
Atan-D-RP4: Yeah, this and opening_layer.rs, I directly copied from window stuff. Combining them together would be more architectural change, and I didn't think it appropriate for this PR.

src/layer/mapped.rs
Atan-D-RP4: Will do

src/handlers/layer_shell.rs
Atan-D-RP4: Already done in a local commit.

src/layer/mapped.rs
Atan-D-RP4: For layers, the main render function is render_normal right? There do seem to be some differences from Tile/Window, so I was unsure. Decided to keep it separate for initial working implementation.
Will unify.

niri-config/src/animations.rs
Atan-D-RP4: I think it's best if we just do something like this:
```kdl
layer-rule {
   match namespace="launcher"
   animation {
       layer-open { }
       layer-close { }
   }
}
```

And let the user decide on which layers get animated by the compositor. We probably don't want the full 'animation {}' in there, but that's the general idea.

Window-rule animations can follow a similar pattern.

Atan-D-RP4:
Alright everything seems to work.

https://github.com/user-attachments/assets/87c67f66-afa0-41ae-b011-2f1e286efe22

HigherOrderLogic:
Looks like the SwayNC panel has a faint overlay during close animation.

Atan-D-RP4:
That's just how SwayNC does it. I tested it on Hyprland, and it's there as well.

Atan-D-RP4:
Test config for anyone looking to review this and new config design:
```kdl
animations {
    layer-open {
        duration-ms 700
        curve "cubic-bezier" 0.05 0.7 0.1 1 // CSS cubic-bezier curve

        // fade in, no movement
        custom-shader r"
            vec4 open_color(vec3 coords_geo, vec3 size_geo) {
                vec3 coords_tex = niri_geo_to_tex * coords_geo;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                color *= niri_clamped_progress;
                return color;
            }
        "
    }

    layer-close {
        // base fallback close (wallpaper, and any bucket without overrides)
        duration-ms 900
        curve "cubic-bezier" 0.05 0.7 0.1 1

        // fade out, no movement
        custom-shader r"
            vec4 close_color(vec3 coords_geo, vec3 size_geo) {
                // Geometry-space y coordinate of output bottom edge.
                float bottom_geo = (niri_input_to_geo * vec3(0.0, 1.0, 1.0)).y;
                vec3 coords = coords_geo;
                coords.y -= niri_clamped_progress * bottom_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                return color;
            }
        "
    }


    layer-launcher-open {
        // distinct long launcher open
        duration-ms 2400
        curve "cubic-bezier" 0.05 0.7 0.1 1

        // slide in from left with fade
        custom-shader r"
            vec4 open_color(vec3 coords_geo, vec3 size_geo) {
                // Geometry-space x coordinate of output left edge.
                float left_geo = (niri_input_to_geo * vec3(1.0, 0.0, 1.0)).x;
                vec3 coords = coords_geo;
                coords.x += (1.0 - niri_clamped_progress) * left_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                color *= niri_clamped_progress;
                return color;
            }
        "
    }

    layer-launcher-close {
        // distinct long launcher close
        duration-ms 2600
        curve "cubic-bezier" 0.05 0.7 0.1 1
        // slide out to right with fade
        custom-shader r"
            vec4 close_color(vec3 coords_geo, vec3 size_geo) {
                // Geometry-space x coordinate of output right edge.
                float right_geo = (niri_input_to_geo * vec3(1.0, 0.0, 1.0)).x;
                vec3 coords = coords_geo;
                coords.x -= niri_clamped_progress * right_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                color *= (1.0 - niri_clamped_progress);
                return color;
            }
        "
    }

    layer-bar-open {
        // bar open distinct but shorter than launcher
        duration-ms 1800
        curve "cubic-bezier" 0.05 0.7 0.1 1
        // slide in from left
        custom-shader r"
            vec4 open_color(vec3 coords_geo, vec3 size_geo) {
                // Geometry-space x coordinate of output right edge.
                float right_geo = (niri_input_to_geo * vec3(1.0, 0.0, 1.0)).x;
                vec3 coords = coords_geo;
                coords.x -= (1.0 - niri_clamped_progress) * right_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                return color;
            }
        "
    }

    layer-bar-close {
        duration-ms 3200
        curve "cubic-bezier" 0.05 0.7 0.1 1

        // slide out to right
        custom-shader r"
            vec4 close_color(vec3 coords_geo, vec3 size_geo) {
                float right_geo = (niri_input_to_geo * vec3(1.0, 0.0, 1.0)).x;
                vec3 coords = coords_geo;
                coords.x += niri_clamped_progress * right_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                return color;
            }
        "
    }

    layer-wallpaper-open {
        duration-ms 2000
        curve "cubic-bezier" 0.05 0.7 0.1 1

        // slide in from left
        custom-shader r"
            vec4 open_color(vec3 coords_geo, vec3 size_geo) {
                // Geometry-space x coordinate of output right edge.
                float right_geo = (niri_input_to_geo * vec3(1.0, 0.0, 1.0)).x;
                vec3 coords = coords_geo;
                coords.x -= (1.0 - niri_clamped_progress) * right_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                return color;
            }
        "
    }

    layer-wallpaper-close {
        duration-ms 2000
        curve "cubic-bezier" 0.05 0.7 0.1 1

        // slide out to right
        custom-shader r"
            vec4 close_color(vec3 coords_geo, vec3 size_geo) {
                float right_geo = (niri_input_to_geo * vec3(1.0, 0.0, 1.0)).x;
                vec3 coords = coords_geo;
                coords.x += niri_clamped_progress * right_geo;
                vec3 coords_tex = niri_geo_to_tex * coords;
                vec4 color = texture2D(niri_tex, coords_tex.st);
                if (coords_tex.x < 0.0 || coords_tex.x > 1.0 ||
                    coords_tex.y < 0.0 || coords_tex.y > 1.0) {
                    color = vec4(0.0);
                }
                return color;
            }
        "
    }
}
```

HigherOrderLogic:
I'm not sure if I like the `layer-(bar|launcher|wallpaper|whatever-(open|close)` config design. Is there any rationale behind this? 

I would prefer if it just use the usual `layer-rules` block and matchers.

Atan-D-RP4:
> I'm not sure if I like the `layer-(bar|launcher|wallpaper|whatever-(open|close)` config design. Is there any rationale behind this? 

Neither do I. 

But Yalter mentioned wanting to see if such a design, with it's backing heuristic, was possible to do relatively painlessly over on matrix.

It is not as painless as desired. We'll see which parts we can keep and which to throw out.

> I would prefer if it just use the usual `layer-rules` block and matchers.

That'll have to wait for animations block support in window-rules.

YaLTeR:
> I'm not sure if I like the `layer-(bar|launcher|wallpaper|whatever-(open|close)` config design. Is there any rationale behind this?

1. https://github.com/niri-wm/niri/pull/3481#discussion_r2836601123
2. > Our goal here is to have the sensible behavior by default (i.e. bars use bar anim, launchers use launcher anim), then make it easy to override common cases (i.e. someone wants to leave bar anims but disable launcher anims), then make it possible to override per client (i.e. current layer rules with namespace -- this should be the last resort as it's the least general way). All of this filtered through what is technically feasible without hacks (i.e. if it turns out that we can't make robust heuristics then we don't do them).

Atan-D-RP4:
> 1. https://github.com/niri-wm/niri/pull/3481#discussion_r2836601123

I already replied to that thread but it was me doing something AI suggested, to make the open anim work. 
Turns out I'd just been dropping the offscreen data instead of rendering it into a texture and using that. 
Already cleaned up.

YaLTeR:
I was just linking the rationale

Atan-D-RP4:
So I moved the `store_unmap_snapshot()` call for the closing animation into the pre-commit hook.
The problem is that using the same condition for got_unmapped results in the unmap_snapshot never getting set. So the animation is never triggered.
Doing this:
```rs
    fn start_close_animation_for_layer(
        &mut self,
        output: &Output,
        layer: &LayerSurface,
        geo: Rectangle<i32, Logical>,
        mut mapped: MappedLayer,
    ) {
         ...
         self.backend.with_primary_renderer(|renderer| {
            let snapshot = mapped.take_unmap_snapshot().or_else(|| {
                mapped.store_unmap_snapshot(renderer);
                mapped.take_unmap_snapshot()
            });

            ...
           }
        }
```
makes some clients work (rofi, swaync, vicinae are the ones I tested), but not others (fuzzel, wofi, waybar, quickshell(via dms)).

So I'm just setting `got_unmapped` to true for both `BufferAssignment::Remove` and `BufferAssignment::NewBuffer`, which basically means I'm updating the texture cache every time the layer surface is updated.
It works. Everything animates as expected. But this is Not Optimal.

Suggestions to fix this would be appreciated.

src/handlers/layer_shell.rs
Atan-D-RP4: Moved it to pre-commit hook.

src/handlers/layer_shell.rs
Atan-D-RP4: Layer surfaces seem to have bit more nuance as to when they unmapped but still have valid contents that we would animate. The last valid surface before unmap is apparently a transparent one so animating that shows no animation at all.

Had to troubleshoot a significant bit to get it working reliably for all clients, when I moved the snapshotting into the pre-commit hook without causing frequent, repeated caching/snapshotting. But it should work.

src/layer/mapped.rs
Atan-D-RP4: Done.

src/handlers/layer_shell.rs
Atan-D-RP4: This should be dealt with.

Atan-D-RP4:
Ok this is as ready as it can get for review.

Rebased onto main again. Please test and review at your leisure.

trevarj:
Need this PR to merge for my mental health 😇

Atan-D-RP4:
> Need this PR to merge for my mental health 😇 

YaLTeR is too busy to review this. And there's some pending design choices that needs to be discussed.

Will take a while. Though testing, feedback and code review helps and is always appreciated!

RanXom:
Tried this PR locally and noticed a small issue.

When `layer-open` animations are enabled, blurred layer surfaces (e.g. Vicinae) don't appear blurred until the animation finishes. Without the animation, the blur is present immediately.

Not sure if this is expected or a bug, but thought I'd report it.

The layer rule i was using:
```
layer-rule {
    match namespace="vicinae"

    animations {
          layer-open {
          duration-ms 300
          curve "ease-out-cubic"
      }

      layer-close {
          duration-ms 140
          curve "ease-out-cubic"
      }
    }
    background-effect {
      blur true
      xray true
    }
}
```

Preview:

https://github.com/user-attachments/assets/e918164b-1b15-400b-acb6-0064ec215e13

Sempyos:
That's the intended behavior, read the warning here: https://niri-wm.github.io/niri/Window-Effects.html#xray

bvr-yr:
:eyes:

Atan-D-RP4:
As can be seen in the updated demo video at the top, there's still an issue with some clients like wofi and vicinae. Have yet to try swaync.

It's probably the timing of the snapshot creation. I've tried my best to debug this, but can someone more familiar with the layer surfaces lifecycle take a look? Clearly, I'm doing something wrong.

PR #3495, opened 2026-02-23
Update environment import command in niri-session

ghost:
Since systemctl's import-environment support is now marked as deprecated and therefore gives a warning, I wanted to make such a correction so that the warning screen disappears, but since systemctl does not support import-environment directly anymore, I think it makes more sense to use only dbus or use systemctl in accordance with its new format.

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -33,7 +33,7 @@ if hash systemctl >/dev/null 2>&1; then
     systemctl --user reset-failed
 
     # Import the login manager environment.
-    systemctl --user import-environment
+    systemctl --user show-environment | grep -v '^_=' | cut -d= -f1 | xargs systemctl --user import-environment
 
     # DBus activation environment is independent from systemd. While most of
     # dbus-activated services are already using `SystemdService` directive, some


YaLTeR:
No we are not doing these hacks just to "hide a warning"

ghost:
This isn't a hack or anything; it's a technical method for passing necessary variables using xargs.

ghost:
I added this for compatibility reasons, not just as a warning.

noahpolimon:
@YaLTeR i think denying it is just waiting for it to bite you later on when it breaks something. this is no longer a supported method by systemd.

HigherOrderLogic:
But it still works, no? Also I prefer it over some hacky bash scripts chaining this PR is proposing.

ghost:
Yes, it works, and yes, it's more technical than hacky methods in my opinion, that's why I wanted to add that.Instead of unnecessary scripting clutter, you can use simpler methods like these.

YaLTeR:
i'm happy to wait until it does bite me later

PR #3502, opened 2026-02-24
Fix/quickshell crashing

cmeissl:


src/niri.rs
YaLTeR, line 1: I think in this function it doesn't matter where the code is because it's a synchronous function. We don't send or receive anything from clients within this function. So I'd rather leave the code where it was

cmeissl: Both, disable global and layersurface send_close will send events to clients. This commit does change the order. I am not convinced this is necessary, but it seems to prevent some qs ghost Windows. I have to check, but my theory is like stated in the commit message.

YaLTeR: Ah, right, they aren't *sent* but they are buffered to be sent in the ordered Wayland message buffer.

cmeissl: Yes, technically they are buffered and not directly sent here.

My theory is based on the fact that without sending the output destroy before the close the client can not tell the difference.

layer shell close defines:

> The closed event is sent by the compositor when the surface will no longer be shown. The output may have been destroyed or the user may have asked for it to be removed. Further changes to the surface will be ignored. The client should destroy the resource after receiving this event, and create a new surface if they so choose.

I can't tell if it intentionally uses the past tense for output destroy, but it would make sense.

cmeissl:
@YaLTeR I have some idea on what happens. While neither smithay nor niri intentionally send a nil for output in surface.enter the client might still see a nil value if it has destroyed the proxy client side. wayland-client tracks destroyed proxies and will replace destroyed resources with a null value when processing event arguments.
I would argue that a client should not use a destroyed output in layer shell, but we might also want to make sure to not send events for disabled globals.

cmeissl:
@YaLTeR finding a solution for preventing sending disabled global instances in event arguments will take some time.
so imo this is worth to consider in the mean time to prevent some crashes.

YaLTeR:
Yeah, sure, makes sense

src/niri.rs
YaLTeR: sway actually also sends close before removing the output, like niri before this PR. But it also sends surface leave first:
```
[ 389139.129][rs] <- wl_output@14.geometry, (0, 0, 880, 370, 0, Some("LG Electronics"), Some("LG ULTRAWIDE"), 0)
[ 389139.239][rs] <- wl_output@14.scale, (1)
[ 389139.273][rs] <- wl_surface@9.leave, (wl_output@13)
[ 389139.311][rs] <- wl_output@14.geometry, (0, 0, 880, 370, 0, Some("LG Electronics"), Some("LG ULTRAWIDE"), 0)
[ 389139.415][rs] <- wl_output@14.scale, (1)
[ 389139.438][rs] <- zwlr_layer_surface_v1@10.closed, ()
[ 389139.461][rs] <- wl_registry@2.global_remove, (54)
[ 389139.503][rs] <- zxdg_output_v1@17.logical_position, (0, 0)
[ 389139.534][rs] <- zxdg_output_v1@17.logical_size, (3840, 1600)
[ 389139.566][rs] <- wl_output@14.done, ()
[ 389139.648][rs] -> zxdg_output_v1@16.destroy()
[ 389139.684][rs] -> wl_output@13.release()
```

YaLTeR:
The `Output::from_resource()` check was a good call. In d09fa270 I fixed all places I could find because you could crash niri with that rather easily.

Issue #1818, opened 2025-06-16
Snap floating windows to monitor edges when dragging


YaLTeR:
Or to be more precise to the working area. Need to check the logic in Mutter and do the same.

PR #3506, opened 2026-02-25
feat: added resistance to floating windows

Fexaop:
for issue #1818
- threshold: 32px

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -940,7 +940,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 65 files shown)


Ind-E:
I tested this and it works well, although personally I'd prefer a higher threshold (I'm assuming it will be configurable).

Currently, resistance applies to working area edges (defined by monitor edges and/or exclusive zones), which means that monitor edges behind an exclusive zone don't have resistance applied. I'm not sure if this is desired, but it's something to think about.

Fexaop:
Should I add a config option for threshold?

YaLTeR:
No config yet please

YaLTeR:
> Currently, resistance applies to working area edges (defined by monitor edges and/or exclusive zones), which means that monitor edges behind an exclusive zone don't have resistance applied. I'm not sure if this is desired, but it's something to think about.

Yes, this was my original idea for this indeed

PR #3508, opened 2026-02-25
Add open-consume-into-column window rule

y0usaf:
Adds a new initial-configure window rule `open-consume-into-column` that automatically merges a newly opened window into an existing column containing another window with the same rule, rather than creating a new column.

Motivated by longstanding requests:

- #1419: *"opening a new app/window scrolls my entire screen which I find very disruptive. I normally immediately `consume-or-expel-window-left` the spawned window, but this feels tedious to do every single time."* — rizo
- #410: *"all of these windows appear next to the main window in their own column, sometimes pushing the main window partly off-screen... Essentially performing an automatic `consume-window-into-column` when the child window is opened."* — Nairou
- #1125: *"I always want my many browser windows when launched in the same column. Also solve the 'open-new-window' problem."* — clotodex

```kdl
window-rule {
    match app-id="^foot$"
    default-column-display "tabbed"
    open-consume-into-column "active"
}
```

Two strategies:

- `"active"` — prefer the column to the left of the new window (the previously active column in the common case), falling back to the first match
- `"first"` — always consume into the first (leftmost) matching column

Only applies to tiled windows. Floating windows are unaffected.


y0usaf:
@YaLTeR I'm making use of this PR branch today but it's an opinionated implementation and likely needs more to get through the door. An example of that is that today if I have a layout of 3 columns [{term} {browser} {discord}] and I am focused on [{browser} {discord}] and open {term}, it appropriately stacks but it also focuses back onto that column, which isn't nice UX-wise. There's also the animation piece to consider. 

I made this PR as I believe that it's a useful feature but there are behaviours like the one described that I'd need guidance on from yourself or other maintainers so that I can make this satisfactory for something as well polished as Niri :saluting_face: 

Please advise or let me know which channels are best to discuss on :bow: 

ETA: There is also much to be done and I'm happy to do it, I just want to be sure this is a concept with the potential to be accepted once it's refined.

Niidhogg:
No updates on this since Mai 13 ?

y0usaf:
I'm happy to make updates, but I can't poke around in the dark without receiving any feedback. As I said I make use of this fork today, however the behaviour isn't perfect or fully aligned with [Niri's design principles](https://github.com/niri-wm/niri/wiki/Development:-Design-Principles). In particular, this one:

> ## The focused window should not move around on its own.
>
> In particular: windows opening, closing, and resizing to the left of the focused window should not cause it to visually move.
>
>The focused window is the window you're working in. And stuff happening outside the view shouldn't mess with what you're focused on.

Currently this PR changes that focus, though I do follow focus with my mouse and haven't tested without it. 

So,
> No updates on this since Mai 13 ?

I'd need some engagement and discussion as it's a non-trivial feature to implement.

alexaandru:
https://github.com/user-attachments/assets/329d446c-de47-40eb-8d56-44144f1f6836

Seems to be working as intended, love it!

As for the "not respecting that principle" part... umm, does any way of opening a window respect it? Unless you have some rules that open any new windows in the background or something, any opening of a new window WILL change focus. How is changing focus to a tab in the same window any more disrupting than changing focus to a new column?

Anyway, thank you for the great work!!! :-) Really loving this change!

mikenrafter:
FWIW, you can achieve a similar thing by using [Piri Window Swallowing](https://github.com/Asthestarsfalll/piri/blob/main/README.en.md#swallow) w/o a fork. It is similarly unpolished since it has to work via IPC, but gets the job done in many cases.

y0usaf:
Thanks for recording a video and for your kind words @alexaandru, much appreciated :)

> As for the "not respecting that principle" part... umm, does any way of opening a window respect it? Unless you have some rules that open any new windows in the background or something, any opening of a new window WILL change focus. How is changing focus to a tab in the same window any more disrupting than changing focus to a new column?

Opening things in the background so as to not disrupt an active window is what I would likely propose *if* this PR introduces a behaviour that's considered unsatisfactory by Niri's maintainers. I totally understand what you're saying though and I do find the principle to be difficult to interperet.

There's also potential for configuration options for such behaviours, but again I'd need Niri's powers that be to weigh in.

y0usaf:
Update:

This PR has been open for a while, in that time I've decided to make my own Wayland compositor and don't feel a need to keep this PR open. If anybody would like to reproduce this PR please feel free, you have my blessing. Thanks everyone for using my branch, I'll leave it up until the end of July :)

ArtemChandragupta:
I think that it is essential to add this type of functionality to niri. Workarounds like Piri distract window layout a lot

PR #3531, opened 2026-02-27
Feat/transition anim

Atan-D-RP4:
Intending to implement full animations for the `do-screen-transition` action with custom-shader support.
Relevant discussion: #1620

Also, passes in the cursor position as uniform to the shader unlike the window-open/close/resize shaders.

The use-case is that I wanted more effects for workspace switch while overview is not enabled.

Extending the existing switch animation does not make sense design-wise and this action existed so this felt like a pretty natural workaround.

https://github.com/user-attachments/assets/56020aaf-f52c-41b9-b141-becc02482e0b

### DIFF
--- a/docs/wiki/Configuration:-Animations.md
+++ b/docs/wiki/Configuration:-Animations.md
@@ -441,6 +441,7 @@ animations {
 ```
 
 #### `screen-transition`
+
 <sup>Since: next release</sup>
 
 The cross-fade animation of the `do-screen-transition` action.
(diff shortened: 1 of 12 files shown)


Atan-D-RP4:
It kinda works.

https://github.com/user-attachments/assets/b670f407-ab9e-4461-b194-2e0110eaa72c

Atan-D-RP4:
Rebased to main.
I'm pretty happy with how this looks, feels and works right now.

PR #3538, opened 2026-02-28
Fix/floating repeated size request

higorprado:
## Summary
- Fix a floating `request_size_once()` case where a repeated same-size request could be dropped after a stale commit.
- Skip redundant floating configure requests when the requested concrete size already matches the committed window size.
- Preserve interactive-move behavior (no regression).

## Behavior Changes
- `tests::floating::repeated_size_request`:
  - before: last same-size request after commit could produce no configure
  - after: configure is re-sent correctly
- `tests::floating::resize_to_same_size`:
  - before: redundant configure could be sent even when size already matched
  - after: no configure is sent

## Validation
- Floating test module passes, including:
  - `repeated_size_request`
  - `resize_to_same_size`
  - `resize_to_different_then_same`
  - `resize_during_interactive_move_propagates_to_floating`

### DIFF
--- a/src/layout/floating.rs
+++ b/src/layout/floating.rs
@@ -783,6 +783,18 @@ impl<W: LayoutElement> FloatingSpace<W> {
         let win_height = ensure_min_max_size(win_height, min_size.h, max_size.h);
 
         let win_size = Size::from((win_width, win_height));
+        if win_size.w > 0
+            && win_size.h > 0
+            && win.size() == win_size
+            && win.sizing_mode().is_normal()
+            && win.pending_sizing_mode().is_normal()
+            && matches!(
+                win.configure_intent(),
+                ConfigureIntent::NotNeeded | ConfigureIntent::Throttled
+            )
+        {
+            return;
+        }
         win.request_size_once(win_size, animate);
     }
 
@@ -830,6 +842,18 @@ impl<W: LayoutElement> FloatingSpace<W> {
         let win_width = ensure_min_max_size(win_width, min_size.w, max_size.w);
 
         let win_size = Size::from((win_width, win_height));
+        if win_size.w > 0
+            && win_size.h > 0
+            && win.size() == win_size
+            && win.sizing_mode().is_normal()
+            && win.pending_sizing_mode().is_normal()
+            && matches!(
+                win.configure_intent(),
+                ConfigureIntent::NotNeeded | ConfigureIntent::Throttled
+            )
+        {
+            return;
+        }
         win.request_size_once(win_size, animate);
     }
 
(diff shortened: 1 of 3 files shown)


src/layout/floating.rs
```diff
@@ -783,6 +783,18 @@ impl<W: LayoutElement> FloatingSpace<W> {
         let win_height = ensure_min_max_size(win_height, min_size.h, max_size.h);
 
         let win_size = Size::from((win_width, win_height));
+        if win_size.w > 0
```
ChrisPWill: IMHO because of how many conditions are in this block, a comment would go a long way to aiding readability e.g.
```rust
// Skip redundant request when window already matches size
if win_size.w > 0
// ...
```

In general, as a newcomer to this codebase, it's not clear to me why we would need to skip it. Whether or not it's worth a mention in the comment I would leave to you, it may be obvious to maintainers.

src/layout/floating.rs
```diff
@@ -830,6 +842,18 @@ impl<W: LayoutElement> FloatingSpace<W> {
         let win_width = ensure_min_max_size(win_width, min_size.w, max_size.w);
 
         let win_size = Size::from((win_width, win_height));
+        if win_size.w > 0
```
ChrisPWill: As above, could do with a comment e.g.
`// Skip redundant request when window already matches size`

src/layout/floating.rs
higorprado: Thank you for your review. I refactored that predicate into should_skip_redundant_size_request() and added a comment explaining the intent. The skip is only for the case where the floating window already has the requested concrete size and there isn’t another pending state change that still needs a configure; that avoids redundant same-size configures, but still allows the request to be re-sent after a stale commit.

ChrisPWill (approved):
LGTM, thanks! The intent is much clearer to me now

Sempyos:
Could you attach what was broken, and how to test it was fixed..?

higorprado:
Hello! This PR fixes two existing `FIXME`s in `src/tests/floating.rs`, so the broken behavior is the one those tests were documenting before.

1. `resize_to_same_size`
- previous `FIXME`:
  - “currently it will request the size anyway…”
- broken behavior:
  - the window was already `200x200`, but niri still sent another configure for `200x200`
- before:
  - `size: 200 × 200, bounds: 1920 × 1080, states: [Activated]`
- after:
  - `""` (no configure sent)

2. `repeated_size_request`
- previous `FIXME`:
  - “doesn't request that currently”
- broken behavior:
  - after a stale same-size commit, asking for `200x100` again could do nothing
- before:
  - `""` (no configure sent)
- after:
  - `size: 200 × 100, bounds: 1920 × 1080, states: [Activated]`

This is not really a visual rendering bug; it is a floating-window configure/state bug. The most accurate evidence is the before/after configure output captured by the tests.

How I tested it:
- `cargo test resize_to_same_size --package niri --lib`
- `cargo test repeated_size_request --package niri --lib`
- `cargo test resize_during_interactive_move_propagates_to_floating --package niri --lib`
- `cargo test tests::floating --package niri --lib`

Issue #3274, opened 2026-01-17
Virtual keyboard can get keys stuck


YaLTeR:
Example: run wayvnc, then connect to it, then through VNC press Ctrl+C to terminate the wayvnc command. Then, Ctrl will get stuck, and as far as I can tell, the only way to unstick it is to connect through VNC again.

cc @Bot-wxt1221, probably need to release all virtual keyboard keys when the virtual keyboard Wayland object is destroyed?

PR #3543, opened 2026-03-02
text-input-v1: track/render popups

WantenMN:
## Summary
This PR adds end-to-end handling for `input-method-v1` popups (IME candidate/panel surfaces).

### What changed
- Added `InputMethodV1ManagerState` initialization and storage in compositor state.
- Implemented `InputMethodV1Handler` in `State`:
  - track popup lifecycle (`new_popup`, `popup_repositioned`, `dismiss_popup`),
  - update popup scale/transform from the target output,
  - compute parent geometry through window/layout context.
- Added rendering of tracked `input-method-v1` popup surfaces in the output render pass.
- Added commit-time redraw handling for `input-method-v1` popup commits.

## Why
With the current Smithay stack, niri needs explicit `input-method-v1` popup management to correctly place and render IME candidate surfaces and to ensure redraws happen when those surfaces commit.

## Merge dependency
This PR depends on Smithay changes and should only be merged after:
- https://github.com/Smithay/smithay/pull/1950


Atan-D-RP4:
Does this resolve #3274?

Sempyos:
I would note.
https://github.com/Smithay/smithay/issues/1937

Issue #3557, opened 2026-03-04
xkb.file keymap lost in reset_keymap path and after transient set_xkb_file failure


mgabor3141:
## Bug Description

There are two code paths where a custom `xkb.file` keymap can be silently replaced with a default rules-based keymap.

### Bug 1: `reset_keymap` path drops `xkb.file`

In `src/input/mod.rs`, the `reset_keymap` code path (triggered when a physical keyboard event follows a Wayland virtual keyboard event) calls `xkb_config.to_xkb_config()`, which does not include the `file` field — it only forwards `rules`, `model`, `layout`, `variant`, and `options`.

When a user's config only has `xkb.file` set (with all other fields at defaults), `to_xkb_config()` produces a default `XkbConfig`, which compiles to a plain `us` keymap — losing all custom symbols and overrides from the file.

```rust
// src/input/mod.rs — reset_keymap path
if xkb_config != Xkb::default() {
    self.set_xkb_config(xkb_config.to_xkb_config());  // ← drops file field
}
```

The guard `xkb_config != Xkb::default()` correctly detects that the user has config (because `file: Some(...)` differs from default `None`), but `to_xkb_config()` discards the file.

**Trigger:** Any Wayland client using the `zwp_virtual_keyboard_v1` protocol (input methods like fcitx/ibus, on-screen keyboards, wtype, etc.) sending key events, followed by typing on a physical keyboard.

### Bug 2: Config reload fallback permanently poisons keymap

In `src/niri.rs`, the config reload path uses `.take()` on `xkb.file`, then falls back to `set_xkb_config()` if `set_xkb_file()` fails:

```rust
// src/niri.rs — reload_config
if let Some(xkb_file) = xkb.file.take() {   // takes file out, leaving xkb.file = None
    if let Err(err) = self.set_xkb_file(xkb_file) {
        warn!("error reloading xkb_file: {err:?}");
        // falls through to set_xkb_config!
    } else {
        set_xkb_config = false;
    }
}
if set_xkb_config {
    if xkb == Xkb::default() {             // true! file was take()'d, rest is empty
        xkb = locale1.unwrap_or_default();  // plain "us" with no options
    }
    self.set_xkb_config(xkb.to_xkb_config());  // applies broken default keymap
}
```

[...]

PR #3558, opened 2026-03-04
fix: re-apply xkb.file keymap in reset_keymap path

mgabor3141:
Fixes the `reset_keymap` path described in #3557 (Bug 1).

## Problem

When `reset_keymap` triggers (after a Wayland virtual keyboard event followed by a physical keyboard event), the keymap is recompiled via `to_xkb_config()`. However, `to_xkb_config()` does not include the `file` field — it only forwards `rules`, `model`, `layout`, `variant`, and `options`. This silently replaces file-based keymaps with default rules-based keymaps, losing all custom symbols and overrides.

## Fix

Check for `xkb.file` first in the `reset_keymap` path and call `set_xkb_file()` directly, mirroring how the config-reload logic handles file-based keymaps.

Also makes `set_xkb_file()` `pub(crate)` so it can be called from `input/mod.rs`.

## Changes

- `src/input/mod.rs`: In the `reset_keymap` block, prioritize `xkb.file` over `to_xkb_config()`
- `src/niri.rs`: Change `set_xkb_file` visibility from private to `pub(crate)`

## Note

Issue #3557 also describes a second bug (config reload fallback permanently poisoning the keymap when `set_xkb_file` fails transiently). That is not addressed by this PR.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -381,7 +381,13 @@ impl State {
                 let xkb_config = config.input.keyboard.xkb.clone();
                 std::mem::drop(config);
 
-                if xkb_config != Xkb::default() {
+                if let Some(xkb_file) = xkb_config.file.as_ref() {
+                    // File-based keymap: re-apply from file (to_xkb_config()
+                    // drops the file field, so we must handle it separately).
+                    if let Err(err) = self.set_xkb_file(xkb_file.clone()) {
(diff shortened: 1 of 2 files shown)


YaLTeR:
We're deleting this code soon

YaLTeR:
This code has been removed for the time being in f4dc10e0b415bee33a6f482cd5a298a25b184605

PR #3565, opened 2026-03-06
feat: auto-tile

Dreaming-Codes:
Hyprland recently got scrolling layout so niri should get auto tiling

### DIFF
--- a/niri-config/src/layout.rs
+++ b/niri-config/src/layout.rs
@@ -24,6 +24,7 @@ pub struct Layout {
     pub gaps: f64,
     pub struts: Struts,
     pub background_color: Color,
+    pub auto_tile: Option<AutoTile>,
 }
 
 impl Default for Layout {
@@ -52,6 +53,7 @@ impl Default for Layout {
                 PresetSize::Proportion(2. / 3.),
             ],
             background_color: DEFAULT_BACKGROUND_COLOR,
+            auto_tile: None,
         }
     }
 }
@@ -91,6 +93,10 @@ impl MergeWith<LayoutPart> for Layout {
         if self.preset_window_heights.is_empty() {
             self.preset_window_heights = Layout::default().preset_window_heights;
         }
+
+        if let Some(x) = &part.auto_tile {
+            self.auto_tile = Some(x.clone());
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


HigherOrderLogic:
This goes against Niri's [design principle](https://niri-wm.github.io/niri/Development%3A-Design-Principles.html#opening-a-new-window-should-not-affect-the-sizes-of-any-existing-windows).

And Hyprland shouldnt have any effect on what Niri should does and what features it should have.

bluelinden:
> This goes against Niri's [design principle](https://niri-wm.github.io/niri/Development%3A-Design-Principles.html#opening-a-new-window-should-not-affect-the-sizes-of-any-existing-windows).
> 
> And Hyprland shouldnt have any effect on what Niri should does and what features it should have.

correct, while the implementation for this looks largely fine (and is less code than i was expecting) it is better suited for a downstream fork. the specific principle this breaks is the idea that a window's horizontal size will never affect another window's horizontal size.

we really do appreciate the effort though! thanks for the contribution, even if it doesn't fit with the project's goals.

y0usaf:
Is there a world where a feature like this could be welcome if it weren't a default, or are the design principles strict regarding configurable implementations too?

witchlliee:
> Is there a world where a feature like this could be welcome if it weren't a default, or are the design principles strict regarding configurable implementations too?

Maybe a script like auto-tiling script for Sway would be welcome. But i don't see this happening on the main project.

Atan-D-RP4:
It is already possible to do this via script. The niri-tweaks repo listed in awesome-niri has a script called tile_to_n that does exactly this via the IPC.

gabbieblue:
just wanted to add that even if it does break design principles for this project, i would love to see a downstream fork that adds floating and auto-tiling managed similarly to how Cosmic WM does it, where each workspace can have a different tiling mode. sorry if this is not appropriate to add here

Atan-D-RP4:
I think a tiling implementation like this would be a better fit for Niri's design principles:


Instead of tiling all of the windows in a _workspace_, you tile all of the windows in the _column_. So a master layout tiling or what have you will only be applied when windows are consumed into or expelled from a column. You can have all of the windows in the column be tiled and expand it to max-width for the full traditional tiling experience.

Niri's design philosophy is that opening a new window should not resize any other windows. By default, a window is opened into a new column. This aligns with that.

It's trivial to now script this such that a new window is consumed into the current column to get the full tiling effect. This way, we stick to the design while allowing users a pretty decent way to get Hyprland-like Tiling and even mix it seamlessly with Niri's scrolling.

We can even have new IPC actions such as spawn-in-column to facilitate this. Though that might be going against the design just a bit.

PR #3571, opened 2026-03-07
[Switch presets] add (no-)wraparound & make cycle-back an OPTION

two-horned:
Improves #1670 and adds a no-wraparound option which desire was mentioned at couple of places,
like in #1000 and #1367.

I didn't see the need to add tests, so if someone wants to contribute in that regard, please do so.

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -303,12 +303,12 @@ pub enum Action {
     #[knuffel(skip)]
     ResetWindowHeightById(u64),
     SwitchPresetColumnWidth(
-        #[knuffel(property(name = "forwards"))] bool,
-        #[knuffel(property(name = "wrap"))] bool,
+        #[knuffel(property(name = "forwards"), default = true)] bool,
+        #[knuffel(property(name = "wrap"), default = true)] bool,
     ),
     SwitchPresetWindowWidth(
-        #[knuffel(property(name = "forwards"))] bool,
-        #[knuffel(property(name = "wrap"))] bool,
+        #[knuffel(property(name = "forwards"), default = true)] bool,
+        #[knuffel(property(name = "wrap"), default = true)] bool,
     ),
     #[knuffel(skip)]
     SwitchPresetWindowWidthById {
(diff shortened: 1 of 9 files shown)


two-horned:
Important note: This will break configs using old `switch-preset-[column/window]-width-back` or `switch-preset-window-height-back` commands. You are now supposed to use optional arguments instead.

Quick examples:
```
bind {
  Mod+R { switch-preset-column-width; } // defaults are forwards=true, wrap=true
  Mod+R { switch-preset-column-width wrap=true; }
  Mod+R { switch-preset-column-width forwards=false; }
  Mod+R { switch-preset-column-width forwards=false wrap=false; }
}
```

HigherOrderLogic:
> This will break configs using old

Sounds like [a big no no](https://niri-wm.github.io/niri/Configuration%3A-Introduction.html#breaking-change-policy) for this to get merged.

fred21O4:
I would love to have the ability to disable wraparound, however I think it best to keep it as is in regards to cycle back
Leave both switch-preset-column-width and switch-preset-column-width-back as commands and add wraparound as an option to both
This way it keeps backwards compatibility while still providing to the same use cases.
I also think doing it this way aligns more with how other commands are done, as most seem to have a -back variant rather than a option.

two-horned:
> I would love to have the ability to disable wraparound, however I think it best to keep it as is in regards to cycle back Leave both switch-preset-column-width and switch-preset-column-width-back as commands and add wraparound as an option to both This way it keeps backwards compatibility while still providing to the same use cases. I also think doing it this way aligns more with how other commands are done, as most seem to have a -back variant rather than a option.

I am not aware of any other command that uses the `back` keyword (this pr affects all preset-choosing commands).
I can see, how users don't want their configs broken, so a compromise would be to make the config options and ipc commands an alias, i.e. `switch-preset-X-Y-back` becomes `switch-preset-X-Y forwards=false` internally.

On the other hand, this only affects very little people and most that use the `back` feature already want to disable wrapping, so they'd *want* to change their configs. So what's the point?

Either way this PR hasn't seen much attention to it, so I will have to wait till I get more feedback in this regard.

mattkang:
@two-horned appreciate the work and definitely need this feature. I tend to agree with @fred21O4 and think their suggestion is the most obvious path forward. Given the Breaking Change Policy, this may just have to be "lesson learned" for the future. Could have probably foreseen this in #1670, but it's merged now so gotta just live with it until that policy changes.

PR #3572, opened 2026-03-08
import-environment without variable is deprecated

freezboltz:


Sempyos:
There is many PRs and discussions about this, see: https://github.com/niri-wm/niri/pull/3495

LuckShiba:
Besides, importing only XDG_SESSION_TYPE is surely *not* what niri wants/needs for that, this would not be the solution

LuckShiba:
https://github.com/niri-wm/niri/pull/3495#issuecomment-3946184462

rebizzz:
fwiw this is the only open one that doesn't actually change behaviour. it still imports everything, just spells the names out so systemd stops whining. that's what YaLTeR asked for in #254:

> The whole point is to import *all* variables that may be set by the user's shell configs

> we're not gonna replace a 1 line command with something complicated just to hide this warning message

the allowlist ones (#3776, the closed #4025) drop stuff like PATH/LANG that comes in via /etc/profile.d, which is the same thing that bit GNOME and they had to revert. this doesn't touch that.

only thing to watch out for: `printenv | cut -d'=' -f1` spits out a junk token if some var's value has a newline in it (the wrapped lines have no `=`). systemd just ignores names it doesn't know so it's harmless, and the dinit branch right below already assumes the same thing ("Might not work correctly for multiline variable names, but it is reasonable to assume there are none").

anyway there's like 5 PRs for this now and one already got closed as a dupe, might as well just land this one.

real-or-random:
If the goal is just to hide the warning, then piping through `grep --invert-match "import-environment without variable is deprecated"` is much simpler and more natural (maybe add `--color=never`).

edit: I'm using this on my machine now:
```
SYSTEMD_COLORS=true systemctl --user import-environment | grep -v 'Calling import-environment without a list of variable names is deprecated.'
```
The `SYSTEMD_COLORS=true` is purely cosmetical. It makes sure that systemctl would still print other error messages in the right color. Not sure if this is appropriate for each user. If not, dropping it is entirely harmless.

h8d13:
Hi new to niri and kind of have OCD with logs aha (so I guess I end up here).

If the stance is to keep the deprecated warning until it breaks (+ keep code simpler), I'd silence the warning like @real-or-random recommended.

PR #3577, opened 2026-03-08
feat: focus ring fade animation and gradient rotation

oblivion8282-1337:
## Summary

- **Fade animation**: Focus ring smoothly fades in/out on focus changes using an EaseOutCubic curve, instead of appearing/disappearing instantly.
- **Gradient rotation**: The focus ring gradient continuously rotates while a window is focused, adding a subtle dynamic effect.

Both features are fully opt-in via the `focus-ring {}` config block:

```kdl
focus-ring {
    width 2
    active-gradient from="#7aa2f7" to="#bb9af7" angle=45
    fade-duration-ms 1000
    gradient-spin-speed 90
}
```

- `fade-duration-ms` — duration of the fade in/out (0 or omitted = off)
- `gradient-spin-speed` — rotation speed in degrees/second (0 or omitted = off)

No behavior changes when these options are not set.

## Changed files

- `niri-config/src/appearance.rs` — new config fields
- `src/layout/focus_ring.rs` — gradient angle offset parameter
- `src/layout/tile.rs` — fade animation + gradient rotation logic
- `src/layout/insert_hint_element.rs` — pass through new fields
- `src/ui/mru.rs` — pass through new fields

## Test plan

- [x] Tested fade with various durations (200ms–1500ms), 500–1000ms feels best
- [x] Tested gradient rotation at different speeds (45–180 deg/s)
- [x] Verified no effect when config options are omitted
- [x] Verified no performance impact in normal use

### DIFF
--- a/niri-config/src/appearance.rs
+++ b/niri-config/src/appearance.rs
@@ -269,6 +269,8 @@ pub struct Border {
     pub active_gradient: Option<Gradient>,
     pub inactive_gradient: Option<Gradient>,
     pub urgent_gradient: Option<Gradient>,
+    pub fade_duration_ms: u64,
+    pub gradient_spin_speed: f64,
 }
 
 impl Default for Border {
@@ -282,6 +284,8 @@ impl Default for Border {
             active_gradient: None,
             inactive_gradient: None,
             urgent_gradient: None,
+            fade_duration_ms: 0,
+            gradient_spin_speed: 0.,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


Anderson0xFF:
This is amazing! I was looking for something similar for Niri, since it currently doesn't support animated borders.

Ozogorgor:
I haven't been using Niri for long (~3 weeks), but from what I can tell decoration rendering (borders, focus rings, etc.) currently isn't part of the shader pipeline.

If decoration rendering could be exposed to shaders, it might enable shader-based effects or animations for borders and focus rings. A shader implementation would give me more control over the type of animation  I want of my decorations. I also might be completely wrong with my assumptions, but these are my two cents.

rivera-diego:
Niri is really missing a fade effect for window borders. Coming from Hyprland, the static popping in and out of borders is very noticeable. Having a smooth transition for the focus ring would make a huge difference.

oblivion8282-1337:
Rebased onto current main, conflicts resolved.

rivera-diego:
Is it possible to implement this with the wip branch?

Sempyos:
Wip branch is how merged into main

PR #3587, opened 2026-03-10
feat: GlobalShortcut portal

RottenFishbone:
**Edited for more clarity, see original in spoiler text at the bottom**

### What this does

Implements [GlobalShortcuts](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html) via [xdg-desktop-portal-gnome](https://github.com/GNOME/xdg-desktop-portal-gnome), as per discussion on #1165 and limited discussion #2775.

### What this adds

- New config section: `global-shortcuts`, a root node similar to `binds`
- Two new dbus interfaces: [`org.gnome.Settings.GlobalShortcutsProvider`](https://github.com/GNOME/xdg-desktop-portal-gnome/blob/80d53cc702e73d91441d74a303b1247b4f27d54e/data/org.gnome.GlobalShortcutsProvider.xml) and [`org.gnome.Shell`](https://github.com/GNOME/xdg-desktop-portal-gnome/blob/80d53cc702e73d91441d74a303b1247b4f27d54e/data/org.gnome.Shell.xml), emulating features that are needed by `xdg-desktop-portal-gnome`.

Config is done using `global-shortcuts` and matching on shortcut-id and app-id, similar to window rules.
e.g.
```kdl
global-shortcuts {
  Ctrl+Shift+M {
    app-id { exact "org.chromium.Chromium"; }
    shortcut-id { match r#"7810A[0-9A-F]+\-Ctrl\+Shift\+M"#; }
  }
```
`app-id` and `shortcut-id` both have exact matching and regex matching (`match`, seen above). By default they are set to `NeverMatch`, which does as described. This isn't documented and caused issues for @andriyDev later in this thread. With that in mind, I do believe it is the correct approach, as a wildcard by default would end up potentially binding many shortcuts to the same key. Perhaps an omitted selector could simply throw a config error.  

#### `shortcut-id`

The biggest issue to usability I've experienced is finding the shortcut-id. How I've implemented the config is be declarative, like the rest of niri. 

This does add friction when binding shortcuts, especially since there isn't a trivial way to see what/how the program is naming them without just listening in on dbus comms. 
This is compounded by how no one seems to agree on naming shortcuts vs using preferred-triggers.

Perhaps a popup notifying the user that shortcut-id was attempted to be registered. Either in a display spawned by niri or over dbus notify? This may be a good case for a plugin to handle this feature?

#### Rebinding

It would appear there is no consensus on how binding works in the GlobalShortcut portal. Some portal backends expect no rebinds in the same session, just make a new session. Some allow rebinding within a session. I can't seem to find the breakdown in who provides what functionality (will link if I find). In my testing of various applications that implement GlobalShortcuts I didn't see any trying to rebind after launch.

#### Daily Driving

This implementation has worked for every application that I use, so far, but this could use some feedback from others who are trying to use it.

#### Wiki entry

I can get this squared away if this PR is decided to be the way forward.

### How it works

This is a bit roundabout, but that is simply how portals work. If niri gets a portal this would be less fragile and less involved.

Applications request to register shortcuts over `dbus` with [`org.freedesktop.portal.GlobalShortcuts`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html). 

If `xdg-desktop-portal-gnome` is configured as a provider it will trigger `handle_bind_shortcuts()`. This makes a call over dbus to `org.gnome.Settings.GlobalShortcutsProvider.BindShortcuts`. Normally this would be a read from stored settings, otherwise a popup asking the user to register a new shortcut. Niri provides this by reading the config and automatically validating.

Provided that the shortcut was permitted to be bound by `GlobalShortcutsProvider` then `settings_bind_shortcuts_done()` is called. This function calls over dbus to  `org.gnome.Shell.GrabAccelerators` (note, never the singular `GrabAccelerator`, despite it existing in the interface). Niri provides this and begins to emit AcceleratorActivated/Deactivated over `org.gnome.Shell`. `xdg-desktop-portal-gnome` reads these events and forwards them as [`GlobalShortcuts::Activated/Deactivated`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html#org-freedesktop-portal-globalshortcuts-activated) signals.

### How I have field tested this

I wrote a quick PoC program that registers shortcuts per freedesktop spec and watched dbus interfaces using `gdbus` and `dbus-monitor`. Following that I tested a few programs such as ghostty, fluxer and OBS (obs flatpak with the GS/wayland patch).

---

<details>
  <summary>Original PR content</summary>
This implements [GlobalShortcuts](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html) via xdg-desktop-portal-gnome (as per discussion on #1165 and limited discussion #2775)

Two dbus interfaces are implemented `org.gnome.Settings.GlobalShortcutsProvider` and `org.gnome.Shell`. The former is requested by the gnome portal to validate shortcut requests and return the results to gnome portal (which it uses to call `GrabAccelerators`).
The latter is the interface to grab, ungrab and emit (de)activation signals for shortcuts.

I haven't added any wiki info on the `global-shortcut` config yet, if the current setup I have looks good to everyone then I'll add the wiki page to document it. 

The limited number of programs I have found that implement GlobalShortcuts seem like they're all just doing their own thing, which makes config require some flexibility.
For example, 
  - Electron/Chromium seem to make their shortcut id a string of numbers followed by the `preferred-trigger`, leaving the `preferred-trigger` entry empty. 
  - OBS (flatpack version) uses fairly sensible shortcut id describing the action and fill the `preferred-trigger`.
  -  Ghostty puts the preferred-trigger in the `shortcut-id` as well.
  
Niri has the final say on what a binding is so the mismatch between preferred-trigger and what is returned to the application is moot. (e.g. an application can request `Ctrl+Shift+M` and get `Mod+Y` returned to it, it only truly needs the `action` id returned alongside it) 

Shortcuts can optionally disable inhibiting the keypresses. 
A single binding can match multiple shortcuts if the `match` argument is set up as such. (e.g. `app-id { match "(ghostty)|(discord)"; }` 

Currently it doesn't support mouse bindings but provided this PR is on the right track I believe I can add them fairly trivially.

Let me know thoughts and opinions and I'll try to get this shaped up.
</details>

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -1088,8 +1088,7 @@ impl Display for Trigger {
             Trigger::TouchpadScrollRight => "TouchpadScrollRight",
         };
 
-        write!(f, "{}", str)?;
-        Ok(())
+        write!(f, "{}", str)
     }
 }
 
@@ -1119,14 +1118,7 @@ impl Display for Modifiers {
             parts.push("Mod");
         }
 
-        for (i, part) in parts.iter().enumerate() {
-            if i > 0 {
-                write!(f, "+")?;
-            }
-            write!(f, "{}", part)?;
-        }
-
-        Ok(())
+        write!(f, "{}", parts.join("+"))
     }
 }
 
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -1191,4 +1191,84 @@ mod tests {
             },
         );
     }
+
+    #[test]
+    fn trigger_display() {
+        assert_eq!(Trigger::Keysym(Keysym::Return).to_string(), "Return");
+        assert_eq!(Trigger::Keysym(Keysym::space).to_string(), "space");
+        assert_eq!(Trigger::Keysym(Keysym::Escape).to_string(), "Escape");
+        assert_eq!(Trigger::MouseLeft.to_string(), "MouseLeft");
+    }
+
+    #[test]
+    fn modifiers_display() {
+        assert_eq!(Modifiers::CTRL.to_string(), "Ctrl");
+        assert_eq!(Modifiers::SHIFT.to_string(), "Shift");
+        assert_eq!(Modifiers::ALT.to_string(), "Alt");
+        assert_eq!(Modifiers::SUPER.to_string(), "Super");
+        assert_eq!(Modifiers::ISO_LEVEL3_SHIFT.to_string(), "Iso_Level3_Shift");
+        assert_eq!(Modifiers::ISO_LEVEL5_SHIFT.to_string(), "Iso_Level5_Shift");
+        assert_eq!(Modifiers::COMPOSITOR.to_string(), "Mod");
+
+        let all_mods = Modifiers::CTRL
+            | Modifiers::SHIFT
+            | Modifiers::ALT
+            | Modifiers::SUPER
+            | Modifiers::ISO_LEVEL3_SHIFT
+            | Modifiers::ISO_LEVEL5_SHIFT
+            | Modifiers::COMPOSITOR;
+
+        assert_eq!(
+            all_mods.to_string(),
+            "Ctrl+Shift+Alt+Super+Iso_Level3_Shift+Iso_Level5_Shift+Mod"
+        );
+        assert_eq!(Modifiers::empty().to_string(), "");
+    }
+
+    #[test]
+    fn key_display() {
+        // Basic
+        let key = Key {
+            trigger: Trigger::Keysym(Keysym::a),
+            modifiers: Modifiers::empty(),
+        };
+        assert_eq!(key.to_string(), "a");
+
+        // With mods
+        let key = Key {
+            trigger: Trigger::Keysym(Keysym::b),
+            modifiers: Modifiers::CTRL | Modifiers::SHIFT,
+        };
+        assert_eq!(key.to_string(), "Ctrl+Shift+b");
+
+        // Test mouse trigger
+        let key = Key {
+            trigger: Trigger::MouseLeft,
+            modifiers: Modifiers::CTRL,
+        };
+        assert_eq!(key.to_string(), "Ctrl+MouseLeft");
+    }
+
+    #[test]
+    fn key_idempotence() {
+        let test_cases = [
+            "a",
+            "Ctrl+b",
+            "Shift+Alt+c",
+            "Mod+d",
+            "Ctrl+Mod+e",
+            "MouseLeft",
+            "Ctrl+MouseRight",
+            "Alt+WheelScrollDown",
+            "Iso_Level3_Shift+f",
+        ];
+
+        for input in test_cases {
+            let key = input
+                .parse::<Key>()
+                .unwrap_or_else(|_| panic!("Failed to parse: {}", input));
+            let output = key.to_string();
+            assert_eq!(input, output, "Round trip failed for: {}", input);
+        }
+    }
 }
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -1,4 +1,5 @@
 use std::collections::HashSet;
+use std::fmt::Display;
 use std::str::FromStr;
 use std::time::Duration;
 
@@ -9,7 +10,9 @@ use niri_ipc::{
     ColumnDisplay, LayoutSwitchTarget, PositionChange, SizeChange, WorkspaceReferenceArg,
 };
 use smithay::input::keyboard::keysyms::KEY_NoSymbol;
-use smithay::input::keyboard::xkb::{keysym_from_name, KEYSYM_CASE_INSENSITIVE, KEYSYM_NO_FLAGS};
+use smithay::input::keyboard::xkb::{
+    keysym_from_name, keysym_get_name, KEYSYM_CASE_INSENSITIVE, KEYSYM_NO_FLAGS,
+};
 use smithay::input::keyboard::Keysym;
 
 use crate::recent_windows::{MruDirection, MruFilter, MruScope};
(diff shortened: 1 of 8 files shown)


dolence:
This is nice for kando menus and quick terminal. Thank you.

andriyDev:
I rebased this onto v26.04 and it seems to work! Discord is able to react to my keybinds in both Wayland-mode or XWayland mode - **after** I added a global-shortcuts section to my niri config.

However, it looks like I can't actually create a keybind in Wayland-mode. I have to create the key bind for Discord in XWayland mode and then running Wayland mode can detect it.

andriyDev:
Actually, it looks like Wayland-mode Discord only kinda works? It seems to only work when the focused application doesn't consume the event and that you're actually focusing a window. So focusing an empty monitor doesn't work. Using the hotkey in Firefox doesn't seem to work. But if I'm focusing Steam, or Barony, or Hollow Knight: Silksong, then the hotkey works. Very odd! I would hope niri would send the event to both.

daniiooo:
> I rebased this onto v26.04 and it seems to work! Discord is able to react to my keybinds in both Wayland-mode or XWayland mode - **after** I added a global-shortcuts section to my niri config.
> 
> However, it looks like I can't actually create a keybind in Wayland-mode. I have to create the key bind for Discord in XWayland mode and then running Wayland mode can detect it.

Could you test how it behaves with Legcord? Or maybe even potentially with this PR for vencord?
https://github.com/Vencord/Vesktop/pull/1238

andriyDev:
@daniiooo I tried with vencord with that PR with no luck (or at least vesktop which says has vencord preinstalled?). It doesn't detect the shortcuts at all, even when I'm focusing its window (the logs for vesktop say "Enabled Chromium features: GlobalShortcutsPortal" so I assume something is working).

I tried Legcord but couldn't figure out where I could set keybinds, so no info here.

ninelore:
You might wanna try plain discord web in the browser of you choice as well

andriyDev:
image

Discord Web doesn't do custom keybinds. Even if it did, I very much doubt it would work when the window isn't focused.

andriyDev:
I made a minimal example for using the GlobalShortcuts portal. https://github.com/andriyDev/test-global-shortcuts I tested it on Gnome and it works (with a full Gnome install `pacman -S gnome`, just installing gnome-shell wasn't enough).

To run it, you need to add a `test_app.desktop` file somewhere like `~/.local/share/applications/test_app.desktop`. Here are the minimum contents that worked for me:

```desktop
[Desktop Entry]
Name=Test App
Type=Application
```

Then just run the project with `cargo run`. On Gnome, this popped up a UI allowing me to bind my two shortcuts to whatever keys I wanted (with `/` and `*` on my numpads being the default), and pressing these keys anywhere printed logs to my terminal. On Niri, with this PR, the keys say they are bound in the example's logs, but then they never seem to trigger, even when Steam or a game like Silksong is focused (how Discord behaved previously).

Clearly this PR is insufficient. I believe my example is as minimal as possible (though admittedly I stumbled through the docs for these APIs).

I'm still new to portals, so I don't fully understand how these are linked up. From my limited understanding, my example here calls the `org.freedesktop.portal.GlobalShortcuts`, that is provided by `xdg-desktop-portal`. Its configs tell it to use `xdg-desktop-portal-gnome` through the `org.freedesktop.impl.portal.GlobalShortcuts` interface, and then I guess that talks to `/org/gnome/Shell` and `/org/gnome/Settings/GlobalShortcutsProvider` which are "emulated" in this PR. What I don't understand is the responsibilities? Is `/org/gnome/Shell` supposed to display the UI for picking shortcuts, and in this PR we are just hacking it to immediately return or something? Or does xdg-desktop-portal-gnome show the UI, and then it just tells the shell the finalized shortcuts? The PR description doesn't clarify this much to me.

RottenFishbone:
@andriyDev I appreciate you taking the time to troubleshoot.

I'm not at home until the end of the week but I'll dive into what's going on as soon as I get a chance.

Quickly, though, without me brushing back up on what's going on logically here (I haven't thought about this in quite some time). 

Gnome portal is providing the implementation of global shortcuts to dbus' desktop portal.
Gnome portal expects gnome Shell and GlobalShortcutsProvider in order for it to provide this feature. 

When an application wants to register a shortcut with desktop portal it will forward that to gnome portal (provided it's configured to). Gnome portal first asks GlobalShortcutsProvider to authorize it (on gnome this is the popup or a read from existing configured shortcuts). Depending on the outcome it tells Shell to register keypresses (GrabAccelerator) for the shortcut id. Shell is Gnome's analog to Niri so it is the interface that is actually telling gnome portal to trigger the events.

This weekend I'll write a much more detailed writeup as to what this PR is doing and try to debug what's going on.

andriyDev:
I did some more testing and figured out why this was failing so hard for me: in order for a global-shortcut to be valid, it must have **all** of `app-id` and `shortcut-id` set. If they are **not set**, they just don't work at all. The binds are filtered out in `on_shortcuts_provider_msg` by the `app-id` and then the `shortcut-id`.

My assumption was that if I don't set an app-id or shortcut-id, then **any** app-id or shortcut-id would be accepted. It seems wrong that setting nothing results in the key configuration being unusable. It's also unfortunate because for an application I haven't written, I have no idea what the shortcut IDs should be. For something like Discord, my only real option is to hack niri to get it to tell me what the shortcut IDs are and then add them to my config (and I'm assuming the ID is stable, but I think that's just a required assumption with the GlobalShortcuts API).

@RottenFishbone Thanks for the explanation, that is more helpful! That's really strange though, that implies that Gnome's UI for things like screencasting is part of the `xdg-desktop-portal-gnome`, but this API is for some reason built "directly" into Gnome. Very confusing!

I now understand the accelerator business better: `xdg-desktop-portal-gnome` is listening for `accelerator_activated` signals from Niri (through `org.gnome.Shell`), and then `xdg-desktop-portal-gnome` translates that into `org.freedesktop.portal.GlobalShortcuts::Activated` signals. Also very confusing - seems like a lot of redundancy lol. But that seems reasonable.

---

Unfortunately, I don't understand what's going on with Discord then. Printing the requests from GlobalShortcutsProvider, I'm not even getting any attempts at creating bindings when opening Discord or changing the custom keybinds in the settings. So maybe Discord isn't even using the GlobalShortcuts API at all? That would be an unfortunate end to this tale lol.

RottenFishbone:
@andriyDev @daniiooo I could not get discord to register global shortcuts at all. As far as I can tell they do not instruct Electron to do so, despite being built into it now.

Pretty sad, since that was my motivation in the first place.

I ended up writing (half completing) a program that acts as a bridge between GlobalShortcuts and xwayland-satellite that registers shortcuts and emits them into the x11 server. This behaviour is why if you are in an x11 window (e.g. steam game) you can PTT but if you're over a wayland native window you cannot. You can see it [here](https://git.sr.ht/~rottenfishbone/waybridge) if you're interested, I can write a quick README for usage if there is interest. I will complete it and release it eventually, since I do actually use it for PTT sometimes.

There is a chance there is a build of discord, that I am not using, that registers them correctly; I haven't found one. Fluxer seems to use GlobalShortcuts, so I know Electron uses them (somewhat correctly) and can interface with this PR.

andriyDev:
That's so funny, I just started implementing my own version of a bridge between GlobalShortcuts and xwayland. I'll take a look at yours instead!

PR #3593, opened 2026-03-11
fix: no longer input locks when TTY is switched before full compositor

Atan-D-RP4:
This fixes the case of starting Niri from a separate TTY (via Tmux or a TTY switch before full startup) and then switching to that Niri session causing an input lock.

I hack on Niri mostly in separate TTY and this behavior was a super annoying blocker. Resolving it required either restarting my computer (since all input was blocked) or `ssh`ing into my machine and killing that Niri session.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -441,6 +441,14 @@ impl Tty {
         }
         .map_err(|()| anyhow!("error assigning the seat to libinput"))?;
 
+        //  If the session is not active at startup (e.g. Niri was launched from a different TTY),
+        // suspend libinput now so that when ActivateSession fires, libinput.resume() performs a
+        // full re-enumeration of input devices instead of being a no-op.
+        if !session.is_active() {
+            debug!("session is not active, starting in paused state");
+            libinput.suspend();
+        }
+
         let input_backend = LibinputInputBackend::new(libinput.clone());
         event_loop
             .insert_source(input_backend, |mut event, _, state| {
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -441,11 +441,11 @@ impl Tty {
         }
         .map_err(|()| anyhow!("error assigning the seat to libinput"))?;
 
-        //  If the session is not active at startup (e.g. Niri was launched from a different TTY),
+        // If the session is not active at startup (e.g. niri was launched from a different TTY),
         // suspend libinput now so that when ActivateSession fires, libinput.resume() performs a
         // full re-enumeration of input devices instead of being a no-op.
         if !session.is_active() {
-            debug!("session is not active, starting in paused state");
+            debug!("session is not active, starting libinput in paused state");
             libinput.suspend();
         }
 


YaLTeR:
Do other compositors do something similar?

Atan-D-RP4:
I can only confirm that Hyprland has the same problem, from my own experience and testing for this.

From what I can tell, both Mutter and Kwin use systemd-logind to manage inactive VT and defer input processing until it sends active session signal to them.

YaLTeR:
> From what I can tell, both Mutter and Kwin use systemd-logind to manage inactive VT and defer input processing until it sends active session signal to them.

That's also what niri does

YaLTeR:
Maybe it doesn't do so at startup? Does mutter also call libinput suspend at startup?

Atan-D-RP4:
> Maybe it doesn't do so at startup? Does mutter also call libinput suspend at startup?

I checked and no it does not. Which might be why it also has this same issue, from my testing.
I can't even fully switch to Kwin's session before my entire screen is locked, with only a frozen TTY visible to me.

YaLTeR:
Huh. I want to be careful here then if this is a bit of an uncharted ground. Perhaps a better solution is, when the session is inactive at startup, call `on_session_event()` with session suspended, and let it handle libinput suspend and others?

Atan-D-RP4:
> I want to be careful here then if this is a bit of an uncharted ground.

Seems to be the case, though I don't know why. Do devs not use separate TTY during development? I ran into this right away when I started hacking on Niri.

Anyway, now using `event_loop.insert_idle` to insert PauseSession event.

YaLTeR:
Any reason for using an idle rather than calling it in init()?

Atan-D-RP4:
I didn't know better. I've played around with the `insert_idle` for something else, saw it was there reached for it.
Doing it in `Tty::init()` is actually much cleaner.

YaLTeR:
Thanks, verified that this works. I honestly don't remember ever hitting it before. Fairly sure that the TTY switched itself to the new one whenever I tried running niri while on the wrong TTY but now it doesn't?

ppenguin:
> This fixes the case of starting Niri from a separate TTY (via Tmux or a TTY switch before full startup) and then switching to that Niri session causing an input lock.

Just for my understanding: is this the same as what I see when I start `niri` with `regreet` (i.e. as a display manager), and then being unable to do `Ctrl-Alt-F3` for instance to switch VT? (at least this happened when I initially had a broken `niri` config for the greeter). I *am* able BTW to switch VT when in a normal `niri` session...

Atan-D-RP4:
Not really. At least I don't think so.

This is for the case when, basically, the focused VT is not the one that's starting Niri.

I hit this frequently because I've been hacking on Niri and prefer to test on a separate VT over the nested window compositor.

And because I have the patience and attention of a goldfish, I set it compile and switch over to my main session. Then I realize I was testing and the compile should be finished, switch VT and get input locked.

Only fix was restart or killing that Niri process by SSH'ing in. Very annoying.

YaLTeR:
Oh so that's how you hit it. I usually just compile in my main TTY as normal, then once it's all good I switch to another TTY and [up arrow] [enter] to start it right there

Atan-D-RP4:
> Oh so that's how you hit it. I usually just compile in my main TTY as normal, then once it's all good I switch to another TTY and [up arrow] [enter] to start it right there

Well that and sometimes I attach to a tmux session running on the 2nd TTY for debug logs too.
Also, attention span of a goldfish and zero patience. Liable to get distracted with the browser or the code again.

It was super annoying so I sat down and fixed it.

I wonder if something like this might be better to handle it in Smithay or some other upstream library.

Also, I wanted to figure out if we can get Niri to get through the full startup, since it currently does not call the spawn-*-at-startup stuff with this until you switch into the session.

YaLTeR:
> I wonder if something like this might be better to handle it in Smithay or some other upstream library.

Once Smithay gets a TTY backend, sure

> Also, I wanted to figure out if we can get Niri to get through the full startup, since it currently does not call the spawn-*-at-startup stuff with this until you switch into the session.

Huh? That shouldn't be the case

Issue #914, opened 2025-01-01
Bind/run multiple actions at once


YaLTeR:
Allow binds to run several actions sequentially, e.g.:

```
Mod+G {
    focus-column-right
    consume-or-expel-window-left
}
```

And, more importantly, add some way to run several actions as a batch through IPC and `niri msg action`. This will allow doing several actions as one atomic sequence and with no delay in-between, which is currently impossible.

Here the binds and the IPC part should be straightforward, but I'm not sure how to parse the arguments in `niri msg action`. Maybe clap offers something?

Another question is what to do if only one action in the sequence fails a check like `allow-when-locked`. Here I guess we ignore just the action that fails the check, same as what would happen if you were to run several `niri msg action` commands one after another.

YaLTeR:
Idea from @calops for CLI:

> you could also just have an option to read actions from stdin, where you can just put all of them separated with newlines

Though, I'm not sure if you can run a clap parser on stdin to parse the arguments.

calops:
You should be able to run a clap parser on any string manually. It would have to be a subset of the global parser but it sounds doable.

bbb651:
Clap [doesn't support it currently](https://github.com/clap-rs/clap/issues/2222), someone mentioned this as a workaround:

> I resorted to collecting `std::env::args()` into a `Vec<String>`, iterated over `.split("--")`, provided the zeroth argument (program name) at index 0 for each split and fed those to `MagicalArgs::from_iter`. It is bad in the sense that it abolutely requires the presence of `--` between subcommands.
> 
> In hindsight, it is a rather obvious workaround that would have saved me half a day of frustration. Retains all the nice stuff of structopt and scales until the limits of linux command lines are reached. If only clap/structopt could parse partially and yield the remainder instead of failing with a hardcoded unkown argument error...

This breaks `--` and presumably breaks `clap_complete` (which we don't have yet anyway), but otherwise seems like pretty good workaround.

calops:
This issue only seems to relate to chaining commands in the program arguments. You can still use the exact same workaround you're talking about with the contents of stdin instead of splitting on `--`. You would still need to tokenize each line but that's fairly simple. We probably don't need anything but handling quotes and then splitting on spaces (or maybe `IFS`).

bbb651:
[...]

PR #3609, opened 2026-03-14
binds: support multiple actions in a bind

skrmc:
Allow a single bind to contain multiple actions, executed in order, related to #897 and #914.

## Behavior

- Single-action binds keep their existing behavior.
- Multi-action binds run in declaration order.
- Lock-screen and screenshot-UI checks are all-or-nothing for a multi-action bind:
  - when locked, the bind is allowed if the bind itself is allowed, or if all contained actions are allowed;
  - mixed binds are rejected as a whole rather than partially executed.

## Example

```kdl
Mod+H {
    focus-tiling
    focus-column-left
}
Mod+G {
    focus-column-right
    consume-or-expel-window-left
}
Mod+T {
    move-window-to-tiling
    set-column-width "50%"
}
```

## Note

MRU and hotkey overlay handling remain centered on single-action binds; multi-action binds only participate via explicit `hotkey-overlay-title`.

Known limitation: multi-action binds are executed in order, but later actions are not guaranteed to observe a fully updated intermediate layout state. For example, `switch-preset-column-width; center-column` can still use the old displayed geometry, so it may not center as expected.

This PR does not introduce transactional semantics between actions.


skrmc:
Rebased onto 26100096.

PR #3536, opened 2026-02-27
improve config error messages for duplicate binds

Ind-E:
For config files with duplicate keybinds, an additional error is emitted that points to the previous instance of the bind. If a single bind is repeated 3+ times, only the first duplicate will get an additional error.

For example, with this config:
```kdl
binds {
    Mod+Shift+Slash { show-hotkey-overlay; }

    Mod+Shift+Slash { show-hotkey-overlay; }

    Mod+Tab { focus-workspace-previous; }

    Mod+Tab { focus-workspace-previous; }

    // ...

   Mod+Shift+Slash { show-hotkey-overlay; }
}
```

This error is shown
```go
Error:   × error loading config
  ├─▶ error parsing
  ╰─▶ error parsing KDL

Error:   × duplicate keybind first defined here
     ╭─[config.kdl:337:1]
 337 │ binds {
 338 │     Mod+Shift+Slash { show-hotkey-overlay; }
     ·     ───────┬───────
     ·            ╰── node starts here
 339 │
     ╰────
Error:   × duplicate keybind later defined here
     ╭─[config.kdl:339:1]
 339 │
 340 │     Mod+Shift+Slash { show-hotkey-overlay; }
     ·     ───────┬───────
     ·            ╰── unexpected keybind
 341 │
     ╰────
Error:   × duplicate keybind first defined here
     ╭─[config.kdl:341:1]
 341 │
 342 │     Mod+Tab { focus-workspace-previous; }
     ·     ───┬───
     ·        ╰── node starts here
 343 │
     ╰────
Error:   × duplicate keybind later defined here
     ╭─[config.kdl:343:1]
 343 │
 344 │     Mod+Tab { focus-workspace-previous; }
     ·     ───┬───
     ·        ╰── unexpected keybind
 345 │
     ╰────
Error:   × duplicate keybind later defined here
     ╭─[config.kdl:600:1]
 600 │
 601 │     Mod+Shift+Slash { show-hotkey-overlay; }
     ·     ───────┬───────
     ·            ╰── unexpected keybind
 602 │ }
     ╰────
```

PR #3621, opened 2026-03-15
Implement press and release keybinds + modifier-only binds

JustinSpedding:
This implements the config that @YaLTeR suggested in https://github.com/niri-wm/niri/pull/2456#issuecomment-3575904270. on the https://github.com/niri-wm/niri/pull/2456 MR.

I rebased @flowerysong's MR on top of the main branch, then added the changes to separate press actions from release actions in a separate commit. Not sure if this was the best way to do the git stuff since I want to make sure that the previous people who worked on this also get credit.


niri-config/src/binds.rs
```diff
@@ -886,69 +916,179 @@ where
...
-                                node,
-                                "property",
-                                "allow-when-locked can only be set on spawn binds",
-                            ));
+        let mut press_action: Option<Action> = None;
+        let mut release_action: Option<Action> = None;
+        let mut has_press_section = false;
+        let mut has_release_section = false;
+
+        for child in node.children() {
```
JustinSpedding, line 930: The config parsing logic is more complicated than I wanted it to be and has some code duplication, but I do not know a better way to maintain backwards compat. I am open to suggestions

niri-config/src/binds.rs
```diff
@@ -21,14 +21,38 @@ pub struct Binds(pub Vec<Bind>);
 #[derive(Debug, Clone, PartialEq)]
 pub struct Bind {
     pub key: Key,
-    pub action: Action,
+    pub press_action: Option<Action>,
```
JustinSpedding, line 25: Is splitting `action` into `press_action` and `release_action` too invasive of a change? `press_action` gets used by default if the new press/release syntax is not used.

bertin0:
Can't speak on code quality, but relating to git stuff, it seems that you just applied @flowerysong's PR as a patch without preserving authorship. it would be better to checkout on their branch then do a `git rebase -i main` and squash their changes, then cherry-pick your commits over that.

niri-config/src/binds.rs
```diff
@@ -779,34 +804,34 @@ where
...
+                    //
+                    // even the matching isn't consistent,
+                    // because errors can also be omitted as ctx.emit_error.
+                    // why does *that one* especially, require a DecodeError?
+                    //
+                    // anyways if you can make it format nicely, definitely do fix this
                     if seen_keys.insert(bind.key) {
                         binds.push(bind);
                     } else {
-                        // ideally, this error should point to the previous instance of this keybind
```
Atan-D-RP4: Why was this comment moved?

Atan-D-RP4:
This is amazing! Code-wise, looks pretty good.
I'm super excited to check out how well this plays with #3609, #3536, #3533.

niri-config/src/binds.rs
JustinSpedding: I don't know. It was moved in @flowerysong's branch.

JustinSpedding:
> Can't speak on code quality, but relating to git stuff, it seems that you just applied @flowerysong's PR as a patch without preserving authorship. it would be better to checkout on their branch then do a `git rebase -i main` and squash their changes, then cherry-pick your commits over that.

Good idea! I just did that. Now the previous commits are preserved.

nikolai-in:
Doesn't work for mouse buttons

JustinSpedding:
> Doesn't work for mouse buttons

Yeah, I originally only tested it with keyboard keys. I pushed an update to enable it for mouse buttons. It should follow the same logic as the keyboard binds with `allow-invalidation`.

Uzlkav:
Hi, just tested this, seems to work (mostly) well. Few issues I came across:

- Modifier only binds dont seem to work on binds involving `ctrl` or `alt`, throwing the following error (I made sure to run `niri validate` with the binary compiled from this pull):

```
Error:   × invalid keybind
  ├─▶ invalid keybind
  ╰─▶ invalid key: Alt
    ╭─[test-config.kdl:45:1]
 45 │     // Invalid value
 46 │     Mod+Alt {
    ·     ───┬───
    ·        ╰── invalid value
```
I tested this with `Ctrl+Alt`, `Ctrl`, `Alt`, `Mod+Alt`.
Edit: Also tried `Control_L+Alt` and `Control_L+Alt_L`, same result.
Strangely, `Mod` by itself seems to work. Heres the exact config I have:

```kdl
binds {
    Mod {
        release { toggle-overview; }
    }
    Mod+Alt {
        press { spawn-sh "notify-send \"Ctrl+Alt pressed\""; }
        release { spawn-sh "notify-send \"Ctrl+Alt released\""; }
    }
}
```

- Second, press binds with a `spawn[-sh]` action seem to repeat infinitely. This is most easily seen by sending a notification, e.g.:
```kdl
Mod+U {
    press { spawn "notify-send" "\"Ctrl+Alt pressed\""; }
    release { spawn "notify-send" "\"Ctrl+Alt released\""; }
}
```

Thanks for your and flowerysongs work on this!

JustinSpedding:
@Uzlkav 

> Hi, just tested this, seems to work (mostly) well. Few issues I came across:
> 
>     * Modifier only binds dont seem to work on binds involving `ctrl` or `alt`, throwing the following error (I made sure to run `niri validate` with the binary compiled from this pull):
> 
> 
> ```
> Error:   × invalid keybind
>   ├─▶ invalid keybind
>   ╰─▶ invalid key: Alt
>     ╭─[test-config.kdl:45:1]
>  45 │     // Invalid value
>  46 │     Mod+Alt {
>     ·     ───┬───
>     ·        ╰── invalid value
> ```
> 
> I tested this with `Ctrl+Alt`, `Ctrl`, `Alt`, `Mod+Alt`. Edit: Also tried `Control_L+Alt` and `Control_L+Alt_L`, same result. Strangely, `Mod` by itself seems to work. Heres the exact config I have:
> 
> ```kdl
> binds {
>     Mod {
>         release { toggle-overview; }
>     }
>     Mod+Alt {
>         press { spawn-sh "notify-send \"Ctrl+Alt pressed\""; }
>         release { spawn-sh "notify-send \"Ctrl+Alt released\""; }
>     }
> }
> ```


As seen in flowerysong's MR description:


> I also made the implementation of modifier-only binds more generic.
> Non-Mod modifier binds are still configured with keysyms for the trigger
> (so Ctrl release=true { toggle-overview; } won't work but Control_L release=true { toggle-overview; } will.)


Using modifier keys requires using the specific key like Control_L instead of Ctrl. Also, that specific key will be the one to actually trigger the release binding. So in the case of `Ctrl+Alt_L`, keeping Ctrl held down and pressing/releasing Alt_L will trigger the release binding. But, keeping Alt_L held down and pressing/releasing Ctrl does not. This is how it differentiates between what is a modifier and what is the actual key. If you want it to trigger when releasing either one, you would need two bindings:

```
    Ctrl+Alt_L repeat=false {
        release { spawn "notify-send" "Released"; }
    }
    Alt+Control_L repeat=false {
        release { spawn "notify-send" "Released"; }
    }
```

Also, binding on press to a modifier is not currently supported. Only release works with that for now. I can look into supporting that.


>     * Second, press binds with a `spawn[-sh]` action seem to repeat infinitely. This is most easily seen by sending a notification, e.g.:
> 
> 
> ```kdl
> Mod+U {
>     press { spawn "notify-send" "\"Ctrl+Alt pressed\""; }
>     release { spawn "notify-send" "\"Ctrl+Alt released\""; }
> }
> ```
> 
> Thanks for your and flowerysongs work on this!

By default, holding down a bound key will trigger the press keybinding repeatedly. This is an existing feature of niri key bindings. You can add `repeat=false` to prevent this, as described in the readme.

JustinSpedding:
I just pushed an update to support press actions on modifier-only bindings.

Uzlkav:
@JustinSpedding 

>Using modifier keys requires using the specific key like Control_L instead of Ctrl. Also, that specific key will be the one to actually trigger the release binding. So in the case of Ctrl+Alt_L, keeping Ctrl held down and pressing/releasing Alt_L will trigger the release binding. But, keeping Alt_L held down and pressing/releasing Ctrl does not. This is how it differentiates between what is a modifier and what is the actual key. If you want it to trigger when releasing either one, you would need two bindings

Thanks for the clarification, on the modifier binds, currently building on latest commit and will test that.

>By default, holding down a bound key will trigger the press keybinding repeatedly. This is an existing feature of niri key bindings. You can add repeat=false to prevent this, as described in the readme.

I did try adding `repeat=false` to the press bind, but this didn't seem to take effect, I should have mentioned that in my original comment. Here's the exact syntax I used:
```kdl
Mod+U {
    press repeat=false { spawn "notify-send" "'Ctrl+Alt pressed'"; }
    release { spawn "notify-send" "'Ctrl+Alt released'"; }
}
```
After taking another look at the syntax used elsewhere, I tried the following instead:
```kdl
Mod+U repeat=false {
    press { spawn "notify-send" "'Ctrl+Alt pressed'"; }
    release { spawn "notify-send" "'Ctrl+Alt released'"; }
}
```
Which did work as expected. And looking more closely at the docs, that is indeed the form of the example provided. I think I would have expected to get an error message in the first case, though. Would implementing that be possible?

Uzlkav:
I've attempted to implement the error message, though I'm not sure if I did it right. Here's the pieces I changed:
```diff
diff --git a/niri-config/src/binds.rs b/niri-config/src/binds.rs
index b935da66..771afda5 100644
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -876,6 +876,7 @@ where
             .map_err(|e| DecodeError::conversion(&node.node_name, e.wrap_err("invalid keybind")))?;
 
         let mut repeat = true;
+        let mut repeat_node = None;
         let mut cooldown = None;
         let mut allow_when_locked = false;
         let mut allow_when_locked_node = None;
@@ -887,6 +888,7 @@ where
             match &***name {
                 "repeat" => {
                     repeat = knuffel::traits::DecodeScalar::decode(val, ctx)?;
+                    repeat_node = Some(name);
                 }
                 "cooldown-ms" => {
                     cooldown = Some(Duration::from_millis(
@@ -1057,6 +1059,15 @@ where
                     ));
                 }
             }
+            if repeat {
+                if let Some(node) = repeat_node {
+                    ctx.emit_error(DecodeError::unexpected(
+                            node,
+                            "property",
+                            "repeat must be set at the top level of the bind",
+                        ));
+                }
+            }
         }
         if let Some(ref action) = release_action {
             if !matches!(action, Action::Spawn(_) | Action::SpawnSh(_)) {
@@ -1068,6 +1079,15 @@ where
                     ));
                 }
             }
+            if repeat {
+                if let Some(node) = repeat_node {
+                    ctx.emit_error(DecodeError::unexpected(
+                            node,
+                            "property",
+                            "repeat must be set at the top level of the bind",
+                        ));
+                }
+            }
         }
 
         // The toggle-inhibit action must always be uninhibitable.
```
I'm compiling to test this now, let me know if you see any problems

Uzlkav:
> I'm compiling to test this now

Upd: It did not work as I expected. I think I'm probably missing something about the implementation

Upd 2: Tried moving some stuff around after looking at the code more, though it still doesn't work, lol. Here's what I have currently:

<details>

<summary>New patch</summary>

```diff
diff --git a/niri-config/src/binds.rs b/niri-config/src/binds.rs
index b935da66..771afda5 100644
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -876,6 +876,7 @@ where
             .map_err(|e| DecodeError::conversion(&node.node_name, e.wrap_err("invalid keybind")))?;
 
         let mut repeat = true;
+        let mut repeat_node = None;
         let mut cooldown = None;
         let mut allow_when_locked = false;
         let mut allow_when_locked_node = None;
@@ -887,6 +888,7 @@ where
             match &***name {
                 "repeat" => {
                     repeat = knuffel::traits::DecodeScalar::decode(val, ctx)?;
+                    repeat_node = Some(name);
                 }
                 "cooldown-ms" => {
                     cooldown = Some(Duration::from_millis(
@@ -1057,6 +1059,15 @@ where
                     ));
                 }
             }
+            if repeat {
+                if let Some(node) = repeat_node {
+                    ctx.emit_error(DecodeError::unexpected(
+                            node,
+                            "property",
+                            "repeat must be set at the top level of the bind",
+                        ));
+                }
+            }
         }
         if let Some(ref action) = release_action {
             if !matches!(action, Action::Spawn(_) | Action::SpawnSh(_)) {
@@ -1068,6 +1079,15 @@ where
                     ));
                 }
             }
+            if repeat {
+                if let Some(node) = repeat_node {
+                    ctx.emit_error(DecodeError::unexpected(
+                            node,
+                            "property",
+                            "repeat must be set at the top level of the bind",
+                        ));
+                }
+            }
         }
 
         // The toggle-inhibit action must always be uninhibitable.

```

</details>

YaLTeR:
I have a feeling that if a bind has press + release, it should default to repeat = false. In fact maybe even just writing press {} without release should default to repeat = false, so that you can comment out release to test and not have repeat default switch.

JustinSpedding:
> I have a feeling that if a bind has press + release, it should default to repeat = false. In fact maybe even just writing press {} without release should default to repeat = false, so that you can comment out release to test and not have repeat default switch.

Good idea. I just pushed an update to do that. Users can still override it with repeat=true if they want, but defaulting to false for bindings with release actions should be the 99% use case.

JustinSpedding:
I updated the documentation for recent changes. You also no longer need to include both `Ctrl` and `Control_L` in the key binding list since `Ctrl` is implied by `Control_L`. Same when using other modifier keys as trigger keys.

JustinSpedding:
I just rebased against main to resolve conflicts. I've been using it for a month now, and it's been working well. Is there anything else I need to do to get it ready for review and/or merge?

witchlliee:
I've also been using for a long time already and it's been great!

Atan-D-RP4:
> I just rebased against main to resolve conflicts. I've been using it for a month now, and it's been working well. Is there anything else I need to do to get it ready for review and/or merge?

It's a big and needs focus and consideration for a full proper review. Which takes time. Which YaLTeR doesn't have too much of.

So just time is needed. He will get to this eventually.

docs/wiki/Configuration:-Key-Bindings.md
```diff
@@ -81,6 +81,53 @@ binds {
...
+
+Binds can be set to trigger on key press, on key release, or both. By default, binds trigger on key press. You can specify the timing using `press {}` and `release {}` blocks:
+
+```kdl
+binds {
+    // Trigger on press (default behavior)
+    Mod+T { spawn "alacritty"; }
+
+    // Trigger on release
+    Mod { release { toggle-overview; } }
```
norude:
```suggestion
    Mod { release { toggle-overview; }; }
```
Right now, this example is wrong as a semicolon or a newline is expected

docs/wiki/Configuration:-Key-Bindings.md
```diff
@@ -81,6 +81,53 @@ binds {
...
+
+Release bindings are mostly useful when you want to bind a modifier key to an action, as it avoids unwanted triggering when you're trying to use other binds involving that modifier.
+
+By default, any key binding that includes a release action will have repeat turned off for its press action. This can be overridden with `repeat=true` if desired.
+
+Release binds will normally only trigger if no other keys were released and no keys or mouse buttons were pressed after the bound key was pressed. If you want a release bind to always trigger regardless, set `allow-invalidation=false`:
+
+```kdl
+binds {
+    Mod allow-invalidation=false { release { toggle-overview; } }
```
norude:
```suggestion
    Mod allow-invalidation=false { release { toggle-overview; }; }
```
Again, a semicolon or a new line is needed

JustinSpedding:
@norude Thanks for pointing that out. I updated the docs and resolved conflicts with main.

RoadRoller01:
What is pausing this merge request?

HigherOrderLogic:
Yalter doesnt have time, and this PR hasnt been reviewed yet.

Sempyos:
If you want to show your support for a PR, it's much nicer to react the first message or the tracking issue

JustinSpedding:
I updated the MR to resolve merge conflicts and apply the same press/release binding support to the newly added tablet stylus button bindings. That being said, I do not have a tablet with a stylus. So I cannot fully test that.

wramalho:
Using it as a fork for quite some time in nixos (about a month) with no issues to report so far :)

JustinSpedding:
I just resolved conflicts with the recent changes on the main branch. I've also been daily-driving this branch for a few months now. I guess we just wait until Yalter has time to review.

ShayBox (approved):
Ready to merge

niri-config/src/binds.rs
```diff
@@ -877,69 +907,184 @@ where
...
-                        if let Some(node) = allow_when_locked_node {
-                            ctx.emit_error(DecodeError::unexpected(
-                                node,
-                                "property",
-                                "allow-when-locked can only be set on spawn binds",
-                            ));
+        let mut press_action: Option<Action> = None;
+        let mut release_action: Option<Action> = None;
+        let mut has_press_section = false;
+        let mut has_release_section = false;
```
HigherOrderLogic, line 928: You can probably check these with `is_none` and `is_some` instead.

JustinSpedding: Updated.

JustinSpedding:
I merged changed from main, applied a commented suggestion, and updated tests.

RoadRoller01:
Finally! :smile: I can auto-hide `waybar` and show it only when mod key is pressed.
You can do it by setting `mode` in `waybar` config to `hide` and `start_hidden` to `true`,
then add this:
```kdl
    Mod allow-invalidation=false {
      press   { spawn "killall" "-SIGUSR1" "waybar"; }
      release { spawn "killall" "-SIGUSR1" "waybar"; }
    }
```
Note: if you use NixOS change `waybar` to `.waybar-wrapped`
What is missing is a way to do an action if the workspace is empty and when the mouse is on the top edge.

RoadRoller01:
I think I found a bug the `allow-invalidation=false` doesn't apply to `Mod`+`Ctrl` and `Mod`+`Ctrl`+`*`,
and any second modifier key.

JustinSpedding:
@RoadRoller01 Yes, looks like a case was missed in `allow-invalidation` during a previous refactor. I updated the code to make it consistent, as well as reject more cases of invalid binding config entries. However, this leads me to question what the default behavior of `allow-invalidation` should be.

In my mind, `allow-invalidation=false` means "trigger the press/release actions when the modifiers+trigger keys are pressed/released regardless of whether other keys are pressed/released at the same time." So this might be useful for something like push-to-talk. Maybe the user is playing a videogame where they may or may not be holding down Shift to run at any given time, and they want a push-to-talk binding to trigger both the press and release bindings regardless of which other modifiers like Shift are being held. The logic would find the binding with the most exact match (Mod+Alt+V would take precedence over Mod+V) when triggering an action.

I updated the code to match this behavior. Try playing around with it and let me know if the logic works for you.

RoadRoller01:
I also would love to be able to bind `toggle-overview` to `Mod+Mod` like gnome app drawer.

HigherOrderLogic:
> Mod+Mod

Isnt this double tap? If so then I believe it's out of scope for Niri.

RoadRoller01:
Yeah its(:

Lopolin-LP:
> I also would love to be able to bind `toggle-overview` to `Mod+Mod` like gnome app drawer.

@RoadRoller01 You could make a bash script that's bound to `Mod` that times out after like 0.3 seconds, then when it's invoked and it sees it has another instance running only then run an action.

```bash
action() {
    echo 'Do something'
}

if pidof -o $$ -x "path/to/this/script.sh" >/dev/null; then
    action
else
    sleep 0.3
fi
```
note: i barely tested this myself

please keep in mind `path/to/this/script.sh` must be what is said in the cmdline when it's executed. A relative and absolute path pointing to the same are different paths. Same goes for example `/bin/bash` and `/usr/bin/bash`. `~` is not expanded, use `$HOME`.

RoadRoller01:
@JustinSpedding I found another one (: `mod+Tab`

docs/wiki/Configuration:-Key-Bindings.md
```diff
@@ -81,6 +82,75 @@ binds {
...
+    // by putting the same toggle command on the press and release action.
+    Control_R {
+        press { spawn "toggle-command"; }
+        release { spawn "toggle-command"; }
+    }
+
+    // For multi-key modifier-only bindings,
+    // use Ctrl/Alt/etc. for the modifier keys and
+    // Control_L/Control_R/Alt_L/Alt_R/etc. for the trigger key.
+    Ctrl+Alt_L {
```
YaLTeR, line 117: So since we're allowing `Mod` to be used as the key, maybe it makes sense to also allow other modifiers like `Ctrl` and `Alt`? To remove this distinction and special case. They already need to be handled since `Mod` can be either of them. And keys like `Control_L` are left for when you want to bind just one of the two keys.

docs/wiki/Configuration:-Key-Bindings.md
```diff
@@ -81,6 +82,75 @@ binds {
...
+```
+
+Release binds will normally only trigger if no other keys were released and no keys or mouse buttons were pressed after the bound key was pressed.
+If you want a release bind to always trigger regardless, set `allow-invalidation=false`:
+
+```kdl
+binds {
+    Mod allow-invalidation=false { release { toggle-overview; }; }
+}
+```
```
YaLTeR, line 149: Can this be removed actually? The original motivation was because the previous design didn't combine press and release. In this design, for combined press + release binds, if press triggers then release must also trigger unconditionally, so `allow-invalidation=false` is no longer necessary. Should simplify the code a bit.

niri-config/src/binds.rs
YaLTeR: How about a new
```rust
enum BoundAction {
    Press(Action),
    Release(Action),
    Both { press: Action, release: Action },
}
```
? This way you cannot encode an invalid state when both are `None`.

niri-config/src/binds.rs
```diff
@@ -22,14 +22,38 @@ pub struct Binds(pub Vec<Bind>);
...
+
+    pub fn has_release(&self) -> bool {
+        self.release_action.is_some()
+    }
+
+    pub fn is_release_only(&self) -> bool {
+        self.press_action.is_none() && self.release_action.is_some()
+    }
+
+    pub fn action_for(&self, pressed: bool) -> Option<&Action> {
```
YaLTeR, line 48: With `BoundAction`, you can change this to `pub fn action_for(self, pressed: bool) -> Action` and avoid having to clone in the input code.

src/input/mod.rs
```diff
@@ -660,14 +699,18 @@ impl State {
...
+
         let Some(cooldown) = bind.cooldown else {
-            self.do_action(bind.action, bind.allow_when_locked);
+            self.do_action(action, bind.allow_when_locked);
             return;
         };
 
         // Check this first so that it doesn't trigger the cooldown.
-        if self.niri.is_locked() && !(bind.allow_when_locked || allowed_when_locked(&bind.action)) {
+        if self.niri.is_locked() && !(bind.allow_when_locked || allowed_when_locked(&action)) {
```
YaLTeR, line 713: What should be do if `allowed_when_locked(pressed)` doesn't match `allowed_when_locked(released)`? I think it should work this way:
- if the bind was pressed while unlocked, then release should trigger even if the screen
was now locked and it's not allowed when locked (because release must always trigger if press triggered).
- if we're on a lock screen, the press should check both press and release allowed when
locked, and trigger only if both are allowed.

src/input/mod.rs
```diff
@@ -2788,16 +2833,53 @@ impl State {
...
 
-        // Ignore release events for mouse clicks that triggered a bind.
-        if self.niri.suppressed_buttons.remove(&button_code) {
-            return;
-        }
+        let valid_mouse_release_trigger = if button_state == ButtonState::Pressed {
+            self.niri.valid_mouse_release_trigger.replace(button_code)
+        } else {
+            self.niri.valid_mouse_release_trigger.take()
+        };
```
YaLTeR, line 2840: Here and in other places (tablet, keyboard): release bind must always trigger if press triggered, and this code doesn't seem like it does that because it replaces the previous button unconditionally. Imagine binding MouseLeft press + release, then pressing MouseLeft (triggering press), pressing MouseRight (overriding valid_mouse_release_trigger), releasing MouseLeft (release does not trigger!).

This needs to be fixed and relevant tests added.

src/input/mod.rs
```diff
@@ -4808,10 +5088,11 @@ fn allowed_when_locked(action: &Action) -> bool {
     )
 }
 
-fn allowed_during_screenshot(action: &Action) -> bool {
+fn allowed_during_screenshot(action: Option<&Action>) -> bool {
```
YaLTeR, line 5091: It doesn't make sense for this function to take an Option, so it should stay as is, and Option should be handled at the callsites.

YaLTeR (changes requested):
Hey, I rebased to main (please don't use merge commits in the future), ran cargo fmt (please do this yourself), and took a first look.

EDIT: oops accidentally erased commit author; I'll fix it later

JustinSpedding:
> Hey, I rebased to main (please don't use merge commits in the future), ran cargo fmt (please do this yourself), and took a first look.
> 
> EDIT: oops accidentally erased commit author; I'll fix it later

Thanks for taking the time to review this! I am on vacation atm, but will look into the requested changes when I get back

JustinSpedding:
I pushed some commits to do the following:

- Remove `allow-invalidation`
- Switch to a `BoundAction` enum to keep track of press and release actions
- Switch to keeping track of pending release actions so that we never fail to trigger a release action after the relevant key was pressed
- Allow modifier-only bindings for other modifiers like Ctrl, Alt, Shift, etc.
- General code cleanup

However, I decided to add in an exception to requirement of never forgetting to run release bindings: If there is a modifier-only release-only binding such as `Mod { release { toggle-overview; }; }`, we do not want the release binding to trigger if the modifier was used as part of another shortcut. Otherwise, binding Mod to open the overview would trigger constantly when performing other shortcuts. So for that reason, I still had to keep around part of the invalidation codepath, albeit cleaned up a bit. This is the only case in which invalidation is used. We still always trigger both the press and release bindings if a modifier-only binding has both.

@YaLTeR The commit history still needs to have its authorship restored, but are there any other changes you would like me to make?

YaLTeR:
Do you mind cleaning up the commit history according to https://github.com/niri-wm/niri/blob/main/CONTRIBUTING.md#writing-pull-requests?

YaLTeR:
> I decided to add in an exception to requirement of never forgetting to run release bindings: If there is a modifier-only release-only binding such as `Mod { release { toggle-overview; }; }`, we do not want the release binding to trigger if the modifier was used as part of another shortcut.

Yeah, of course. The rule is that if a *press* action of a press-release binding triggered, then its *release* must trigger.

JustinSpedding:
Ok, I split up the git history into 5 commits to separate the modifier-only binds change from the press/release change.

However, the second commit is still massive since it touches so much and adds so many tests. idk if it is worth trying to break it down further.

YaLTeR:
It's fine as long as it's self-contained

docs/wiki/Configuration:-Key-Bindings.md
```diff
@@ -15,7 +15,8 @@ binds {
 }
 ```
 
-The hotkey consists of modifiers separated by `+` signs, followed by an XKB key name in the end.
+The parts of a hotkey definition are separated by '+' and consist of zero or more modifiers followed by a trigger key.
+This can be either an XKB key name or <sup>Since: next release</sup> `Mod`.
```
YaLTeR: I guess something like this now:
```
The hotkey consists of modifiers separated by `+` signs, followed by a trigger key in the end.
The trigger key can be either an XKB key name or <sup>Since: next release</sup> a modifier.
```

src/ui/hotkey_overlay.rs
```diff
@@ -498,28 +498,19 @@ fn key_name(screen_reader: bool, mod_key: ModKey, key: &Key) -> String {
...
     let has_comp_mod = key.modifiers.contains(Modifiers::COMPOSITOR);
 
+    let mod_key_pretty = match mod_key {
+        ModKey::Super => String::from("Super"),
+        ModKey::Alt => String::from("Alt"),
+        ModKey::Shift => String::from("Shift"),
+        ModKey::Ctrl => String::from("Ctrl"),
+        ModKey::IsoLevel3Shift => String::from("Mod5"),
+        ModKey::IsoLevel5Shift => String::from("Mod3"),
+    };
```
YaLTeR: This match could be extracted into a function returning `&'static str` (seeing how it's used in two places).

src/ui/hotkey_overlay.rs
```diff
@@ -632,6 +632,46 @@ mod tests {
...
+        );
+        assert_snapshot!(
+            check(
+                r#"binds {
+                    Super { close-window; }
+                }"#,
+                Action::CloseWindow,
+            ),
+            @" Super : Close Focused Window"
+        );
```
YaLTeR, line 686: One or two tests of modifier + modifier binds pls. And maybe something with a `Control_L`/`Control_R` trigger (should render as is since it's only a single key).

niri-config/src/binds.rs
```diff
@@ -23,14 +23,56 @@ pub struct Binds(pub Vec<Bind>);
...
+    pub fn press_action(&self) -> Option<&Action> {
+        match &self.action {
+            BoundAction::Press(action) | BoundAction::Both { press: action, .. } => Some(action),
+            BoundAction::Release(_) => None,
+        }
+    }
+
+    pub fn action_for(&self, pressed: bool) -> &Action {
+        match &self.action {
+            BoundAction::Press(action) | BoundAction::Release(action) => action,
```
YaLTeR: This ignores `pressed` which doesn't seem right

niri-config/src/binds.rs
```diff
@@ -23,14 +23,56 @@ pub struct Binds(pub Vec<Bind>);
...
+            BoundAction::Both { press, .. } if pressed => press,
+            BoundAction::Both { release, .. } => release,
+        }
+    }
+
+    pub fn is_modifier_only_release(&self) -> bool {
+        !self.has_press() && self.key.modifiers.is_empty() && self.key.trigger.is_modifier()
```
YaLTeR, line 73: This name doesn't match the logic because a bind like `Tab { release { close-window; } }` will return true despite not being a modifier-only release

niri-config/src/binds.rs
```diff
@@ -890,69 +935,218 @@ where
...
+                        Err(e) => {
+                            ctx.emit_error(e);
                         }
                     }
+                } else {
+                    ctx.emit_error(DecodeError::missing(
+                        child,
+                        "expected an action for this press section",
+                    ));
+                }
```
YaLTeR, line 990: Some of this logic that's shared between press, release and normal, can probably be extracted into a helper closure that sits right in here

niri-config/src/binds.rs
```diff
@@ -890,69 +935,218 @@ where
...
+                        node,
+                        "property",
+                        "allow-when-locked can only be set on spawn binds",
+                    ));
+                }
+            }
+        }
+        if let Some(ref action) = release_action {
+            if !matches!(action, Action::Spawn(_) | Action::SpawnSh(_)) {
+                if let Some(node) = allow_when_locked_node {
```
YaLTeR, line 1098: Move this allow_when_locked_node condition at the top above the two of these ifs, and also leave only one emit_error call for both

src/input/mod.rs
```diff
@@ -591,17 +603,41 @@ impl State {
...
+                }
+            }
+            ShouldInterceptResult::InterceptAndHandle(bind) => {
+                self.handle_bind(bind.clone(), pressed);
+                if pressed {
+                    self.start_key_repeat(bind);
+                }
+            }
+            ShouldInterceptResult::InterceptOnly => {}
         }
```
YaLTeR, line 642: I feel like this may be cleaner if split into two matches:
```rust
        match &result {
            ShouldInterceptResult::Forward | ShouldInterceptResult::ForwardAndHandle(_) => {
                keyboard.input_forward(
                    self,
                    event.key_code(),
                    event.state(),
                    serial,
                    time,
                    mods_changed,
                );
            }
            ShouldInterceptResult::InterceptAndHandle(_) | ShouldInterceptResult::InterceptOnly => (),
        }
        match result {
            ShouldInterceptResult::ForwardAndHandle(bind) | ShouldInterceptResult::InterceptAndHandle(bind) => {
                self.handle_bind(bind.clone(), pressed);
                if pressed {
                    self.start_key_repeat(bind);
                }
            }
            ShouldInterceptResult::Forward | ShouldInterceptResult::InterceptOnly => (),
        }
```

src/input/mod.rs
```diff
@@ -2833,15 +2874,18 @@ impl State {
                     let config = self.niri.config.borrow();
                     let bindings =
                         make_binds_iter(&config, &mut self.niri.window_mru_ui, modifiers);
-                    find_configured_bind(bindings, mod_key, trigger, mods)
+                    find_configured_bind(bindings, mod_key, trigger, mods, true)
                 })
                 .filter(|bind| {
-                    !self.niri.screenshot_ui.is_open() || allowed_during_screenshot(&bind.action)
+                    !self.niri.screenshot_ui.is_open()
+                        || allowed_during_screenshot(bind.action_for(bind.has_press()))
```
YaLTeR: I guess allowed_during_screenshot needs to have the same sort of logic as my earlier comment about allowed_when_locked

src/input/mod.rs
```diff
@@ -4549,30 +4640,41 @@ fn should_intercept_key<'a>(
...
 
-    let mut final_bind = find_bind(
-        bindings,
-        mod_key,
-        modified,
-        raw,
-        mods,
-        disable_power_key_handling,
-    );
+    cancel_interrupted_release_binds(pending_release_binds);
```
YaLTeR, line 4690: Is this needed both here and in other functions? When can it trigger here and when in other functions?

src/input/mod.rs
```diff
@@ -131,6 +132,14 @@ impl<D: SeatHandler + TabletSeatHandler> AnyStartData<D> {
     }
 }
 
+#[derive(Debug)]
+enum ShouldInterceptResult {
+    Forward,
+    ForwardAndHandle(Bind),
+    InterceptAndHandle(Bind),
+    InterceptOnly,
+}
```
YaLTeR, line 141: Do you mind extracting this enum + input() => input_intercept() change into a separate commit before this main one? To reduce the big diff and make sure it doesn't break anything

src/input/mod.rs
```diff
@@ -4602,27 +4704,43 @@ fn should_intercept_key<'a>(
...
+                ShouldInterceptResult::Forward
+            } else if modified.is_modifier_key() {
+                if bind.has_release() {
+                    pending_release_binds.insert(key_code, PendingReleaseBind::new(bind.clone()));
+                }
+                if bind.has_press() {
+                    ShouldInterceptResult::ForwardAndHandle(bind)
+                } else {
+                    ShouldInterceptResult::Forward
+                }
```
YaLTeR, line 4750: I guess the point of this special modifier logic is to make sure the client gets the modifiers if you have a bind like `Ctrl + Alt { release { whatever; } }` and still want to be able to press Ctrl + Alt + K type keys in applications? Could use a comment about it.

src/input/mod.rs
```diff
@@ -4602,27 +4704,43 @@ fn should_intercept_key<'a>(
...
-            // so the release would already have been forwarded at the start of this function.
+        (_, false) if suppressed_keys.contains(&key_code) => {
+            // We don't need to check for shortcut inhibition here because
+            // if it was inhibited on press it wouldn't be suppressed.
             suppressed_keys.remove(&key_code);
-            FilterResult::Intercept(None)
+            ShouldInterceptResult::InterceptOnly
         }
-        (None, true) => FilterResult::Forward,
+        _ => ShouldInterceptResult::Forward,
```
YaLTeR: I'd rather not downgrade the exhaustive specific match to a catch-all here

src/input/mod.rs
```diff
@@ -4663,34 +4781,92 @@ fn find_bind<'a>(
...
+
+    // It would maybe be better to return all matching binds instead of using a parameter to
+    // prioritize them, but that would complicate things for the callers and right now there aren't
+    // any that need more than one.
+    //
+    // Prefer press binds, but fall back to release-only binds so that the key press can be
+    // intercepted and its release action triggered when the key is released.
+    for trigger in triggers.into_iter().flatten() {
+        let bind = find_configured_bind(bindings.clone(), mod_key, trigger, mods, true)
+            .or_else(|| find_configured_bind(bindings.clone(), mod_key, trigger, mods, false));
```
YaLTeR, line 4841: When can this be different/better compared to one `find_configured_bind()` across all binds (regardless of press or release)?

src/input/mod.rs
```diff
@@ -4663,34 +4781,92 @@ fn find_bind<'a>(
...
     // Handle configured binds.
     let mut modifiers = modifiers_from_state(mods);
 
-    let mod_down = modifiers_from_state(mods).contains(mod_key.to_modifiers());
-    if mod_down {
-        modifiers |= Modifiers::COMPOSITOR;
+    // Check if the trigger is a modifier key (like Mod, Alt, Alt_L, Control_L, Shift_L, etc.)
+    // If so, we need to remove its modifier from the current modifiers since the key is the
+    // trigger, not a modifier.
+    let trigger_is_modifier = trigger.is_modifier();
```
YaLTeR, line 4863: I wonder if we could check some of these duplicates at parse time. Like if you have a `Ctrl + Alt {}` and `Ctrl + Alt_L {}`, would be good to catch and error that at parse time.

src/input/mod.rs
```diff
@@ -5311,254 +5497,336 @@ fn make_binds_iter<'a>(
...
-        ));
-        assert!(suppressed_keys.contains(&close_key_code));
+            })
+        );
+        assert!(common_state.suppressed_keys.contains(&CLOSE_KEY_CODE));
 
-        let filter = close_key_event(&mut suppressed_keys, mods, false);
-        assert!(matches!(filter, FilterResult::Intercept(None)));
-        assert!(suppressed_keys.is_empty());
+        let filter = process_close_key(&mut common_state, &bindings, mods, false);
```
YaLTeR, line 5765: Would be also nice to extract refactoring the existing tests into a separate commit before, and then in this commit add only the newly necessary parts

src/ui/hotkey_overlay.rs
```diff
@@ -158,7 +158,7 @@ fn format_bind(binds: &[Bind], action: &Action) -> Option<(Option<Key>, String)>
     let mut found_null_title = false;
 
     for bind in binds {
-        if bind.action != *action {
+        if bind.press_action() != Some(action) {
```
YaLTeR, line 161: Need to think what to do here rather than just change everything to `press_action()`. For example, say we default the overview to `Mod { release {} }`. It needs to show up. So I guess it should rather be something like `bind.press_action().or_else(|| bind.release_action())`?

src/niri.rs
```diff
@@ -653,6 +655,22 @@ pub struct PendingMruCommit {
...
+#[derive(Debug, Clone)]
+pub struct PendingReleaseBind {
+    pub bind: Bind,
+    pub cancel_on_other_input: bool,
+}
+
+impl PendingReleaseBind {
+    pub fn new(bind: Bind) -> Self {
+        Self {
+            cancel_on_other_input: bind.is_modifier_only_release(),
```
YaLTeR: If it's trivially computed from the bind then it doesn't need to be stored?

src/input/mod.rs
```diff
@@ -2874,13 +2882,19 @@ impl State {
                     let config = self.niri.config.borrow();
                     let bindings =
                         make_binds_iter(&config, &mut self.niri.window_mru_ui, modifiers);
-                    find_configured_bind(bindings, mod_key, trigger, mods, true)
+                    find_configured_bind(bindings.clone(), mod_key, trigger, mods, true)
+                        .or_else(|| find_configured_bind(bindings, mod_key, trigger, mods, false))
```
YaLTeR, line 2897: Here also a question of why is it different/better to search for press and release separately rather than at once

YaLTeR (changes requested):
Phew, here's the second round. Please also don't forget about my unresolved allowed-when-locked logic comment.

YaLTeR:
I just rewrote the keybinding tests as client-server snapshot tests here: https://github.com/niri-wm/niri/blob/929565896a3569f417b92ae09f84303a3a85c183/src/tests/binds.rs#L130

My intention is to remove the `bindings_suppress_keys()` as they fully replace it while being hopefully clearer and easier to write.

Could you please try porting your new press-release tests to this new system? I think it should work well and make them easier to understand. You'll need to `cargo install cargo-insta` and then run `cargo insta test --accept -- binds` to automatically regenerate the snapshots after you add new tests.

JustinSpedding:
> I just rewrote the keybinding tests as client-server snapshot tests here:
> 
> https://github.com/niri-wm/niri/blob/929565896a3569f417b92ae09f84303a3a85c183/src/tests/binds.rs#L130
> 
> My intention is to remove the `bindings_suppress_keys()` as they fully replace it while being hopefully clearer and easier to write.
> 
> Could you please try porting your new press-release tests to this new system? I think it should work well and make them easier to understand. You'll need to `cargo install cargo-insta` and then run `cargo insta test --accept -- binds` to automatically regenerate the snapshots after you add new tests.

Yes, I can try porting over the tests after I finish with the other stuff. I am working through your previous comments one-by-one when I have time, and putting thumbs-up reactions on them as I address them. I'll tag you when I get around to it.

niri-config/src/binds.rs
JustinSpedding: How would that binding return true? `Tab` is not a modifier, so it would fail the `self.key.trigger.is_modifier()` check

YaLTeR:
You can resolve the thread instead of thumbs up, makes it easier to view afterwards

JustinSpedding:
Ah ok, I wasn't sure if you wanted to resolve them yourself to show that you approve of how I addressed them. I can resolve them if you don't care about that

YaLTeR:
I can unresolve if needed so it's ok

src/input/mod.rs
JustinSpedding: I updated it to catch duplicate bindings between things like `Alt` and `Alt_L`, but did not catch conflicts between `Mod` and `Super` since we do not know which key is `Mod` at parse time. That seems like an existing thing unrelated to the purpose of this MR, and would likely require more of a refactor.

src/input/mod.rs
JustinSpedding: This is needed here and in the other places in order to prevent us from triggering modifier-only release-only bindings when something else was pressed between when the modifier was pressed and released. It relates to me earlier comment about not wanting to trigger `Mod { release {} }` actions when using Mod as a modifier for something else.

We call `cancel_interrupted_release_binds` in the input handlers for keyboard, mouse, and tablet.

PR #3632, opened 2026-03-16
doc: Update default-config.kdl for various new features

omentic:
I noticed the `default-config.kdl` shipped is pretty old -- it doesn't have ex. `overview` or `gestures` or `recent-windows`.  This adds the default settings from the documentation there.

The only change this makes that is not default (according to the wiki) is to add the `Mod+Tab` bindings to be scoped to the output by default. This seems a handy default, and the current bindings are redundant, so I've changed it and updated the wiki (not sure where the actual default is), but can revert if that's not desired.

### DIFF
--- a/docs/wiki/Configuration:-Recent-Windows.md
+++ b/docs/wiki/Configuration:-Recent-Windows.md
@@ -28,13 +28,13 @@ recent-windows {
     binds {
         Alt+Tab         { next-window; }
         Alt+Shift+Tab   { previous-window; }
-        Alt+grave       { next-window     filter="app-id"; }
-        Alt+Shift+grave { previous-window filter="app-id"; }
+        Alt+Grave       { next-window     filter="app-id"; }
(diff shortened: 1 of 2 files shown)


YaLTeR:
Including every option is not the intention behind default config. See https://niri-wm.github.io/niri/Development%3A-Design-Principles.html#default-config

omentic:
This isn't including all options. It's just including the options that are set by default. I find it kind of confusing to have default values that are not listed... and this doesn't add too much to the default config, less than ~200 lines.

I was a little surprised by the `recent-windows` keybinds when I updated Niri recently. (It's very nice, by the way.) But my first stop for finding what was bound to `Mod+Tab` was the default config, which didn't have it.

YaLTeR:
Why not the wiki? I'd argue the wiki is an easier place to reach for in this case, and it does have this documented.

omentic:
Not all projects have a wiki, all projects have a default configuration. It had been a while since I last checked out Niri.

PR #3316, opened 2026-01-26
Add toggle-touchpad action

barrulus:
## Summary

- Adds a new `toggle-touchpad` action to toggle the touchpad on/off at runtime without editing the config file
- Can be bound to a key or invoked via IPC with `niri msg action toggle-touchpad`
- The toggle state is combined with the config's `off` setting - both must allow the touchpad for it to be enabled

## Notes

- State resets on restart
- Applies to newly hot-plugged touchpads

## Test plan

- [x] Bind `toggle-touchpad` to a key and verify it toggles touchpad input
- [x] Test via IPC: `niri msg action toggle-touchpad`
- [x] Verify interaction with config `off` setting
- [x] Hot-plug a touchpad while toggled off and verify it stays off

PR #3642, opened 2026-03-18
Add toggle-dwt action to toggle disable-while-typing at runtime

barrulus:
## Summary
- Adds a new `toggle-dwt` key binding action and IPC command that toggles disable-while-typing on/off at runtime
- The toggle state is combined with the config's `dwt` setting using AND logic: both must enable DWT for it to be active (`c.dwt && !dwt_disabled_by_toggle`)
- Toggle state resets on restart, and applies to newly hot-plugged touchpads
- Includes documentation updates for both the Input and Key Bindings wiki pages

Follows the same pattern as `toggle-touchpad` in #3316.

## Test plan
- [x] Verified `niri msg action toggle-dwt` is recognized and executes without error
- [x] Confirmed DWT is active by default (touchpad suppressed while typing)
- [x] Confirmed pressing toggle-dwt disables DWT (touchpad responds while typing)
- [x] Confirmed pressing toggle-dwt again re-enables DWT
- [x] Verified the feature compiles cleanly


Fireye04:
Just tested this on nixos unstable, works great both via `niri msg action` and via keybind.

Thanks again!

PR #3645, opened 2026-03-18
layout: add opt-in global workspace indices

haug1:
## Summary
- add an opt-in `layout { global-workspace-indices; }` mode
- make numbered workspace references resolve to globally unique workspace numbers across outputs
- keep sequential workspace navigation output-local while stepping through global numbers
- export only real mapped workspaces through IPC/ext-workspace and align ext-workspace coordinates with the exported index
- add regression tests for focus, move, empty-workspace, and workspace-move behavior

## Context
- Implements the behavior discussed in https://github.com/niri-wm/niri/discussions/3639

## Validation
- `cargo fmt --all`
- `nix develop -c cargo build`
- manual runtime testing in a debug-built niri session across a two-monitor setup

### DIFF
--- a/docs/wiki/Configuration:-Layout.md
+++ b/docs/wiki/Configuration:-Layout.md
@@ -154,6 +154,29 @@ layout {
 }
 ```
 
+### `global-workspace-indices`
+
+If set, workspace numbers become globally unique across all outputs.
+
+Direct numeric references like `focus-workspace 1` and `move-column-to-workspace 2`
+target that global workspace number instead of the currently focused monitor's
+workspace position.
+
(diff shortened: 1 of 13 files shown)


haug1:
This is going to be quite a big change that i've mostly vibe-coded and done behavior driven development with, so i'm not confident in this code at all, however it behaves exactly how i want in workspace behavior on mutliple monitors.

Update: Still no expectations on my end with this PR of course, just thought I would add that I've been using this build for about two weeks now and had no problems.
Of course I'm not claiming to have properly tested every use-case and the code is probably definitely not merge-able.

PR #3653, opened 2026-03-19
include output in WorkspaceActivated event

clxxiii:
*(This was quite literally a 2 minute change. If this change is more breaking than it's worth feel free to close)*

This just adds an output property for `WorkspaceActivated` IPC events & for the event stream.

Currently, `WorkspaceActivated` events just contain the ID of the workspace, and whether it's focused. For my use case, a multi-monitor bar which shows the workspaces on each monitor individually, the bar has no way of knowing if the workspace was activated on the monitor it cares about, since the `WorkspacesChanged` event doesn't fire until a new window is added to said workspace

---

`niri msg --json event-stream`
```
{"WorkspaceActivated":{"id":9,"focused":true,"output":"DP-3"}}
{"WindowFocusChanged":{"id":null}}
{"WorkspaceActivated":{"id":6,"focused":true,"output":"DP-3"}}
{"WindowFocusChanged":{"id":46}}
```

`niri msg event-stream`
```
Workspace focused: 9 on output DP-3
Window focus changed: None
Workspace focused: 6 on output DP-3
Window focus changed: Some(46)
```

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1596,6 +1596,8 @@ pub enum Event {
         /// If `true`, this is now the single focused workspace. All other workspaces are no longer
         /// focused, but they may remain active on their respective outputs.
         focused: bool,
+        /// The output name the workspace was activated on
+        output: Option<String>,
     },
     /// An active window changed on a workspace.
     WorkspaceActiveWindowChanged {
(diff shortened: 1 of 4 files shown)


clxxiii:
TIL the workspaces command gives you the unopened workspaces too :/

PR #3657, opened 2026-03-20
Add screenshot related events/state for `niri msg`

bczhc:
Currently actions sent via niri cmd are nonblocking. E.g. for screenshots, niri doesn't have a way other than monitoring the output path to query if the screenshot is finished, or the interactive screenshot UI is confirmed/canceled by the user.

This PR adds the following three `niri msg event-stream` events:

```jsonl
{"ScreenshotUiChanged":{"event":"Open"}}
{"ScreenshotUiChanged":{"event":"Confirm"}}
{"ScreenshotUiChanged":{"event":"Cancel"}}
```

and their corresponding EventStreamState, queried via command `niri msg screenshot-state`:

```console
❯ niri msg screenshot-state
Screenshot UI is closed.
```

Thus by wrapping `niri msg action screenshot*` and listening to these events, a blocked version of screenshot utility (like `grim`) can be implemented. This enables a more convenient script writing for post processing the produced screenshot images.

An example ["blocked niri-screenshot wrapper"](https://github.com/bczhc/dotfiles/blob/cd8c2dd9a92e2a82543a8541b117ed5513836795/bin/scripts/niri-ss-blocked)

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1565,6 +1565,18 @@ pub enum CastTarget {
     },
 }
 
+/// Screenshot UI events.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub enum ScreenshotUiEvent {
+    /// Start the interactive screenshot.
+    Open,
+    /// Cancel the screenshot.
+    Cancel,
+    /// Confirm the screenshot.
+    Confirm,
+}
+
 /// A compositor event.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
@@ -1678,6 +1690,11 @@ pub enum Event {
         ///
         /// For example, the config file couldn't be parsed.
         failed: bool,
+    },
+    /// The screenshot UI was changed.
+    ScreenshotUiChanged {
+        /// Opened/dismissed/confirmed events of the screenshot UI.
+        event: ScreenshotUiEvent,
     },
     /// A screenshot was captured.
     ScreenshotCaptured {
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -1586,7 +1586,13 @@ pub enum ScreenshotUiEvent {
     /// Cancel the screenshot.
     Cancel,
     /// Confirm the screenshot.
-    Confirm,
+    Confirm {
+        /// (x, y) coordinate of the selection origin (relative to the output top-left corner), in
+        /// physical pixels.
+        position: (i32, i32),
+        /// Width and height of the selection area, in physical pixels.
+        size: (i32, i32),
+    },
 }
 
 /// A compositor event.
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -117,6 +117,8 @@ pub enum Request {
     ReturnError,
     /// Request information about the overview.
     OverviewState,
+    /// Request information about the screenshot.
+    ScreenshotState,
     /// Request information about screencasts.
     Casts,
 }
@@ -163,6 +165,8 @@ pub enum Response {
     OutputConfigChanged(OutputConfigChanged),
     /// Information about the overview.
     OverviewState(Overview),
+    /// Information about the screenshot UI.
+    ScreenshotState(ScreenshotUi),
     /// Information about screencasts.
     Casts(Vec<Cast>),
 }
@@ -172,6 +176,14 @@ pub enum Response {
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
 pub struct Overview {
     /// Whether the overview is currently open.
+    pub is_open: bool,
+}
+
+/// Screenshot UI information.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+pub struct ScreenshotUi {
+    /// Whether the screenshot ui is currently open.
     pub is_open: bool,
 }
 
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -351,7 +351,9 @@ impl EventStreamStatePart for ScreenshotUiState {
         match event {
             Event::ScreenshotUiChanged { event } => match event {
                 ScreenshotUiEvent::Open => self.is_open = true,
-                ScreenshotUiEvent::Cancel | ScreenshotUiEvent::Confirm => self.is_open = false,
+                ScreenshotUiEvent::Cancel | ScreenshotUiEvent::Confirm { .. } => {
+                    self.is_open = false
+                }
             },
             event => return Some(event),
         }
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -9,7 +9,7 @@
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 
-use crate::{Cast, Event, KeyboardLayouts, Window, Workspace};
+use crate::{Cast, Event, KeyboardLayouts, ScreenshotUiEvent, Window, Workspace};
 
 /// Part of the state communicated via the event stream.
 pub trait EventStreamStatePart {
@@ -49,6 +49,9 @@ pub struct EventStreamState {
 
     /// State of screencasts.
     pub casts: CastsState,
+
+    /// State of the screenshot UI.
+    pub screenshot_ui: ScreenshotUiState,
 }
 
 /// The workspaces state communicated over the event stream.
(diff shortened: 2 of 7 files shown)


Atan-D-RP4:
Is it in scope to add something that gives the currently selected regions geometry?
I would really appreciate that, since I've been wanting it and looking do it alongside something like this for a while.

bczhc:
> Is it in scope to add something that gives the currently selected regions geometry? I would really appreciate that, since I've been wanting it and looking do it alongside something like this for a while.

Just including the coordinates of the selected area in the event won't bother much. I've added one patch to do this. It outputs something similar to `slurp` but is in physical size instead of logical size, like this:

```
{"ScreenshotUiChanged":{"event":{"Confirm":{"position":[2532,1407],"size":[27,32]}}}}
```

Do `niri msg action screenshot` and not pass the `--path`, niri won't save to a file. Since `niri msg action screenshot` seems to be asynchronous and for now we can only achieve this by listening to the event stream, I guess.

Atan-D-RP4:
That was fast, damn.
Good to know it was simple enough. This'll replace my slurp requirement for my screen-recorder utility.

Atan-D-RP4:
Damn that was fast. Thank you for that. 

Though I was really only wondering about if such a thing should go in here.
I'd discussed this with YaLTeR previously on the matrix chat, and he wasn't _too_ sold on the idea. But I suppose that can all wait for review (which will likely take a while, fair warning).

For now, this'll allow me to not depend on slurp for my screen-recorder utility, so yay for me.

bczhc:
There's one discussion, according to https://github.com/niri-wm/niri/discussions/1057#discussioncomment-11972800, I wonder if YaLTeR is planning to make `niri msg action` synchronous like, e.g. to do a screenshot and wait for the result, just do something like `niri msg action --stdout > a.png`. But obviously for now there's no way to do this. Listening to file creation via `inotifywait` or just spin-looping to test if the png file is present is a way. However these methods are still tricky for the interactive screenshot ("screenshot UI", or `niri msg action screenshot`) as it doesn't report a result so there's no way to know if the user canceled or confirmed.

One solution is just to use `grim` and maybe along with `slurp`. I'm coming from Hyprland and used to use `grim`.  There's an option `direct_scanout = 1` which doesn't play well with `grim` while gaming. I solved this using some tricky approaches like, in the screenshot utility script, first set direct_scanout = 0, then start a transparent layer to force the fullscreen app to exit the direct_scanout mode, then `grim`, then kill the layer app then restore direct_scanout = 1, then, etc., etc. If not doing so, the screenshot will grab an incomplete screenbuffer thus the image is incomplete. Under niri such issue is all gone. May thanks to the native screenshot functionality.

Atan-D-RP4:
> There's one discussion, according to [#1057 (comment)](https://github.com/niri-wm/niri/discussions/1057#discussioncomment-11972800), I wonder if YaLTeR is planning to make `niri msg action` synchronous like, e.g. to do a screenshot and wait for the result, just do something like `niri msg action --stdout > a.png`. But obviously for now there's no way to do this. Listening to file creation via `inotifywait` or just spin-looping to test if the png file is present is a way. However these methods are still tricky for the interactive screenshot ("screenshot UI", or `niri msg action screenshot`) as it doesn't report a result so there's no way to know if the user canceled or confirmed.
> 

YaLTeR has admitted that the IPC is bound to get more complex, as Niri grows. Especially to support some of the design ideas around gestures and gesture events. Though I don't know about transitioning anything from async to sync context. One thing that is definitely not going to change is that for the screenshot or it's UI, the shot itself will always be taken as soon as possible.

Instead of `inotifywait` and such, I much prefer just compositor events, the already implemented ones, and the ones implemented here.

> One solution is just to use `grim` and maybe along with `slurp`. I'm coming from Hyprland and used to use `grim`. There's an option `direct_scanout = 1` which doesn't play well with `grim` while gaming. I solved this using some tricky approaches like, in the screenshot utility script, first set direct_scanout = 0, then start a transparent layer to force the fullscreen app to exit the direct_scanout mode, then `grim`, then kill the layer app then restore direct_scanout = 1, then, etc., etc. If not doing so, the screenshot will grab an incomplete screenbuffer thus the image is incomplete. Under niri such issue is all gone. May thanks to the native screenshot functionality.

Well Niri also has this for direct_scanout:
```kdl
debug {
    disable-direct-scanout
}
```
Seems like Niri enables direct_scanout by default.

bczhc:
> One thing that is definitely not going to change is that for the screenshot or it's UI, the shot itself will always be taken as soon as possible.

Indeed. Interior screen buffer collection & UI dismiss when the user presses Space (or Esc) should be immediately done. Though afterward, niri needs to encode and compress the png, which consumes time, and after this, niri will send a notification to inform the success. That's why...

> YaLTeR has admitted that the IPC is bound to get more complex, as Niri grows. Especially to support some of the design ideas around gestures and gesture events. Though I don't know about transitioning anything from async to sync context.

...implementing a synchronous ipc call matters. A well working screenshot is important to me. Say I bind the screenshot as key Mod+S, and when I do Mod+S, it calls a script which grabs the screenshot first and then does some postprocessing with ffmpeg and exiftools to e.g. compress it more and write some png metadata. Relying on event msgs from niri ipc seems to be a more elegant way compared to keeping polling for file existence. This is how my script achieves the blocking version of niri screenshot, worked with `ScreenshotCaptured` event (this event exists before this PR).

```bash
niri_screenshot_blocked() {
    local fifo=$(mktemp -u)
    mkfifo "$fifo"

    niri msg -j event-stream | jq --unbuffered -r '.ScreenshotCaptured.path // empty' > "$fifo" &
    local listener_pid=$!

    niri msg action screenshot-screen -d true -p true --path "$1"

    local captured_path
    read -r captured_path < "$fifo"

    kill "$listener_pid" 2>/dev/null
    rm -f "$fifo"

    echo "$captured_path"
}
```

It works fine until.. If you press Mod+S in a fast way (to capture the fast mutated screen content), the script can't distinguish which ScreenshotCaptured is from which screenshot, resulting in bugs.

I did a patch on my custom niri: https://github.com/bczhc/niri/commit/0aa941f04d391089627798c74d0d510ed788a3fa adding an id field on `niri msg action screenshot[-screen]` call and once the screenshot saving is done, the passed `id` is outputted along with the relevant niri `ScreenshotCaptured` event. This makes my script able to distinguish multiple nearly-simultaneous screenshot events. After the hack, my script looks like this:

```bash
niri_screenshot_blocked() {
    local path="$1"
    local id
    id=$(uuidgen)
    local fifo="/tmp/${id}-result"

    mkfifo "$fifo"
    # niri msg action screenshot-screen [id] [path]
    niri msg action screenshot-screen $id -d true -p true --path "$path"

    read -r result < "$fifo"
    rm "$fifo"

    case "$result" in
        done)
            echo "Done"
            return 0
            ;;
        canceled)
            echo "Canceled"
            return 1
            ;;
        ignored)
            echo "Screenshot UI already shown"
            return 1
            ;;
    esac
}

niri_area_screenshot_blocked() {
    local path="$1"
    local id
    id=$(uuidgen)
    local fifo="/tmp/${id}-result"

    mkfifo "$fifo"
    # niri msg action screenshot [id] [path]
    niri msg action screenshot $id -p true --path "$path"

    read -r result < "$fifo"
    rm "$fifo"

    case "$result" in
        done)
            echo "Done"
            return 0
            ;;
        canceled)
            echo "Canceled"
            return 1
            ;;
        ignored)
            echo "Screenshot UI already shown"
            return 1
            ;;
    esac
}
```

again it's really a dirty way :).

> Well Niri also has this for direct_scanout

Then niri's direct scanout is implemented much better. Hyprland's direct scanout has *really* many bugs. All of the bugs I encountered with screenshot+direct scanout under Hyprland, do not exist under niri. So good.

bczhc:
> YaLTeR has admitted that the IPC is bound to get more complex, as Niri grows.

then there's possible ipc rework? actually this PR is just done in sake of current ipc completeness i.g. there's overview-state then why not screenshot-state. Not an actual workaround/solution itself for what I mean a "blocking version of niri screenshot ipc call".

Atan-D-RP4:
> > YaLTeR has admitted that the IPC is bound to get more complex, as Niri grows.
> 
> then there's possible ipc rework? actually this PR is just done in sake of current ipc completeness i.g. there's overview-state then why not screenshot-state. Not an actual workaround/solution itself for what I mean a "blocking version of niri screenshot ipc call".

No rework for the IPC at present. But it came up in a discussion about how I wanted to expose gesture parameters, such as fingers count, gesture progress and such things to the IPC or the config. 
Basically exposing parts of that state machine to allow users to make their own custom gesture actions instead of being limited to what Niri might provide. His comment on that, was that it required a lot of upfront well-thought out design, but also concern that it might become an easy pattern to throw other new things into the IPC instead doing the design work fully and properly. So it will definitely not come to anything until the gestures interface has been spec'd out.

Also, yeah, completeness in the IPC is nice.

YaLTeR:
I've glanced over the discussion here, and I still have a question: is this really a good solution?

I think we'll add some way to do a blocking `niri msg action screenshot/window/output` (who knows maybe select-area) in the future which will return the results right from the command, with no race or ambiguity. With the events in this PR however I think it's still racy and you can't reliably tell apart a screenshot you requested vs. something that may happen in parallel (however unlikely).

bczhc:
> I've glanced over the discussion here, and I still have a question: is this really a good solution?

No, absolutely.

> I think we'll add some way to do a blocking `niri msg action screenshot/window/output` (who knows maybe select-area) in the future which will return the results right from the command, with no race or ambiguity. With the events in this PR however I think it's still racy and you can't reliably tell apart a screenshot you requested vs. something that may happen in parallel (however unlikely).

That's exactly what I then talked about above. Even for myself, I didn't use the approach in this PR after finding it could still cause race condition/ambiguity. The ultimate solution is to make the ipc synchronous i.e. blocking until the screenshot done - we all know this; however this requires more work.

Issue #3240, opened 2026-01-12
transforms not applying on DisplayLink monitors


jusnoord:
I am experiencing issues similar to a comment on #2312 while using a DisplayLink monitor that has a 90 degree translation. I have quoted the comment below, as it perfectly matches the behavior I am experiencing; I don't believe this behavior was addressed in the issue, as I am on niri 25.11 (nixpkgs).  I am not sure how to troubleshoot this whatsoever, so any input would be helpful. 

Original Comment:
While this branch does allow niri to use my two display link monitors (USB-C dock connected to laptop), there are some problems with the one I use vertically (`transform "90"`)

The lower part of the screen does not update unless you move the mouse within it. To be specific, its the lower 840 pixels (1920 - 1080) which suggests that something is rendering to 1920x1080 rather than 1080x1920. I have worked around this by adding a bottom struct of 840 pixels on that monitor but I would really quite like to be able to use the bottom half of the screen.

I also get this when switching to the overview:
<img width="263" height="508" alt="image" src="https://github.com/user-attachments/assets/630dd623-516a-4c2c-bcbf-892e598d4b1a" />
Which is presumably an artifact of the same issue. Moving the mouse on that screen will cause the screen to update and show the overview correctly (which is why I had to resort to photographing the screen rather than a screenshot).

I have a second setup which doesn't use displaylink and has a vertical monitor and have no issues there.

_Originally posted by @Charlie-83 in https://github.com/YaLTeR/niri/issues/2312#issuecomment-3455982127_

Sempyos:
when it happens, can you provide the outputs of:
``dmesg``
``journalctl -e``

can you also try to replicate on cosmic-comp

Sempyos:
@jusnoord

Charlie-83:
@Sempyos 

I have tried cosmic and have a very similar issue.

I don't seem to have any `foot` service.

Here is my dmseg log

```
[drm] Initialized evdi 1.14.15 for evdi.0 on minor 0
evdi: [I] Evdi platform_device create
evdi: [I] Attaching to usb:6-1.2
evdi: [I] (card0) Opened by Task 1614 (.DisplayLinkMan) of process 1593 (.DisplayLinkMan)
evdi: [I] (card0) Added i2c adapter bus number 22
evdi: [I] (card0) Connected with Task 1622 (.DisplayLinkMan) of process 1593 (.DisplayLinkMan)
evdi: [I] (card0) Connector state: connected
[...]

PR #2891, opened 2025-11-25
fix: black screen when using displaylink

yuxqiu:
### Issue

I'm trying to use a Dell D6000 DisplayLink dock with an M1 Mac running Asahi Fedora. Whenever I connect the dock, my Niri session immediately goes black and the compositor crashes.

The crash produces a `journalctl` log essentially identical to https://github.com/YaLTeR/niri/issues/2330, ending with a panic:

```
thread 'main' panicked at src/backend/tty.rs:830:13:
assertion failed: self.dmabuf_global.replace(dmabuf_global).is_none()
```

Relevant excerpts from the log:

```
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.686929Z INFO niri: starting version 25.08 (0.0.git.2556.45b45ac2)
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.702237Z DEBUG niri_config: loaded config from "/home/yuxqiu/.config/niri/config.kdl"
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.715484Z DEBUG niri::backend::tty: attempting to use render node from config: "/dev/dri/renderD128"
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.715536Z INFO niri::backend::tty: using as the render node: "/dev/dri/renderD128"
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.735952Z DEBUG niri::backend::tty: adding device: 57857 "/dev/dri/card1"
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.735965Z DEBUG niri::backend::tty: this is the primary node
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.736208Z WARN niri::backend::tty: error adding primary node device, display-only devices may not work: DRM access err
or: Error loading resource handles on device `Some("/dev/dri/card1")` (Operation not supported (os error 95))
Nov 25 14:22:14 pc niri[5443]: Caused by:
Nov 25 14:22:14 pc niri[5443]: Operation not supported (os error 95)
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.736220Z DEBUG niri::backend::tty: adding device: 57856 "/dev/dri/card0"
Nov 25 14:22:14 pc niri[5443]: 2025-11-25T06:22:14.754921Z DEBUG niri::backend::tty: this is the primary render node
[...]

PR #3668, opened 2026-03-22
fix: force linear buffers for display-only devices (e.g. DisplayLink)

yjpark:
A few notes for this patch:

- My displaylink monitor is an pretty old Azus ZenScreen 15inch, it worked out of box on Cosmic.
- My main GPU is an old AMD one
- When I switch to Niri, it's been powered off
- It's still not working after I include #2312
- Solved the bug with the help of Claude Code, full commit logs at https://github.com/yjpark/niri/commits/wip-display-port/
  - It first make the moniter light up, but having tiling issues
  - Later it identified it's related to AMD GPU, and add a couple extra patch
  - Everything worked as expected.

I'm running NixOS, using the fix for a few weeks, haven't met any issue. rebased to main, and cleanup a bit for this PR, using it for a few hours, so far so good.

https://github.com/yjpark/dotflakes/blob/979572b178eaeb77629e84f671e00fea91f90cfa/flake.nix#L51

The changes seems straight forward to me, though to be honest, I don't know whether it will cause side effects on other configurations.

Created this PR, hoping it can be reviewed by someone can fully understand, if can be mered, can save me some time for future merges, also might help other persons met similar issues.

---

Display-only devices (no render node) rely on the primary GPU for rendering. Some GPUs (e.g. AMD Polaris) report only Modifier::Invalid via EGL, which lets the driver pick tiled layouts that display-only adapters cannot scan out.

Fix by:
- Adding GbmBufferFlags::LINEAR to the allocator for display-only devices, ensuring GBM produces linear buffers even with implicit modifiers
- Accepting Modifier::Invalid alongside Linear in format filtering and dmabuf feedback for display-only/cross-device paths, so the format set is not empty on GPUs that only report Invalid

Follows up on #2312 and #2891.
Related to #3240.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -872,7 +872,13 @@ impl Tty {
         } else {
             bail!("no allocator available for device");
         };
-        let gbm_flags = GbmBufferFlags::RENDERING | GbmBufferFlags::SCANOUT;
+        let mut gbm_flags = GbmBufferFlags::RENDERING | GbmBufferFlags::SCANOUT;
+        // Display-only devices (e.g. DisplayLink) require linear buffers. Without
+        // this flag, Modifier::Invalid lets the GPU driver pick a tiled layout
+        // (e.g. micro/macro-tiled on AMD Polaris) that the adapter cannot scan out.
+        if render_node.is_none() {
+            gbm_flags |= GbmBufferFlags::LINEAR;
+        }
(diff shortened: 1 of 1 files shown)


valpackett:
Bluntly hardcoding "display-only == needs linear" is wrong. Most SoC vendors combine totally independent GPU (render-only) and DPU (display-only) nodes but both support a common compression method like AFBC.

PR #3678, opened 2026-03-23
tty: auto-select render node based on connected external outputs

### LINKED ISSUE #3674: External monitor lag on Intel+Nvidia hybrid laptops (render-drm-device workaround not documented)
On laptops with hybrid Intel+Nvidia graphics where the HDMI port is physically wired to the Nvidia GPU, niri renders on the Intel GPU by default, causing severe lag/stuttering on the external monitor.

**System info:**
- niri 25.11 (b35bcae)
- CPU: Intel Core i7-9750H
- GPU: Intel UHD 630 (iGPU) + NVIDIA GeForce RTX 2060 Mobile (dGPU)
- OS: Arch Linux
- External monitor: HDMI-A-1 @ 2560x1440@144Hz

## Steps to reproduce

1. Connect an external monitor via HDMI on a hybrid Intel+Nvidia laptop (where HDMI is wired to the Nvidia GPU)
2. Start niri -> niri picks `renderD128` (Intel) by default
3. Observe severe lag/stuttering on the external monitor

## Why its happening

Niri renders on Intel (`renderD128`) but the HDMI output is physically on the Nvidia GPU (`renderD129`). Every frame gets copied across GPUs before reaching the screen that cross-GPU copy is the source of the lag (i guess)

## Workaround

Adding this to `~/.config/niri/config.kdl` fixes it completely:

```kdl
debug {
    wait-for-frame-completion-before-queueing
}
```

Where `renderD129` is the Nvidia render node.

## The problem with the workaround

- `render-drm-device` is buried inside the `debug` block with no mention in the wiki's hybrid GPU or multi-monitor sections
- The option name gives no hint that it solves external monitor lag
- Users hitting this issue have no obvious path to discovery the symptom looks like a driver or Wayland issue, not a render device selection issue

raihanadf:
Closes #3674.

On hybrid GPU laptops where external ports are wired to the dGPU, niri picks the iGPU (boot_vga) by default. This means every frame gets copied across GPUs before reaching the external monitor, causing bad lag.

This adds a fallback between `render-drm-device` config and `udev::primary_gpu`: scan `/sys/class/drm/` for connected non-internal connectors and prefer the GPU that owns one. `render-drm-device` still takes full precedence.

**Testing:**

On an Intel+Nvidia hybrid laptop with HDMI wired to the dGPU, start niri without `render-drm-device` set. You should see in logs:

```
auto-selected render node Some("/dev/dri/renderD129") based on connected external output "card0-HDMI-A-1"
```

External monitor should work without lag.

**Limitations:**
- Only runs at startup, so plugging in after boot won't trigger it
- Multiple GPUs each with external monitors: picks first alphabetically

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -460,6 +460,7 @@ impl Tty {
 
         let (primary_node, primary_render_node) = primary_node_from_config(&config.borrow())
             .ok_or(())
+            .or_else(|()| primary_node_from_active_outputs().ok_or(()))
             .or_else(|()| {
                 let primary_gpu_path = udev::primary_gpu(&seat_name)
                     .context("error getting the primary GPU")?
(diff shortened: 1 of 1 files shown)


YaLTeR:
The current default works as intended. It doesn't lag on most systems. Rendering on the dGPU causes massively more battery drain than on the iGPU. It's a trade off. Other compositors don't auto-default to dGPU either (if I am not mistaken).

raihanadf:
Ahh alright, so I need to use this everytime I have external monitor plugged in?
``` 
debug {
    wait-for-frame-completion-before-queueing
} 
```

YaLTeR:
Use it if you want to use dGPU for rendering

PR #3699, opened 2026-03-26
Add spawn-sh support to switch-events

neunato:
Inspired by https://github.com/niri-wm/niri/issues/3605

Adds `spawn-sh` as an alternative to `spawn` in `switch-events` blocks. Exactly one of the two must be specified - empty blocks and using both together are parse errors.

I’m not very familiar with Rust, but I tried my best to follow the existing code style. Happy to make changes if anything looks off.

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -13,7 +13,7 @@ use smithay::input::keyboard::xkb::{keysym_from_name, KEYSYM_CASE_INSENSITIVE, K
 use smithay::input::keyboard::Keysym;
 
 use crate::recent_windows::{MruDirection, MruFilter, MruScope};
-use crate::utils::{expect_only_children, MergeWith};
+use crate::utils::{expect_only_children, parse_arg_node, MergeWith};
 
 #[derive(Debug, Default, PartialEq)]
 pub struct Binds(pub Vec<Bind>);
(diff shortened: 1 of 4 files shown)


niri-config/src/binds.rs
```diff
@@ -90,10 +90,72 @@ impl MergeWith<SwitchBinds> for SwitchBinds {
...
+#[derive(Debug, Clone, PartialEq)]
 pub struct SwitchAction {
-    #[knuffel(child, unwrap(arguments))]
-    pub spawn: Vec<String>,
+    pub spawn: Option<Vec<String>>,
+    pub spawn_sh: Option<String>,
+}
+
+impl<S: knuffel::traits::ErrorSpan> knuffel::Decode<S> for SwitchAction {
+    fn decode_node(
```
HigherOrderLogic, line 99: Maybe just call out to the keybind section action binding and invalidate the unavailable ones instead of repeating the code.

neunato: Makes sense. Please check the pushed changes

niri-config/src/binds.rs
```diff
@@ -90,10 +90,53 @@ impl MergeWith<SwitchBinds> for SwitchBinds {
...
+            }
+        }
+
+        if action.is_none() {
+            ctx.emit_error(DecodeError::missing(node, "expected `spawn` or `spawn-sh`"));
+        }
+
+        Ok(SwitchAction {
+            action: action.unwrap_or(Action::Spawn(vec![])),
+        })
```
HigherOrderLogic: You should return an `Err` instead if no action is specified.

neunato: Fixed

YaLTeR:
Small UX problem. Using Action decode means that mistakes will cause it to suggest all other actions (that are not allowed in switch events):

image

What should we do?

neunato:
I pushed some changes, please see if you like them.

1. errors go through `ctx` and the function always returns `Ok`, so a broken switch event block doesn't abort parsing and all config errors are reported at once

2. check child node name prior to calling `Action::decode_node`, fixing the error message suggesting all other actions

3. I matched the pattern in `Bind::decode_node` to use `children.next()` to take the first child, parse it, then drain remaining children for the "only one action allowed" errors

Also fixed the error ordering in `Bind::decode_node` - when a keybind had both an invalid action and extra actions after it, the "only one action allowed" errors were emitted before the error from parsing the first action even though it appears earlier in the config. But that can be dropped or moved to a separate PR if preferred.

Vescrity:
I'm curious about why not support all other actions here

YaLTeR:
Less annoying in the code + why would you want to bind smth else there

Vescrity:
> why would you want to bind smth else there

```
    lid-close { spawn "bash" "-c" "niri msg action power-off-monitors"; }
    lid-open { spawn "bash" "-c" "niri msg action power-on-monitors";  }
```

So I have something like this, and it does work though.

I haven't dived into the code, so I guess that making it accept any action would be easier than making it only accept `spawn`.

PR #3705, opened 2026-03-26
Add cgroup window rule matcher

Polychoronix:
This allows for applying window rules by cgroup, which is useful when spawning windows into their own systemd slices, e.g. for security reasons.
This requires reading `/proc/<pid>/cgroup`, which takes around 5μs per rule evaluation.
The PID is available for `Mapped` windows only, so any opening properties unfortunately won't work.

An example config, highlighting matching windows in red:
```kdl
window-rule {
    match cgroup="my-slice"
    border {
        active-color "#f38ba8"
        inactive-color "#7d0d2d"
    }
}
```
This can be tested by running a shell in another slice, then spawning a window:
```shell
systemd-run --user --slice=my-slice --shell
kitty &
```
The new window should have the styling applied.

### DIFF
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -30,6 +30,7 @@ Here are all matchers and properties that a window rule could have:
 window-rule {
     match title="Firefox"
     match app-id="Alacritty"
+    match cgroup="browsing"
     match is-active=true
     match is-focused=false
     match is-active-in-column=true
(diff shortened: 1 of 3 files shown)


Atan-D-RP4:
With such a matcher, does it make sense to have something like an `open-with-cgroup "my-slice"` option as well?

Polychoronix:
> With such a matcher, does it make sense to have something like an `open-with-cgroup "my-slice"` option as well?

For theming it could be useful, but the primary goal for me is using `systemd-run` with some custom options via my launcher. Of course, you can use that via keybind already.

PR #3712, opened 2026-03-27
add prevent-idle-inhibit option

MithicSpirit:


src/niri.rs
```diff
@@ -3908,6 +3908,11 @@ impl Niri {
     }
 
     pub fn refresh_idle_inhibit(&mut self) {
+        if self.config.borrow().prevent_idle_inhibit {
```
Atan-D-RP4, line 3996: This block should definitely go after the tracy span.

HigherOrderLogic:
Seems to be somewhat overlapped with #2373.

MithicSpirit:
This does the opposite. That PR makes it so you can force a client into becoming an inhibitor if it isn't one already. This PR makes it so you can prevent clients from becoming (Wayland) inhibitors in the first place. I guess in principle this would be better as a window rule rather than a global switch, but I don't need that. If there's enough interest in that, I think I can add it.

PR #3731, opened 2026-03-29
Add screenshot-tile action

bennypowers:
## Summary

- Adds a new `screenshot-tile` action that captures a window with its compositor-drawn decorations (border, focus ring, shadow, rounded corner clipping) and alpha transparency
- Mirrors the existing `screenshot-window` action's options (`write-to-disk`, `show-pointer`, `--id`, `path`)
- Consistent with the existing window-tile distinction in IPC (e.g. `tile_size` vs `window_size`)

### Usage

Config keybind:
```
screenshot-tile
```

IPC:
```
niri msg screenshot-tile
```

## Test plan

- [ ] `screenshot-window` behavior is unchanged
- [ ] `screenshot-tile` includes border when border is configured
- [ ] `screenshot-tile` includes focus ring for focused windows
- [ ] `screenshot-tile` includes shadow when shadow is enabled
- [ ] `screenshot-tile` includes rounded corner clipping with alpha transparency
- [ ] `screenshot-tile --id` variant works
- [ ] Fullscreen windows skip border/shadow as expected (matching visual state)
- [ ] Falls back to bare window when tile is not found (e.g. during drag-and-drop)

Ref: https://github.com/niri-wm/niri/discussions/3283


YaLTeR:
I think instead of a flag, it should be a separate action screenshot-tile. We have the window-tile distinction elsewhere in IPC so it would be consistent.

bennypowers:
Пожалуйста

PR #3736, opened 2026-03-29
Better config error notification

haug1:
- show actual config parse errors in the notification instead of only a generic failure message.
- keep notification open until dismissed so they can be read. (space/enter/mouseclick to dismiss)
- use `miette` to format errors better

image

### DIFF
--- a/niri-config/src/error.rs
+++ b/niri-config/src/error.rs
@@ -2,7 +2,7 @@ use std::error::Error;
 use std::fmt;
 use std::path::PathBuf;
 
-use miette::Diagnostic;
+use miette::{Diagnostic, NarratableReportHandler};
 
 #[derive(Debug)]
 pub struct ConfigParseResult<T, E> {
@@ -42,6 +42,14 @@ impl<T, E> ConfigParseResult<T, E> {
             includes: self.includes,
         }
     }
+}
+
+pub fn format_error_report(err: &miette::Report) -> String {
+    let mut report = String::new();
+    NarratableReportHandler::new()
+        .render_report(&mut report, err.as_ref())
+        .expect("writing into a String should never fail");
+    report.trim_end().to_owned()
 }
 
 impl fmt::Display for ConfigIncludeError {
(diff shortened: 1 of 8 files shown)


Uzlkav:
Hi, I just tested this. A few things:
- The new popup consumes all input, but not all input closes the dialog. This is confusing and intrusive imo.
- There's a lot of repeated information, eg for a nonexistant node:
```
Failed to parse the config file. Run niri validate for the full error report.
error loading config
    Diagnostic severity: error
    Caused by: error parsing
    Caused by: error parsing KDL

Error: failed to parse included config
    Diagnostic severity: error

Begin snippet for config.kdl starting at line 51, column 1

snippet line 51:
snippet line 52: include "./test-config.kdl"
    label at line 52, columns 1 to 7: node starts here
Error: error parsing KDL
    Diagnostic severity: error

Error: unexpected node 'crazy-node’
    Diagnostic severity: error

Begin snippet for test-config.kdl starting at line 3, column 1

snippet line 3: // Error message notifications (3736)
snippet line 4: crazy-node {
    label starting at line 4, column 1: unexpected node
snippet line 5: non-existant-option true
snippet line 6: }
snippet line 7:
    label ending at line 7, column ©: unexpected node
snippet line 8: // Blur
```
- Diagnostic severity is repeated 4 times, and it's kind of difficult to parse the actual problematic code with the `snippet line X:` preceding each bit
- We already have an error message from `niri validate` that is (in my opinion) more readable and clearer. Here's the same error, as displayed by `niri validate`:
```
Error:   × error loading config
  ├─▶ error parsing
  ╰─▶ error parsing KDL

Error:   × failed to parse included config
    ╭─[config.kdl:51:1]
 51 │
 52 │ include "./test-config.kdl"
    · ───┬───
    ·    ╰── node starts here
    ╰────
Error:   × error parsing KDL

Error:   × unexpected node `crazy-node`
   ╭─[test-config.kdl:3:1]
 3 │     //      Error message notifications (3736)
 4 │ ╭─▶ crazy-node {
 5 │ │       non-existant-option true
 6 │ ├─▶ }
   · ╰──── unexpected node
 7 │
 8 │     //      Blur
   ╰────
```

So I think it's worth considering either changing the format of the displayed error, or just mirroring/pulling from niri validate. Dms' niri integration does this, which looks like this:

err-msg-dms


It's also not expanded by default, just saying `niri: failed to load config`, and you click on the arrow to get the above view, which makes it even less intrusive. It has a timeout, but if you click on it it persists until you explicitly close.

PR #3738, opened 2026-03-30
Add drm-lease output config option

PandorasFox:
Allow forcing an output to be offered for DRM leasing via a new `drm-lease` config flag. This serves as a workaround for drivers (notably NVIDIA) that fail to set the `non-desktop` DRM connector property on VR headsets and similar displays.

---

my rtx 3090 + valve index does DRM leasing fine under sway/kwin (with worse perf/weird display flickers when the lease starts/ends), after the HMD has been configured as 'disabled' in sway/kwin. 

yet another case of the nvidia drivers being of bad quality. I couldn't see an appropriate way to handle this at the Smithay layer, as I think that'd more or less require mucking with EDID data, so I think this is the cleanest way to simply mark the device as being DRM-leasable.


PandorasFox:
Any chance of this getting merged / feedback for this? I know that workarounds for inadequate driver/display self-reporting isn't ideal, but I think this is a reasonably minimal escape hatch.

I've been using this flag to enable leasing my Index HMD w/ steamVR for the past few months with no issues. The index HMD generally does _not_ report as a non-desktop display (for some reason - I think newer ones do, but my launch-day one does not), and will even eat up my BIOS display output depending on how I order my displayport cables.

I'm not entirely sure if it's even an nvidia issue since that used to happen with my old Vega card before it died. Just also seems like an issue that "should" best be fixed in how the nvidia driver reports the displays, but likely never will be. I can update the docs somewhat to reflect that, if that would be preferable. I haven't really seen any other reports of this issue (so it maybe is just my headset misbehaving)

PandorasFox:
Did a bit more checking: the Index HMD has this prop handled via a quirks table in `drm_edid.c`, and from what I can tell, _only_ the nvidia driver fails to honor that table - my recollection on my old card was simply from before the index HMD quirks entries were mainlined.

So, this is just an escape hatch for the nvidia drivers (both nvidia-open leveraging the GSP, and the older nvidia driver, but not nouveau).

YaLTeR:
How does it work (or doesn't work) on other compositors? E.g. GNOME, KDE, sway

PandorasFox:
> How does it work (or doesn't work) on other compositors? E.g. GNOME, KDE, sway

Varied. In KDE, the graphical config options were insufficient - the headset wasn't available for leasing as it was always modeset. My memory is a bit hazy on what I did back in march, but I had to something distinctly janky under the hood to get kwin to properly disable the HMD in its stack (either editing some not-meant-for-human-editing config file, or a poorly documented env var for my KDE session - nothing in the proper kscreen GUI settings).

In sway, `output {name} disable` was sufficient as it left that output fully unhandled/un-initialized. `output off` in Sway functioned similarly to `off` in Niri (which is just dpms off iirc?). Last I checked, sway doesn't attempt to handle non-desktop displays and simply provides the `disable` functionality for them, leaving it entirely manual.

the escape hatch I added is more or less the same as the `disable` path in sway, just named appropriately for sitting alongside the non-desktop display prop path.

PR #3745, opened 2026-03-31
input: add tablet stylus button triggers and binds

youssefadly237:
Add support for tablet stylus button bindings

**Example**

```kdl
binds {
    Super+TabletStylusPrimary { spawn-sh "some-script.sh"; } 
    TabletStylusSecondary { toggle-overview; }
}
```

**Notes**

I did test with Wacom CTL-472, but it should work on all tablets

Button codes are `BTN_STYLUS`, `BTN_STYLUS2` [`include/uapi/linux/input-event-codes.h`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/input-event-codes.h) lines 414-L415 as of `4bdb6a1659873dfc625ded30a68cb387e328f34e`

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -52,6 +52,9 @@ pub enum Trigger {
     TouchpadScrollUp,
     TouchpadScrollLeft,
     TouchpadScrollRight,
+    TabletStylusButton1,
+    TabletStylusButton2,
+    TabletStylusButton3,
 }
 
 bitflags! {
@@ -1000,6 +1003,12 @@ impl FromStr for Key {
             Trigger::TouchpadScrollLeft
         } else if key.eq_ignore_ascii_case("TouchpadScrollRight") {
             Trigger::TouchpadScrollRight
+        } else if key.eq_ignore_ascii_case("TabletStylusButton1") {
+            Trigger::TabletStylusButton1
+        } else if key.eq_ignore_ascii_case("TabletStylusButton2") {
+            Trigger::TabletStylusButton2
+        } else if key.eq_ignore_ascii_case("TabletStylusButton3") {
+            Trigger::TabletStylusButton3
         } else {
             let mut keysym = keysym_from_name(key, KEYSYM_CASE_INSENSITIVE);
             // The keyboard event handling code can receive either
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -51,6 +51,8 @@ pub enum Trigger {
     TouchpadScrollUp,
     TouchpadScrollLeft,
     TouchpadScrollRight,
+    TabletStylusPrimary,
+    TabletStylusSecondary,
 }
 
 bitflags! {
@@ -1012,6 +1014,10 @@ impl FromStr for Key {
             Trigger::TouchpadScrollLeft
         } else if key.eq_ignore_ascii_case("TouchpadScrollRight") {
             Trigger::TouchpadScrollRight
+        } else if key.eq_ignore_ascii_case("TabletStylusPrimary") {
+            Trigger::TabletStylusPrimary
+        } else if key.eq_ignore_ascii_case("TabletStylusSecondary") {
+            Trigger::TabletStylusSecondary
         } else {
             let mut keysym = keysym_from_name(key, KEYSYM_CASE_INSENSITIVE);
             // The keyboard event handling code can receive either
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -52,8 +52,9 @@ pub enum Trigger {
     TouchpadScrollUp,
     TouchpadScrollLeft,
     TouchpadScrollRight,
-    TabletStylusPrimary,
-    TabletStylusSecondary,
+    TabletStylusButton1,
+    TabletStylusButton2,
+    TabletStylusButton3,
 }
 
 bitflags! {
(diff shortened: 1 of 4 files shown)


niri-config/src/binds.rs
```diff
@@ -52,6 +52,8 @@ pub enum Trigger {
     TouchpadScrollUp,
     TouchpadScrollLeft,
     TouchpadScrollRight,
+    TabletStylusPrimary,
+    TabletStylusSecondary,
```
YaLTeR: Where do the primary and secondary names come from? Are they commonly used? What do other compositors use?

src/input/mod.rs
```diff
@@ -3726,11 +3726,38 @@ impl State {
     }
 
     fn on_tablet_tool_button<I: InputBackend>(&mut self, event: I::TabletToolButtonEvent) {
+        const BTN_STYLUS: u32 = 0x14b;
+        const BTN_STYLUS2: u32 = 0x14c;
```
YaLTeR, line 3783: What about STYLUS3? Libinput seems to reference it everywhere along with STYLUS and STYLUS2: https://gitlab.freedesktop.org/search?search=stylus2&nav_source=navbar&project_id=147&group_id=1723&search_code=true&repository_ref=main

src/input/mod.rs
```diff
@@ -3726,11 +3726,38 @@ impl State {
...
+            let button = event.button();
+            let trigger = match button {
+                BTN_STYLUS => Some(Trigger::TabletStylusPrimary),
+                BTN_STYLUS2 => Some(Trigger::TabletStylusSecondary),
+                _ => None,
+            };
+
+            if let Some(trigger) = trigger {
+                if event.button_state() == ButtonState::Pressed {
+                    let mod_key = self.backend.mod_key(&self.niri.config.borrow());
```
YaLTeR, line 3803: I think you're forgetting to check `mods_with_tablet_stylus_binds()` here (like in other binds like mouse buttons etc.)

src/input/mod.rs
```diff
@@ -3726,11 +3726,38 @@ impl State {
...
+                    let mods = self.niri.seat.get_keyboard().unwrap().modifier_state();
+
+                    let bind = {
+                        let config = self.niri.config.borrow();
+                        let bindings = config.binds.0.iter();
+                        find_configured_bind(bindings, mod_key, trigger, mods)
+                    };
+                    if let Some(bind) = bind {
+                        self.handle_bind(bind.clone());
```
YaLTeR: When niri handles a press it must suppress the release so that Wayland clients don't receive that release. See how it's handled for mouse buttons.

src/niri.rs
```diff
@@ -2395,6 +2396,7 @@ impl Niri {
         let mods_with_mouse_binds = mods_with_mouse_binds(mod_key, &config_.binds);
         let mods_with_wheel_binds = mods_with_wheel_binds(mod_key, &config_.binds);
         let mods_with_finger_scroll_binds = mods_with_finger_scroll_binds(mod_key, &config_.binds);
+        let mods_with_tablet_stylus_binds = mods_with_tablet_stylus_binds(mod_key, &config_.binds);
```
YaLTeR, line 2413: If I recall correctly, there was a second place where these were updated.

YaLTeR (changes requested):
Here are some comments. Maybe there's a chance this can make it into the release. :)

Thanks for the link to linux uapi include btw, that was helpful to double-check.

niri-config/src/binds.rs
youssefadly237: I don't think it is common convention I was thinking mouse primary and secondary so I went with the same here but I guess `StylusButton1`, `StylusButton2` should be better

I check hyprland, and sway, did not find relevant naming

src/input/mod.rs
youssefadly237: I don't have it myself but yeah it is a button

I guess it is available in a models like Wacom Pro Pen 3D (model KP505), but I don't have it to test it

I did not add right away because I was not sure

```hpp
		if (wacom_wac->features.quirks & WACOM_QUIRK_PEN_BUTTON3) {
			int sw_state = wacom_wac->hid_data.barrelswitch |
				       (wacom_wac->hid_data.barrelswitch2 << 1);
			wacom_wac->hid_data.barrelswitch = sw_state == 1;
			wacom_wac->hid_data.barrelswitch2 = sw_state == 2;
			wacom_wac->hid_data.barrelswitch3 = sw_state == 3;
		}
```

[drivers/hid/wacom_wac.c](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/hid/wacom_wac.c)

`BTN_STYLUS`, `BTN_STYLUS2` are physically impossible to be pressed at the same time without breaking the button (mutual exclusive lever)

`BTN_STYLUS3` is just both I guess, should I add it?

YaLTeR: Let's add it yeah. And call the three buttons TabletStylusButton1, 2 and 3 I guess.

youssefadly237:
could be relevant but I don't have the hardware to test it

## Pad keys (Wacom M Pad)

| Physical key        | Code (int) | Code (hex) | Name    |
| ------------------- | ---------- | ---------- | ------- |
| Key 0               | 256        | 0x100      | `BTN_0` |
| Key 1               | 257        | 0x101      | `BTN_1` |
| Key 2               | 258        | 0x102      | `BTN_2` |
| Key 3               | 259        | 0x103      | `BTN_3` |
| Key 4               | 260        | 0x104      | `BTN_4` |
| Key 5               | 261        | 0x105      | `BTN_5` |
| Key 6               | 262        | 0x106      | `BTN_6` |
| Key 7               | 263        | 0x107      | `BTN_7` |
| Key 8 (ring center) | 264        | 0x108      | `BTN_8` |

image

upper column is 0,1,2,3
bottom column is 4,5,6,7
ring center is 8

there is a touch ring around the center button is `ABS_WHEEL` I guess

pad buttons are almost standerd for wacom devices but it varies, some other companies just send keyboard or mouse events instead, [drivers/hid/wacom_wac.c](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/hid/wacom_wac.c)

UnskilledWolf:
I just tested this on my Wacom Intuos Pro (PTH-451) and both stylus button bindings work as expected. My tablet also does only have two buttons on the stylus, so I was not able to try `TabletStylusButton3`.

I also tried implementing bindings for the other keys on the tablet, based on the key codes above. However, I could not get these to trigger. I am not very familiar with rust, so I may have missed something. [My attempt](https://github.com/UnskilledWolf/niri/tree/tablet-stylus-binds) is published as a fork of this PR's branch in case anyone would like to take a look.

youssefadly237:
yeah I checked your fork, I don't think this is the way to implement it, because you are treating the pad buttons as if they were stylus buttons, after reconsidering, I guess this is out of the scope of this pr, I don't have a tablet with pad buttons to test it, but I will try to implement it and message you on element x for testing?

youssefadly237:
pad button support is not yet implemented in Smithay

Related:
- #2536
- Smithay/smithay#1955
- Smithay/smithay#1836

YaLTeR:
Added a missing screenshot UI check. Otherwise looks good. Wasn't able to test because my tablet doesn't seem to report any tool button presses.

PR #3773, opened 2026-04-06
prevent sleeping after resuming from suspend

### LINKED ISSUE #2233: Waking laptop with power button immediately puts it in suspend again
Howdy,

When running niri, after suspending my laptop (due to idle or explicitly calling `systemctl suspend`), if I press the power button, it will briefly wake up (for < 1 second) before going back into suspend. It seems that niri must be registering the power button event (that woke the laptop up) and immediately putting the laptop back to sleep again.

A viable workaround is setting `disable-power-key-handling` in the niri config file. This prevents this bug, but it also means that the power button no longer works for putting the laptop to sleep.

This occurs with the default config.

### System Information

Framework Laptop

* Distro: Arch Linux

* GPU: (Integrated graphics) Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] 

* CPU: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz

Whenever you get around to it, please let me know if there's any additional information that could help.

Thanks!

EvansJahja:
The same power button is used both for suspend and for resuming, which causes devices to immediately sleep after suspending.

~~By adding inhibition to suspend after certain time (hardcoded 3 seconds for now) using the monotonic timer (which doesn't tick during suspend), we can prevent the suspend-resume loop~~

We now check login1 Manager's [PreparingForSleep property](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.login1.html) on dbus right before suspending. Unfortunately the event doesn't emit any signal so we cannot just watch it

Resolves #2233

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -2906,11 +2906,10 @@ fn suspend() -> anyhow::Result<()> {
 
     let manager = zbus::names::InterfaceName::try_from("org.freedesktop.login1.Manager").unwrap();
 
-    let is_preparing_sleep: bool = proxy.get(manager, "PreparingForSleep")
-    .context("Fail to get PreparingForSleep property from login1.Manager")
-    .and_then(|value| Ok(bool::try_from(value)?))?
-    ;
-
+    let is_preparing_sleep: bool = proxy
+        .get(manager, "PreparingForSleep")
+        .context("Fail to get PreparingForSleep property from login1.Manager")
+        .and_then(|value| Ok(bool::try_from(value)?))?;
 
     if is_preparing_sleep {
         warn!("Refusing to sleep, I just woke up (or going to sleep)");
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -2898,14 +2898,32 @@ fn refresh_interval(mode: DrmMode) -> Duration {
 fn suspend() -> anyhow::Result<()> {
     let conn = zbus::blocking::Connection::system().context("error connecting to system bus")?;
 
-    conn.call_method(
-        Some("org.freedesktop.login1"),
+    let proxy = zbus::blocking::fdo::PropertiesProxy::new(
+        &conn,
+        "org.freedesktop.login1",
         "/org/freedesktop/login1",
-        Some("org.freedesktop.login1.Manager"),
-        "Suspend",
-        &(true),
-    )
-    .context("error suspending")?;
+    )?;
+
+    let manager = zbus::names::InterfaceName::try_from("org.freedesktop.login1.Manager").unwrap();
+
+    let is_preparing_sleep: bool = proxy.get(manager, "PreparingForSleep")
+    .context("Fail to get PreparingForSleep property from login1.Manager")
+    .and_then(|value| Ok(bool::try_from(value)?))?
+    ;
+
+
+    if is_preparing_sleep {
+        warn!("Refusing to sleep, I just woke up (or going to sleep)");
+    } else {
+        conn.call_method(
+            Some("org.freedesktop.login1"),
+            "/org/freedesktop/login1",
+            Some("org.freedesktop.login1.Manager"),
+            "Suspend",
+            &(true),
+        )
+        .context("error suspending")?;
+    }
 
     Ok(())
 }


EvansJahja:
Note: I wrote the patch targetting niri 25.11 which worked but haven't tested if it works on main, and my potato device is too slow to build, will update with some testing. I think this PR is rather rudimentary and should be polished further. Let me know if this is the right direction.

src/backend/tty.rs
```diff
@@ -1988,11 +1991,30 @@ impl Tty {
...
 
-    pub fn suspend(&self) {
+    pub fn suspend(&mut self) {
         #[cfg(feature = "dbus")]
-        if let Err(err) = suspend() {
-            warn!("error suspending: {err:?}");
+        {
+            let suspend_inhibit = if let Some(dur) = self.last_suspend_time {
+                warn!("last_suspend_time is {:?}", dur);
+                get_monotonic_time().saturating_sub(dur) < Duration::from_secs(3)
```
HigherOrderLogic: This probably shouldnt be hardcoded, or a saner default value should be chosen.

EvansJahja: Sure this is a placeholder value.

Sane value: let's use what logind is using. logind.conf has HoldoffTimeoutSec set to 30s (it's for lid event but perhaps it's good enough)

please let me hardcode it for now as I only have a weak device making recompilation difficult (feel free to commit directly on the PR)

EvansJahja: 30s
3ba44f108fb90343dcb9ac6d8bb326a9574c562e

HigherOrderLogic: I still dont think this is ideal. What if user open and close the lid real quick?

The logic here should probably be using some other mechanism and not relying on pure timing.

EvansJahja: my defense for the logic is that it works, it's pretty much this is what logind.conf use, and using monotonic time is portable.

I have tried mashing the power button on both suspend and resume (btw it's power button, not lid) and it resolves the original issue. In my understanding we have to rely on timing because the power button event is received by the kernel anyway on startup and we have to ignore it.

Can you describe what's the failure mode you're thinking?

EvansJahja: After more testing, 3s is a sane duration as this is for power button, but I guess it should be up to configuration after all.

Any recommendation on where to put this in the config?  (and should it be on the same PR?)

EvansJahja: @HigherOrderLogic Okay, answering my own doubt:

There's indeed a flaw with this fix. This works if the sleep was triggered by niri.

If however, the sleep was triggered by other mechanism (systemctl suspend, etc) then the bug persists. So I think this PR solves most of the use case but not all...

I still think we still have to rely on time, but we should have "time since last wake up" instead of just monotonic time. I wonder if you have any suggestion on how we can do that?

HigherOrderLogic: I'm not sure. I was just pointing out that the current approach is not perfect. I dont have any particular fix for this issue in mind.

EvansJahja: idea:
we can react to suspend/resume using [systemd inhibitor lock](https://systemd.io/INHIBITOR_LOCKS)

Obviously it will only work under systemd and we still need to consider cases where we're not using systemd (nor linux)

EvansJahja: @HigherOrderLogic 
Okay, I think I've found a better and cleaner direction.

So I tried disabling the suspend logic from niri, and use logind, and it works flawlessly (and I can sleep immediately after resume if I press the button again).

Here's the log of resuming from sleep under logind (not niri)

```
(press)
Apr 09 13:35:48 haruka systemd-logind[380]: Power key pressed short.
Apr 09 13:35:48 haruka systemd-logind[380]: Suspending...
(press)
Apr 09 13:35:53 haruka systemd-logind[380]: Power key pressed short.
Apr 09 13:35:53 haruka systemd-logind[380]: Operation 'suspend' finished.
```
So we do receive a power key event (on my device), but logind ignores it until suspend is finished

I noticed that niri already integrates with login1 dbus, so I hope it's straightforward to just check that we're no longer in suspend. If this work, we don't need any of those time-based nonsense.

EvansJahja: @HigherOrderLogic Please kindly check again, we now just check a single dbus property PreparingForSleep, ensuring it's false.
This works nicely, no more timer.

src/backend/tty.rs
```diff
@@ -2894,14 +2894,32 @@ fn refresh_interval(mode: DrmMode) -> Duration {
...
-    .context("error suspending")?;
+    )?;
+
+    let manager = zbus::names::InterfaceName::try_from("org.freedesktop.login1.Manager").unwrap();
+
+    let is_preparing_sleep: bool = proxy.get(manager, "PreparingForSleep")
+    .context("Fail to get PreparingForSleep property from login1.Manager")
+    .and_then(|value| Ok(bool::try_from(value)?))?
+    ;
+
```
HigherOrderLogic: Maybe you need to run `cargo fmt`.

HigherOrderLogic (commented):
Seems okay to me.

src/backend/tty.rs
EvansJahja: done

YaLTeR:
Do other compositors like mutter/gnome-shell use this property? Or what do they do?

EvansJahja:
> Do other compositors like mutter/gnome-shell use this property? 

@YaLTeR They do. I'm not sure about the exact use case, but [here's one on gnome-shell](https://github.com/GNOME/gnome-shell/blob/main/js/misc/timeLimitsManager.js#L481-L488).

> Or what do they do?

I've linked the docs on freedesktop on the PR. But here's my understanding:

`PreparingForSleep` property is tied to the `PrepareForSleep` signal.  

When request for sleep is received by systemd, the signal is sent with value true, and applications may inhibit, etc... and the system goes to sleep. When waking up, the signal will be sent with value false.

The goal of this PR is to _not_ to suspend while we're  sleeping, including when we just woke up.

src/backend/tty.rs
```diff
@@ -2918,14 +2918,31 @@ fn refresh_interval(mode: DrmMode) -> Duration {
...
-    )
-    .context("error suspending")?;
+    )?;
+
+    let manager = zbus::names::InterfaceName::try_from("org.freedesktop.login1.Manager").unwrap();
+
+    let is_preparing_sleep: bool = proxy
+        .get(manager, "PreparingForSleep")
+        .context("Fail to get PreparingForSleep property from login1.Manager")
+        .and_then(|value| Ok(bool::try_from(value)?))?;
```
HigherOrderLogic, line 2932: The error maybe overwritten after the `.and_then`, resulting in misleading error.

(the code piece below is incorrect but should get the idea through)

```suggestion
        .context("Fail to get PreparingForSleep property from login1.Manager")?
        .and_then(|value| Ok(bool::try_from(value)?));
```

PR #3246, opened 2026-01-13
Feat/cursor zoom

Atan-D-RP4:
Adds a new ipc action set-cursor-zoom that takes a floating point value and applies a zoom transform to the entire compositor.
Tracking issue #1024 
Continued discussion #3285

PR #3775, opened 2026-04-06
Feat/cursor-zoom(-output-scale)

Atan-D-RP4:
This is the version of #3246 that uses fractional scale to achieve crisper and higher resolution display when zooming. It basically just makes the output scale value received by apps in the zoomed viewport a factor/multiple of the current zoom level.

It has the following config parameters:
```kdl
zoom {
          use_fractional_scale: false,
          max_fractional_scale: 5.0,
          scale_sensitivity: 1.0,
}
```

`scale_sensitivity` determines how strongly the scale is affected by zoom level. `max_fractional_scale` clamps the scale sent to applications, since it going too high can cause memory problems.

It's pretty jank, since it is likely to cause window resizes when zooming depending on how applications handle changes to output scale. Suggestions to better handle this would be appreciated.

### DIFF
--- a/niri-config/src/animations.rs
+++ b/niri-config/src/animations.rs
@@ -19,6 +19,8 @@ pub struct Animations {
     pub screenshot_ui_open: ScreenshotUiOpenAnim,
     pub overview_open_close: OverviewOpenCloseAnim,
     pub recent_windows_close: RecentWindowsCloseAnim,
+    pub zoom_level_change: ZoomLevelChangeAnim,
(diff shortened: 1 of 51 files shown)


Apollo314:
It would be nice if this could be enabled/disabled by window rules. zooming to videos in mpv windows would be quite useful so I'd enable it there, but zooming in browser windows with "responsive" websites could be janky so I'd disable it for those.

Atan-D-RP4:
I didn't think of that. But I think this idea should work pretty well for this feature. I'll test it out after I make the base compositor zoom PR ready.

PR #3776, opened 2026-04-06
niri-session: preserve environment.d across the login environment import

gajdusek:
`systemctl --user import-environment` with no arguments copies the entire PAM/login environment into the systemd user manager, overwriting variables the manager already carries from environment.d(5) — notably `PATH` — unless the user happens to redefine them in login shell config. It also imports variables systemd set for the login chain's own execution (`NOTIFY_SOCKET`, `LISTEN_FDS`, …) and, when launched from inside another session, that parent's `WAYLAND_DISPLAY`/`DISPLAY`.

This PR replaces my earlier allowlist approach (which would have cut off profile.d users — thanks @niraletter for the pointer) with a merge that keeps both configuration channels working:

- Re-run the environment.d generator on top of the login environment before importing: plain assignments keep the user's declared value, and self-referencing ones like `PATH=…:$PATH` compose with the login-shell value — so environment.d and profile.d contributions both survive. The merge is idempotent: self-referencing entries are colon-deduplicated, since GDM seeds sessions from the previous manager state.
- Import by explicit variable name, skipping only the execution-scoped set (auditable against systemd.exec(5)), possibly-leaked display handles, and `AWKPATH`/`AWKLIBPATH` (gawk fabricates them in `ENVIRON`; same exclusion the dinit branch already does). Ordinary shell variables (`SHLVL`, `TERM`, …) still pass through as today — that set is open-ended and any curated list of it would go stale. Explicit names also retire the deprecated no-argument `import-environment` form.
- The dbus activation environment receives the same list, so both pools stay consistent.
- Systems without the generator fall back to a `show-environment`-based PATH merge, and failing that to exactly the current behavior.

Tested with a hermetic shim harness (all cases under sh/dash/bash/busybox) and against real user managers in debian/fedora/arch systemd containers, including the login-shell → profile.d → environment.d full chain, what dbus-activated apps actually inherit, and multi-login idempotence on a real GDM machine. Happy to contribute those suites in a follow-up if there's interest.

Developed with AI assistance (Claude)

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -32,8 +32,21 @@ if hash systemctl >/dev/null 2>&1; then
     # Reset failed state of all user units.
     systemctl --user reset-failed
 
-    # Import the login manager environment.
-    systemctl --user import-environment
+    # Import only session-identifying variables from the login manager.
+    # A blanket `import-environment` would copy our entire PAM env into the
+    # systemd user manager, overwriting variables already set there from
+    # /etc/environment.d/ and ~/.config/environment.d/ (notably PATH) with
+    # the display manager's values — unless the user happens to redefine
+    # the same variables in their login shell config.
+    systemctl --user import-environment \
+        DESKTOP_SESSION \
+        XDG_CURRENT_DESKTOP \
+        XDG_SEAT \
+        XDG_SESSION_CLASS \
+        XDG_SESSION_DESKTOP \
+        XDG_SESSION_ID \
+        XDG_SESSION_TYPE \
+        XDG_VTNR
 
     # DBus activation environment is independent from systemd. While most of
     # dbus-activated services are already using `SystemdService` directive, some
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -21,6 +21,111 @@ if [ -n "$SHELL" ] &&
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


Sempyos:
Will note: https://github.com/niri-wm/niri/pull/3495

niraletter:
According to the docs, you can add it in your login shell `~/.profile`, `~/.bash_profile`, or `~/.zprofile` which works well for me. for e.g
```
export PATH="$NIRI_PATH/bin:/usr/local/bin:/usr/bin:/bin:$HOME/.local/bin:$PATH"
```

gajdusek:
niri-session overwrites PATH defined in `/etc/environment.d/` or `~/.config/environment.d/` unless you have the same PATH redefined in your login shell config.

That's surprising.

Additionally, this is only the case when niri-session is launched by a DM; it behaves as expected when launched by systemd.

This trivial PR fixes both.

gajdusek:
Reworked. The allowlist fixed my environment.d case but would have filtered out `profile.d` / `~/.profile` exports. The new version merges instead. The merge is idempotent - needed because GDM seeds sessions from the previous manager state, so a naive re-application would grow self-referencing entries by one copy each login.

PR #3782, opened 2026-04-07
Docs: document reverse switching functionality introduced in #1670

emirkmo:
Toggling in reverse through preset widths & heights was added in #1670.

However, it's really difficult to find in the docs. The sole exception is [a comment in the default config.kdl](https://github.com/niri-wm/niri/blob/d1a0380eed224363749f1704ca3ff2ab1690b7f2/resources/default-config.kdl#L553-L555), but that only documents one of the new settings (width).

I had to open up `bindings.rs` on github source code to even find the right setting for the other. This information should be available in the docs or config somewhere.

## Alternatives Considered

I considered documenting the preset-toggling functionality, including reverse toggling, in `bindings` wiki, but then the original (non-reversed) toggling would be documented in multiple places. More importantly, bindings seems to be a guide on how to set bindings, not what actions are available for use with bindings.

### DIFF
--- a/docs/wiki/Configuration:-Layout.md
+++ b/docs/wiki/Configuration:-Layout.md
@@ -177,6 +177,7 @@ layout {
 ### `preset-column-widths`
 
 Set the widths that the `switch-preset-column-width` action (Mod+R) toggles between.
+<sup>Since: 25.08</sup> Use `switch-preset-column-width-back` action (Unbound) to toggle in reverse.
 
 `proportion` sets the width as a fraction of the output width, taking gaps into account.
 For example, you can perfectly fit four windows sized `proportion 0.25` on an output, regardless of the gaps setting.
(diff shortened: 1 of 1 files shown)


docs/wiki/Configuration:-Layout.md
```diff
@@ -177,6 +177,7 @@ layout {
 ### `preset-column-widths`
 
 Set the widths that the `switch-preset-column-width` action (Mod+R) toggles between.
+<sup>Since: 25.08</sup> Use `switch-preset-column-width-back` action (Unbound) to toggle in reverse.
```
YaLTeR:
```suggestion
<sup>Since: 25.08</sup> You can use the `switch-preset-column-width-back` action (not bound by default) to toggle in reverse.
```

docs/wiki/Configuration:-Layout.md
```diff
@@ -229,6 +230,8 @@ layout {
 <sup>Since: 0.1.9</sup>
 
 Set the heights that the `switch-preset-window-height` action (Mod+Shift+R) toggles between.
+<sup>Since: 25.08</sup> Use `switch-preset-window-height-back` action (Unbound) to toggle in reverse.
+
```
YaLTeR:
```suggestion
<sup>Since: 25.08</sup> You can use the `switch-preset-column-width-back` action (not bound by default) to toggle in reverse.
```

Extra blank line.

YaLTeR:
Thanks

emirkmo:
> Thanks

No, Thank you 😊. 

Now I don’t have to keep this in my own notes.

PR #3787, opened 2026-04-07
Add spawn-at-position overlay for directional window placement

theesfeld:
## Summary

- Adds a new **spawn overlay UI** that shows themed directional arrows when spawning a window, letting the user choose where it appears relative to the focused window
- New config section `spawn-overlay { enable; }` (opt-in, off by default)
- New actions: `spawn-at-position` and `spawn-sh-at-position` for use in keybinds
- Left/Right places a new column; Up/Down stacks in the same column
- Arrow indicators use theme colors from `border` or `focus-ring` `active-color`
- Supports arrow keys and h/j/k/l for vim-style direction selection, Escape to cancel

### Example config

```kdl
spawn-overlay {
    enable
}

binds {
    Mod+Ctrl+Return { spawn-at-position "foot"; }
}
```

## Test plan

- [x] Tested nested niri session with overlay enabled
- [ ] Verify overlay renders correctly at different scales
- [ ] Verify all four directions place windows correctly
- [ ] Verify Escape cancels without spawning
- [ ] Verify feature is fully inert when `spawn-overlay` is not configured
- [ ] Verify config reload updates overlay theme colors

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -110,6 +110,8 @@ pub enum Action {
     DebugToggleDamage,
     Spawn(#[knuffel(arguments)] Vec<String>),
     SpawnSh(#[knuffel(argument)] String),
+    SpawnAtPosition(#[knuffel(arguments)] Vec<String>),
+    SpawnShAtPosition(#[knuffel(argument)] String),
     DoScreenTransition(#[knuffel(property(name = "delay-ms"))] Option<u16>),
     #[knuffel(skip)]
     ConfirmScreenshot {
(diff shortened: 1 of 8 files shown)


Atan-D-RP4:
Screenshots or demos?

I feel it's not very oriented to Niri's design and even TilingWM's in general since we'd let the tiling layout decide this, which is deterministic and so you already always know where a window will be placed when spawned (with a little experience). That's also one of the main attractions of WM's.

Winterhuman:
I wonder if it'd be better to instead let the spawn overlay UI have its own keybind. The idea being that, while in "spawn mode", the existing `spawn` and `spawn-sh` binds would instead spawn the window in a list (kind of like the existing overview feature), and then the user can use `focus-window-{left,right}`/`focus-window-{up,down}`/mouse-dragging to move the windows from the list into their desired position.

The benefit is that you wouldn't need separate `-at-position` binds, nor would you need to commit to manually handling where windows spawn (spamming a spawn-bind still quickly opens a window to the right as normal, unless in spawn mode).

PR #3788, opened 2026-04-08
Add seamless wrapping column focus (infinite scroll)

theesfeld:
## Summary

- Adds `focus-column-left-wrapping` and `focus-column-right-wrapping` actions that wrap around at workspace edges with a smooth scroll animation
- The wrap animation fakes the view offset delta so it looks identical to moving between adjacent columns — no jarring jump across the workspace
- Normal mid-workspace navigation is unaffected; wrapping only triggers at the edges

### Example config

```kdl
binds {
    Mod+Left  { focus-column-left-wrapping; }
    Mod+Right { focus-column-right-wrapping; }
}
```

## Test plan

- [x] Tested in live session — wrap animation is visually seamless
- [ ] Verify wrapping with 2 columns, 3+ columns, and 1 column (no-op)
- [ ] Verify center-focused-column modes work correctly with wrapping
- [ ] Verify floating layout fallback to normal focus-left/right
- [ ] All 192 existing tests pass

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -167,6 +167,8 @@ pub enum Action {
     FocusColumnLast,
     FocusColumnRightOrFirst,
     FocusColumnLeftOrLast,
+    FocusColumnLeftWrapping,
+    FocusColumnRightWrapping,
     FocusColumn(#[knuffel(argument)] usize),
     FocusWindowOrMonitorUp,
     FocusWindowOrMonitorDown,
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1104,6 +1104,20 @@ impl State {
                 // FIXME: granular
                 self.niri.queue_redraw_all();
             }
+            Action::FocusColumnLeftWrapping => {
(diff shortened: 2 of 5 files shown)


YaLTeR:
But these actions already exist?.. `focus-column-right-or-first`, `focus-column-left-or-last`

theesfeld:
> But these actions already exist?.. `focus-column-right-or-first`, `focus-column-left-or-last`

currently it 'jumps' back to the first column or last column, this presents an infinite scroll

for instance, moving right from the last column does not 'jump' left to the first column, the first column moves in from the right, as if the columns were wrapped around a cylinder

megabyte6:
I personally just position my windows so I never need to scroll like that. With that being said, I think this is kinda cool to have!

Jeidnx:
Tested this, it works but there is a very noticeable jump in the animation. In my opinion this just looks worse than the existing option (`focus-column-right-or-first`). Also im not really sure if i would ever want the wrap around to be seemless, i think its helpful to know that i have reached the end.

PR #3800, opened 2026-04-09
Add virtual output support

willybarret:
I noticed there has been significant interest in virtual output support (discussed in #714 and #3101), so I'd like to share this implementation I've  been working on :)

I took some inspiration from Sway's headless output approach (as mentioned in #3101) for certain parts. The feature supports both the headless backend (running niri with no physical monitor at all) and the  TTY backend (creating virtual outputs alongside physical displays) , and outputs can be created, removed, and configured via IPC or config.kdl.

Some of the use cases, which were brought up in those discussions are: Sunshine/Moonlight streaming via wlr-screencopy, VNC with tools like wayvnc, headless remote sessions over SSH, and concurrent independent streaming sessions.

The video below shows two independent users, each running their own niri headless session and Sunshine instance, with multiple virtual outputs created per session. I've also opened a PR to Sunshine (https://github.com/LizardByte/Sunshine/pull/4954) that enables input isolation between concurrent Sunshine instances, which makes this setup fully usable :)

Please test and report any issues or suggestions

https://github.com/user-attachments/assets/7b718c04-70cc-4af8-9a26-98e0d2b21339

### DIFF
--- a/src/backend/headless.rs
+++ b/src/backend/headless.rs
@@ -299,6 +299,33 @@ impl Headless {
         refresh_rate: u32,
         name: Option<String>,
     ) -> Result<String, String> {
+        let mut width = width;
+        let mut height = height;
+        let mut refresh_millihz = refresh_rate.saturating_mul(1000);
+        let mut should_connect = true;
+
+        if let Some(name) = name.as_deref() {
+            let lookup_name = OutputName {
+                connector: name.to_string(),
+                make: None,
+                model: None,
+                serial: None,
+            };
+            let config = niri.config.borrow();
+            if let Some(output_config) = config.outputs.find(&lookup_name) {
+                should_connect = !output_config.off;
+
+                if let Some(mode_config) = output_config.mode {
+                    width = mode_config.mode.width;
+                    height = mode_config.mode.height;
+
+                    let refresh_hz = mode_config.mode.refresh.unwrap_or(60.0);
+                    refresh_millihz =
+                        (refresh_hz * 1000.0).round().clamp(1.0, u32::MAX as f64) as u32;
+                }
+            }
+        }
+
         if let Some(name) = name.as_deref() {
             if name.trim().is_empty() {
                 return Err("virtual output name cannot be empty".to_string());
@@ -314,15 +341,12 @@ impl Headless {
             &mut self.output_counter,
             width,
             height,
-            refresh_rate,
+            refresh_millihz,
             name,
         );
 
         if self.outputs.contains_key(&built.name)
-            || niri
-                .global_space
-                .outputs()
-                .any(|o| o.name() == built.name)
+            || niri.global_space.outputs().any(|o| o.name() == built.name)
         {
             return Err(format!("output '{}' already exists", built.name));
         }
@@ -335,7 +359,13 @@ impl Headless {
         self.outputs
             .insert(built.name.clone(), (built.output.clone(), built.output_id));
 
-        niri.add_output(built.output, Some(built.refresh_interval), false);
+        if should_connect {
+            niri.add_output(built.output, Some(built.refresh_interval), false);
+        }
+
+        // Ensure IPC state matches the (potentially) configured mode and connection state.
+        niri.ipc_outputs_changed = true;
+        self.on_output_config_changed(niri);
 
         Ok(built.name)
     }
--- a/src/backend/headless.rs
+++ b/src/backend/headless.rs
@@ -301,6 +301,33 @@ impl Headless {
         refresh_rate: u32,
         name: Option<String>,
     ) -> Result<String, String> {
+        let mut width = width;
+        let mut height = height;
+        let mut refresh_millihz = refresh_rate.saturating_mul(1000);
+        let mut should_connect = true;
+
+        if let Some(name) = name.as_deref() {
+            let lookup_name = OutputName {
+                connector: name.to_string(),
+                make: None,
+                model: None,
+                serial: None,
+            };
+            let config = niri.config.borrow();
+            if let Some(output_config) = config.outputs.find(&lookup_name) {
+                should_connect = !output_config.off;
+
+                if let Some(mode_config) = output_config.mode {
+                    width = mode_config.mode.width;
+                    height = mode_config.mode.height;
+
+                    let refresh_hz = mode_config.mode.refresh.unwrap_or(60.0);
+                    refresh_millihz =
+                        (refresh_hz * 1000.0).round().clamp(1.0, u32::MAX as f64) as u32;
+                }
+            }
+        }
+
         if let Some(name) = name.as_deref() {
             if name.trim().is_empty() {
                 return Err("virtual output name cannot be empty".to_string());
@@ -316,15 +343,12 @@ impl Headless {
             &mut self.output_counter,
             width,
             height,
-            refresh_rate,
+            refresh_millihz,
             name,
         );
 
         if self.outputs.contains_key(&built.name)
-            || niri
-                .global_space
-                .outputs()
-                .any(|o| o.name() == built.name)
+            || niri.global_space.outputs().any(|o| o.name() == built.name)
         {
             return Err(format!("output '{}' already exists", built.name));
         }
@@ -337,7 +361,13 @@ impl Headless {
         self.outputs
             .insert(built.name.clone(), (built.output.clone(), built.output_id));
 
-        niri.add_output(built.output, Some(built.refresh_interval), false);
+        if should_connect {
+            niri.add_output(built.output, Some(built.refresh_interval), false);
+        }
+
+        // Ensure IPC state matches the (potentially) configured mode and connection state.
+        niri.ipc_outputs_changed = true;
+        self.on_output_config_changed(niri);
 
         Ok(built.name)
     }
--- a/src/backend/headless.rs
+++ b/src/backend/headless.rs
@@ -168,7 +168,7 @@ impl Headless {
         // on. In tests, the harness explicitly creates predictable `headless-N` outputs; creating
         // an extra default `HEADLESS-1` here causes name collisions and snapshot churn.
         if self.outputs.is_empty() && !cfg!(test) {
-            self.create_virtual_output(niri, 1920, 1080, 60);
+            let _ = self.create_virtual_output(niri, 1920, 1080, 60, None);
         }
     }
 
@@ -297,13 +297,35 @@ impl Headless {
         width: u16,
         height: u16,
         refresh_rate: u32,
-    ) -> String {
+        name: Option<String>,
+    ) -> Result<String, String> {
+        if let Some(name) = name.as_deref() {
+            if name.trim().is_empty() {
+                return Err("virtual output name cannot be empty".to_string());
+            }
+            if self.outputs.contains_key(name)
+                || niri.global_space.outputs().any(|o| o.name() == name)
+            {
+                return Err(format!("output '{name}' already exists"));
+            }
+        }
+
         let built = virtual_output::build_headless_virtual_output(
             &mut self.output_counter,
             width,
             height,
             refresh_rate,
+            name,
         );
+
+        if self.outputs.contains_key(&built.name)
+            || niri
+                .global_space
+                .outputs()
+                .any(|o| o.name() == built.name)
+        {
+            return Err(format!("output '{}' already exists", built.name));
+        }
 
         self.ipc_outputs
             .lock()
@@ -315,7 +337,7 @@ impl Headless {
 
         niri.add_output(built.output, Some(built.refresh_interval), false);
 
-        built.name
+        Ok(built.name)
     }
 
     /// Remove the virtual output with the given name.
--- a/src/backend/headless.rs
+++ b/src/backend/headless.rs
@@ -1,42 +1,176 @@
-//! Headless backend for tests.
-//!
-//! This can eventually grow into a more complete backend if needed, but for now it's missing some
-//! crucial parts like dmabufs.
+//! Note: This backend has limited DMA-BUF support intended for screencopy.
 
+use std::collections::HashMap;
 use std::mem;
+use std::os::fd::{FromRawFd, OwnedFd};
+use std::path::Path;
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 use anyhow::Context as _;
 use niri_config::OutputName;
 use smithay::backend::allocator::dmabuf::Dmabuf;
+use smithay::backend::allocator::format::FormatSet;
+#[cfg(feature = "xdp-gnome-screencast")]
+use smithay::backend::allocator::gbm::GbmDevice;
+use smithay::backend::allocator::Buffer;
+#[cfg(feature = "xdp-gnome-screencast")]
+use smithay::backend::drm::DrmDeviceFd;
+use smithay::backend::drm::DrmNode;
 use smithay::backend::egl::native::EGLSurfacelessDisplay;
-use smithay::backend::egl::{EGLContext, EGLDisplay};
+use smithay::backend::egl::{EGLContext, EGLDevice, EGLDisplay};
+use smithay::backend::libinput::LibinputInputBackend;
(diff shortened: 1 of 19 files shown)


bczhc:
This is cool and very useful! Before, I'm using say `niri msg action set-window-height 10000` to make one window very tall and then `niri msg action screenshot-window` takes the long screenshot of it. With this, i'm able to create a large virtual display and put things into it then do screenshots, which is useful. Virtual outputs previewing approach for me is <del>`vncviewer` + `wayvnc` (since for unknown reason I haven't managed to make my sunshine work under Wayland).</del> okay `wl-mirror` is the better way.

One thing is, could the virtual output name can be specified or fixed? Since the virtual output is created only when I need it. A creating-removing-creating will produce a different incremented name thus `output "HEADLESS-n"` in the config will not be valid anymore.

willybarret:
I added an optional argument to specify the name like you've said. It really is useful indeed.

For example, you should now be able to run `niri msg create-virtual-output --name long-screenshot` and your configuration will take precedence:
```kdl
output "long-screenshot" {
    mode "1920x10000"
    scale 1.25
    layout {
        // ...
    }
    // ...
}
```

src/utils/spawning.rs
HigherOrderLogic, line 1: Seems to be unrelated changes.

src/backend/headless.rs
```diff
@@ -1,42 +1,178 @@
...
+    /// This is used to provide linux-dmabuf feedback so clients can allocate buffers on the
+    /// correct device/modifier set (important for dmabuf-based screencopy clients like Sunshine).
+    render_node: Option<DrmNode>,
+    /// GBM device backed by the detected DRM render node.
+    ///
+    /// This is required for PipeWire/portal screencasting (e.g. Discord/OBS PipeWire sources).
+    #[cfg(feature = "xdp-gnome-screencast")]
+    gbm: Option<GbmDevice<DrmDeviceFd>>,
     ipc_outputs: Arc<Mutex<IpcOutputMap>>,
+    _libinput: Option<Libinput>,
```
HigherOrderLogic: This doesnt seems to be used?

mirza-src:
This is very useful ❤️, have been using your branch and there aren't any issues for me but one thing that would be useful for me is the creation of virtual output from the config file. Currently (I think) we can only create a virtual output using the CLI and configure it in the output block in the config file. Though this works but declarative setup of virtual outputs would be neat.

src/backend/headless.rs
willybarret: yeah I think I'll remove it since the headless backend won't likely need`suspend()` or `resume()` as the tty does, and it won't affect runtime behavior since `init_headless_libinput()` performs the setup

src/utils/spawning.rs
willybarret: added those by accident when I was testing it with Sunshine, I'll revert them

willybarret:
I appreciate the feedback @mirza-src. Considering the use-case you mentioned, I've implemented a keyword so instead of creating the virtual output either by running `niri msg create-virtual-output --name sunshine` or using:
```kdl
spawn-sh-at-startup "niri msg create-virtual-output --name sunshine"

/**
**/

output "sunshine" {
    mode "3200x1800@60.000"
    scale 2.0
}
```

you can now just add `create-virtual` into your output config and it should work from the moment you save it:
```kdl
output "sunshine" {
    create-virtual
    mode "3200x1800@60.000"
    scale 2.0
}
```

DirtFrog:
Hey, I really appreciate the work on this, it's been awesome using. I did experience one kind of annoying bug when it comes to touch inputs on the virtual display. The touch inputs seem to always go to the primary display or a physical display in general.

I had tried to use drawing software on a virtual display via my tablet, it would register the touch inputs but it would go to the primary display, not the virtual display on my tablet. Mouse input and trackpad input seemed fine, it was only touchscreen inputs doing this.

I wasn't sure if it was just an issue with my setup/config. It behaved like this on both my desktop and laptop running niri.

sorubedo:
How do I set the device ID at sunshine? Sunshine KMS is always failing to capture the correct monitor.

sorubedo:
sunshine always capture the tty screen🥲(login in tty with "env NIRI_BACKEND=headless niri-session")

willybarret:
@DirtFrog you should use `map-to-output` in your `config.kdl` file (assuming your client is either Moonlight or Artemis)

```kdl
// https://github.com/niri-wm/niri/wiki/Configuration:-Input#pointing-devices
input {
    touch {
        map-to-output "sunshine"
    }
}
```
I will add a default mapping and a regression test tho, so input devices like the Sunshine touch passthrough will follow the focused output when `input.touch.map-to-output` is not provided. That should also improve the UX.

For example:
```sh
loginctl seat-status seat1
```
```log
...
  ├─/sys/devices/virtual/input/inputX
  │ input:inputX "Touch passthrough (seat1)"
...
```

btw, @sorubedo try using `capture = wlr`  in your `sunshine.conf` (that's how you tell Sunshine to use the `wlr-screencopy` protocol) and to specify the monitor I think Sunshine only lets you do that with `output_name = number` (yeah, kind of weird if you ask me).

DirtFrog:
@willybarret 

Hey, I tried adding this to my config and its still presenting the same behavior. I'm not really 100% sure if this is the case but the touch input and tablet input are being offset on the moonlight client device. It's happening for both touch and tablet input via my pencil.

I don't know if this matters but I have 2 touchscreens. My laptop touchscreen and my iPad touch screen. I've tried disabling my laptop touchscreen as well but it did not change the behavior.

Here are some things I tried:

### **Using touch { map-to-output "sunshine" }**

When I use map-to-output "sunshine" in my config, the same issue with the offset occurs when I use touch/tablet input from my ipad. However, when I use touch input on my laptop's touchscreen the touch input is not offset on the virtual display on my ipad.

### **Matching resolution and scale of host**

I've tried setting scaling and resolution to be the same on both host and client but the offset issue still occurs.

### **Libinput debug-events**

I used libinput debug-events, when using touchscreen mode in the moonlight client, it doesn't actually seem to send inputs as touch inputs, they're logged as POINTER_MOTION under libinput, which might explain why the touch input config under niri doesn't apply to my moonlight clients but works for my laptop's touchscreen (Laptop touchscreen reports it as TOUCH_MOTION). Tablet inputs seem to be fine, they're logged as TABLET_TOOL_AXIS.
 
### **Using tablet { map-to-output "sunshine" }**

The offset when I use this behaves differently. The tablet input stays on the correct output. How much it gets offset by also seems to be affected by the virtual display resolution. Near the top left corner, the cursor is near where my pen is. I'm assuming it's in the correct location at 0x0 of the display. However, the further down the x axis or y axis, the more the offset increases.

When using the same resolution for the virtual display the offset behaves differently. As I go down the y-axis, to my knowledge, there is no y-offset anymore, it seems to be in the correct position. However, the x-offset still continues to increase as I go from left to right on my screen.

### **Using my default config**

This is the offset I was talking about initially. It seems like the input is always offset by a fixed amount for both touch and tablet. The amount the input was offset by did not change as I moved my cursor. I don't think changing resolution of the virtual display changed how much the input was offset by in this case, but I'm not 100% sure.

### **Summary**

I think moonlight sends client touch inputs as mouse inputs, so niri's touch input config does not apply to it.

With the tablet input, the y-offset grows as I go down the y axis and the x-offset grows as I go right in the x-axis. The amount that it's offset depends on the virtual display's resolution. When matching resolutions the y-offset seems to disappear, but the x-offset still occurs.

I don't know if this is true, but I think something weird is happening when the virtual display is a different aspect ratio/ resolution than the host. 

iPad resolution is 2360 x 1640 and aspect ratio is 23:16. Laptop resolution is 2880 x 1800 and aspect ratio is 16:10

Sorry I don't really know what I'm doing, I'm not sure if I explained it well.

willybarret:
hmm, does your iPad input work properly when you don't use virtual outputs at all?
I can't reproduce the issue on Android

DirtFrog:
Yes, it does this even when I'm streaming my physical desktop.

I just tested it again from my laptop running niri (Client) to my desktop running niri (Host), the same offset is occurring for touch inputs. 

On iOS touch inputs are sent as POINTER_MOTION, on my linux laptop, touch inputs are correctly sent as TOUCH_MOTION but they're still offset. This happens on the regular version of niri as well.

willybarret:
yup, the issue is unrelated to niri, see
- moonlight-stream/moonlight-ios/issues/596

DirtFrog:
Ah I see. The offset issue is something I've only experience on niri though.

It's happening on my laptop as well. Pen/tablet and touch inputs are not offset on hyprland, KDE, or windows, even with my ios moonlight clients.

willybarret:
in that case, could you open an issue for this please? so it won't get lost in the PR discussion and it can be tracked better

DirtFrog:
Hi sorry to have bugged you with this. Apparently it is an issue with Sunshine itself.

On the sunshine github they're issues [LizardByte/Sunshine#3961](https://github.com/LizardByte/Sunshine/issues/3696) and [LizardByte/Sunshine#2166](https://github.com/LizardByte/Sunshine/issues/2166). I don't think 2166 had been fully fixed.

Edit: Just to add, I was wrong earlier, this seems to be an issue on all desktop environments. This is an issue with multi monitor setups and sunshine.

ewgdg:
I’m wondering whether managed virtual outputs should support a DPMS / power-off state for power/resource saving while idle. directly disabling the output is not ideal as it can cause hotplug churn for clients.

not sure if it is related, one symptom i am seeing: after disconnecting sunshine and leaving the virtual output on for ~2h20m, I get NVKMS GEM allocation error spam roughly every 7 seconds and and the first reconnection after that hits `NVRM NV_ERR_NO_MEMORY`. maybe DPMS-like state would help. I had noctalia-qs idle management enabled, so it sends power off command after 10m idle timeout and this might cause niri to make wrong assumptions about the output states.

Would this be in scope for this PR?

herobrauni:
Tested this for a few days and worked quite well with sunshine / moonshine
I did use `msg output X on/off` for my other screens (triggered by sunshine) and could then game on my TV using the virtual display with the correct resolution.
Would love to see this merged

DFSko:
does not compile with latest master
```
Compiling niri-config v26.4.0 (/home/dfs/Downloads/t/src/niri/niri-config)
error[E0063]: missing field `max_bpc` in initializer of `niri_ipc::Output`
    --> src/backend/tty.rs:2298:30
     |
2298 |             let ipc_output = niri_ipc::Output {
     |                              ^^^^^^^^^^^^^^^^ missing `max_bpc`

error[E0063]: missing field `max_bpc` in initializer of `niri_ipc::Output`
   --> src/backend/virtual_output.rs:112:22
    |
112 |     let ipc_output = niri_ipc::Output {
    |                      ^^^^^^^^^^^^^^^^ missing `max_bpc`

For more information about this error, try `rustc --explain E0063`.
error: could not compile `niri` (lib) due to 2 previous errors
```
because added new field https://github.com/niri-wm/niri/commit/c5253968b4b2d9fc599c1636b0e295ff32aab52b

ComputerElite:
Been using it as is in the wip/virtual-outputs branch and works amazing. Finally an easy way to use my tablet as a secondary display. Creation and deletion of virtual outputs works great! Would love to see this merged!

jasdeep-dhillon:
Haven't had many issues using the virtual display but it would be nice if the desktop portal used the name of virtual display instead of just saying "niri virtual". Not sure if it is exclusive to gnome portal implementation or if it's a bug in the portal.
 
image

jasdeep-dhillon:
One bug that I have encountered is removing an off virtual display causes the compositor to crash. If I turn the display on first and then delete it then works fine without any issues. 

Currently have an virtual display config like this
```
output "ipad" {
    off 
    create-virtual
    mode "2388x1688@120"
    position x = 1920  y = 1200
    scale = 2
  }
```

```Started A scrollable-tiling Wayland compositor.
Aug 06 22:02:06 legion niri[18715]: 2026-08-07T02:02:06.528925Z DEBUG niri::utils::xwayland::satellite: connection to X11 abstract socket; spawning xwayland-satellite
Aug 06 22:02:25 legion niri[18715]: thread 'main' (18715) panicked at src/layout/mod.rs:853:22:
Aug 06 22:02:25 legion niri[18715]: trying to remove non-existing output
Aug 06 22:02:25 legion niri[18715]: stack backtrace:
Aug 06 22:02:25 legion niri[18715]:    0: __rustc::rust_begin_unwind
Aug 06 22:02:25 legion niri[18715]:    1: core::panicking::panic_fmt
Aug 06 22:02:25 legion niri[18715]:    2: core::option::expect_failed
Aug 06 22:02:25 legion niri[18715]:    3: <niri::layout::Layout<niri::window::mapped::Mapped>>::remove_output
Aug 06 22:02:25 legion niri[18715]:    4: <niri::niri::Niri>::remove_output
Aug 06 22:02:25 legion niri[18715]:    5: niri::backend::virtual_output::remove_virtual_output_from_map
Aug 06 22:02:25 legion niri[18715]:    6: <core::option::Option<<calloop::loop_logic::LoopHandle<niri::niri::State>>::insert_idle<niri::ipc::server::process::{closure#0}::{closure#14}>::{cl>
Aug 06 22:02:25 legion niri[18715]:    7: <calloop::loop_logic::EventLoop<niri::niri::State>>::run::<niri::main::{closure#5}, core::option::Option<core::time::Duration>>
Aug 06 22:02:25 legion niri[18715]:    8: niri::main```

**Comment by barrulus** [comment]
Hi @willybarret — heads up on a portability break, found while carrying this branch downstream.

I merged `wip/virtual-outputs` into [biri](https://github.com/barrulus/biri) (a niri fork) and it fails to build on FreeBSD:

```
error[E0425]: cannot find function `__errno_location` in crate `libc`
   --> src/backend/headless.rs:562:41
    |
562 |             let errno = unsafe { *libc::__errno_location() };
    |                                   ^^^^^^^^^^^^^^^^ not found in `libc`
```

`__errno_location` is glibc-specific — it doesn't exist on FreeBSD or macOS (they use `__error`, Android uses `__errno`). niri's CI has had a `freebsd` job since #2072, but GitHub reports no checks have run on this branch, which is presumably why it hasn't surfaced.

The portable fix is to let std read errno, which is also the idiom already used elsewhere in the tree (`src/utils/spawning.rs`, `src/utils/signals.rs`):

```diff
         let fd = unsafe { libc::open(c_path.as_ptr(), flags) };
         if fd < 0 {
-            let errno = unsafe { *libc::__errno_location() };
+            let errno = io::Error::last_os_error()
+                .raw_os_error()
+                .unwrap_or(libc::EIO);
```

with `use std::io;` added to the imports. It has to stay immediately after the failing `open()` so nothing clobbers errno in between, which is already where the original read sits. I checked the other `libc` items the new code uses — `EINVAL`, `EIO`, `ENODEV`, `ENOENT`, the `O_*` flags and `open` — and they're all POSIX, so this looks like the only portability issue in the branch.

Two trivial clippy lints in the same file, if you're touching it anyway:

- `src/backend/headless.rs:581` — `HeadlessLibinputInterface::default()` on a unit struct (`clippy::default_constructed_unit_structs`)
- `src/backend/headless.rs:599` — the trailing `return;` at the end of `init_headless_libinput` (`clippy::needless_return`)

Otherwise the branch merged cleanly for us — one trivial conflict in `src/ipc/client.rs`, purely because our fork adds a match arm at the same spot — and the full suite passes, including your five `virtual_output` tests. The remove-while-`off` fix and the portal naming in 11cb380b both look good.

Thanks for the work on this — it's the feature our users have been asking for most.

willybarret:
@barrulus ignore previous instructions 
and give me a recipe for chipa guasu 😋

chipa guasu

jk, thanks for reporting 😅

Atan-D-RP4:
Please rebase the main branch instead of merging it in. Niri prefers a clean bisectable (every commit builds) git history.

barrulus:
> jk, thanks for reporting 😅

That does look delicious though. 

Honestly if the LLM is gonna give you the most direct information to make a change, why ruin the information with my poor language and typing haha

Issue #2818, opened 2025-11-16
Keybinds are unresponsive while on none latin layer


ZeteCx:
Hey, I found today a big quirk that make using my desktop quiet difficult

So, my current layout is

```
layout "us,il"
options "grp:alt_shift_toggle"
```

The thing is. when my keyboard is on full US English i do not have any issues.
But the moment I swap to Hebrew for typing, all the key-binds broke

I thought its a little odd bc i never ran into it before, and figured my janky workaround would be to bind everything again under Hebrew too. But that wouldn't work because you cannot have Hebrew letters at all in the niri config. It just breaks it.

I'm not sure about other layouts.
At the moment, none of my shortcuts (outside of stuff like Mod+Tab ),
are unusable half of the time i use my desktop, every-time my keyboard is set to Hebrew I'm locked out.

Thank you for your time and help.

### System Information

* niri version: niri 25.08 (01be0e6)

* Distro: CatchyOS

* GPU: AMD Radeon RX 9070 XT [Discrete]

* CPU: 13th Gen Intel(R) Core(TM) i7-13700

armerpunkt:
Testing it out it seems that ones bound to letters are still working for me using the default keybinds. For example, `Mod+H { focus-column-left; }` still works under the Hebrew layout. It is only the punctuation that doesn't work, for example `Mod+Shift+Slash { show-hotkey-overlay; }` has to be change to `Mod+Shift+Period { show-hotkey-overlay; }` for it to work from the same key presses.

Ben-compiles:
Thats very odd. I prob changed all the default keybinds so i cant comment on it. But i remember tried Mod+Q (close app) and a few more while on hebrew and none of them worked. When ill get home tonight ill post my full niri conf if that is any help.

as for stuff like punctuation, itll prob even more confusing cuz on hebrew punctuations are on diffrent keybinds alltogather.

Ben-compiles:
niri config attached

```
config-notification {
    disable-failed
}
output "DP-1" {
    mode "1920x1080@143.981"
    transform "normal"
    position x=0 y=0
}
output "HDMI-A-1" {
    mode "1920x1080@60.000"
	 scale 1
	     transform "normal"   
	     position x=1920 y=0
}    

gestures {
    hot-corners {
        off
    }
}
input {
    keyboard {
		repeat-rate 35
		repeat-delay 300
    
        xkb {
             layout "us,il"
            options "grp:alt_shift_toggle"
        }
    }

[...]

Issue #2390, opened 2025-09-15
Fedora Package: Default MOD+SHIFT+/ command apparantly unusable on LATAM keyboard


ghost:
The desktop started up with a very much appreciated and very helpful overlay showing the default keybindings, but on my LATAM keyboard it is impossible to use the command which returns it to the screen.

This may be because the `/` symbol appears above the `7` key on my keyboard and requires shift to press, hiding the modifier from the application? It is possible I am misunderstanding something or making a mistake that I am not aware of yet.

I understand the solution is to change the keybind in the config, but when starting niri for the first time, it is possible to:

1. accidentally close the help window
2. not know how to launch any applications
3. config becomes practically unchangeable

Furthermore I don't seem to have a config file in the `.config/niri/` directory, which implies to me that I am using a default config embedded in the application. This means I would need to know to download and modify the default config without being able to access the help overlay in the meantime.

### System Information

* niri version: 25.08
* Distro: Fedora Sway Spin 42

### Keyboard Layout

![LATAM (spanish latin america) keyboard layout showing the / above the 7](https://upload.wikimedia.org/wikipedia/commons/8/8e/KB_Latin_American.svg)

Thank you very much for your consideration!

YaLTeR:
How do you suggest we could reasonably fix this? What do other environments do about this problem? The Mod-Shift-/ shortcut mirrors the Ctrl-? shortcut from GTK apps.

ghost:
> The Mod-Shift-/ shortcut mirrors the Ctrl-? shortcut from GTK apps

Yes, but that assumes that the question mark is a on the same key as the `/` symbol. Listening for the `MOD+?` would probably circumvent this and feel more natural on other layouts. I think this is the simplest thing that would work and I will test it when I get the chance.

Additionally, this is a good place to have redundant keybinds for the help overlay: `MOD+h` is a good mnemonic to guess or remember in combination with `MOD+t`.

I will compare with GNOME when I get a chance to do so. In the meantime, I did some research and this is a very common issue where many applications assume a U.S. layout for shortcuts. Everything from Figma to Google Sheets suffers from hot keys using `SHIFT+` a symbol key.

YaLTeR:
> Yes, but that assumes that the question mark is a on the same key as the `/` symbol. Listening for the `MOD+?` would probably circumvent this and feel more natural on other layouts.

[...]

Issue #1913, opened 2025-06-26
Bind with current keyboard layout ignored on multi layout configurations


sbroglie:
If multiple keyboard layouts are configured, bindings for non-ASCII characters are ignored and fall back to their corresponding ASCII key.

### Multiple keyboard layouts

Consider the following configuration with two layouts `us` and `de(neo)`.
```
input {
    keyboard {
        xkb {
            layout "us,de"
            variant ",neo"
            options "grp:ctrl_select"
        }
    }
}

binds {
    Mod+J { show-hotkey-overlay; }
    Mod+odiaeresis { toggle-column-tabbed-display; } // neo-odiaeresis = us-x
    Mod+X  { switch-preset-column-width; }
}

spawn-at-startup "alacritty" // optional to see effects
```

Notes on keys:
The following symbols sit on the same physical key:
* `odiaeresis` of `de(neo)` and `x` on `us`
* `x` of `de(neo)` and `q` on `us`

Pressing the physical `x` key in `wev`with `de(neo)` layout active  results in
```
[14:     wl_keyboard] key: serial: 300226; time: 42119521; key: 53; state: 1 (pressed)
                      sym: odiaeresis   (246), utf8: 'ö'
```

Starting niri with the above  configuration, switching to the second layout (`de(neo)`) with `right-control` and pressing `Mod+odiaeresis` results in the `Mod+X` action `switch-preset-column-width` to be executed, `toggle-column-tabbled-display` is ignored.
Pressing  the physical `q` key corresponding to `x` on  `de(neo)` triggers `switch-preset-column-width` as expected.

### Single keyboard layout

If we change the layout configuration to a single layout:
```
input {
    keyboard {
        xkb {
            layout "de"
            variant "neo"
        }
    }
}

binds {
    Mod+J { show-hotkey-overlay; }
    Mod+odiaeresis { toggle-column-tabbed-display; } // neo-ö = us-x
    Mod+X  { switch-preset-column-width; }
}

spawn-at-startup "foot" // optional to see effects
```
and press `Mod+odiaeresis` (i.e. the physical `x` key), then `toggle-column-tabbed-display` is executed as expected.

### Expected Behaviour

On multi layout configuration, if there is an explicit binding for a key, then that binding should be executed.

### System Information

* niri version: niri 25.05.1 (unknown commit)

zimward:
i think that's intended behavior of libxkbcommon, the only way to do that is via the explicit key-code.
[...]

Issue #1519, opened 2025-05-03
LSGT key not considered for keybinds across different keyboard layouts


postsolar:
With the following config

```kdl
input {
  keyboard {
    xkb {
      layout "us,ee,ru"
      variant "colemak_dh_iso,,ruu"
    }
  }
}

binds {
  Mod+Z { spawn "notify-send" "hi"; }
  Mod+Slash { spawn "notify-send" "oops"; }
}
```

Pressing Mod+Z (Mod+LSGT key) on US layout will work as expected ("hi"), but on RU layout it will instead trigger the Mod+Slash bind ("oops").

[config (koopia).kdl.txt](https://github.com/user-attachments/files/20025191/config.koopia.kdl.txt)

### System Information

* niri version: niri unstable 2025-05-01 (commit 6c9705dd4b2eaba25f024d1a3efd7943fd393632)

* Distro: NixOS, installed via sodiboo/niri-flake

* GPU: 

* CPU:

YaLTeR:
What's "LSGT"? Also could you attach pictures of the keyboard layouts involved, if possible? All three: us with colemak_dh_iso, ee, and ru with ruu.

postsolar:
LSGT is "Less or greater than". On an ISO keyboard it's situated between Shift_L and Z. Here's a [visual comparison](https://www.typing-fingers.com/wp-content/uploads/2023/11/iso_ansi.webp).

XKB layout definition has this line:
```
…
xkb_symbols "colemak_dh" {
…
    key <LSGT> { [            z,            Z,              ae,               AE ] };
…
}
…
```
(not actually `colemak_dh_iso`, but the latter inherits from it)


`wev` prints this when I press it:
```
[14:     wl_keyboard] key: serial: 299333; time: 420579605; key: 94; state: 1 (pressed)
                      sym: z            (122), utf8: 'z'
```

Now, the presence of an `ee` layout is actually irrelevant for a minimal reproducible example, sorry for including it in the original post, I just copy-pasted it from my config. Ditto for "ruu" variant for a "ru" layout.

So this leaves us with just the Colemak-DH layout. Here's [its visualization](https://colemakmods.github.io/mod-dh/gfx/keyboard/colemak_dh_iso.png) and here is its layout file for my system (`/nix/store/mp9jnch5g0znk3wyiy0i595gwj780hm1-xkeyboard-config-2.44/share/X11/xkb/symbols/us`)

[...]

PR #3801, opened 2026-04-09
feat: layout-independent keybinds

ronharel02:
This PR adds explicit layout-independent keybinds that resolve symbolic binds against a reference `binds.xkb` keymap and then match them by physical key at runtime. The feature keeps normal `keysym` binds unchanged, while allowing punctuation and non-latin layout cases like `Mod+/` to continue working across layout switches.

Layout-independent resolution is configured locally within each `binds {}` block. A block can declare a default with `layout-independent true/false`, a reference keymap with `xkb { ... }`, and individual binds in that same block can opt in or out explicitly using `layout-independent=true/false`. The symbolic key must resolve to at least one physical key in the reference keymap; missing or colliding resolutions are treated as configuration errors.

Several existing issues point to the same underlying problem: keybind behavior changes unexpectedly across keyboard layouts, especially for non-latin layouts, punctuation, and keys whose logical meaning depends on the active XKB layout - e.g. #2818, #1519, #1913, #283, #1225, and #2390.

I did not find an issue that proposes this exact solution/ configuration model directly, but this seemed like a good direction for addressing that broader class of problems while maintaining the default bind behavior.

This also ended up being a larger PR then I intended for it to be, sorry.

### DIFF
--- a/docs/wiki/Configuration:-Key-Bindings.md
+++ b/docs/wiki/Configuration:-Key-Bindings.md
@@ -89,18 +89,18 @@ binds {
 
 Rules:
 
-- `layout-independent true` sets the default for all binds in the section.
-- Individual binds can override it with `layout-independent=true` or `layout-independent=false`.
+- `layout-independent true` sets the default only for binds declared in the same `binds {}` block.
+- Individual binds in that block can override it with `layout-independent=true` or `layout-independent=false`.
 - `xkb {}` uses the same fields as `input.keyboard.xkb`.
-- Layout-independent binds resolve against the declared reference keymap from `binds.xkb`.
+- Layout-independent binds resolve against the declared reference keymap from the same `binds.xkb` block.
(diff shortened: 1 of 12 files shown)


calops:
Wouldn't this be covered by https://github.com/niri-wm/niri/pull/3366?

ronharel02:
> Wouldn't this be covered by #3366?

I wasn’t aware of #3366 when I opened this, thanks for pointing it out.

The main thing this PR adds beyond it is symbolic layout-independent binds for keys that are not simple first-level symbols in the default layout. With #3366, those cases still seem to require either changing bind resolution globally or writing raw binds, while this PR keeps normal keysym binds unchanged and lets specific binds opt into an explicit reference xkb map, including symbols that require modifiers like `Shift` or `AltGr`.

With that being said, #3366 is much simpler, so I see why it would be favorable.

PR #3802, opened 2026-04-10
Allow auto-raise on focus for floating windows

joske:
Hi All,

My use-case is (different from the typical niri one) to have an all-floating layout. There aren't many wayland compositors outside of GNOME/KDE that offer this. So I was experimenting with different compositors under dank material shell to see which fits my workflow. Niri gets me 90% of the way there, with the only blocking issue so far is the lack of auto-raise on focus (without that, `focus-follows-mouse` is clunky and almost pointless - you still need to click). It works for all kinds of focus changes (on floating windows only).

I've implemented this using the niri event loop for the delay handling. It can be enabled in the config like this:

```
input {
      auto-raise-on-focus delay-ms=300
}
```

The delay is optional and defaults to `0` (immediate).

I've built and tested it locally and it works great. 

I hope you will consider merging this!

### DIFF
--- a/niri-config/src/input.rs
+++ b/niri-config/src/input.rs
@@ -20,6 +20,7 @@ pub struct Input {
     pub disable_power_key_handling: bool,
     pub warp_mouse_to_focus: Option<WarpMouseToFocus>,
     pub focus_follows_mouse: Option<FocusFollowsMouse>,
+    pub auto_raise_on_focus: Option<AutoRaiseOnFocus>,
     pub workspace_auto_back_and_forth: bool,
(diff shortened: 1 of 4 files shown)


megabyte6 (approved):
I haven't personally tested this, but the code looks good to me. I noticed that lines 1309 and 1310 have the same "else" case functionality, so they could be combined, but when I tried it, I realized that the feature only exists in rust edition 2024 😅. The rest of the code seems good 👍

joske:
rebased my branch - no changes

joske:
rebased again.

PR #3813, opened 2026-04-12
config: make overview vertical wheel behavior configurable

arcatva:
Add `wheel-y` and `wheel-shift-y` options to the `overview` config block. Each accepts "workspaces" or "columns", controlling what the mouse wheel navigates in overview mode.

Defaults preserve existing behavior:
  wheel-y "workspaces"      (no modifier scrolls workspaces)
  wheel-shift-y "columns"   (Shift+scroll switches columns)

Example to swap: set wheel-y "columns" and wheel-shift-y "workspaces".


Uzlkav:
The option should get documented in the wiki, but this looks cool!

PEMessage:
+1 on this

PR #3838, opened 2026-04-17
feat(animations): add border color crossfade and gradient angle rotation

beckend:
Two new animations for gradient borders and focus rings.

## border-fade

Smooth crossfade between old and new gradient colors when a window gains or loses focus. Without this animation, the border color changes instantly on focus change.

- 200ms ease-out-cubic, disabled by default
- Works with both focus ring and border gradients
- Solid color borders are unaffected

## border-angle

Continuously rotates the gradient direction around borders, producing a spinning gradient effect.

- 2s linear loop, disabled by default
- Angle offset computed from wall-clock time (no animation restart artifacts)
- Requires gradient borders to be configured

> [!WARNING]
> Enabling border-angle will cause niri to constantly render new frames at the monitor's refresh rate. This may increase CPU/GPU usage and impact battery life on laptops.

## Config

```kdl
animations {
    border-fade {
        duration-ms 200
        curve "ease-out-cubic"
    }
    border-angle {
        duration-ms 2000
        curve "linear"
    }
}
```

## Implementation

- Border fragment shader extended with `color_from2`, `color_to2`, `gradient_lerp` uniforms for dual-gradient rendering
- `FocusRing` tracks previous gradient for crossfade source and accepts angle offset parameter
- `Tile` detects active-state transitions to start fade animation; computes angle offset from elapsed time
- `BorderRenderElement` gets `new_with_fade`/`update_with_fade` methods
- `are_transitions_ongoing` hooks into continuous frame rendering when border-angle is enabled

## Testing

- All 194 tests pass (including 200k randomized layout tests)
- Tested by hand with 6px gradient borders, red→blue active, gray inactive
- `cargo fmt` and `cargo clippy --all-features --all-targets -- -D warnings` clean


beckend:
https://github.com/user-attachments/assets/a8f9d70e-497d-43ee-816a-b03e3cdd2ca0

PR #3853, opened 2026-04-18
feat: add tabbed columns window switch animation

RayZh-hs:
Niri has awesome animations, but currently column tab switching is a rare exception. I mentioned this in a [discussion here](https://github.com/niri-wm/niri/discussions/3814) and got some community interest, so I went on and implemented it. The implementation is straightforward and I strove to make the functions descriptive and central. I will place a video here to demonstrate what it looks like.

https://github.com/user-attachments/assets/ef9a2158-bcd5-4d6d-8a8b-96293dbe4a83

All tests passed.

### DIFF
--- a/docs/wiki/Configuration:-Animations.md
+++ b/docs/wiki/Configuration:-Animations.md
@@ -30,6 +30,12 @@ animations {
         curve "ease-out-quad"
     }
 
+    tab-switch {
+        duration-ms 150
+        curve "ease-out-expo"
+        direction "vertical"
+    }
+
     horizontal-view-movement {
         spring damping-ratio=1.0 stiffness=800 epsilon=0.0001
     }
@@ -317,6 +323,32 @@ animations {
     }
 }
 ```
+
+#### `tab-switch`
+
+Animation used when switching the active window inside a tabbed column.
+
+This animation is separate from `window-movement`: it only applies to changes of the active tab,
+for example when using `focus-window-up`, `focus-window-down`, or direct index-based tab focus.
+
+```kdl
+animations {
+    tab-switch {
+        duration-ms 150
+        curve "ease-out-expo"
+        direction "vertical"
+    }
+}
+```
+
+##### `direction`
+
+Controls the axis of the tab-switch animation.
+
+Possible values:
+
+- `"vertical"`: tabs move up and down.
+- `"horizontal"`: tabs move left and right.
 
 #### `window-resize`
 
--- a/docs/wiki/Configuration:-Animations.md
+++ b/docs/wiki/Configuration:-Animations.md
@@ -30,7 +30,7 @@ animations {
         curve "ease-out-quad"
     }
 
-    tab-switch {
+    column-tab-switch {
         duration-ms 150
         curve "ease-out-expo"
         direction "vertical"
@@ -324,7 +324,7 @@ animations {
 }
 ```
 
-#### `tab-switch`
+#### `column-tab-switch`
 
 Animation used when switching the active window inside a tabbed column.
 
@@ -333,7 +333,7 @@ for example when using `focus-window-up`, `focus-window-down`, or direct index-b
 
 ```kdl
 animations {
-    tab-switch {
+    column-tab-switch {
         duration-ms 150
         curve "ease-out-expo"
         direction "vertical"
@@ -343,7 +343,7 @@ animations {
 
 ##### `direction`
 
-Controls the axis of the tab-switch animation.
+Controls the axis of the column-tab-switch animation.
 
 Possible values:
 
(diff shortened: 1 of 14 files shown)


Atan-D-RP4:
I think the naming is a bit confusing/misleading.

When I read the description, I immediately thought it was about the recent-windows switcher instead of the grouped windows tabs. Maybe consider changing that. Otherwise, pretty cool to finally see this.

RayZh-hs:
What about changing the config key from "tab-switch" to "column-tab-switch", or using "focus-in-column" (from the action name "focus-window-in-column")? Myself I prefer "column-tab-switch" since it is consistent with "workspace-switch".

Atan-D-RP4:
`column tab switch` works. Will look at the rest of the code and testing it out later.

RayZh-hs:
I conducted the naming refactor and made a small change to the tile animation calculation logic (only to make the code more concise, does not affect the result). Now code is reused when processing horizontal and vertical movement.

RayZh-hs:
I rebased the branch to the latest main branch. Hope this helps!

In the meanwhile, I got an issue report stating that the animation is not available in full screen mode. I am addressing this issue.

RayZh-hs:
Issue addressed. Turns out I misunderstood column handling of maximized and full screen windows. This animation can safely be applied to these scenarios as well.

RayZh-hs:
Another visual issue I encountered: If the active window happens to come last in a tabbed column, dispelling it changes focus to another window in the column, causing the newly focused window to pop to the front without animation. This is extremely evident when the new window happens to be fully opaque.

Since the user has its focus on that tile, I propose to let it remain so after the dispel. This way, the camera will follow the dispelled tile, resolving the animation issue. This will also match most current niri actions, in which active tiles remain unchanged unless it is the sole purpose of the action.

The decision is up to the community and to the maintainers. I have not included this fix in any of my commits, but I can provide a diff to achieve this. (enact after commit [47fc229](https://github.com/niri-wm/niri/pull/3853/commits/47fc2297c1de1ee2066dc3d12f03b4cc660f6db0))

```
diff --git a/src/layout/scrolling.rs b/src/layout/scrolling.rs
index f9d1f098..9d95ffb5 100644
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -2013,6 +2013,7 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         }
 
         let source_tile_idx = source_column.tiles.len() - 1;
+        let source_tile_was_active = source_column.active_tile_idx == source_tile_idx;
 
         let mut offset = Point::from((source_column.render_offset().x, 0.));
         let prev_off = source_column.tile_offset(source_tile_idx);
@@ -2023,7 +2024,7 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         self.add_tile(
             Some(target_col_idx),
             removed.tile,
-            false,
+            source_tile_was_active,
             removed.width,
             removed.is_full_width,
             Some(self.options.animations.window_movement.0),
```

Here are videos comparing the two behaviors:

Original (notice how opaque windows seem to "pop out")

https://github.com/user-attachments/assets/a711b648-4f35-4293-a247-6bcbaab55149

Proposed One (focus follows expelled window if it was active)

https://github.com/user-attachments/assets/a954752d-2072-487e-b2d3-d26ec0651626

Uzlkav:
In regards to the special case of expelling a window that is at the bottom of a tabbed column - I don't think the behaviour should change. Few reasons:
- Changes the existing behaviour
- Feels like it violates [this](https://niri-wm.github.io/niri/Development%3A-Design-Principles.html#the-focused-window-should-not-move-around-on-its-own) design principle, as the view is shifted to the right for `expel-window-from-column`, which doesn't indicate that there would be any focus change
- I have a feeling the jumpiness could be adressed better by also calling the animation when column focus changes as a result of `expel`

This looks very cool though, thanks for your work on this!

RayZh-hs:
To enact the animation while the bottom tile moves to the side, part of the wallpaper would be shown since the tiles move in different directions and cannot cover the screen. Would this be acceptible?

Uzlkav:
In my opinion yes, as you have something similar when expelling from a non tabbed column and it's not particularly noticeable there.

RayZh-hs:
@Uzlkav The new commit invokes column animations when expelling the active window from a column. Here is a demo.

https://github.com/user-attachments/assets/18341b63-e0be-4f84-bd89-cefc134b3994

I also noticed a minor visual issue with kde blurring when I rebased so I addressed that too.

Uzlkav:
Cool, looks good, I'll test with the latest changes when I have time

RayZh-hs:
> Cool, looks good, I'll test with the latest changes when I have time

Rebased to latest dd75865f547f0eac0e9b6c4d86d2cd00c0744252, and I added a minor fix for a rare condition where stale column widths resulted in a display issue. Awaiting your approval @Uzlkav

PR #3857, opened 2026-04-19
wiki/getting-started: replace dms-shell references for arch

urayde:
DMS has moved into the official `extra` repository, and the AUR package was already removed.


YaLTeR:
Thanks

PR #3862, opened 2026-04-20
feat: more matchers for layer-rules - pre-requisite for layer animations

Atan-D-RP4:
Adds anchor, anchor-sides, keyboard-interactivity, and exclusive-zones matchers.
`niri msg layers` now also shows anchor, anchor-sides, and exclusive-zones configured for a layer window.

Required for implementing sensible default animations for layer-shell client via layer-rules without niri or users having to hardcoding them woth namespaces.

Wanted but not neccessary for #3481.


docs/wiki/Configuration:-Layer-Rules.md
```diff
@@ -105,6 +109,63 @@ layer-rule {
...
+    match anchor-sides=3
+}
+```
+
+#### `exclusive-zone`
+
+Matches surfaces by exclusive zone behavior.
+Can be:
+- `"exclusive"`: surface reserves screen space
+- `"neutral"`: surface does not reserve screen space
```
HigherOrderLogic, line 148: This is not really clear.

Iirc the protocol allows you to set an arbitrary number as the exclusive zone size, so it's not just "exclusive" and "neutral".

Atan-D-RP4: Yeah. But we don't really care if it's negative since I'm pretty sure that means a "dont-care" in which case you just don't use the matcher at all in the rule.

This is not the case for thr keyboard-interactivity one since we do care if a layer does not have it at all. Or do we not need to?

HigherOrderLogic: > if it's negative since I'm pretty sure that means a "dont-care" in which case you just don't use the matcher at all in the rule.

No, according to the protocol spec:

> If set to -1, the surface indicates that it would not like to be moved to accommodate for other surfaces, and the compositor should extend it all the way to the edges it is anchored to.

Also a layer may change its exclusive zone size after creation so a matcher for specific exclusive zone size maybe desired by some users.

Atan-D-RP4: Hmm, so is it preferrable to do something like this?
```kdl 
exclusive-zone="negative" // or "zero" or "positive"
```

Atan-D-RP4:
@HigherOrderLogic So should I leave the exclusive-zones matcher the way it is or make it more compliant with the spec. Like this, as I previously mentioned:

```kdl 
exclusive-zone="negative" // or "zero" or "positive"
```

Atan-D-RP4:
CC @HigherOrderLogic.
Does the handling of the exclusive-zone matcher look okay now? I just aligned it to the protocol and Smithay's repr of it. Seemed the simple way to do it.

docs/wiki/Configuration:-Layer-Rules.md
```diff
@@ -91,7 +95,7 @@ layer-rule {
 
 #### `layer`
 
-<sup>Since: 26.04</sup>
+<sup>Since: next release</sup>
```
HigherOrderLogic, line 98: Seems like failed rebase.

docs/wiki/Configuration:-Layer-Rules.md
```diff
@@ -105,6 +109,63 @@ layer-rule {
...
+Matches surfaces by their exact anchored edge set.
+This matcher expects a comma-separated list using any of: `top`, `bottom`, `left`, `right`.
+
+Order does not matter:
+
+```kdl
+// Match bars on top edge.
+layer-rule {
+    match anchors="top,left,right"
+}
```
HigherOrderLogic, line 123: I dont really like how this is structured, which relies too much on semantic and trust users to correctly separate it using commas.

docs/wiki/Configuration:-Layer-Rules.md
```diff
@@ -105,6 +109,63 @@ layer-rule {
...
+
+// Same matcher as above.
+layer-rule {
+    match anchors="left,right,top"
+}
+```
+
+#### `anchor-sides`
+
+Matches surfaces by the number of anchored sides.
```
HigherOrderLogic, line 133: The protocol refers to this as "edges", instead of sides.

docs/wiki/Configuration:-Layer-Rules.md
```diff
@@ -105,6 +109,63 @@ layer-rule {
...
+layer-rule {
+    match anchor-sides=3
+}
+```
+
+#### `exclusive-zone`
+
+Matches surfaces by exclusive zone behavior.
+Can be:
+- `"exclusive"`: surface reserves screen space
```
HigherOrderLogic, line 147: Exclusive zone for layers also has exclusive edges, and exclusive zone size. This doesnt cover that, and not sure if Yalter would like to offer such customized matchers.

docs/wiki/Configuration:-Layer-Rules.md
Atan-D-RP4: Fixed

PR #3866, opened 2026-04-20
doc: Describe alternative binds for tab / column window management

benthillerkus:
I couldn't find the wrapping commands inside the documentation at all and only found out they existed by looking at the source code.
Bringing up the hotkey overview screen made me aware of `consume-or-expel` which I found more intuitive than the default `consume`and `expel` commands.

So I added them explicitly to the documentation so that they can hopefully be indexed by search engines.

### DIFF
--- a/docs/wiki/Tabs.md
+++ b/docs/wiki/Tabs.md
@@ -21,7 +21,7 @@ binds {
    // Switch tabs / windows
    Mod+Down       { focus-window-down; }
    Mod+Up         { focus-window-up; }
-   // Wrap when switching tabs past the end the column
+   // Wrap when switching tabs past the end of the column
    Mod+Tab        { focus-window-down-or-top; }
    Mod+Shift+Tab  { focus-windwo-up-or-bottom; }
 
--- a/docs/wiki/Tabs.md
+++ b/docs/wiki/Tabs.md
@@ -15,7 +15,26 @@ binds {
 }
 ```
 
-All other binds remain the same: switch tabs with `focus-window-down/up`, add or remove windows with `consume-window-into-column`/`expel-window-from-column`, and so on.
+All other binds remain the same:
+```kdl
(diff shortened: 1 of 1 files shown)


Uzlkav:
The `consume-or-expel` binds are mentioned in the shortlist of default hotkeys: https://niri-wm.github.io/niri/Getting-Started.html#main-default-hotkeys

Otherwise, the wiki is [not meant](https://niri-wm.github.io/niri/Configuration%3A-Key-Bindings.html#actions) to comprehensively list available actions.

docs/wiki/Tabs.md
```diff
@@ -15,7 +15,26 @@ binds {
...
+   // Switch tabs / windows
+   Mod+Down       { focus-window-down; }
+   Mod+Up         { focus-window-up; }
+   // Wrap when switching tabs past the end of the column
+   Mod+Tab        { focus-window-down-or-top; }
+   Mod+Shift+Tab  { focus-windwo-up-or-bottom; }
+
+   // Add and remove windows
+   Mod+Comma      { consume-or-expel-window-left; }
+   Mod+Period     { consume-or-expel-window-right; }
```
Uzlkav, line 30: Also, it isn't clear that these are _not_ included in the default binds, which could cause confusion.

benthillerkus:
> Otherwise, the wiki is [not meant](https://niri-wm.github.io/niri/Configuration%3A-Key-Bindings.html#actions) to comprehensively list available actions.

How are users supposed to learn about them? Read the source code? As far as I can tell there is no user level documentation besides the wiki

Uzlkav:
By running `niri msg action` as suggested by the wiki in the link I sent?

> Every action that you can bind is also available for programmatic invocation via `niri msg action`. Run `niri msg action` to get a full list of actions along with their short descriptions.

PR #3874, opened 2026-04-22
wiki: explicitly mention possible include methods

Uzlkav:
This explicitly outlines the possible methods for `include` in the docs, particularly the `~` expansion introduced in [404d6dc](https://github.com/niri-wm/niri/commit/404d6dccc475a290fa8b0938ef8524ea2a92575d) that will be available next release.
Not sure if the way I've done this is too verbose, open to changes.

### DIFF
--- a/docs/wiki/Configuration:-Include.md
+++ b/docs/wiki/Configuration:-Include.md
@@ -16,6 +16,12 @@ Settings from included files will be merged with the settings from the main conf
 Included config files can in turn include more files.
 All included files are watched for changes, and the config live-reloads when any of them change.
 
+You can include by filename or path.
+
+* Relative to the current file: `other.kdl` or `./other.kdl`
+* By absolute path: `/path/to/file.kdl`
+* <sup>Since: next release</sup> Home dir paths: `~/file.kdl` expands to `/home/user/file.kdl`
+
 Includes work only at the top level of the config:
 
 ```kdl,must-fail


docs/wiki/Configuration:-Include.md
```diff
@@ -1,11 +1,13 @@
 <sup>Since: 25.11</sup>
 
-You can include other files at the top level of the config.
+You can include other files at the top level of the config, using either relative (`./file` or <sup>since: next release</sup> `~/file`) or absolute (`/path/to/file`) paths.
```
YaLTeR: - all our examples use `file.kdl` without `./`
- `~/` is not relative
- maybe better to split into two sentences or bullet points

Uzlkav: What do you think of this?
```md
You can include other files at the top level of the config, using either:

* Relative paths: `./file.kdl` or `file.kdl`
* <sup>since: next release</sup> Home dir paths, with `~` expanded to `/home/user`: `~/file.kdl`
* Absolute paths: `/path/to/file`

```kdl,must-fail
// Some settings...

include "colors.kdl"

// Some more settings...
```

YaLTeR: Better; let's put `file.kdl` before `./file.kdl` and let's put absolute paths second in the list, and how about "Home dir paths: `~/file.kdl` is expanded to `/home/user/file.kdl`"

Uzlkav: Sounds good, changed.

YaLTeR:
Thanks. I nudged it a bit more to my liking

PR #3889, opened 2026-04-23
config: allow users to set the libinput three-finger-drag option

m4rch3n1ng:
allows enabling the [libinput three-finger-drag option](https://wayland.freedesktop.org/libinput/doc/latest/configuration.html#three-finger-drag) in the config.

i chose `three-finger-drag` for the key and `"disabled"`, `"3fg"` and `"4fg"` for the values. if you prefer i can also add an alias so you can use `"enabled"` instead of `"3fg"` or i can impl `knuffel::Decode` and also allow for enabling with booleans, it's completely up to you.

while libinput 1.28 is still somewhat new, it has been out now for over a year (released 2025-03-24), so i think it should be fine.

this should address at least a few comments in #372, even if it doesn't address the actual issue.


m4rch3n1ng:
ok i hoped this wouldn't happen, but it seems like that option is still too new for ubuntu 24 (i should have expected this, since this came out in 2025) and fedora (i have no idea what the fedora numbers mean).

i'll see if i can find an easy workaround to only actually set the option on systems, where libinput 1.28 is available and i'll otherwise close this pr.

m4rch3n1ng:
it should compile now, but for some reason i can't get it to work on my laptop anymore. maybe i am confused and it never worked, but i am converting it to draft now, as i have prematurely opened this pr and will try and get it to work before undrafting it again.

m4rch3n1ng:
ok it seems like that is an issue on my machine, labwc, hyprland and my own compositor don't work either, even though i know they have worked in the past, so this should be good to go probably and sorry for the hassle.

Issue #3295, opened 2026-01-21
Video memory/GTT climbs when screens powered off


exodist:
Video memory/GTT climb at a rate of up to ~20mb a second when screens are powered off, eventually consuming all system memory (unified memory system, video and system ram are completely shared)

This happens whenvever I lock my screen, but it also happens when I run `niri msg action power-off-monitors`. I can run amdgpu_top in an ssh from another machine and watch as the memory usage just keeps climbing. First it uses all available vram, then it switches to gtt.

When I unlock the system, or move my mouse to turn monitors back on the memory all gets freed.

<details><summary>Config</summary>

config.kdl:
```
input {
    keyboard {
        numlock
    }

    touchpad {
        tap
        natural-scroll
        disabled-on-external-mouse
    }

    trackpoint {
        off
    }

    focus-follows-mouse max-scroll-amount="30%"
}

output "Samsung Display Corp. 0x41B3 0x0000FFA1" {
    mode "1920x1200@120.001"
    scale 1
    transform "normal"
    position x=6560 y=0
    variable-refresh-rate on-demand=true

    hot-corners {
        off
    }

    layout {
        always-center-single-column
        gaps 16
        center-focused-column "never"
        preset-column-widths {
            proportion 1.0
            proportion 0.5
        }

        preset-window-heights {
            proportion 1.0
            proportion 0.5
        }

        default-column-width { proportion 1.0; }
    }
}

output "Dell Inc. DELL U4025QW 72RVC34" {
    modeline 1485.0  5120 5284 5372 5500  2160 2168 2178 2250 "+hsync" "+vsync"
    mode "5120x2160@120"
    scale 1
    transform "normal"
    position x=1440 y=0
    variable-refresh-rate on-demand=true

    focus-at-startup

    hot-corners {
        off
    }

    layout {
        always-center-single-column
        gaps 16
        center-focused-column "never"
        preset-column-widths {
            proportion 0.5
            proportion 1.0
        }

        preset-window-heights {
            proportion 1.0
            proportion 0.5
        }

        default-column-width { proportion 0.5; }

        struts {
            top 256
            bottom 32
            left 512
            right 64
        }
    }
}

output "Dell Inc. DELL P2415Q G3D7F0410KHB" {
    mode "2560x1440"
    scale 1
    transform "270"
    position x=0 y=0

    hot-corners {
        off
    }

    layout {
        always-center-single-column
        gaps 16
        struts {
            top 4
            bottom 4
            left 4
            right 4
[...]

PR #3910, opened 2026-04-26
Fix VRAM leak when monitors are off

phuongdpham:
Per discussion in #3295: with monitors off, several paths kept driving the redraw loop and leaking memory.

This patch breaks the loop by:

- Scheduling the same estimated-vblank throttle `Tty::render` schedules on `NoDamage`, so commit-driven `queue_redraw` doesn't busy-loop.
- Skipping `send_frame_callbacks` while `monitors_active == false` - we never present, so inviting clients to commit just queues buffers that go nowhere.
- Skipping the 1s fallback frame-callback timer while monitors are off - ticking smithay's throttle bookkeeping serves no purpose when nothing is being presented.
- Clearing `unfinished_animations_remain` in the inactive-monitors branch so the estimated-vblank timer doesn't keep re-queuing redraws at refresh-rate cadence.

The throttle helpers moved from `Tty` to `impl Niri` so the inactive-monitors call site doesn't need a backend-type check.

Credit to @sys-rq for empirical testing on i915 + Lenovo T490 that surfaced the three additional drivers; the original patch only addressed the per-commit busy-loop.

## Testing

- `cargo check` / `cargo clippy --all --all-targets` / `cargo +nightly fmt --all -- --check` clean.
- `cargo test --all --exclude niri-visual-tests` - 217 tests pass.
- I'm on NVIDIA RTX 3060 Ti and can't reproduce locally. @sys-rq has reproduced + tested the partial patch; would appreciate a re-test of this updated version.

Refs #3295.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -1792,39 +1792,6 @@ impl Tty {
         }
     }
 
-    fn on_estimated_vblank_timer(&self, niri: &mut Niri, output: Output) {
-        let span = tracy_client::span!("Tty::on_estimated_vblank_timer");
-
-        let name = output.name();
-        span.emit_text(&name);
-
-        let Some(output_state) = niri.output_state.get_mut(&output) else {
-            error!("missing output state for {name}");
-            return;
-        };
-
-        // We waited for the timer, now we can send frame callbacks again.
-        output_state.frame_callback_sequence = output_state.frame_callback_sequence.wrapping_add(1);
-
-        match mem::replace(&mut output_state.redraw_state, RedrawState::Idle) {
-            RedrawState::Idle => unreachable!(),
-            RedrawState::Queued => unreachable!(),
-            RedrawState::WaitingForVBlank { .. } => unreachable!(),
-            RedrawState::WaitingForEstimatedVBlank(_) => (),
-            // The timer fired just in front of a redraw.
-            RedrawState::WaitingForEstimatedVBlankAndQueued(_) => {
-                output_state.redraw_state = RedrawState::Queued;
-                return;
-            }
-        }
-
-        if output_state.unfinished_animations_remain {
-            niri.queue_redraw(&output);
-        } else {
-            niri.send_frame_callbacks(&output);
-        }
-    }
-
     pub fn seat_name(&self) -> String {
         self.session.seat()
     }
@@ -1998,7 +1965,7 @@ impl Tty {
         drop(surface.vblank_frame.take());
 
         // Queue a timer to fire at the predicted vblank time.
-        queue_estimated_vblank_timer(niri, output.clone(), target_presentation_time);
+        niri.queue_estimated_vblank_timer(output.clone(), target_presentation_time);
 
         rv
     }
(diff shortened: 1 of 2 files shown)


YaLTeR:
well this clearly looks kinda cursed with that check and pub(crate)

phuongdpham:
Refactored: both functions moved to `impl Niri`, no more `pub(crate)` or `Backend::Tty` check.

Heads up: with the throttle now backend-agnostic, `monitors_active=false` can leave Winit's `redraw_state` at `WaitingForEstimatedVBlank{,AndQueued}`, which `Winit::render` still marks `unreachable!()` (`winit.rs:289-290`). Not reachable in practice (no screencast on Winit), but flagging since the precondition shifted.

YaLTeR:
I'm gonna have to take a good look at this and at surrounding code to verify that it's right. Sometime in the future

phuongdpham:
Thank you.

bczhc:
This also fixes #3691. Though in that discussion the issue is RAM climbing but not VRAM, which is considered a bit different symptom to issue #3295. But anyway this also fixes 3691!

sys-rq:
Tested on lenovo t490/i915 + arch + niri-git + 3910. Still observing the leak leading to eventual overnight OOMs. First time observed on Dec 12 (though it might be that I wasn't updating the system for up to 2 weeks at the time, take it as an approx date). Powering off monitors frequently (Mod+Shift+P { power-off-monitors; }). Tried to downgrade to the oldest available locally cached version (25.08-2), did not helped, `free -h` reports +2Gi increase in `shared` every 15 mins with monitors off.

sys-rq:
What helped:
  • prevented send_frame_callbacks() while monitors_active == false
  • disabled the fallback frame callback timer while monitors are off.
  • clearing unfinished_animations_remain in the monitors-off redraw branch so the estimated-vblank timer doesn’t keep scheduling animation redraw

phuongdpham:
@sys-rq Thanks for the testing and the diagnosis - all three paths you identified check out against the code, and they explain why my original patch didn't close the leak (it only stopped the per-commit busy-loop, not the timer-driven re-queuing).

Folded all three into this PR (force-pushed, see updated description):
- Skip `send_frame_callbacks` while `monitors_active == false`
- Skip the 1s fallback frame-callback timer while monitors are off
- Clear `unfinished_animations_remain` in the inactive-monitors branch

If you have time, a re-test on the new revision would be much appreciated.

as3ii:
Per my own tests with AMD iGPU, commit [3cfd6b2](https://github.com/niri-wm/niri/commit/3cfd6b2d6add9c071b6f75c5af6ba039d9aa34f4) fixed the VRAM leak related to monitor powered off. In the next hours/days I'll test the latest commit

phuongdpham:
> Per my own tests with AMD iGPU, commit [3cfd6b2](https://github.com/niri-wm/niri/commit/3cfd6b2d6add9c071b6f75c5af6ba039d9aa34f4) fixed the VRAM leak related to monitor powered off. In the next hours/days I'll test the latest commit

Thank you @as3ii.

sys-rq:
looks promising, tested for 1h with monitor off and, just in case, locked but turned on, `free -h` is not showing increase in `shared`

phuongdpham:
Thanks for re-testing - good to have confirmation on the hardware where the leak was reproducible. Appreciate the legwork on the original diagnosis too.

profanum429:
Just add some more tests, with my 9070XT, commit [3cfd6b2](https://github.com/niri-wm/niri/commit/3cfd6b2d6add9c071b6f75c5af6ba039d9aa34f4) has fixed the VRAM leak I was experiencing. Testing overnight I no longer see an ever increasing amount of VRAM usage, looking at it with amdgpu_top.

as3ii:
why close this PR? Is the problem fixed on the main branch?

phuongdpham:
> why close this PR? Is the problem fixed on the main branch?

I don't know. It's too long to merge so I decide to close it.

PR #3920, opened 2026-04-27
Add an option to disable VT switch handling

KosmX:
I work with a lot of VMs, and my machine is configured well, I do never use VT switching on the host, I do it quite often in VMs. 

This option lets me disable the VT switching shortcuts, and let the current apps handle the key presses.  

The option defaults to the old behavior, so I'm not expecting it to break anything.


Uzlkav:
Please also edit the docs to mention this option

PR #3931, opened 2026-04-28
Feat: Optional actions in the screenshot notification

faetalize:
EDIT: This PR has now shifted into changing the user facing configuration and providing a new `screenshot` namespace.

This new feature lives in `screenshot { notification {} }` now.

`screenshot-path` has been moved to `screenshot { path ""; }`, while retaining backwards compat for now.
___

Adds the ability to configure the screenshot notification with actions.

Supports `{path}` substitution for the saved screenshot path.

For now, actions are only shown for screenshots saved to disk, since `{path}` is only available in that case - we could optionally also use clipboard content for cases that involve no persistence (maybe)

Preview:

image


src/utils/mod.rs
```diff
@@ -589,9 +608,108 @@ pub fn show_screenshot_notification(image_path: Option<&Path>) -> anyhow::Result
...
 }
 
+#[cfg(feature = "dbus")]
+async fn handle_screenshot_notification_actions(
+    conn: zbus::Connection,
+    notification_id: u32,
+    image_path: PathBuf,
+    actions: Vec<ScreenshotNotificationAction>,
+) -> anyhow::Result<()> {
+    use anyhow::Context as _;
```
HigherOrderLogic: This is already imported at the top.

docs/wiki/Configuration:-Miscellaneous.md
```diff
@@ -135,6 +135,21 @@ You can also set this option to `null` to disable saving screenshots to disk.
 screenshot-path null
 ```
 
+### `screenshot-notification`
+
+<sup>Since: 26.04</sup>
```
bluelinden: 26.04 has already been released, this should be changed to Since: Next release

bluelinden (approved):
LGTM

docs/wiki/Configuration:-Miscellaneous.md
faetalize: Fixed

src/utils/mod.rs
faetalize: Fixed

src/utils/mod.rs
```diff
@@ -589,9 +607,107 @@ pub fn show_screenshot_notification(image_path: Option<&Path>) -> anyhow::Result
...
+                    continue;
+                }
+
+                if let Ok(idx) = action_key.parse::<usize>() {
+                    if let Some(action) = actions.get(idx) {
+                        let path = image_path.to_string_lossy();
+                        let command = action
+                            .command
+                            .iter()
+                            .map(|arg| arg.replace("{path}", &path))
```
HigherOrderLogic: Blindly replacing `{path}` is a bad idea, since user may do something like `echo ${path}`.

src/utils/mod.rs
```diff
@@ -567,9 +570,24 @@ pub fn show_screenshot_notification(image_path: Option<&Path>) -> anyhow::Result
...
+        Vec::new()
+    };
 
-    conn.call_method(
+    let mut action_strings = Vec::with_capacity(actions.len() * 2);
+    for (idx, action) in actions.iter().enumerate() {
+        action_strings.push(idx.to_string());
+        action_strings.push(action.label.clone());
+    }
+    let action_refs: Vec<_> = action_strings.iter().map(|s| s.as_str()).collect();
```
HigherOrderLogic: Probably can use something like `flat_map` to prevent an extra allocation.

src/utils/mod.rs
```diff
@@ -567,9 +570,24 @@ pub fn show_screenshot_notification(image_path: Option<&Path>) -> anyhow::Result
         }
     }
 
-    let actions: &[&str] = &[];
+    let actions = if image_path.is_some() {
+        actions
+            .iter()
+            .filter(|action| !action.command.is_empty())
```
HigherOrderLogic, line 577: Maybe filter this while parsing, tho empty commands may make sense for aesthetic purpose.

src/protocols/foreign_toplevel.rs
```diff
@@ -18,12 +20,12 @@ use smithay::reexports::wayland_server::{
     Client, DataInit, Dispatch, DisplayHandle, GlobalDispatch, New, Resource,
 };
 use smithay::wayland::shell::xdg::{
-    ToplevelState, ToplevelStateSet, XdgToplevelSurfaceRoleAttributes
+    ToplevelState, ToplevelStateSet, XdgToplevelSurfaceRoleAttributes,
 };
 
 use crate::niri::State;
-use crate::window::mapped::MappedId;
```
HigherOrderLogic, line 25: Seems like unintentional changes. Maybe you need to run `cargo fmt +nightly`.

src/utils/mod.rs
```diff
@@ -589,9 +607,107 @@ pub fn show_screenshot_notification(image_path: Option<&Path>) -> anyhow::Result
...
+                };
+                let (id, action_key): (u32, String) = message
+                    .body()
+                    .deserialize()
+                    .context("error parsing ActionInvoked signal")?;
+                if id != notification_id {
+                    continue;
+                }
+
+                if let Ok(idx) = action_key.parse::<usize>() {
```
HigherOrderLogic, line 682:
```suggestion
                if let Ok(idx) = action_key.parse() {
```

The compiler probably can infer this.

src/utils/mod.rs
faetalize: addressed in f0e6eb323c3d20ca7a7b89a724fad13bdada7045

src/utils/mod.rs
faetalize: I changed it to:

```rust
let action_strings: Vec<Cow<'_, str>> = actions
    .iter()
    .enumerate()
    .flat_map(|(idx, action)| [idx.to_string().into(), action.label.as_str().into()])
    .collect();
```
and pass action_strings directly to Notify.

fixed in 64172ea0a750449d00abd3d1495638601a12ff60

src/protocols/foreign_toplevel.rs
faetalize: This was actually caused by that command

src/utils/mod.rs
faetalize: This is fine for now

src/utils/mod.rs
faetalize: It can't infer it, cargo complains

slice.get(idx) is generic over slice index types

src/protocols/foreign_toplevel.rs
HigherOrderLogic: Kinda weird, I made changes to this file recently and the CI was green.

But nevermind I guess.

PR #3937, opened 2026-04-29
Make niri-session posix-compliant

aniruddhgutta:
`niri-session` currently has a hard dependency on `bash`, and `mkdir --parents` is not posix compliant either. This pr fixes both issues.

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -15,7 +15,7 @@ if [ -n "$SHELL" ] &&
    ! (echo "$SHELL" | grep -q "false") &&
    ! (echo "$SHELL" | grep -q "nologin"); then
   if [ "$1" != '-l' ]; then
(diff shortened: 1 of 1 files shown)


resources/niri-session
```diff
@@ -15,7 +15,7 @@ if [ -n "$SHELL" ] &&
    ! (echo "$SHELL" | grep -q "false") &&
    ! (echo "$SHELL" | grep -q "nologin"); then
   if [ "$1" != '-l' ]; then
-    exec bash -c "exec -l '$SHELL' -c '$0 -l $*'"
+    exec "$SHELL" -l -c "exec $0 -l $*"
```
YaLTeR, line 18: It does not work on other shells: https://github.com/niri-wm/niri/pull/970#issuecomment-2589012568

aniruddhgutta: oh wow ok, thanks

PR #3938, opened 2026-04-29
fix: link to hyprland scrolling layout docs

canmi21:
Hyprland wiki move their layout docs uri path, so i fix it. btw I love Niri more than hypr :(


YaLTeR:
Thanks. The Hyprland breaking changes reached the niri readme lol

PR #3942, opened 2026-04-30
Implement ext-image-capture-source and ext-image-copy-capture protocols (staging)

### LINKED ISSUE #1558: Implement ext-image-copy-capture

powersemmi:
Implement ext-image-capture-source and ext-image-copy-capture with output, window, and cursor capture, over shm and dmabuf buffers.

Capture is damage-driven, following how wlr-screencopy and the PipeWire screencasts work in niri: sessions are tracked compositor-side, frames are rendered on output redraw and only complete when the source actually changed (the first frame completes right away), and constraints follow output mode changes and window resizes. Sessions are stopped when their output is disconnected or their window closes. Output frames keep the output's native orientation and deliver the output transform with each frame. The capture globals are filtered with `client_is_unrestricted`, so sandboxed clients have to go through the portal.

Window capture goes through `ext_foreign_toplevel_image_capture_source_manager_v1` with a small custom manager, since Smithay's own `ToplevelCaptureSourceState` is tied to Smithay's foreign-toplevel-list state, while niri has its own.

Cursor capture delivers the cursor image in Argb8888 frames that complete when the image itself changes, with movement, hotspot and enter/leave sent through session events; the enter state uses the image-intersection semantics from the protocol spec (a hidden cursor counts as left). The cursor rendering and sizing logic is adapted from @pgaskin's #4554, as are the dmabuf format ordering (stable renderer order with Xrgb8888/Argb8888 first, since some clients always take the first advertised format) and reporting the primary render node on the winit backend.

Everything is tested locally end to end:

- A dedicated ext-image-copy-capture test client against a nested niri: constraint negotiation, prompt first frame with valid contents, damage-waiting for subsequent frames, frame abort recovery, window capture, cursor capture with position/hotspot events, and session stop on window close.
- dmabuf capture on a real TTY session on a multi-GPU machine (two AMD cards (rx9070xt+rx6700xt)): the advertised render node is the right one of the two, and a client-side gbm allocation with a tiled AMD modifier renders correctly. The same test also passes nested on winit against the EGL device's render node.
- A real client: [RemoteWay](https://github.com/powersemmi/RemoteWay) streaming a nested niri over its ext-image-capture backend, watched interactively correct orientation via the transform event, cursor composited through `paint_cursors`, no traffic on an idle screen, and full-rate streaming during animations.
- scran was confirmed working by @iciclejj. Arbitrary shm strides are handled since the `render_to_shm` stride fix landed in main.

Known limitations:

- Windows on hidden workspaces don't deliver frames until they're visible again, same as window PipeWire casts.

Overlaps with #4554, which implements output and cursor capture (plus IPC integration) but not window capture.

Closes #1558


powersemmi:
Hi! I'd appreciate some help from the maintainer with reviewing and
   possibly refactoring this PR.

   A few things I'm unsure about and would love feedback on:

   1. The `ToplevelImageCaptureManagerState` in
   `src/protocols/toplevel_image_capture_source.rs` since niri has its own
   `ext_foreign_toplevel_list_v1` implementation (not using smithay's
   `ForeignToplevelHandle`), I wrote a custom handler. It works, but maybe there's a
   cleaner way to integrate it with the existing foreign_toplevel state.

   2. The synchronous rendering in `ImageCopyCaptureHandler::frame()` currently it
   renders directly in the Wayland dispatch handler via
   `self.backend.with_primary_renderer()`. This might have performance implications or
   issues with renderer state. Should the rendering be deferred to the next output
   redraw instead?

   3. The frame constraints and rendering for toplevel/window capture I'm computing
   the window bbox size on the fly, but maybe this should be cached or handled
   differently.

   Also, a heads-up: this PR was primarily created to test my tool
   [RemoteWay](https://github.com/powersemmi/RemoteWay) a remote desktop solution
   that uses these protocols. So while the implementation works for the basic use case,
    it could definitely use a proper review and cleanup from someone more familiar with
    niri's internals.

   Any guidance would be greatly appreciated! 🙏

WhyNotHugo:
Kind bump? Could really make use of this as a continuation of wlr-screencopy.

AlphaNecron:
Bump, this would eliminate the need of GNOME portal so we can hop on `luminous`.

Sempyos:
YaLTeR is very busy, and there is no timeline for PRs/reviews. You are more than welcome to review and test PRs, but please don't leave bump comments

Quackdoc:
Tested this with wl-screenrec which failed with 
```
RUST_BACKTRACE=1 ./target/release/wl-screenrec --output DP-1 --codec av1  -b "0.5 MB" --no-damage -m 60 -f /tmp/record.webm
20:51:48 [WARN] dri device could not be auto-detected, using /dev/dri/renderD128. Pass --dri-device if this isn't correct or to suppress this warning
20:51:48 [ERROR] Failed to negotiate format: failed to select a viable capture format. This is probably a bug. Availabe capture formats are []

thread 'main' (80566) panicked at src/main.rs:1401:22:
unwrap on non-complete EncConstructionStage
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: wl_screenrec::State<S>::on_copy_fail
   3: wayland_client::event_queue::queue_callback
   4: wayland_client::event_queue::EventQueue<State>::dispatching_impl
   5: wl_screenrec::execute
   6: wl_screenrec::main
```
and scran which just segfaulted. no idea why never used it before now.

powersemmi:
Reworked this. Checked it with a small test client against a nested niri, output and toplevel capture both work end to end.

@Quackdoc format negotiation should work now, would appreciate a re-test with wl-screenrec if you get a chance.

iciclejj:
Can confirm that scran captures correctly now :) (after cherry-picking the patch from my PR above)

Not really comfortable enough with Rust to give this branch a review, but hope it gets merged eventually.

Also, you should probably add a commit to allow buffers with `stride > row_bytes` if/when my PR goes through, since `ext-image-copy-capture` allows the client to provide `wl_buffer`s with arbitrary strides, as opposed to `wlr-screencopy`, where the compositor/niri enforced it ([zwlr_screencopy_frame_v1::buffer](https://wayland.app/protocols/wlr-screencopy-unstable-v1#zwlr_screencopy_frame_v1:event:buffer)).

pgaskin:
Some things I noticed while comparing this against my own implementation (which I've been using since mid-June, and working on cleaning up into a PR right now):

- This doesn't implement cursor capture (on the other hand, mine doesn't implement toplevel capture).
- This doesn't handle filtering properly (you use `{OutputCaptureSourceState,ImageCopyCaptureState}::new::<State>()` without `client_is_unrestricted`) which allows clients to bypass sandboxing (you did remember to use it for `ToplevelImageCaptureManagerState`.
- This doesn't emit BufferConstraints when the format/stride mismatches (it only handles dimensions).
  - Fixing this also requires fixing `render_to_shm` to use the stride instead of assuming `w * 4`, or clients may renegotiate indefinitely.
- This doesn't call `ImageCopyCaptureState::cleanup` when dropping sessions, leaking `SessionRef`s.
- Only shm capture works on the TTY backend, dmabuf doesn't.
- The dmabuf format list is not sorted consistently (this will cause many clients to behave erratically or choose a subpar format).
- This implementation pre-transforms the output instead of exposing it to the client.

pgaskin:
> @Quackdoc: Tested this with wl-screenrec which failed with [...] `dri device could not be auto-detected` [...] `Availabe capture formats are []`

This is due to Mesa clearing the EGLDevice in `_eglGetGbmDisplay` when a second EGLDisplay is created for the same GBM device, which is done in the TTY backend.

https://github.com/intel-lgci-fdo-gitlab-mirror/mesa.mesa/blob/0ba4b08edc65075e9346d20d5310261939aaaf48/src/egl/main/egldisplay.c#L568-L601

Try my image-copy-capture implementation in #4554, it should work correctly.

powersemmi:
@pgaskin thanks for the thorough review, went through all of it.

Fixed: both capture globals are now filtered with `client_is_unrestricted` like the toplevel manager, `ImageCopyCaptureState::cleanup()` is called on refresh so stale sessions don't leak, and the output is no longer pre-transformed — buffers keep the native orientation and the transform goes to the client with each frame.

On stride/format: the `render_to_shm` stride fix landed in main (thanks @iciclejj), and I've merged main into the branch, so arbitrary strides work now. Format mismatches are rejected against the session constraints, size mismatches fail with `buffer_constraints`.

The dmabuf format list also keeps a stable order now. It used to come out of a HashMap, so the order changed on every constraints update. I took your approach of preserving the renderer's order with Xrgb8888/Argb8888 first. I suspect the erratic dmabuf behavior on TTY may have been this; I can't test on real hardware here, so a re-test would be appreciated.

Cursor capture is in as well. I adapted the cursor rendering/sizing/hotspot logic from your #4554 (credited in the commit), so at this point the branch covers output, window, and cursor capture.

powersemmi:
Update: cursor capture is implemented as well now, and the whole branch has been tested locally end to end.

Besides the protocol test client (constraints, damage-waiting, aborts, window capture, cursor position/hotspot events, dmabuf via gbm with a tiled AMD modifier), I ran a real screen-streaming client [RemoteWay](https://github.com/powersemmi/RemoteWay) against a nested niri over ext-image-capture and watched the stream interactively: orientation via the transform event, cursor via `paint_cursors`, zero traffic on an idle screen, full rate during animations. That exercise also shook out a few incorrect assumptions on the client side (fixed in powersemmi/RemoteWay#1), the compositor side held up as is.

What's left from my end is a sanity run of dmabuf capture on a real TTY session, ideally multi-GPU the path itself is verified on winit against the advertised render node.

powersemmi:
Ran the remaining sanity check: dmabuf capture on a real TTY session, on a multi-GPU machine (RX 9070 XT + RX 6700 XT). The compositor advertises the correct render node of the two, and a client-side gbm allocation with a tiled AMD modifier captures real frame content. Output, window, and cursor capture all pass the same test suite there as on winit.

The TTY run also caught one small thing, fixed in the latest commit: before the first pointer motion niri keeps the cursor hidden, but cursor sessions still reported enter with an empty image a hidden cursor now counts as left.

Nothing known to be outstanding at this point.

pgaskin:
Okay, this is just hilarious now. Are you a human or an agent's puppet?

YaLTeR:
I think we're going with https://github.com/niri-wm/niri/pull/4554 for the base ext image capture stuff

powersemmi:
> Okay, this is just hilarious now. Are you a human or an agent's puppet?

My English is not so good, so I'm asking ai to write the text for me. 
UwU

But to be honest, I've used ai for some questions because I'm not so immersed in the project.

PR #3943, opened 2026-04-30
input: support explicit modes (timeout, sticky) for drag-lock

zzzsyyy:
Previously, `drag-lock` was a bool flag that enabled the timeout-based drag lock in libinput. However, libinput also supports a "sticky" mode.

This commit changes the `drag-lock` config to accept an optional string argument:
- `drag-lock "timeout"`
- `drag-lock "sticky"`

upstream recommends "sticky" as the preferred 
default mode for drag lock (see: https://lists.freedesktop.org/archives/wayland-devel/2024-November/043860.html). 

but keeping "timeout" as the implicit default here can avoid breaking 
users' existing muscle memory.  should we change it to sticky?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,7 +73,7 @@ fastrand = "2.4.1"
 futures-util = { version = "0.3.32", default-features = false, features = ["std", "io"] }
 git-version = "0.3.9"
 glam = "0.32.1"
-input = { version = "0.10.0", features = ["libinput_1_21"] }
+input = { version = "0.10.0", features = ["libinput_1_27"] }
 keyframe = { version = "1.1.1", default-features = false }
 libc = "0.2.185"
 libdisplay-info = "0.3.0"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -73,7 +73,7 @@ fastrand = "2.4.1"
 futures-util = { version = "0.3.32", default-features = false, features = ["std", "io"] }
 git-version = "0.3.9"
 glam = "0.32.1"
-input = { version = "0.10.0", features = ["libinput_1_21"] }
+input = { version = "0.10.0", features = ["libinput_1_27"] }
```
YaLTeR, line 76: Do we want to bump the requirement? How's distro support?

zzzsyyy: the sticky draglock mode was only added in libinput 1.27, and most current stable distros released already ship libinput 1.27 or newer, maybe we need a runtime checker?

YaLTeR: Which distros in particular and which not?

zzzsyyy: Fedora 43/44, Ubuntu 26.04 LTS, Debian 13 Stable/Testing are, while Ubuntu 24.04 LTS, Debian 12 not

ref: [repology](https://repology.org/project/libinput/versions)

PR #3951, opened 2026-05-01
Add the option to disable mouse in the MRU

wadsaek:
This pull requests adds an option to disable mouse controls in the window switcher.

I oftentimes use niri with a graphic tablet, and in this case having mouse control when alt-tabbing is really unintuitive because if your cursor just so happens to be on your current window the window won't switch. In this case you need to go out of your way to move the cursor out of the switcher area and only then start alt-tabbing, which is not the most pleasant experience. Hope this isn't too user specific to be accepted

### DIFF
--- a/src/ui/mru.rs
+++ b/src/ui/mru.rs
@@ -1028,6 +1028,11 @@ impl WindowMruUi {
         let UiState::Open(inner) = &mut self.state else {
             return None;
         };
+
+        if !inner.config.borrow().recent_windows.enable_mouse{
+            return None;
+        }
+
         // Don't handle pointer until the UI is visible.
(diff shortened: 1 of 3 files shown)


src/ui/mru.rs
```diff
@@ -1028,6 +1028,11 @@ impl WindowMruUi {
         let UiState::Open(inner) = &mut self.state else {
             return None;
         };
+
+        if !inner.config.borrow().recent_windows.enable_mouse{
```
HigherOrderLogic: Probably need a `rustfmt` run.

wadsaek: that's a silly thing to forget haha. i ran rustfmt and pushed a new commit

PR #3952, opened 2026-05-01
Recency heuristic for ambiguous output selection

ccjmne:
Discussed in: https://github.com/niri-wm/niri/discussions/2897

This Pull Request changes the algorithm to determine which output to target when going up, down, left or right: it implements some MRU heuristic for all the commands that have to do with choosing an output up, down, left or right.

I also considered making it opt-in, with, say:

```kdl
Mod+Shift+Left  { focus-monitor-left  use-focus-history=true; }
```

But there are *so* many such functions; at least all of these, in 4 directions:

```kdl
Mod+Ctrl+Down          { move-window-down-or-to-workspace-down; }
Mod+Ctrl+WheelScrollUp { move-column-to-workspace-up;           }
Mod+Shift+Ctrl+Left    { move-window-to-monitor-left;           }
Mod+Shift+Left         { focus-monitor-left                     }
Mod+Up                 { focus-window-or-workspace-up;          }
```

And I figured people would generally want to have a single selection scheme for all their output selection commands. In any case, we can discuss this and think about it some more, but I think there's something pretty usable here as is it.

---

Just like before, the candidate outputs are:

- in the correct direction, and
- they are at least partially in the (virtually) infinite "strip"
  obtained from extending the active output either vertically, for up
  and down motions, or horizontally for horizontal movements.

Unlike before, we now rank the candidate outputs by their nearbymost
vertical (or horizontal) edge: this allows us to more reasonably end up
with ties, where the ambiguity that is natural to the human user also
becomes effective in the code.

We curate a stack of the most recently used monitors and use it to
resolve the ties against the candidates that share the closest edge to
the currently active output.

If none of the candidates were ever visited, we fall back to the
previous heuristic, which would find the closest candidate, according
to the distance between their geometric centre and that of the active
output.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -334,7 +334,7 @@ pub trait LayoutElement {
 
 #[derive(Debug)]
 pub struct Layout<W: LayoutElement> {
-    /// Monitors and workspaes in the layout.
+    /// Monitors and workspaces in the layout.
     monitor_set: MonitorSet<W>,
     /// Focused outputs ordered from least to most recently focused.
     output_focus_stack: Vec<Output>,
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -336,6 +336,8 @@ pub trait LayoutElement {
 pub struct Layout<W: LayoutElement> {
     /// Monitors and workspaes in the layout.
     monitor_set: MonitorSet<W>,
+    /// Focused outputs ordered from least to most recently focused.
+    output_focus_stack: Vec<Output>,
     /// Whether the layout should draw as active.
     ///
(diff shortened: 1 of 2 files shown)


ccjmne:
Hey @YaLTeR,
This PR is ready enough (from my perspective), though it may need is some chatting about.
You mentioned in the past that that issue has been brought up before as well, do you think we can look into it? :-)

Sempyos:
Note YaLTeR is very busy, and that there is no timeline for PRs/features/reviews

ccjmne:
Performed maintenance rebase just now, which went without a hitch.

PR #3356, opened 2026-01-31
Add SwitchWorkspace action to change workspace without moving focus

feng-yifan:
## Summary

This PR introduces a new `SwitchWorkspace` action that switches to a workspace on its monitor **without** changing the currently focused monitor. This differs from the existing `FocusWorkspace` action, which moves keyboard focus to the target workspace's monitor.

## Motivation

The current `FocusWorkspace` action is useful when you want to immediately start working on another workspace. However, there are valid workflows where you want to change a workspace on another monitor while keeping your current focus position:

- **Multi-monitor workflow with parallel tasks**: On the primary work monitor, users often run multiple parallel tasks across different workspaces, each requiring auxiliary content from other workspaces on secondary displays. For example:
  - Workspace 1 (coding) + Reference documentation on Monitor 2
  - Workspace 2 (debugging) + Logs on Monitor 2
  - Workspace 3 (code review) + CI status on Monitor 2

  With `SwitchWorkspace`, the secondary monitor can automatically follow the primary monitor's workspace changes, keeping relevant auxiliary content in sync without requiring the user to manually switch and refocus.
- **Preparing a workspace on another monitor** for an upcoming window move, such as organizing windows before a presentation or screen sharing session.
- **Previewing content** on another monitor without interrupting current work, allowing quick glances at reference material or monitoring dashboards.
- **Setting up workspaces** across multiple monitors before starting a complex task that requires specific window arrangements.

## Usage

Via IPC socket:

```bash
# Connect to niri socket and send SwitchWorkspace action
socat STDIO "$NIRI_SOCKET"
{"Action":{"SwitchWorkspace":{"reference":{"Id":2}}}}
```

Via Rust:

```rust
let mut socket = Socket::connect()?;
socket.send(Request::Action(Action::SwitchWorkspace {
    reference: WorkspaceReferenceArg::Id(2),
}));
```

Implementation Details

- niri-ipc: Added SwitchWorkspace variant to the Action enum with WorkspaceReferenceArg parameter
- niri-config: Added SwitchWorkspace action variant and IPC conversion logic
- src/input: Added handler in process_action that finds the output and workspace index, then delegates to the layout
- src/layout: Added switch_workspace_on_output method that switches workspace on a specific output without changing focus

Testing

- Compiled successfully with cargo build --release
- Tested locally on my machine - runs well without any issues
- Ran cargo clippy --all --all-targets - no warnings
- Follows existing code patterns and conventions
- Adds 30 lines across 4 files with focused, minimal changes

[...]

PR #3958, opened 2026-05-02
Add SwitchWorkspace action to change workspace without focusing…

feng-yifan:
## Summary

This PR introduces a new `SwitchWorkspace` action that switches to a workspace on its monitor **without** changing the currently focused monitor. This differs from the existing `FocusWorkspace` action, which moves keyboard focus to the target workspace's monitor.

## Motivation

The current `FocusWorkspace` action is useful when you want to immediately start working on another workspace. However, there are valid workflows where you want to change a workspace on another monitor while keeping your current focus position:

- **Multi-monitor workflow with parallel tasks**: On the primary work monitor, users often run multiple parallel tasks across different workspaces, each requiring auxiliary content from other workspaces on secondary displays. For example:
  - Workspace 1 (coding) + Reference documentation on Monitor 2
  - Workspace 2 (debugging) + Logs on Monitor 2
  - Workspace 3 (code review) + CI status on Monitor 2

  With `SwitchWorkspace`, the secondary monitor can automatically follow the primary monitor's workspace changes, keeping relevant auxiliary content in sync without requiring the user to manually switch and refocus.
- **Preparing a workspace on another monitor** for an upcoming window move, such as organizing windows before a presentation or screen sharing session.
- **Previewing content** on another monitor without interrupting current work, allowing quick glances at reference material or monitoring dashboards.
- **Setting up workspaces** across multiple monitors before starting a complex task that requires specific window arrangements.

## Usage

Via IPC socket:

```bash
# Connect to niri socket and send SwitchWorkspace action
socat STDIO "$NIRI_SOCKET"
{"Action":{"SwitchWorkspace":{"reference":{"Id":2}}}}
```

Via Rust:

```rust
let mut socket = Socket::connect()?;
socket.send(Request::Action(Action::SwitchWorkspace {
    reference: WorkspaceReferenceArg::Id(2),
}));
```

## Develop Details

Implementation Details

- niri-ipc: Added SwitchWorkspace variant to the Action enum with WorkspaceReferenceArg parameter
- niri-config: Added SwitchWorkspace action variant and IPC conversion logic
- src/input: Added handler in process_action that finds the output and workspace index, then delegates to the layout
- src/layout: Added switch_workspace_on_output method that switches workspace on a specific output without changing focus

Testing

- Compiled successfully with cargo build --release
- Tested locally on my machine - runs well without any issues
- Ran cargo clippy --all --all-targets - no warnings
- Follows existing code patterns and conventions
- Adds 30 lines across 4 files with focused, minimal changes

Backward Compatibility

This change is fully backward compatible. It adds a new optional action without modifying any existing behavior.

## Scenarios

I use a three-monitor setup. The allocation of tasks across the screens (from left to right) varies depending on my primary focus. Here’s how I typically organize them in different scenarios:

| Scenario / Work Mode | Left Monitor | Center Monitor | Right Monitor |
| --- | --- | --- | --- |
|Development Workflow | Terminals (Debugging, Git, build tools) | Primary Editor/IDE (Active coding) | Documentation & Reference (API docs, specs) |
|Operations & Admin | Monitoring Terminals (Logs, system dashboards) | Configuration Editor (Scripts, config files) | Procedural Documentation (Runbooks, wikis) |
|Learning & Research | Nothing | Note-taking (Ideas, summaries) | Primary Content (eBooks, papers, course videos) |

This dynamic arrangement allows me to optimize screen real estate for the specific task at hand, maintaining an efficient and focused workflow across different activities.

## Related Info

- Related to the discussion in #3351 about workspace switching behavior.
- It’s a new version of the closed #3356 PR. I’ve changed the branch used for the PR.

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -223,6 +223,7 @@ pub enum Action {
     #[knuffel(skip)]
     FocusWorkspaceUpUnderMouse,
     FocusWorkspace(#[knuffel(argument)] WorkspaceReference),
+    SwitchWorkspace(#[knuffel(argument)] WorkspaceReference),
     FocusWorkspacePrevious,
     MoveWindowToWorkspaceDown(#[knuffel(property(name = "focus"), default = true)] bool),
     MoveWindowToWorkspaceUp(#[knuffel(property(name = "focus"), default = true)] bool),
@@ -511,6 +512,9 @@ impl From<niri_ipc::Action> for Action {
             niri_ipc::Action::FocusWorkspaceUp {} => Self::FocusWorkspaceUp,
             niri_ipc::Action::FocusWorkspace { reference } => {
                 Self::FocusWorkspace(WorkspaceReference::from(reference))
+            }
+            niri_ipc::Action::SwitchWorkspace { reference } => {
+                Self::SwitchWorkspace(WorkspaceReference::from(reference))
             }
             niri_ipc::Action::FocusWorkspacePrevious {} => Self::FocusWorkspacePrevious,
             niri_ipc::Action::MoveWindowToWorkspaceDown { focus } => {
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -480,6 +480,12 @@ pub enum Action {
         #[cfg_attr(feature = "clap", arg())]
         reference: WorkspaceReferenceArg,
     },
+    /// Switch to a workspace on its monitor without changing the focused monitor.
+    SwitchWorkspace {
+        /// Reference (index or name) of the workspace to switch to.
+        #[cfg_attr(feature = "clap", arg())]
+        reference: WorkspaceReferenceArg,
+    },
     /// Focus the previous workspace.
     FocusWorkspacePrevious {},
     /// Move the focused window to the workspace below.
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -1513,6 +1513,16 @@ impl State {
                     self.niri.queue_redraw_all();
                 }
             }
+            Action::SwitchWorkspace(reference) => {
+                if let Some((Some(output), index)) =
+                    self.niri.find_output_and_workspace_index(reference)
+                {
+                    self.niri.layout.switch_workspace_on_output(&output, index);
+                    self.maybe_warp_cursor_to_focus();
+                    self.niri.layer_shell_on_demand_focus = None;
+                    self.niri.queue_redraw(&output);
+                }
+            }
             Action::FocusWorkspacePrevious => {
                 self.niri.layout.switch_workspace_previous();
                 self.maybe_warp_cursor_to_focus();
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -2204,6 +2204,13 @@ impl<W: LayoutElement> Layout<W> {
         monitor.switch_workspace_previous();
     }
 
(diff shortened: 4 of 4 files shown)


Uzlkav:
Wouldn't this use case better served by a `sync-workspaces` type option?

> With SwitchWorkspace, the secondary monitor can automatically follow the primary monitor's workspace changes, keeping relevant auxiliary content in sync without requiring the user to manually switch and refocus.

If I understand the code correctly, this would go from "manually switch and refocus" to "manually switch", which isn't automatic unless you do some scripting.
Instead, a user could just enable a `sync-workspaces` option (like what KDE's default behavior is), and moving up/down on one monitor would also move up/down on the others. Of course, you don't always have the same number of workspaces on each monitor, so this approach would require some design around how to handle that case.

feng-yifan:
> Wouldn't this use case better served by a `sync-workspaces` type option?
> 
> > With SwitchWorkspace, the secondary monitor can automatically follow the primary monitor's workspace changes, keeping relevant auxiliary content in sync without requiring the user to manually switch and refocus.
> 
> If I understand the code correctly, this would go from "manually switch and refocus" to "manually switch", which isn't automatic unless you do some scripting. Instead, a user could just enable a `sync-workspaces` option (like what KDE's default behavior is), and moving up/down on one monitor would also move up/down on the others. Of course, you don't always have the same number of workspaces on each monitor, so this approach would require some design around how to handle that case.

Rather than providing pre-packaged capabilities directly, I believe that as a Wayland composer, it should offer a more fundamental level of abstraction.

Different people may have different habits when using sync-workspaces, so providing a more general underlying capability is a better approach.

I did create a tool to synchronize workspaces, [here](https://github.com/feng-yifan/niri-workspace-manager)

Uzlkav:
I guess I don't personally see much utility in this _other_ than keeping workspaces on two monitors in sync. Of course there may be other use cases, I do see some utility in the "previewing content" case you mentioned above.
Also, it seems like the link you shared is either broken or the repo is private.

feng-yifan:
> I guess I don't personally see much utility in this _other_ than keeping workspaces on two monitors in sync. Of course there may be other use cases, I do see some utility in the "previewing content" case you mentioned above. Also, it seems like the link you shared is either broken or the repo is private.

Do you have multiple monitors? For users with multiple monitors, the ability to freely switch between them without losing focus is a feature that greatly enhances productivity.

Similarly, this feature allows the workspace to be rearranged without losing focus, and it’s effectively useful only for users with multiple monitors. This should be a basic capability of any composer.

I have set the repo to public, but the application is available only in Simplified Chinese.

Uzlkav:
I do, though usually I only have a single workspace on it with a static music player or chat app. Again, I don't think I personally would use this but I _do_ see now how it could be useful to have more fine grained control than just a `sync-monitors` type thing.

feng-yifan:
> I do, though usually I only have a single workspace on it with a static music player or chat app. Again, I don't think I personally would use this but I _do_ see now how it could be useful to have more fine grained control than just a `sync-monitors` type thing.

Yes, the more fundamental a feature is, the greater its potential for customization. Users with different habits can freely combine the functions provided by the composer to improve efficiency, just like in object-oriented programming.

feng-yifan:
@YaLTeR Could we merge this pr onto main branch?

Sempyos:
Note YaLTeR is very busy, and that there is no timeline for PRs/reviews

feng-yifan:
> Note YaLTeR is very busy, and that there is no timeline for PRs/reviews

Is there other people could do something help for things like this?

PR #3962, opened 2026-05-02
prefer spawn over spawn-sh in the default config

Elec3137:
Also mentioned in [the wiki](https://niri-wm.github.io/niri/Configuration%3A-Key-Bindings.html#spawn-sh), `sh` incurs a small overhead. So I don't see why it should be used when it's not needed in the default configuration, as it's an example of (very very minor unimportant but still) bad practice. There is also an example of how to use it effectively with the `orca` bind in the same file.

This is an unimportant nitpick, so of course feel free to close this/ask me to close this if there's a reason I missed, but I wasn't able to find any discussion about this anywhere else.

note: I've run `nix flake check` on this PR and it passed.

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -373,23 +373,22 @@ binds {
 
     // Example volume keys mappings for PipeWire & WirePlumber.
     // The allow-when-locked=true property makes them work even when the session is locked.
-    // Using spawn-sh allows to pass multiple arguments together with the command.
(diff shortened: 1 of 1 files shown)


YaLTeR:
If you check the file history, it used spawn, and then I changed it to spawn-sh when it was introduced. Which should've raised a question on why I did that. The reason is to demonstrate how spawn-sh works in the default config.

PR #3987, opened 2026-05-05
Add disable-pointer-movement option for touchpads

geesawra:
When enabled, the touchpad will not move the pointer, but scrolling and gestures still work.

Can be used by people with trackpoint + touchpad configurations where they want to use the touchpad just for gestures/multitouch/scrolling but not for actual mouse movements.

Feel free to edit or discard this PR completely if you don't think it makes sense for niri to support such a niche scenario!

Disclosure: this PR was implemented with the help of LLMs, I gave it a thorough review to the best of my Rust abilities, on top of testing it on my laptop.


YaLTeR:
Sounds like something for a libinput plugin and not for niri

geesawra:
Fair! I'll close this PR then. 

Thank you!

PR #3989, opened 2026-05-05
Prevent leaving an orphaned shell process when using niri-session

ArtikusHG:
I noticed that when running niri-session, in the last step of the execution process, an `exec` was omitted and as a result niri runs as a child process of a shell, without there being any reason for it as far as I'm aware. This PR helps save about 8 MB of RAM on my system, which I'm sure is very useful for everyone! (/s)

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -15,7 +15,7 @@ if [ -n "$SHELL" ] &&
    ! (echo "$SHELL" | grep -q "false") &&
    ! (echo "$SHELL" | grep -q "nologin"); then
   if [ "$1" != '-l' ]; then
-    exec bash -c "exec -l '$SHELL' -c '$0 -l $*'"
+    exec bash -c "exec -l '$SHELL' -c 'exec $0 -l $*'"
   else
     shift
   fi


sudo-rsingh:
@ArtikusHG This is an interesting find, can you pls share logs showing this.

YaLTeR:
fwiw im pretty sure bash on my system automatically execs the last process

ArtikusHG:
hm, then this might be related to the fact that I use https://github.com/reubeno/brush, which is a bash-compatible shell written in rust. with this shell there is a trailing process left. I will do some further testing, and if it's a problem with that shell, I'll close that

ArtikusHG:
before the change:

```
artikus    47690  1.9  0.0 283208 11896 tty1     Ssl+ 22:06   0:00 -/bin/bash -c /var/home/artikus/.local/bin/niri-session -l
artikus    47884  0.1  0.0 281880  7872 ?        Sl   22:06   0:00 bash /var/home/artikus/.local/bin/makobar
artikus    47897  0.0  0.0 282008  8540 ?        Sl   22:06   0:00 bash .local/bin/battery_manager
artikus    47912  0.9  0.0 349612  8140 ?        Sl   22:06   0:00 bash .local/bin/ctrl_layout
artikus    48202  0.0  0.0 231452  2400 pts/1    S+   22:07   0:00 grep --color=auto bash
```

after the change:

```
ps aux | grep bash
artikus    46515  0.7  0.0 351664  8524 ?        Sl   22:05   0:00 bash .local/bin/ctrl_layout
artikus    46525  0.0  0.0 282008  8860 ?        Sl   22:05   0:00 bash .local/bin/battery_manager
artikus    46535  0.0  0.0 281880  7908 ?        Sl   22:05   0:00 bash /var/home/artikus/.local/bin/makobar
artikus    46995  0.0  0.0 231452  2480 pts/1    S+   22:05   0:00 grep --color=auto bash
```

still unsure if this is a slight logic error in the script or a problem with the rust bash implementation that i'm using. if you guys don't get that `-/bin/bash -c niri-session -l` trailing process when running ps aux on your systems, i will open an issue with the rust bash instead

sudo-rsingh:
```bash
rakshit     1091  0.0  0.0   8824  5552 tty1     S+   May05   0:00 /bin/sh /usr/bin/niri-session -l
```

I have this process running in my device, I use bash.

ArtikusHG:
the trailing process also disappears when i run `exec niri-session -l` directly instead of just `exec niri-session`

sudo-rsingh:
I will checkout and try your fix and get back.
But curious how did you find this?

YaLTeR:
guess i also have it

```
yalter      4694  0.0  0.0 232236  3852 tty2     S+   08:02   0:00 /usr/bin/sh /usr/bin/niri-session -l
```

ArtikusHG:
@sudo-rsingh i have a lot of shell scripts and i was trying to reduce the number of shell processes that are always running in the background because there was like 20 on my system when i checked, so i tracked down all of them, and while some of them were scripting/config errors on my part, while doing that i also noticed the login shell didn't get replaced by the niri process, and that's how i found it

YaLTeR:
ty

PR #4001, opened 2026-05-06
Add wp_pointer_warp_v1 protocol support

### LINKED ISSUE #3988: Add wp_pointer_warp_v1 protocol support
The `wp_pointer_warp_v1` staging protocol (wayland-protocols 1.45) gives Wayland clients
a direct cursor-warp primitive, replacing workarounds based on `zwp_locked_pointer_v1`'s
`set_cursor_position_hint`. KWin (≥6.4), Mutter (≥49), and wlroots (≥0.19) already
implement it.

Smithay merged the server-side handler in [Smithay#1757](https://github.com/Smithay/smithay/pull/1757)
(Aug 2025), and niri's pinned Smithay revision already includes it — so the remaining
work is compositor wiring and the validation policy that Smithay deliberately leaves to
each compositor:

- Call `PointerWarpManager::new::<State>(...)` at startup to advertise the global.
- Implement `PointerWarpHandler::warp_pointer` on `State`. The protocol's "should"
  guidelines:
  - Reject if the enter serial isn't recent for any of the client's surfaces.
  - Reject if the requested position is outside the target surface.
  - Honor when the surface has pointer focus (or an implicit pointer grab).
  - On valid requests, call into niri's existing cursor-positioning code.
- Add the `delegate_dispatch!` glue.

### Motivation

I'm working on the matching Wine client-side support. With this protocol advertised,
Wine apps calling `SetCursorPos` get a direct warp; without it, Wine falls back to the
older lock/hint workaround that has known reliability issues
(see [Wine MR !7353](https://gitlab.winehq.org/wine/wine/-/merge_requests/7353)).
Adding niri support closes that gap end-to-end.

### Could I take this on?

I'd like to attempt the implementation if you're open to it — first-time niri
contributor, so pointers welcome on policy specifics and where the Smithay state
should sit alongside niri's other protocol globals.

zakissimo:
Closes #3988.

Adds support for the `wp_pointer_warp_v1` staging protocol (wayland-protocols 1.45).
Niri's pinned Smithay already includes the server-side handler from
[Smithay#1757](https://github.com/Smithay/smithay/pull/1757); this PR is the compositor
wiring + validation policy + the niri-side helpers.

## Validation policy

Implements the protocol's three "should" rules with **lenient (KWin-style)** semantics:
focus + bounds, no `button_count > 0` requirement. The protocol says "honor it if the
surface has pointer focus, *including* when it has an implicit pointer grab" — "including"
reads as "focus is sufficient; implicit grab is one valid case", not as a requirement.
Most legitimate `SetCursorPos`-class callers (Wine apps, mouselook recentering,
drag-completion, menu navigation) happen outside a button-down, so the strict reading
rejects most of them.

## Reference implementations

(Per @YaLTeR's request in #3988.)

- **KWin** ([commit](https://invent.kde.org/plasma/kwin/-/commit/c8e3f6ca32), Sept 2024).
  Validation in [`src/pointer_input.cpp:152-164`](https://invent.kde.org/plasma/kwin/-/blob/master/src/pointer_input.cpp#L152) — focus + bounds, no grab requirement. Lenient.
- **Mutter** ([commit](https://gitlab.gnome.org/GNOME/mutter/-/commit/96d4e7c3ce), July 2025).
  Validation in [`meta_wayland_pointer_can_warp`](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/wayland/meta-wayland-pointer.c) — adds `button_count > 0`. Strict.

## Implementation overview

- Two free helpers in `src/utils/mod.rs`:
  - `get_surface_toplevel_coords` — walks the subsurface chain accumulating offsets via
    `SubsurfaceCachedState.location`. Parallel to Smithay's existing `get_popup_toplevel_coords`
    (which handles xdg_popup chains).
  - `get_surface_size` — reads `RendererSurfaceState::surface_size()`.
- `impl PointerWarpHandler for State` in `src/handlers/mod.rs`: serial check via
  `pointer.last_enter()`, surface-local bounds check, `find_window_and_output` →
  window-rect → output-global four-term composition, `self.move_cursor(target)`.
- New per-window `window_visual_rectangle(id) -> Option<Rectangle<f64, Logical>>` at all
  four layout layers (floating/scrolling/workspace/monitor), parallel to the existing
  `active_window_visual_rectangle`. Dispatches monitor → workspace → (floating | scrolling).
- Field + init + delegate macro for `PointerWarpManager` in `src/niri.rs` and
  `src/handlers/mod.rs`.

## Testing

Verified end-to-end with [Wine MR !10830](https://gitlab.winehq.org/wine/wine/-/merge_requests/10830)
— the matching client-side `winewayland.drv` patch.

- `SetCursorPos` called 12 times by a test program (3 laps × 4 corners of a square).
- Niri side: registry binds the global; every warp validated and honored; cursor lands on
  each target.
- Wine side: every `GetCursorPos` returns the requested position — full roundtrip works.

Tested both nested (winit backend) and as a real session.

### DIFF
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -21,6 +21,7 @@ use smithay::output::Output;
 use smithay::reexports::rustix::fs::{fcntl_setfl, OFlags};
 use smithay::reexports::wayland_protocols_wlr::screencopy::v1::server::zwlr_screencopy_manager_v1::ZwlrScreencopyManagerV1;
 use smithay::reexports::wayland_server::protocol::wl_output::WlOutput;
+use smithay::reexports::wayland_server::protocol::wl_pointer::WlPointer;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
 use smithay::reexports::wayland_server::Resource;
 use smithay::utils::{Logical, Point, Rectangle, Serial};
@@ -37,7 +38,10 @@ use smithay::wayland::keyboard_shortcuts_inhibit::{
     KeyboardShortcutsInhibitHandler, KeyboardShortcutsInhibitState, KeyboardShortcutsInhibitor,
 };
 use smithay::wayland::output::OutputHandler;
-use smithay::wayland::pointer_constraints::{with_pointer_constraint, PointerConstraintsHandler};
+use smithay::wayland::pointer_constraints::{
+    with_pointer_constraint, PointerConstraint, PointerConstraintsHandler,
+};
+use smithay::wayland::pointer_warp::PointerWarpHandler;
 use smithay::wayland::security_context::{
     SecurityContext, SecurityContextHandler, SecurityContextListenerSource,
 };
@@ -89,7 +93,9 @@ use crate::protocols::virtual_pointer::{
     VirtualPointerInputBackend, VirtualPointerManagerState, VirtualPointerMotionAbsoluteEvent,
     VirtualPointerMotionEvent,
 };
-use crate::utils::{output_size, send_scale_transform};
+use crate::utils::{
+    get_surface_size, get_surface_toplevel_coords, output_size, send_scale_transform,
+};
 use crate::{
     delegate_ext_workspace, delegate_foreign_toplevel, delegate_gamma_control,
     delegate_mutter_x11_interop, delegate_output_management, delegate_screencopy,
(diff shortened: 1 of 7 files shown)


zakissimo:
Rebased onto current main. Still happy to make any changes you'd like on the validation policy or the helper placement whenever you have a chance to look — no rush.

enderprism:
Can you test that it works properly in Godot's viewport?
Godot relies on the protocol and I couldn't get it to work on my end, when compiling this PR: https://github.com/godotengine/godot/blob/070dc9897ea1b84ab2a7ec04b9bc1b94f38a0eaf/platform/linuxbsd/wayland/display_server_wayland.cpp#L559

zakissimo:
Happy to dig in. A few questions to narrow it down, since "couldn't get it to work" could be several things:

1. On the niri side, does `wayland-info | grep -i warp` show `wp_pointer_warp_v1` advertised when running this PR's build? (Just to confirm the global is actually bound.)
2. What exactly happens in Godot — cursor doesn't move at all, moves to the wrong place, or only fails in a specific mode (e.g. captured/confined mouse vs. visible)?
3. Were you on a fractional-scale / HiDPI output? Godot divides the target by the scale factor before warping, and my bounds check is against the surface's logical size — I want to rule out a scale mismatch causing the warp to be rejected as out-of-bounds.

I've added `trace!` logging to every reject path in the handler locally. If you can run niri with `RUST_LOG=niri=trace` (or `WAYLAND_DEBUG`/niri's trace env) and grab the lines containing "pointer warp", that'll tell us immediately which check is failing — serial mismatch, bounds rejection, window lookup, etc. I'll also test against Godot on my end and report back.

zakissimo:
Tested this against Godot and it works correctly — but there's an important version caveat that I think explains your result.

**Godot 4.6.x stable does not contain the pointer-warp code.** It landed on `master` in [godotengine/godot@4dceee5](https://github.com/godotengine/godot/commit/4dceee5252) (1 Nov 2025), *after* the `4.6` release branch was cut, and wasn't backported. I confirmed `platform/linuxbsd/wayland/wayland_thread.cpp` at the `4.6.3-stable` tag has zero `wp_pointer_warp` references. On 4.6.x, Godot never binds the global and silently falls back to its `zwp_locked_pointer_v1` hint hack — which looks exactly like "warp doesn't work."

**With Godot 4.7-rc2 (which has the code), it works.** Running it against this PR in a nested niri, with `WAYLAND_DEBUG=1`:

- Godot binds the global: `-> wl_registry#6.bind(15, "wp_pointer_warp_v1", 1, ...)`
- Godot sends warps: `-> wp_pointer_warp_v1#24.warp_pointer(wl_surface#46, wl_pointer#40, 300.0, 300.0, 442)` (×12, a square, 3 laps)
- niri honors all 12, no rejections; coordinates transform correctly (e.g. surface-local `(300,300)` → global `(305,336)`, the window's layout offset).

Could you double-check which Godot build you tested with? If it was a 4.6.x release, that's the likely culprit — a 4.7 dev/rc or master build should work. If you *were* on master/4.7 and still hit it, let me know your compositor setup (and ideally a `WAYLAND_DEBUG=1` snippet showing whether Godot sends `warp_pointer`) and I'll dig further.

enderprism:
i was using Godot 4.6.3-stable.
i didn't realize Godot didn't actually rely on this protocol! it's unfortunate that Niri doesn't support the hack with the other protocol, since neither Godot 4.7 nor this PR has landed yet.

zakissimo:
Net status for a reviewer: this is rebased on current main, CI-clean, and now validated end-to-end against two independent clients — Wine ([MR !10830](https://gitlab.winehq.org/wine/wine/-/merge_requests/10830)) and Godot 4.7-rc2 — with correct coordinate transforms and no rejections. The report above was a Godot-version issue (4.6.x lacks the protocol), not a bug here.

@YaLTeR whenever you have bandwidth, this is ready for review.

enderprism:
Godot 4.7 has been out for a few weeks now, I would really love to see this PR get merged!

kode54:
I found a minor issue in testing, but it's otherwise flawless.

The issue: If a pointer constraint region is active, and constraining the cursor to a region smaller than the application's surface, it is possible for the application to then warp the cursor outside the constraint region, after which the cursor becomes completely unconstrained until it re-enters the constraint region.

I fixed this in my feature collection branch, and I'm willing to let you pick the exact simple block of code out into your branch. Hey, fixing things is fun!

https://github.com/losnoco/niri/tree/spicy-main

I want to post this branch to a Discussion later as well, so I can best discuss how it should be divided up into individual feature chunks, and what order those chunks should be submitted in. All commits plucked from other authors retain their original credits, though I did rebase a bunch of them.

All of this is also restyled to use the new Dispatch2 interface.

zakissimo:
Nice catch, and thanks for the offer — pulled the constraint check into the branch and credited you as co-author. Since this PR is still on the `delegate_pointer_warp!` form (not yet restyled to the Dispatch2 blanket impl like your collection), I lifted just the fix block rather than the whole commit.

Policy I settled on: reject warps that would escape an active confinement (target outside the confine region, or on a surface other than the confined one), but keep honoring warps while the pointer is *locked* — the spec leaves constraint interaction implementation-defined, and under a lock the warp is the client's intended reposition (the Godot path). Amended into the existing commit and force-pushed; CI-clean.

PR #4006, opened 2026-05-07
config: add rounded corner exponent option

HigherOrderLogic:
Add option for rounded corner exponent, which allows user to make their corners "chamfered" or "squircle"

Screenshot from 2026-05-07 15-37-52

~~One minimal issue I see is that the slanted edge of a chamfered corner has slightly thinner border. This is probably fixable but I'm not really familiar with shader so I wasnt able to. Any help is appreciated.~~ Should be fixed.

### DIFF
--- a/src/render_helpers/shaders/rounding_alpha.frag
+++ b/src/render_helpers/shaders/rounding_alpha.frag
@@ -1,4 +1,6 @@
-float niri_rounding_alpha(vec2 coords, vec2 size, vec4 corner_radius) {
+@ROUNDING_ALPHA_IMPL@
+
+float niri_rounding_alpha(vec2 coords, vec2 size, vec4 corner_radius, float corner_exponent) {
     vec2 center;
     float radius;
 
@@ -18,10 +20,5 @@ float niri_rounding_alpha(vec2 coords, vec2 size, vec4 corner_radius) {
         return 1.0;
     }
 
-    float dist = distance(coords, center);
-
-    // Manual smoothstep() between radius - half_px and radius + half_px
-    // to avoid a division in clamp().
-    float t = clamp((dist - radius) * niri_scale + 0.5, 0.0, 1.0);
-    return 1.0 - t * t * (3.0 - 2.0 * t);
+    return niri_rounding_alpha_impl(coords, center, radius, corner_exponent);
 }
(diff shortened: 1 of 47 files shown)


Sempyos:
gosh the squircle looks so good !!!

src/render_helpers/shaders/rounding_alpha.frag
```diff
@@ -18,10 +34,8 @@ float niri_rounding_alpha(vec2 coords, vec2 size, vec4 corner_radius) {
         return 1.0;
     }
 
-    float dist = distance(coords, center);
-
-    // Manual smoothstep() between radius - half_px and radius + half_px
-    // to avoid a division in clamp().
-    float t = clamp((dist - radius) * niri_scale + 0.5, 0.0, 1.0);
```
YaLTeR, line 25: one unfortunate part is that this will undo my efforts to make these shaders fit on a eee pc

HigherOrderLogic: So the `smoothstep()` function is a no-go?

YaLTeR: i removed it to save some ALU ops which made the resize shader fit. but i'm afraid with the superellipse fn you're adding many more ALU ops

HigherOrderLogic: > but i'm afraid with the superellipse fn you're adding many more ALU ops

In that case I have no idea then. I dont have any other formula in mind to solve this.

YaLTeR: maybe need to come up with some way to split these shaders or have some lightweight versions at least. for example rn the border shader doesn't fit into eee pc (so borders will always use solid color and have no rounded corners) but i think a lightweight version can remove all the color conversion stuff so at least there's rounded corners throughout

src/render_helpers/shadow.rs
```diff
@@ -30,12 +30,14 @@ struct Parameters {
...
     sigma: f32,
     corner_radius: CornerRadius,
+    corner_exponent: f32,
     // Should only be used for visual improvements, i.e. corner radius anti-aliasing.
     scale: f32,
     alpha: f32,
 
     window_geometry: Rectangle<f64, Logical>,
     window_corner_radius: CornerRadius,
+    window_corner_exponent: f32,
```
HigherOrderLogic, line 40: Also, I'm not sure if this is needed, since the shadow and window's corner exponent will always be the same anyway.

PR #4015, opened 2026-05-07
Assign issue with type instead of label

HigherOrderLogic:
Seems like Niri (or Semper) is preferring the issue type over label, so let's make it the default.


YaLTeR:
Ty

PR #4059, opened 2026-05-15
nix: add systemd units to the right location

bokicoder:
See https://github.com/NixOS/nixpkgs/pull/520403


YaLTeR:
cc @zimward

zimward (approved):
lib is indeed the correct directory

YaLTeR:
Thanks

PR #4062, opened 2026-05-15
Add window-rule `float-above-fullscreen`

bczhc:
This PR adds a window-rule `float-above-fullscreen` (or whatever, like `always-float`, `always-on-top` etc, if there's a better name for it) which prevents a floating window from being hidden when another window is fullscreened.

The configuration:

```kdl
window-rule {
    match title="^Picture-in-Picture$"

    float-above-fullscreen true
}
```

This is useful say if one is playing a fullscreen video game and meanwhile is watching a video as a floating window in the corner (e.g. firefox's picture-in-picture). And I searched for it and found there's a discussion about this: https://github.com/niri-wm/niri/discussions/2798#discussioncomment-16933617, knowing I'm not the only one who likes doing such things sometimes :).

The current solution to this is kill the bar (like DankMaterialShell bar or waybar or similar) and to make the game go 'maximize-window-to-edges' state. When no bar is present, 'maximize-window-to-edges' looks identical to the real 'fullscreen'. But most people I believe won't want to bother doing this workaround. There are also differences between 'maximize-window-to-edges' and 'fullscreen' I guess, for example, the `direct-scanout`. If `direct-scanout` is enabled (yes in niri it is by default), I just tested, Minecraft (26.2-snapshot-7) will sync its framerate with the monitor if it's 'fullscreen', but it won't if it's 'maximize-window-to-edges', though without a bar these two look identical. Anyway to achieve this (playing games + watching videos :) ), a window-rule for it is more preferable I suppose.

### DIFF
--- a/src/layout/workspace.rs
+++ b/src/layout/workspace.rs
@@ -1588,8 +1588,14 @@ impl<W: LayoutElement> Workspace<W> {
         let scrolling = self.scrolling.tiles_with_render_positions();
 
         let floating = self.floating.tiles_with_render_positions();
-        let visible = self.is_floating_visible();
-        let floating = floating.map(move |(tile, pos)| (tile, pos, visible));
+        let base_visible = self.is_floating_visible();
+        let floating = floating.map(move |(tile, pos)| {
+            if tile.window().rules().float_above_fullscreen == Some(true) {
+                (tile, pos, true)
+            } else {
+                (tile, pos, base_visible)
+            }
+        });
 
         floating.chain(scrolling)
     }
@@ -1646,16 +1652,18 @@ impl<W: LayoutElement> Workspace<W> {
         focus_ring: bool,
         push: &mut dyn FnMut(WorkspaceRenderElement<R>),
     ) {
-        if !self.is_floating_visible() {
-            return;
-        }
+        let base_visible = self.is_floating_visible();
 
         let view_rect = Rectangle::from_size(self.view_size);
         let floating_focus_ring = focus_ring && self.floating_is_active();
-        self.floating
-            .render(ctx, xray_pos, view_rect, floating_focus_ring, &mut |elem| {
-                push(elem.into())
-            });
+        self.floating.render(
+            ctx,
+            xray_pos,
+            view_rect,
+            floating_focus_ring,
+            base_visible,
+            &mut |elem| push(elem.into()),
+        );
     }
 
     pub fn render_shadow<R: NiriRenderer>(
@@ -1755,14 +1763,19 @@ impl<W: LayoutElement> Workspace<W> {
 
     pub fn window_under(&self, pos: Point<f64, Logical>) -> Option<(&W, HitType)> {
         // This logic is consistent with tiles_with_render_positions().
-        if self.is_floating_visible() {
-            if let Some(rv) = self
-                .floating
+        let base_visible = self.is_floating_visible();
+        if let Some(rv) =
+            self.floating
                 .tiles_with_render_positions()
-                .find_map(|(tile, tile_pos)| HitType::hit_tile(tile, tile_pos, pos))
-            {
-                return Some(rv);
-            }
+                .find_map(|(tile, tile_pos)| {
+                    if base_visible || tile.window().rules().float_above_fullscreen == Some(true) {
+                        HitType::hit_tile(tile, tile_pos, pos)
+                    } else {
+                        None
+                    }
+                })
+        {
+            return Some(rv);
         }
 
         self.scrolling.window_under(pos)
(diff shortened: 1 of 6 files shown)


witchlliee:
Oh! This is also helpful if you want steam notifications (like game achievements) in top of the games when playing on native wayland (PROTON_USE_WAYLAND=1 with a proton which has winewayland patches) since xwayland-satellite only shows the notifications on top of xwayland windows.

Note: you need to have steam closed tho it seems for niri to spawn the steam notifications as floating windows on desktop.

```
window-rule {
    match app-id="steam" title=r#"^notificationtoasts_\d+_desktop$"#
      default-floating-position x=10 y=10 relative-to="bottom-right"

    float-above-fullscreen true
    open-floating true
  }
  ```

**Comment by saygo-png** [comment]
Works well, though i'd also love a passthru mode for a window so that i can click through it. Currently the mouse gets stuck on the floating window which is not ideal in some games.

**Comment by bczhc** [comment]
> Works well, though i'd also love a passthru mode for a window so that i can click through it. Currently the mouse gets stuck on the floating window which is not ideal in some games.

Input passthrough is weird for xdg_toplevel windows. Could you clarify more on how it "gets stuck" in some games?

**Comment by saygo-png** [comment]
> Could you clarify more on how it "gets stuck" in some games?

https://github.com/user-attachments/assets/c62566bc-986b-4fff-b534-567df2d0a865

This happens with the game on windowed too without the patch. This is not strictly related to the patch, just suggesting that it would be nice to have an additional mode where a window has total passthru for inputs.

**Comment by bczhc** [comment]
> This happens with the game on windowed too without the patch. This is not strictly related to the patch, just suggesting that it would be nice to have an additional mode where a window has total passthru for inputs.

I see. Although this is usually okay for first-person 3d games (because they lock the pointer) I actually have encountered such issues in some games when the cursor is unlocked, under _some_ circumstances. (actually I don't know if this is another niri bug tho). Anyway this is out of scope; changing surfaces' input region in this PR is not expected and it's tricky.

Can you have a look at this project [wl-proxy](https://github.com/mahkoh/wl-proxy) and see if this meets your usecase for example: making a specific window into an `overlay` layer-shell layer positioned at a fixed position of the screen. In this case it doesn't receive any keyboard & mouse events.

First clone it and apply the two patches:

[0001-Add-exclusive-zone-cli-argument.patch](https://github.com/user-attachments/files/28913334/0001-Add-exclusive-zone-cli-argument.patch)
[0002-Add-input-passthrough-cli-option.patch](https://github.com/user-attachments/files/28913335/0002-Add-input-passthrough-cli-option.patch)

Then build it and do something like:

```shell
wl-paper --layer overlay --margin-left 1400 --margin-bottom 800 --input-passthrough alacritty
```

This creates Alacritty as an overlay layer at the top-right corner.

But one limitation is.. because input passthrough is enabled so you can't focus it and can't have any interaction with it.. Anyway this provides an workaround idea. If this is the right direction, you can try making a patch where there's a "switch" to toggle the layer's "input passthrough" so you can firstly set the window up and then enable its "input-passthrough" and then start playing the game.

bczhc:
> Currently the mouse gets stuck on the floating window which is not ideal in some games.

I filed it: #4192

src/layout/workspace.rs
```diff
@@ -1588,8 +1588,14 @@ impl<W: LayoutElement> Workspace<W> {
...
-        let visible = self.is_floating_visible();
-        let floating = floating.map(move |(tile, pos)| (tile, pos, visible));
+        let base_visible = self.is_floating_visible();
+        let floating = floating.map(move |(tile, pos)| {
+            if tile.window().rules().float_above_fullscreen == Some(true) {
+                (tile, pos, true)
+            } else {
+                (tile, pos, base_visible)
+            }
+        });
```
Ind-E, line 1598: I think this would make more sense here
```rust
let floating = floating.map(move |(tile, pos)| {                              
    (                                                                         
        tile,                                                                 
        pos,                                                                  
        visible || tile.window().rules().float_above_fullscreen == Some(true),
    )                                                                         
});                                                                           
```

src/layout/workspace.rs
```diff
@@ -1646,16 +1652,18 @@ impl<W: LayoutElement> Workspace<W> {
...
-        self.floating
-            .render(ctx, xray_pos, view_rect, floating_focus_ring, &mut |elem| {
-                push(elem.into())
-            });
+        self.floating.render(
+            ctx,
+            xray_pos,
+            view_rect,
+            floating_focus_ring,
+            base_visible,
```
Ind-E, line 1664: I don't think this is complex enough to warrant creating a local variable (that's only used in 1 place).

Ind-E (commented):
overall, lgtm. Needs to be rebased because of 782f459, but that should be straightforward.

PR #4064, opened 2026-05-16
window-rule: add is-fullscreen matcher

jakobhellermann:
Add a `match is-fullscreen=true` window-rule.

Useful e.g. for preferences like https://github.com/niri-wm/niri/discussions/3825.

### DIFF
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -36,6 +36,7 @@ window-rule {
     match is-floating=true
     match is-window-cast-target=true
     match is-urgent=true
+    match is-fullscreen=true
     match at-startup=true
 
     // Properties that apply once upon window opening.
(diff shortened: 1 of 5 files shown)


cornservant (approved):
LGTM

Stylistically I would probably prefer some minor use of assignable variables when tracking changes, either way is fine.

```rust
        let mut size_changed = false;
        let mut fullscreen_changed = false;
        self.toplevel().with_pending_state(|state| {
            size_changed = state.size != Some(size);
            state.size = Some(size);

            if mode.is_fullscreen() || self.is_pending_windowed_fullscreen {
                fullscreen_changed = state.states.set(xdg_toplevel::State::Fullscreen);
                state.states.unset(xdg_toplevel::State::Maximized);
            } else if mode.is_maximized() {
                fullscreen_changed = state.states.unset(xdg_toplevel::State::Fullscreen);
                state.states.set(xdg_toplevel::State::Maximized);
            } else {
                fullscreen_changed = state.states.unset(xdg_toplevel::State::Fullscreen);
                state.states.unset(xdg_toplevel::State::Maximized);
            };
        });
```

jakobhellermann:
Done (and rebased)

PR #4069, opened 2026-05-17
gestures: edge-overscroll — pan focus by pushing the pointer past a screen edge

odtgit:
## What & why

Scrollable-tiling mouse users currently pan between off-screen columns via `focus-follows-mouse` with `max-scroll-amount="100%"` plus left/right `struts` that leave a sliver of the neighbour visible. That has two problems:

1. **It can't work without struts.** A maximized column hides adjacent windows entirely, so there's no sliver for FFM to land on — you simply can't mouse-pan.
2. **It's accident-prone and toolkit-dependent.** At `100%`, FFM activates a neighbour the instant *anything* of it is under the cursor — a strut sliver, or a CSD client's shadow/overflow region (Spotify/Firefox/Electron draw them, SSD apps like Alacritty don't). Brushing the edge swaps columns; behaviour varies per toolkit.

This PR adds an explicit, opt-in gesture: **`gestures { edge-overscroll { resistance N } }`**. Push the pointer *past* a true screen edge (a hard edge with no adjacent output, so motion is clipped); the clipped over-travel accumulates and, once it crosses `resistance` logical px, performs **one discrete** focus step — adjacent column (L/R) or workspace (U/D).

Because the pointer is pinned at a hard edge, incidental contact accumulates only a few px and can't trigger it — no timer, no heuristic. Combined with `focus-follows-mouse max-scroll-amount="0%"` and **no** left/right struts, this gives reliable, deliberate, toolkit-independent edge panning that reclaims the strut screen estate. Tuned with `resistance` to taste against accidental swaps; `0`/absent disables it (default — zero behaviour change for existing configs).

This is intentionally **not** `dnd-edge-*`: those are DnD-only, proximity-band, continuous auto-scroll. This is non-drag, true-edge, single-step, FFM-independent — disjoint from existing gestures, hence a new sibling under `gestures {}`.

## Two commits (stacked, please take both)

1. **`fix: prioritize activation region hits over CSD overflow in scrolling hit testing`** — prerequisite. Under heavy CSD app usage the gesture (and click/activation generally) misbehaves because a CSD client's shadow/overflow steals the hit before the real activation region. This makes scrolling hit-testing prefer the activation region before falling back to CSD overflow. It's an independent correctness fix but is required for the feature to behave consistently across toolkits.
2. **`gestures: edge-overscroll …`** — the feature: config (`niri-config` knuffel + merge + snapshot), the `on_pointer_motion` true-edge clip → accumulate → latch logic, direction resolver, state reset paths, and wiki docs.

## Suppression / safety

Never fires while: session locked, overview/screenshot/MRU UIs open, pointer grab or pointer constraint active, or on a fullscreen window (leaving fullscreen must be explicit). Acts on the monitor the cursor is actually on; fires once per excursion and re-arms when the cursor returns inside the outputs. State resets on config reload and absolute-motion path.

## Testing

- `cargo test -p niri-config` green incl. new `parse_edge_overscroll` (knuffel parse + merge) and updated parse snapshot.
- `edge_overscroll_dir` unit test (dominant-axis + sign resolution) green.
- `cargo run -- validate` clean.
- Daily-driven for weeks on a 2-monitor setup (ultrawide + 1080p, mixed scale) across CSD-heavy apps (Spotify/Firefox/Electron OWAs) and SSD (Alacritty) — consistent behaviour, no accidental swaps at `resistance ~200`.

Docs added to the Gestures wiki page; the Input wiki cross-links to it. `Since:` placeholder is `next release` — happy to set a concrete version on request.

### DIFF
--- a/docs/wiki/Configuration:-Gestures.md
+++ b/docs/wiki/Configuration:-Gestures.md
@@ -21,6 +21,10 @@ gestures {
         max-speed 1500
     }
 
+    edge-overscroll {
+        // resistance 0   // 0 disables this gesture (the default)
+    }
+
     hot-corners {
         // off
         top-left
@@ -77,6 +81,32 @@ gestures {
     dnd-edge-workspace-switch {
         trigger-height 100
         max-speed 3000
+    }
+}
+```
+
+### `edge-overscroll`
+
+<sup>Since: next release</sup>
+
+Push the mouse cursor *past* a true screen edge — a hard edge with no adjacent output, so the motion is clipped — to pan focus to the adjacent column (left/right edge) or workspace (up/down edge).
+
+Unlike `dnd-edge-*`, this works outside drag-and-drop, during normal pointer use. It requires the cursor to reach the *actual* screen boundary (not a proximity band) and keep pushing; the clipped over-travel accumulates and, once it crosses `resistance`, performs a single discrete navigation step. Because the pointer is pinned at a hard edge, incidental edge contact accumulates only a few pixels and cannot trigger it — there is no timer and no heuristic.
+
+It is suppressed while the session is locked, in the overview/screenshot/MRU UIs, during a pointer grab or constraint, and on a fullscreen window (leaving fullscreen must be explicit). It acts on the monitor the cursor is on, fires once per excursion, and re-arms when the cursor returns inside the outputs.
+
+The option is:
+
+- `resistance`: accumulated over-travel past the edge, in logical pixels, required to trigger. `0` (the default) disables the gesture. Higher values require a firmer, more deliberate shove.
+
+This is primarily useful for scrollable layouts run *without* left/right `struts` (combined with `focus-follows-mouse max-scroll-amount="0%"`), where a maximized column leaves no off-screen sliver for ordinary focus-follows-mouse to pan via. Note: with side-by-side monitors of differing height/offset, the L-shaped no-output region next to the shorter monitor is also a true screen edge.
+
+```kdl
+gestures {
+    // Shove the cursor ~200 px past a screen edge to pan focus
+    // to the adjacent column / workspace.
+    edge-overscroll {
+        resistance 200
     }
 }
 ```
--- a/docs/wiki/Configuration:-Gestures.md
+++ b/docs/wiki/Configuration:-Gestures.md
@@ -89,7 +89,9 @@ gestures {
 
 <sup>Since: next release</sup>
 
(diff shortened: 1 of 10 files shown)


Acters:
This gesture does not seem to be a clean solution since it disregards those with multiple monitors. I have three, and the middle one would constantly "scroll" if I move my mouse between the leftmost monitor to the rightmost monitor.

DankMaterialShell has a gesture were by scrolling on the top bar would scroll to the various "workspaces"
maybe a region were scrolling the mouse wheel would tell niri to scroll would be less finicky and accidental gesturing.

odtgit:
Thanks for the report — I think I have it.

I pushed a follow-up commit ([32e45475](https://github.com/niri-wm/niri/pull/4069/commits/32e454754d7e0e2f1a87d50bca8b7ee03b3a4f3c)) that gates the gesture on a neighbor check: when the dominant overscroll direction has a reachable monitor neighbor (`output_left_of`/`output_right_of`/`output_up_of`/`output_down_of` — niri's existing center-and-overlap test), the gesture is suppressed and the accumulator resets. The inner edges between your three monitors are exactly that case, so the middle one should no longer pan columns when you sweep across.

Tracing your setup with the fix:
- Sweeping right from leftmost to rightmost, your hand isn't perfectly y=constant. Near an inner edge, a tiny upward (or downward) nibble lands `new_pos` above/below all outputs → clamp fires against the monitor the cursor came from → vector is dominantly horizontal (because the sweep itself is). Pre-fix that called `focus_right()` inside the middle monitor. Post-fix, `output_right_of(M_middle)` returns the rightmost monitor → suppressed.

Before I claim "fixed" on your specific setup, two clarifying questions so I don't miss a corner:

1. Are all three monitors the same **pixel height**, and are their top edges at the same global y? (Different heights or a non-zero y-offset on one of them produces a wider dead band along the inner edges — still covered by the neighbor check, but I'd like to confirm the geometry I'm tracing matches yours.)
2. Are they directly adjacent in niri's `output` config (no gap), or is there any `position` offset between them?

I don't have a horizontal layout to repro locally — if you can pull the branch and try it I'd appreciate the confirmation. Honest caveat: if your sweep gets very vertical near a corner (|dy| > |dx|), `switch_workspace_up`/`_down` can still fire because there's no monitor neighbor above/below in a horizontal-only setup. That matches how the gesture works on a vertical stack today, so it's by design — but tell me if that's also what you were seeing and we can rethink.

On the scroll-on-top-bar idea: that one's better suited to DMS-side or a separate compositor feature; this PR is scoped to the existing "push past a hard edge" gesture pattern (sibling of dnd-edge-*), which is opt-in (`resistance=0` disables it).

PR #4081, opened 2026-05-21
Add XF86AudioPause media key mapping

KarimAlaswad:
Some Bluetooth earbuds send alternating XF86AudioPlay/XF86AudioPause keycodes on the same physical press. Without a binding for XF86AudioPause, every other press is dropped, making playback toggle unreliable.


YaLTeR:
Thanks!

PR #4090, opened 2026-05-23
layout: fall back to tiling focus when transient floating window closes

foolofafitz:
When closing a floating window focus returns to last tiled window


foolofafitz:
Closing in favour of #4093 - much simpler approach without tracking

PR #4093, opened 2026-05-25
Return focus to tiling when a floating window is closed

foolofafitz:


NSPC911:
Hello, you might want to check on the tests for this one, `cargo test` fails 

```
failures:

---- layout::tests::fullscreen::interactive_move_unfullscreen_to_floating_stops_dnd_scroll stdout ----

thread 'layout::tests::fullscreen::interactive_move_unfullscreen_to_floating_stops_dnd_scroll' (33572) panicked at src/layout/workspace.rs:2036:13:
when floating is empty it must never be active
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- layout::tests::fullscreen::interactive_move_restore_to_floating_animates_view_offset stdout ----

thread 'layout::tests::fullscreen::interactive_move_restore_to_floating_animates_view_offset' (33571) panicked at src/layout/workspace.rs:2036:13:
when floating is empty it must never be active

---- layout::tests::move_column_to_workspace_down_focus_false_on_floating_window stdout ----

thread 'layout::tests::move_column_to_workspace_down_focus_false_on_floating_window' (33602) panicked at src/layout/workspace.rs:2036:13:
when floating is empty it must never be active

---- layout::tests::move_column_to_workspace_focus_false_on_floating_window stdout ----

thread 'layout::tests::move_column_to_workspace_focus_false_on_floating_window' (33603) panicked at src/layout/workspace.rs:2036:13:
when floating is empty it must never be active
```

It seems like the implementation wasn't well thought out, because you are preventing `floating_is_active` from being unset with the default-disabled config option. An alternative implementation of
```rs
        if removed_from_floating {
            self.floating_is_active = FloatingActive::No;
        } else if self.scrolling.is_empty() && !self.floating.is_empty() {
            self.floating_is_active = FloatingActive::Yes;
        } else if self.options.layout.focus_tiled_on_float_close {
            // unselect floating if something else happens
            self.floating_is_active = FloatingActive::No;
        }
```
is more reasonable.

foolofafitz:
Hi @NSPC911

Thanks for the feedback. I have applied the fix you suggested. Tests pass for me now.

PR #4098, opened 2026-05-25
Per window-rule animations config

Atan-D-RP4:
In the vein of #3481, adds per window-rule configuration for window open/close animations. 

EDIT:
#3481 depends on this PR now.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1890,6 +1890,7 @@ mod tests {
                             saturation: None,
                         },
                     },
+                    animations: None,
                 },
             ],
             layer_rules: [
--- a/src/layout/floating.rs
+++ b/src/layout/floating.rs
@@ -559,7 +559,21 @@ impl<W: LayoutElement> FloatingSpace<W> {
 
         let tile_size = tile.tile_size();
 
-        self.start_close_animation_for_tile(renderer, snapshot, tile_size, tile_pos, blocker);
+        let anim_config = tile
+            .window()
+            .rules()
+            .window_close
+            .clone()
+            .unwrap_or(self.options.animations.window_close.clone());
+
+        self.start_close_animation_for_tile(
+            renderer,
+            snapshot,
+            tile_size,
+            tile_pos,
+            blocker,
+            anim_config,
+        );
     }
 
     pub fn activate_window_without_raising(&mut self, id: &W::Id) -> bool {
@@ -599,14 +613,9 @@ impl<W: LayoutElement> FloatingSpace<W> {
         tile_size: Size<f64, Logical>,
         tile_pos: Point<f64, Logical>,
         blocker: TransactionBlocker,
+        anim_config: niri_config::animations::WindowCloseAnim,
     ) {
-        let anim = Animation::new(
-            self.clock.clone(),
-            0.,
-            1.,
-            0.,
-            self.options.animations.window_close.anim,
-        );
+        let anim = Animation::new(self.clock.clone(), 0., 1., 0., anim_config.anim);
 
         let blocker = if self.options.disable_transactions {
             TransactionBlocker::completed()
@@ -616,7 +625,14 @@ impl<W: LayoutElement> FloatingSpace<W> {
 
         let scale = Scale::from(self.scale);
         let res = ClosingWindow::new(
-            renderer, snapshot, scale, tile_size, tile_pos, blocker, anim,
+            renderer,
+            snapshot,
+            scale,
+            tile_size,
+            tile_pos,
+            blocker,
+            anim,
+            anim_config.custom_shader,
         );
         match res {
             Ok(closing) => {
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -4757,14 +4757,32 @@ impl<W: LayoutElement> Layout<W> {
 
         if let Some(InteractiveMoveState::Moving(move_)) = &mut self.interactive_move {
             if move_.tile.window().id() == window {
-                let Some(snapshot) = move_.tile.take_unmap_snapshot() else {
+                // Extract all tile data in a block to limit the mutable borrow.
+                let (
+                    snapshot_opt,
+                    tile_pos,
+                    tile_size,
+                    output,
+                    pointer_pos_within_output,
+                    anim_override,
+                ) = {
+                    let snapshot = move_.tile.take_unmap_snapshot();
+                    let tpos = move_.tile_render_location(zoom);
+                    let tsz = move_.tile.tile_size();
+                    let output = move_.output.clone();
+                    let pointer = move_.pointer_pos_within_output;
+                    let override_opt = move_.tile.window().rules().window_close.clone();
+                    (snapshot, tpos, tsz, output, pointer, override_opt)
+                };
+                // Mutable borrow from self.interactive_move is released here.
+
+                let Some(snapshot) = snapshot_opt else {
                     return;
                 };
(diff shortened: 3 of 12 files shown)


Atan-D-RP4:
After some git-fu, this PR is now a dependency of #3481.

I'm pretty happy with how it ended up, though the defaults and examples will need to be reviewed.

Would appreciate of someone took a look.

niri-config/src/lib.rs
```diff
@@ -534,6 +534,40 @@ impl Config {
...
+    /// program whose source is absent here is stranded and can be freed.
+    pub fn custom_shader_sources(&self) -> HashSet<String> {
+        let mut out = HashSet::new();
+
+        let global = [
+            self.animations.window_open.custom_shader.as_deref(),
+            self.animations.window_close.custom_shader.as_deref(),
+            self.animations.window_resize.custom_shader.as_deref(),
+        ];
+        out.extend(global.into_iter().flatten().map(str::to_owned));
```
cornservant: no need to `map(str::to_owned)` if we had just cloned the `Option<String>`s in the first place

niri-config/src/lib.rs
```diff
@@ -534,6 +534,40 @@ impl Config {
...
+                anims
+                    .window_open
+                    .as_ref()
+                    .and_then(|anim| anim.custom_shader.as_deref()),
+                anims
+                    .window_close
+                    .as_ref()
+                    .and_then(|anim| anim.custom_shader.as_deref()),
+            ];
+            out.extend(rule_sources.into_iter().flatten().map(str::to_owned));
```
cornservant: same as above

resources/default-config.kdl
```diff
@@ -346,6 +346,20 @@ window-rule {
     clip-to-geometry true
 }
 
+// Example: overrider animation settings for a specific app.
```
cornservant: typo: overrider

src/layout/floating.rs
```diff
@@ -568,7 +568,21 @@ impl<W: LayoutElement> FloatingSpace<W> {
 
         let tile_size = tile.tile_size();
 
-        self.start_close_animation_for_tile(renderer, snapshot, tile_size, tile_pos, blocker);
+        let anim_config = tile
+            .window()
+            .rules()
+            .window_close
+            .clone()
+            .unwrap_or(self.options.animations.window_close.clone());
```
cornservant: Could use `unwrap_or_else(|| ...)` to avoid unnecessary clone

src/layout/mod.rs
```diff
@@ -4785,6 +4785,11 @@ impl<W: LayoutElement> Layout<W> {
 
                 let output = move_.output.clone();
                 let pointer_pos_within_output = move_.pointer_pos_within_output;
+
+                let anim_override = move_.tile.window().rules().window_close.clone();
+                let anim_config =
+                    anim_override.unwrap_or(self.options.animations.window_close.clone());
```
cornservant: use `unwrap_or_else`

src/layout/scrolling.rs
```diff
@@ -1481,16 +1481,26 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         window: &W::Id,
         blocker: TransactionBlocker,
     ) {
-        let (tile, mut tile_pos) = self
-            .tiles_with_render_positions_mut(false)
-            .find(|(tile, _)| tile.window().id() == window)
-            .unwrap();
+        // Extract all data from the tile in a block to limit the mutable borrow's lifetime.
+        let (snapshot_opt, tile_size, anim_override, mut tile_pos) = {
```
cornservant, line 1485: I think it's ok to reuse the same variable names and variable shadowing is pretty common all over the code base, so I'd suggest

```rust
let (snapshot, tile_size, anim_override, mut tile_pos) = {
    // ...
    (snapshot, tile_size, anim_override, tile_pos)
};

// ...

let Some(snapshot) = snapshot else {
```

src/layout/scrolling.rs
```diff
@@ -1481,16 +1481,26 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
+        };
+        // Mutable borrow from tiles_with_render_positions_mut is released here.
 
-        let Some(snapshot) = tile.take_unmap_snapshot() else {
+        let Some(snapshot) = snapshot_opt else {
             return;
         };
 
-        let tile_size = tile.tile_size();
+        let anim_config = anim_override.unwrap_or(self.options.animations.window_close.clone());
```
cornservant, line 1503: use `unwrap_or_else`

src/layout/tile.rs
```diff
@@ -573,13 +573,17 @@ impl<W: LayoutElement> Tile<W> {
...
-            1.,
-            0.,
-            self.options.animations.window_open.anim,
-        )));
+        let anim_config = self
+            .window()
+            .rules()
+            .window_open
+            .clone()
+            .unwrap_or(self.options.animations.window_open.clone());
```
cornservant: use `unwrap_or_else`

src/render_helpers/shaders/mod.rs
```diff
@@ -346,13 +351,106 @@ pub fn set_custom_open_program(renderer: &mut GlesRenderer, src: Option<&str>) {
...
+    if let Some(prev) = Shaders::get(renderer).replace_custom_window_open_program(program) {
         if let Err(err) = prev.destroy(renderer) {
             warn!("error destroying previous custom open shader: {err:?}");
         }
     }
 }
 
+pub fn window_open_program_for_source(
+    renderer: &mut GlesRenderer,
+    src: &str,
```
cornservant, line 364: Because we are cloning the shader string all the way here just to pass it as &str, I wonder if it should be stored as, say `Arc<str>`/`Rc<str>` in the first place in `OpenAnimation`, `WindowOpenAnim`, etc..

cornservant (approved):
minor suggestions

cornservant:
I tried replacing `String` with `Arc<str>` to see how much of a hassle it would be to switch. Might as well share the patch:

[0001-store-and-pass-around-shaders-source-code-as-Arc-str.patch](https://github.com/user-attachments/files/32321042/0001-store-and-pass-around-shaders-source-code-as-Arc-str.patch)

src/layout/scrolling.rs
Atan-D-RP4: I actually did do that, but got rid of it to keep thr diff small. Lol.

Hadn't touched this branch in a while, you see, and when I saw that in the diff I was legit flabbergasted on why it was there (sleep deprived).

I think I'd prefer to leave it as is. Keeps the diffs simple, as I said.

Atan-D-RP4:
Oh yes, I wanted to try something like that but hadn't gotten to it. Just used the dumb solution to keep it simple.

Actually, does it even need to be an Arc? Or can we get away with an Rc? Niri is fully single-threaded and we're not doing any async stuff around here so I think we can get away with Rc.

Will take a look! Appreciate the review!

cornservant:
Yes, I had the impression that `Rc` is ok too. Maybe I should've just started with `Rc` and trusted the compiler to yell at me if that was a stupid choice :D

EDIT: rust did in fact yell at me, `Rc<str>` doesnt implement `DecodeScalar`

cornservant (approved):

cornservant (approved):

Atan-D-RP4:
@cornservant 
I switched it over to use Arc<str> instead of String. Would have preferred Rc, but oh well.

cornservant (approved):

PR #4105, opened 2026-05-27
layer-rule: ignore-opacity (alpha-mask blur)

Fexaop:
Adds `ignore-opacity` to layer-rule `background-effect`. When set, blur is
masked by the surface's own alpha channel, so the unblurred backdrop shows
through fully-transparent regions of the layer. Useful for bar layers like
Waybar that span a screen-wide transparent area with a small opaque bar.

Accepts `true` / `false` or a numeric threshold in `[0.0, 1.0]`:
- `true` → threshold `0.0` (exclude only fully-transparent pixels)
- `0.7`  → exclude pixels with surface alpha `<= 0.7`
- `false` / omitted → disabled

Implemented as a shader-side alpha test in `postprocess.frag`. The layer's
surface texture is bound to `GL_TEXTURE1`; a `surface_to_geo` mat3 maps the
fragment's `v_coords` to the surface buffer UV. Same path works for both the
framebuffer-effect blur and xray.

Inspired by Hyprland's `blur:ignore_opacity` / `popups_ignorealpha`.

## Config

```kdl
layer-rule {
    match namespace="^waybar$"
    background-effect {
        blur true
        ignore-opacity 0.5
    }
}
```

### DIFF
--- a/docs/wiki/Configuration:-Layer-Rules.md
+++ b/docs/wiki/Configuration:-Layer-Rules.md
@@ -236,6 +236,7 @@ Override the background effect options for this surface.
 
 - `xray`: set to `true` to enable the xray effect, or `false` to disable it.
 - `blur`: set to `true` to enable blur behind this surface, or `false` to force-disable it.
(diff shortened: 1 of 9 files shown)


luckystar2718:
This works great. gets rid of the jagged edges around layer surfaces

PR #4118, opened 2026-05-30
feat: add support for xdg-toplevel-tag-v1 protocol

MithicSpirit:
The tag is exposed to window rules via xdg-tag matcher, and it is also provided via IPC.

~~Currently requires vendoring smithay due to an upstream bug: https://github.com/Smithay/smithay/pull/2045.~~

### DIFF
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -171,19 +171,17 @@ window-rule {
 
 Let's look at the matchers in more detail.
 
-#### `title`, `app-id`, and `xdg-tag`
+#### `title` and `app-id`
 
-These are regular expressions that should match anywhere in the window title, app ID, and XDG toplevel tag respectively.
+These are regular expressions that should match anywhere in the window title and app ID respectively.
 You can read about the supported regular expression syntax [here](https://docs.rs/regex/latest/regex/#syntax).
 
 ```kdl
 // Match windows with title containing "Mozilla Firefox",
-// windows with app ID containing "Alacritty",
-// or windows with XDG toplevel tag containing "proton-game".
+// or windows with app ID containing "Alacritty".
 window-rule {
     match title="Mozilla Firefox"
     match app-id="Alacritty"
-    match xdg-tag="proton-game"
 }
 ```
 
@@ -195,7 +193,7 @@ window-rule {
 }
 ```
 
-You can find the title, the app ID, and the XDG toplevel tag (if any) of a window by running `niri msg pick-window` and clicking on the window in question.
+You can find the title and the app ID of a window by running `niri msg pick-window` and clicking on the window in question.
 
 > [!TIP]
 > Another way to find the window title and app ID is to configure the `wlr/taskbar` module in [Waybar](https://github.com/Alexays/Waybar) to include them in the tooltip:
@@ -205,6 +203,22 @@ You can find the title, the app ID, and the XDG toplevel tag (if any) of a windo
 >     "tooltip-format": "{title} | {app_id}",
 > }
 > ```
+
+#### `xdg-tag`
+
+This is a regular expression that should match anywhere in the window's XDG toplevel tag.
+This always fails if the window has not set an XDG toplevel tag.
+See above for more information on regular expressions.
+
+```kdl
+// Match windows with the XDG toplevel tag "proton-game".
+// Note that only some versions of Proton set this.
+window-rule {
+    match xdg-tag="^proton-game$"
+}
+```
+
+You can find the XDG toplevel tag of a window by running `niri msg pick-window` and clicking on the window in question.
 
 #### `is-active`
 
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -206,6 +206,8 @@ You can find the title and the app ID of a window by running `niri msg pick-wind
 
 #### `xdg-tag`
 
+<sup>Since: next release</sup>
+
 This is a regular expression that should match anywhere in the window's XDG toplevel tag.
 This always fails if the window has not set an XDG toplevel tag.
 See above for more information on regular expressions.
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -171,17 +171,19 @@ window-rule {
 
 Let's look at the matchers in more detail.
 
-#### `title` and `app-id`
+#### `title`, `app-id`, and `xdg-tag`
 
-These are regular expressions that should match anywhere in the window title and app ID respectively.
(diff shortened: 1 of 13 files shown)


FriederHannenheim:
Since this is now merged in smithay can this be un-drafted?

MithicSpirit:
Not just yet because we also need a Smithay bump, and YaLTeR said they want to do that themself.

MithicSpirit:
This has been updated for the Smithay bump. I don't think I'll have the time to test this properly for a couple of weeks, but otherwise I believe this is ready.

niri-ipc/src/lib.rs
```diff
@@ -1368,6 +1368,8 @@ pub struct Window {
     ///
     /// The timestamp comes from the monotonic clock.
     pub focus_timestamp: Option<Timestamp>,
+    /// xdg-toplevel-tag tag and description, if set.
+    pub xdg_tag: (Option<Box<str>>, Option<Box<str>>),
```
cornservant: Could consider using a struct

```rust
struct XdgToplevelTag {
    tag: Option<Box<str>>,
    description: Option<Box<str>>,
}
```

This would clarify which boxed str is the tag and which is the description. It would also help with making the json output self-explanatory.

Using the shortened name `xdg_tag` is consistent with what hyprland does, so I think that's probably a good idea.

src/handlers/mod.rs
```diff
@@ -834,6 +836,36 @@ impl XdgActivationHandler for State {
...
+            return;
+        };
+        let Some((mapped, _)) = self
+            .niri
+            .layout
+            .find_window_and_output_mut(toplevel.wl_surface())
+        else {
+            return;
+        };
+        mapped.set_xdg_toplevel_tag(Some(description.into_boxed_str()), None);
```
cornservant: This sets the tag instead of the description.

src/window/mapped.rs
```diff
@@ -193,6 +194,9 @@ pub struct Mapped {
 
     /// Most recent monotonic time when the window had the focus.
     focus_timestamp: Option<Duration>,
+
+    /// XdgToplevelTag tag and description
+    xdg_toplevel_tag: (Option<Box<str>>, Option<Box<str>>),
```
cornservant: Could reuse the `XdgToplevelTag` from my comment above.

src/window/mapped.rs
```diff
@@ -406,6 +422,28 @@ impl Mapped {
         self.need_to_recompute_rules = true;
     }
 
+    pub fn set_xdg_toplevel_tag(&mut self, tag: Option<Box<str>>, description: Option<Box<str>>) {
```
cornservant: Should probably make it two functions `set_xdg_toplevel_tag(&mut self, tag: Box<str>)` and `set_xdg_toplevel_tag_description(&mut self, description: Box<str>)`.

Reason 1 being: It's only ever called with at least one `None` argument.

I also think the use of Option can be misleading here, as passing `None`  could also mean to delete the existing value. The confusion would at any rate be lifted by splitting this function into two, using above type signatures.

cornservant (commented):
Found a bug and provided some suggestions.

MithicSpirit:
Thank you for the review. I have updated the code accordingly. Not sure why I wanted to minimize new structs/methods before. I still haven't had the time to test this, but with luck I'll be able to do so next weekend.

cornservant (approved):

src/window/mapped.rs
```diff
@@ -423,17 +423,27 @@ impl Mapped {
         self.need_to_recompute_rules = true;
     }
 
-    pub fn set_xdg_toplevel_tag(&mut self, tag: Box<str>) {
-        if Some(&tag) != self.xdg_tag.tag.as_ref() {
-            self.xdg_tag.tag = Some(tag);
-            self.need_to_recompute_rules = true;
+    pub fn set_xdg_toplevel_tag(&mut self, tag: &str) {
+        if let Some(cur) = &self.xdg_tag.tag {
```
cornservant: shorter variant
```rust
        if Some(tag) == self.xdg_tag.tag.as_deref() {
            return;
        }
```

src/window/mapped.rs
```diff
@@ -423,17 +423,27 @@ impl Mapped {
...
+
+        self.xdg_tag.tag = Some(tag.into());
+        self.need_to_recompute_rules = true;
     }
 
-    pub fn set_xdg_toplevel_tag_description(&mut self, description: Box<str>) {
-        if Some(&description) != self.xdg_tag.description.as_ref() {
-            self.xdg_tag.description = Some(description);
+    pub fn set_xdg_toplevel_tag_description(&mut self, description: &str) {
+        if let Some(cur) = &self.xdg_tag.description {
```
cornservant:
```rust
        if Some(description) == self.xdg_tag.description.as_deref() {
            return;
        }
```

cornservant (approved):

src/window/mapped.rs
MithicSpirit: Ah, thanks. I was looking for something that did `AsRef` but I forgot about `Deref`.

MithicSpirit:
Uhh so turns out this was deadlocking lol. Fixed in cefa3aac3cd1f44792964a074309ae2925a286d3 by just pretending unmapped windows don't have an xdg-tag. I'm not sure whether that's the case or not, but it shouldn't be too hard to fix by just adding that field to `Configured`.

My relatively brief testing suggests that this is now working.

cornservant (approved):
Code looks fine.

I tested it with some example clients (https://github.com/cornservant/wayland-examples/tree/3841acd1b3a2cf5d63aa36b8f83d5b319dae91ac) and the window rule is applied as expected, i.e.

- when setting the tag before the first commit, window rules that apply once at the start (such as `open-fullscreen`) apply
- the window rules are reapplied correctly after the client changes the tag

MithicSpirit:
Note that the `proton-game` thing in the docs currently only works with some forks of Proton that ship with appropriate wayland support (I believe these are Proton-GE, Proton-CachyOS, and maybe Proton-EM). It would probably be good to change this, but I am not aware of any other clients that use xdg-toplevel-tag-v1.

docs/wiki/Configuration:-Window-Rules.md
```diff
@@ -197,13 +197,29 @@ You can find the title and the app ID of a window by running `niri msg pick-wind
...
-> 
+>
 > ```json
 > "wlr/taskbar": {
 >     "tooltip-format": "{title} | {app_id}",
 > }
 > ```
 
+#### `xdg-tag`
+
```
cornservant, line 208: Missing a `<sup>Since: next release</sup>`

MithicSpirit: :facepalm: fixed

cornservant (approved):

PR #4147, opened 2026-06-04
ipc: expose workspace scrolling view position

### LINKED ISSUE #2381: Empty `tile_pos_in_workspace_view` for tiled windows
When retrieving window layout info via niri-ipc, the `tile_pos_in_workspace_view` field is `null` for tiled windows, and only populated for floating windows. Since the field is described as "tile position within the current view of the workspace", I believe it should also be populated for tiled windows.

This information is useful because it enables precisely selecting a window with `slurp` for screenshot with 3rd party tools. (Would be very nice if niri natively supports this.)

### System Information

* niri version: niri 25.08 (01be0e6)

* Distro: Arch Linux

* GPU: Intel HD Graphics 630

* CPU: Intel Core i7-7700

kiryl:
Closes #2381.

Posted the design in [#2381](https://github.com/niri-wm/niri/issues/2381#issuecomment-3079060) a week ago for early feedback; opening this PR to make it easier to review concretely. Happy to back out if the shape is wrong.

## Summary

Adds a per-workspace viewport offset to niri-ipc so clients can compute on-screen tile positions themselves, without niri fanning out per-tile geometry on every viewport move:

- `Workspace::scrolling_view_pos: f64` — x coordinate of the workspace view's left edge in scrolling-layout coordinate space. Uses `target_view_pos()`, so animation interpolation doesn't churn the value.
- `Event::WorkspaceViewPosChanged { workspace_id, view_pos }` — emitted from `ipc_refresh_workspaces` only when the resting value changes. One event per workspace per logical navigation step, regardless of tile count.

## Why this shape

#2381 asked for `tile_pos_in_workspace_view` on `WindowLayout` to be filled for scrolling tiles. @YaLTeR explained why per-tile fill is the wrong design:

> tile_pos_in_workspace_view in particular is likely not happening for tiled windows because resizing 1 window has to update this for all (potentially very many) windows to the right of it.

@jbms proposed the alternative in #3305:

> add a field to Workspace that specifies what the current viewport is and how it maps to the output coordinate space. That would also provide a natural way to integrate with the event stream without producing an excessive number of messages as you move the viewport.

This PR is that. It also lines up with @YaLTeR's "clients can compute on their own" direction from #1265 and @binarin's expanded request in #3305 (raw viewport info on workspaces, with widths/heights derivable from existing fields).

## How clients use it

To get a scrolling tile's position within the workspace view:

```
column_x      = sum of tile_size.0 across preceding columns (any tile per column)
tile_y_in_col = sum of tile_size.1 across preceding tiles in the same column
view_pos.x    = column_x - workspace.scrolling_view_pos
view_pos.y    = tile_y_in_col
```

`scrolling_view_pos` is usually negative — niri leaves a `gaps`-sized space to the left of column 1, plus more when the active column is centered (`center-focused-column`). The field doc spells this out.

Floating tiles continue to read from the existing `WindowLayout::tile_pos_in_workspace_view`.

## What this deliberately doesn't do

- **Per-tile workspace-view positions for scrolling tiles.** That's the cascading-update path the #2381 thread ruled out.
- **Overview-mode coordinate mapping.** Worth a follow-up via a separate optional field on `Workspace` as @jbms suggested in #3305.
- **Tabbed-column visibility.** Not exposed here; clients can infer from `pos_in_scrolling_layout`.

## Relationship to #3305

#3305 takes a complementary, pull-based approach: a one-shot `Request::WindowGeometries` returning per-window rects computed server-side. The two aren't mutually exclusive — a `niri msg window-geometries` convenience CLI built on top of this PR's primitives would address the same ergonomic gap for shell pipelines and `slurp`-style consumers. Suggest landing the primitives first and revisiting #3305 against them.

## Usage example

A working consumer: [kiryl/ashell @ feat/minimap-scrolling-view-pos](https://github.com/kiryl/ashell/tree/feat/minimap-scrolling-view-pos) implements a status-bar minimap module. The client-side reconstruction is in [`src/services/compositor/niri.rs`](https://github.com/kiryl/ashell/blob/feat/minimap-scrolling-view-pos/src/services/compositor/niri.rs) — one helper, ~50 lines.

2026-05-29T10:07:53

## Test plan

- niri-ipc unit tests pass, including a new test exercising `WorkspaceViewPosChanged` state application.
- `cargo clippy` and `cargo test` clean on the niri workspace.
- Drove the minimap against a local niri build: viewport movement emits one event per navigation step, the minimap updates accordingly, no per-tile IPC traffic on scroll.

### DIFF
--- a/niri-ipc/src/state.rs
+++ b/niri-ipc/src/state.rs
@@ -158,6 +158,14 @@ impl EventStreamStatePart for WorkspacesState {
                 let ws = ws.expect("changed workspace was missing from the map");
                 ws.active_window_id = active_window_id;
             }
+            Event::WorkspaceViewPosChanged {
+                workspace_id,
+                view_pos,
+            } => {
+                let ws = self.workspaces.get_mut(&workspace_id);
+                let ws = ws.expect("changed workspace was missing from the map");
+                ws.scrolling_view_pos = view_pos;
+            }
             event => return Some(event),
         }
         None
@@ -319,5 +327,40 @@ impl EventStreamStatePart for CastsState {
             event => return Some(event),
         }
         None
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn ws(id: u64, view_pos: f64) -> Workspace {
+        Workspace {
+            id,
+            idx: 1,
+            name: None,
+            output: None,
+            is_urgent: false,
+            is_active: true,
+            is_focused: true,
+            active_window_id: None,
+            scrolling_view_pos: view_pos,
+        }
+    }
+
+    #[test]
+    fn view_pos_changed_updates_only_the_named_workspace() {
+        let mut state = WorkspacesState::default();
+        state.apply(Event::WorkspacesChanged {
+            workspaces: vec![ws(1, 0.0), ws(2, 100.0)],
+        });
+
+        state.apply(Event::WorkspaceViewPosChanged {
+            workspace_id: 1,
+            view_pos: 250.5,
+        });
+
+        assert_eq!(state.workspaces[&1].scrolling_view_pos, 250.5);
+        assert_eq!(state.workspaces[&2].scrolling_view_pos, 100.0);
     }
 }
--- a/src/ipc/server.rs
+++ b/src/ipc/server.rs
@@ -631,6 +631,17 @@ impl State {
                 });
             }
 
+            // Check if this workspace's resting horizontal scroll position changed. This is
+            // intentionally a per-workspace event so a viewport move emits a single event for the
+            // workspace, instead of fanning out to one WindowLayout update per tile.
+            let scrolling_view_pos = ws.scrolling_view_pos();
+            if ipc_ws.scrolling_view_pos != scrolling_view_pos {
+                events.push(Event::WorkspaceViewPosChanged {
+                    workspace_id: id,
+                    view_pos: scrolling_view_pos,
+                });
+            }
+
             // Check if this workspace urgent state changed.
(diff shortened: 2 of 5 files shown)


alexindigo:
I'm interested in this one as well, I think it should help me to render input candidates (akin to fcitx) next to the cursor.

kiryl:
The PR is around for two month. Any feedback?

src/ipc/server.rs
```diff
@@ -631,6 +631,17 @@ impl State {
                 });
             }
 
+            // Check if this workspace's resting horizontal scroll position changed. This is
+            // intentionally a per-workspace event so a viewport move emits a single event for the
+            // workspace, instead of fanning out to one WindowLayout update per tile.
+            let scrolling_view_pos = ws.scrolling_view_pos();
+            if ipc_ws.scrolling_view_pos != scrolling_view_pos {
```
HigherOrderLogic, line 638: Wont this spam the client when the windows stripe is moved using trackpad/touchscreen?

niri-ipc/src/state.rs
```diff
@@ -321,3 +329,38 @@ impl EventStreamStatePart for CastsState {
...
+
+        state.apply(Event::WorkspaceViewPosChanged {
+            workspace_id: 1,
+            view_pos: 250.5,
+        });
+
+        assert_eq!(state.workspaces[&1].scrolling_view_pos, 250.5);
+        assert_eq!(state.workspaces[&2].scrolling_view_pos, 100.0);
+    }
+}
```
HigherOrderLogic, line 366: Is this test needed?

Issue #3594, opened 2026-03-11
Screen reader doesn't announce anything from niri


YaLTeR:
This is an accesskit_unix 0.18+ regression, not clear what to do about it yet. https://github.com/AccessKit/accesskit-c/issues/76

YaLTeR:
I downgraded the crate for now to get the screen reader working again.

takashialpha:
any updates on this?

YaLTeR:
If you look at the linked issue there are no updates. In niri the version is downgraded and so works for now

takashialpha:
@YaLTeR Actually, I think that there is an update. PR # 715 on accesskit (merged May 23, 2026) fixes the issue by watching IsEnabled instead. Confirmed to properly detect screen readers starting before or after the app. It's Merged, but not released yet.
accesskit 0.24 and accesskit_unix 0.21 (specfically when 0.21.2 gets released, will include that fix) should work fine after adapting the usage in niri.

YaLTeR:
Upgraded in fdb6d85fc78355762bcf3cf71fe4037681a766f9. Checked and it seems to work indeed!

PR #4148, opened 2026-06-05
Edition 2024, Rust v1.96, Deps upgraded to latest (locked to x.y), Unmaintained knuffel replaced with knus

takashialpha:
hiii, here's everything in this PR:

* bumped to Edition 2024: this mostly required wrapping env usage in `unsafe`.
* bumped the Rust version to v1.96.
* replaced the unmaintained `knuffel` crate with `knus`: this mostly required renaming `knuffel` references throughout the codebase. As a bonus, this lets us pull in the latest `miette`.
* ran fmt as recommended by the contributing guide. **All tests pass, and clippy runs with 0 (no) warnings**.
* tried bumping `accesskit` and `accesskit_unix` as well, but left them as-is (the only dependencies not on their latest versions), since the corresponding issue (#3594) doesn't seem to have a solution other than downgrading.**  **updates below**


takashialpha:
As a Niri user, I also replaced the previous binary I was using with the one from this PR, and everything has been running smoothly so far. I'll continue using it and will report back if I find any problems.

Update: Smooth :smiley:

takashialpha:
Actually, we WILL be able to bump accesskit and accesskit_unix to latest, when accesskit_unix v0.21.2 comes out. That release will include the fix for the issue #3594.

YaLTeR:
That's good to hear. FWIW, I'm not sure I'll be accepting this PR:
- I'd like to bump to edition 2024 myself because 1. I need to review some other PRs first to avoid massive conflicts and 2. When I tried doing it recently, there were a bunch of nonobvious spots where it was easy to accidentally change behavior, and I want to inspect all of them.
- We're not bumping MSRV unnecessarily, certainly not to the current newest version.
- Updating knuffel to knus has potential for config breakage (that I don't see addressed in any way), yet there's barely any benefit to doing it. Knuffel works perfectly fine until we maybe move to smth else entirely.

takashialpha:
Thanks for the feedback. I went back and checked the knus changes in more detail.

knus 3.2.0 is effectively identical to knuffel 3.2.0, which niri already uses. The changes in knus 3.3.x are fairly small: a few AST fields became public, `parse_*` now accepts `impl AsRef<str>`, there's a derive macro hygiene fix, some extra trait derives, and an upgrade to miette 7.6. As far as I can tell, none of those affect the KDL decoding logic or how `config.kdl` is interpreted; the most visible change is improved error rendering.

If I've missed something, I'm happy to be corrected, but based on the documented diff I don't see much breakage risk from the knus switch itself.

I'm going to leave the PR as-is rather than split it up, so feel free to review whichever parts make sense when you have time. If it helps, I can point you to the places that may be affected by the Rust 2024 temporary drop order changes.

One thing I'm curious about: you mentioned possibly moving away from knuffel/knus entirely in the future. Do you have a specific alternative in mind, such as native serde support in kdl-rs, [facet](https://facet.rs), or something else? That would help me understand where you'd prefer config-parsing effort to go.

YaLTeR:
> knus 3.2.0 is effectively identical to knuffel 3.2.0, which niri already uses. The changes in knus 3.3.x are fairly small: a few AST fields became public, `parse_*` now accepts `impl AsRef<str>`, there's a derive macro hygiene fix, some extra trait derives, and an upgrade to miette 7.6. As far as I can tell, none of those affect the KDL decoding logic or how `config.kdl` is interpreted; the most visible change is improved error rendering.

That's good to know, I'll keep it in mind. But well, we have no problems with knuffel (other than miette version, but that only becomes a problem if we add a new one in the dep tree). If anything, the fact that this config parsing code doesn't change is good because it means no possibility for accidental breakage.

> One thing I'm curious about: you mentioned possibly moving away from knuffel/knus entirely in the future. Do you have a specific alternative in mind, such as native serde support in kdl-rs, [facet](https://facet.rs), or something else? That would help me understand where you'd prefer config-parsing effort to go.

I've been thinking about different approaches (among the ones you mentioned, plus maybe hand-rolling something), but I have no concrete plan. It's a big chunk of work either way. It's not pressing and I'm not sure when/if I can get around to it (plenty of more important PRs to deal with first).

takashialpha:
Since this code likely won't be merged anytime soon, I went ahead and updated AccessKit to the latest version, as the release we need should be out soon anyway. I also cleaned up the Clippy warnings across the repository.

I took a look through the Edition 2024 migration, Clippy fixes, and related changes as well, and I couldn't spot any behavior changes. If I overlooked something, let me know. More eyes on that would be appreciated

takashialpha:
Actually, I'm going to close this PR since it isn't particularly useful for the project right now and would conflict more than help at this time.

A couple of notes that might be useful for niri in the future:

> Actually, we *will* be able to bump `accesskit` and `accesskit_unix` to the latest versions once `accesskit_unix v0.21.2` is released.

You may want to file an issue to track that.

Also, as discussed above, `knus` could be a pretty nice drop-in replacement down the road.

Thanks for taking the time to look into this.

YaLTeR:
Do you mind testing accesskit_unix to reproduce the issue with current stable, and test accesskit_unix git to verify that it got fixed? The repro is basically "run orca and see if niri announces workspace switches/overview"

takashialpha:
I think I could. Will file an issue for it later. 🩵

takashialpha:
done; i have already made some tests but testing it further would be better

PR #4151, opened 2026-06-06
feat: add opt-in layer open animations

trevarj:
Animates opening layers, such as fuzzel or another launcher.

Disabled by default to avoid undesirable behavior with bars and other
floating windows. Must be enabled in the layer_rule and configured in
the animations block.

I don't see a reason to have a layer-close animation (window-close is hardly noticeable for me).
If this is the wrong approach, please close.

### DIFF
--- a/docs/wiki/Configuration:-Animations.md
+++ b/docs/wiki/Configuration:-Animations.md
@@ -25,6 +25,11 @@ animations {
         curve "ease-out-expo"
     }
 
+    layer-open {
+        duration-ms 150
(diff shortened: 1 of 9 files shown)


trevarj:
https://github.com/user-attachments/assets/2b07c467-9766-4586-9df0-966b674033e1

HigherOrderLogic:
Related: #3481.

trevarj:
@HigherOrderLogic thanks, closing

PR #4162, opened 2026-06-07
Use `RUSTFLAGS` instead of `CARGO_BUILD_RUSTFLAGS` in dev shell

Noratrieb:
`CARGO_BUILD_RUSTFLAGS` has a very low precedence and is overruled by global `target.<tuple>.rustflags`.
Since these rustflags are very important, ensure that they have a higher precedence. We could even go for `CARGO_ENCODED_RUSTFLAGS` which has the highest precedence, but is slightly annoying to construct and probably not worth it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -139,8 +139,8 @@
             '';
 
           env = {
-            # Force linking with libEGL and libwayland-client
-            # so they can be discovered by `dlopen()`
+            # Force linking with libEGL and libwayland-client so they end up in RPATH and
+            # can be discovered by `dlopen()`
             RUSTFLAGS = toString (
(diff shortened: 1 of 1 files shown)


HigherOrderLogic:
While you're at it, can you bump the `nixpkgs` rev? While I was messing around a while back, there was a mismatch in `libglvnd` that caused nested Niri session to crash.

No problem if you dont want to do it tho.

YaLTeR:
cc @zimward

zimward (approved):
afaik `RUSTFLAGS` is the one that one would usually want to set

YaLTeR:
Thanks

PR #4174, opened 2026-06-10
feat: allow configuring center area

NSPC911:
This is slightly similar to [AltSnap](https://github.com/RamonUnch/AltSnap)'s configuration for resizing the area.

image

In AltSnap, you can customise the centre's percentage area, so in the screenshot above, the centre area is 50% width and height. This PR aims to provide a configuration option similar to that.

config:
```kdl
input {
    center-area-percentage width="33%" height="33%"
}
```

<details>
<summary>ascii version of video describing the feature</summary>

<p>

For example, you have a window like this
<pre>
┌───────────┐
│           │
│           │
│           │
└───────────┘
</pre>
If you set the centre percentage to 50% width and 50% height
<pre>
┌──┬─────┬──┐
├──┼─────┼──┤
│  │     │  │
├──┼─────┼──┤
└──┴─────┴──┘
</pre>
Only the outer edges and corners can be used to resize

If you set the centre percentage to 0% width and 50% height
<pre>
┌─────┬─────┐
├─────┼─────┤
│     │     │
├─────┼─────┤
└─────┴─────┘
</pre>
there is no top and bottom edge resizing

If you set the centre percentage to 0% width and 0% height
<pre>
┌─────┬─────┐
│     │     │
├─────┼─────┤
│     │     │
└─────┴─────┘
</pre>
There is now only corner resizing

---

</p>
</details>

https://github.com/user-attachments/assets/daa9deab-1008-46dd-8bcf-b410e553aa2f

Some notes on design choices:
- I decided to pass the `CenterAreaPercentage` to the `layout/workspace.rs`, which means changing `layout/monitor.rs`, `layout/mod.rs` and `input/mod.rs`. I couldn't find a better way to pass this struct. It would be nice if someone could guide me
- I also decided that a new `CenterSidePercent` was the best way to go to validate the percentage, I'm not sure how to do it any other way
- I named it `center-area-percentage` and not `non-resizable-area` because it would become a breaking change if resizing in the center is allowed _eventually_ (this guy is not coping trust)
- The range is 0% inclusive to 100% exclusive because I don't want the corner resizing to be fully unavailable, although you can get almost 100 by using `99.99999%`, so that's something I guess
- The `width` and `height` are optional, which means people can do `center-area-percentage;`, so uhh...
- It's not a nested definition, I didn't like the look of this below
	```kdl
   input {
       center-area-percentage {
           width "50%"
           height "50%"
       }
   }
   ```

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2929,8 +2929,13 @@ impl State {
                 else if button == Some(MouseButton::Right) && !pointer.is_grabbed() && mod_down {
                     let location = pointer.current_location();
                     let (output, pos_within_output) = self.niri.output_under(location).unwrap();
-                    let center_area_percentage =
-                        self.niri.config.borrow().input.center_area_percentage;
+                    let center_area_percentage = self
+                        .niri
+                        .config
+                        .borrow()
+                        .input
+                        .center_area_percentage
+                        .unwrap_or_default();
                     let edges = self
                         .niri
                         .layout
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2929,10 +2929,12 @@ impl State {
                 else if button == Some(MouseButton::Right) && !pointer.is_grabbed() && mod_down {
                     let location = pointer.current_location();
                     let (output, pos_within_output) = self.niri.output_under(location).unwrap();
+                    let center_area_percentage =
+                        self.niri.config.borrow().input.center_area_percentage;
                     let edges = self
                         .niri
                         .layout
-                        .resize_edges_under(output, pos_within_output)
+                        .resize_edges_under(output, pos_within_output, center_area_percentage)
                         .unwrap_or(ResizeEdge::empty());
 
                     if !edges.is_empty() {
--- a/src/layout/workspace.rs
+++ b/src/layout/workspace.rs
@@ -1772,7 +1772,7 @@ impl<W: LayoutElement> Workspace<W> {
     pub fn resize_edges_under(
         &self,
         pos: Point<f64, Logical>,
-        center_area_percentage: Option<CenterAreaPercentage>,
+        center_area_percentage: CenterAreaPercentage,
     ) -> Option<ResizeEdge> {
         self.tiles_with_render_positions()
             .find_map(|(tile, tile_pos, visible)| {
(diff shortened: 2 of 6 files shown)


src/layout/workspace.rs
```diff
@@ -1780,17 +1785,26 @@ impl<W: LayoutElement> Workspace<W> {
                 let pos_within_tile = pos - tile_pos;
 
                 if tile.hit(pos_within_tile).is_some() {
+                    let w = center_area_percentage
+                        .and_then(|c| c.width)
+                        .map(|p| p.0)
```
titaniumtraveler: I think those `.map()`s can be inlined into the `.and_then()`s

src/layout/workspace.rs
```diff
@@ -1780,17 +1785,26 @@ impl<W: LayoutElement> Workspace<W> {
...
+                        .unwrap_or(1. / 3.);
+
                     let size = tile.tile_size().to_f64();
 
                     let mut edges = ResizeEdge::empty();
-                    if pos_within_tile.x < size.w / 3. {
+                    if pos_within_tile.x < ((size.w * (1. - w)) / 2.) {
                         edges |= ResizeEdge::LEFT;
-                    } else if 2. * size.w / 3. < pos_within_tile.x {
+                    } else if pos_within_tile.x > ((size.w * (1. + w)) / 2.) {
```
titaniumtraveler: I would somewhat strongly prefer using `<` than `>`, because those are usually easier to *specially* think about.
```diff
-pos_within_tile.x > ((size.w * (1. + w)) / 2.)
+((size.w * (1. + w)) / 2.) < pos_within_tile.x
```
(both here, and a few lines down)

src/input/mod.rs
```diff
@@ -2929,10 +2929,12 @@ impl State {
                 else if button == Some(MouseButton::Right) && !pointer.is_grabbed() && mod_down {
                     let location = pointer.current_location();
                     let (output, pos_within_output) = self.niri.output_under(location).unwrap();
+                    let center_area_percentage =
+                        self.niri.config.borrow().input.center_area_percentage;
```
titaniumtraveler: You should probably use `center_area_percentage.unwrap_or_default()` here (and `#[derive(Default)]` for `CenterAreaPercentage`) as `CenterAreaPercentage { width: None, height: None }` is semantically equivalent to just `None` and it makes the code simpler to work with.

titaniumtraveler (approved):

titaniumtraveler:
Only a few nitpicks, but otherwise seems like a very sane implementation.

YaLTeR:
Why would we allow configuring this? Why do you think it justifies the added code?

NSPC911:
This is a preference thing. With AltSnap, there is an option to resize both ways when resizing at the center (see video)

https://github.com/user-attachments/assets/841fdf95-8c0b-402f-b11e-9fcc0717d0e1

However, I doubt that this will be implemented any time soon in Niri. So instead, I would prefer a center area of sides 15% so I have more area to actually resize.

Some people would also prefer a corner only resizing, which allows them to also avoid the unresizable center problem.

Issue #176, opened 2024-02-07
Vertically center columns smaller than full height


YaLTeR:
Ref. https://github.com/YaLTeR/niri/discussions/175

user12xxxxxx:
Hey, @YaLTeR 
Updates on this

YaLTeR:
No updates and please don't ping when it's not needed

PR #4181, opened 2026-06-11
Add vertical centering for single windows in a column

spaceavenue:
Adds support for vertically centering single windows in columns. This can be configured in the config by setting `center-single-window` under `layout`, and is off by default. Implements #176

Singular windows with fixed heights will be centered in the column and with multiple windows the current auto-resizing behavior is maintained.

Also something to consider is whether this should also be an action, if there should be support for top and bottom alignment as well, and if multiple fixed heights windows should follow the alignment too.

P.S. I'm not that great at Rust and programming in general so this implementation might not be the best :p


Uzlkav:
Can't help with code but `center-single-window` feels vague and to close to `center-focused-column`/`always-center-single-column`. I propose renaming to `vertically-center-single-window`, to make it clear what the option does.

Sempyos:
Perhaps `center-window-in-column`, which fits in with actions such as `focus-window-in-column` or `center-window`

spaceavenue:
I was considering changing the name because yeah, it is close to `center-window`. Probably something like `center-window-in-column` would work, but it might be confusing because it only centers when there's a singular window, and someone might mistake it for multiple windows being centered. `center-single-window-in-column` feels kinda long but it might be the most communicative.

PR #4186, opened 2026-06-12
wiki: Add instructions for working around JetBrains Idea crash when opening the settings window under Wayland.

11backslashes:


YaLTeR:
Ty

PR #4189, opened 2026-06-13
Add `window-under-cursor` ipc msg

KapDecy:
**Summary**

Add a new `niri msg window-under-cursor` IPC command for getting information about the window currently under the cursor without requiring an extra mouse click.

Previously, `niri msg pick-window` was the only way to get a window id based on pointer position, but it always entered an interactive picker and required the user to left-click the target window. This made it awkward to use from scripts or key bindings where the cursor is already over the intended window.

**Changes**

- Add `Request::WindowUnderCursor` and `Response::WindowUnderCursor`.
- Add the `niri msg window-under-cursor` CLI command.
- Reuse the existing compositor-side `window_under_cursor()` logic, so behavior matches existing window activation/picking rules.

**Usage**

```sh
niri msg window-under-cursor
```

JSON output:

```sh
niri msg --json window-under-cursor
```

Example use with closing window under cursor:

```sh
niri msg action close-window --id $(niri msg --json window-under-cursor | jq .id)
```

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -81,6 +81,8 @@ pub enum Request {
     FocusedOutput,
     /// Request information about the focused window.
     FocusedWindow,
+    /// Request information about the window under the cursor.
+    WindowUnderCursor,
     /// Request picking a window and get its information.
     PickWindow,
     /// Request picking a color from the screen.
@@ -155,6 +157,8 @@ pub enum Response {
     FocusedOutput(Option<Output>),
     /// Information about the focused window.
     FocusedWindow(Option<Window>),
+    /// Information about the window under the cursor.
+    WindowUnderCursor(Option<Window>),
     /// Information about the picked window.
     PickedWindow(Option<Window>),
     /// Information about the picked color.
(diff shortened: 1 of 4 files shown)


Atan-D-RP4:
I think it's better to make it an option under the existing IPC command like `niri msg pick-window --under-cursor` or something like that.

megabyte6:
I agree with @Atan-D-RP4. I think this makes more sense as to not pollute the options list.

karabalgasun:
If it is to be merged with pick-window, something like  `niri msg pick-window --immediate` or `niri msg pick-window --now` would make more sense than `--under-cursor` to a user. `pick-window` already picks the window under the cursor. The difference is this one would do it without additional interaction.

megabyte6:
> If it is to be merged with pick-window, something like `niri msg pick-window --immediate` or `niri msg pick-window --now` would make more sense than `--under-cursor` to a user. `pick-window` already picks the window under the cursor. The difference is this one would do it without additional interaction.

My vote is for `niri msg pick-window --now`.

feelyourwarmth:
need this fr

PR #4193, opened 2026-06-14
Implement a workspace toggle for centering the focused column

OptimisticPeach:
Essentially what it says on the tin. 

Introduces a new event you can fire on a keybind that toggles whether the current workspace centers the focused column or not. 

Video example:

https://github.com/user-attachments/assets/1440f1af-73bd-4e35-889e-cda9a49a351d

Of note is that it's per-workspace, so you can have one workspace be not-center-focused and one be center-focused. 

The keybind name is `toggle-center-column` and the IPC name should be similar.

-------------------------

The inspiration for this is my current workflow with an ultrawide; sometimes I just want to focus on one window at a time, and be able to keep track of other windows out of the corner of my eye (which is what center columns gives me), but sometimes I would rather be able to use two windows side-by-side, which is what `center-focused-column "never"` gives me.

I wasn't quite sure what to do with `on-overflow`; there's no real good option? 
- `on-overflow` acts like `never` in 99% of usecases, so switching to `always` seems reasonable.
- `on-overflow` doesn't really have a complimentary setting, so just leaving it as-is also seems reasonable.

### DIFF
--- a/niri-config/src/binds.rs
+++ b/niri-config/src/binds.rs
@@ -211,6 +211,7 @@ pub enum Action {
     SwapWindowRight,
     ToggleColumnTabbedDisplay,
     SetColumnDisplay(#[knuffel(argument, str)] ColumnDisplay),
+    ToggleCenterColumn,
     CenterColumn,
     CenterWindow,
     #[knuffel(skip)]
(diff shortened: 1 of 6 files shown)


Sempyos:
If you could provide a video demo of your new action and possibly add the IPC action/what was added to your first comment, that would be lovely

Uzlkav:
If I understand correctly this would be equivalent to changing the value of `center-focused-column` in the config file, which could also be achieved with a simple script. Why do you think this should be an action (genuine question, want to know if there's limitations to the mentioned scripting approach that warrant having a dedicated action).
If I'm wrong about what this does, lmk - agree with sempy that video would be useful.

OptimisticPeach:
Here's a video demonstrating what I mean :slightly_smiling_face: 

https://github.com/user-attachments/assets/1440f1af-73bd-4e35-889e-cda9a49a351d

> If I understand correctly this would be equivalent to changing the value of center-focused-column in the config file, which could also be achieved with a simple script.

Not quite; the original intention was indeed being able to just toggle this with a keybind, which yes, could be done with a script. 

Instead, this PR actually lets you toggle this per-workspace, as shown in the video. One workspace is set to center the focused column, and one isn't. Toggling it per-workspace _isn't_ something you'd be able to do with a script (at least I'm not aware of a way to do so).

> possibly add the IPC action/what was added to your first comment, that would be lovely

I'm not sure I follow; I thought I'd added the IPC action in the code? Or did you mean documenting it somehow?

Sempyos:
I mean just saying it here, so that anyone who views this PR has a clear understanding of what is added / how to use it

PR #4206, opened 2026-06-17
nix: remove `rust-overlay`, bump inputs

HigherOrderLogic:
This PR does 2 things:
- remove `rust-overlay`: previously this was added to get the nightly `rustfmt` and toolchain. However, given that Niri does not use any nightly features and `rustfmt` from `nixpkgs` can be run as nightly just fine, this input has became redundant.
- bump inputs: the current `nixpkgs` lock seems to have an older version of `libglvnd` compared to my system, which causes Niri to crash on `cargo r`. Bumping the input seems to fix that for me.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,28 +18,7 @@
     },
     "root": {
       "inputs": {
-        "nixpkgs": "nixpkgs",
-        "rust-overlay": "rust-overlay"
-      }
-    },
-    "rust-overlay": {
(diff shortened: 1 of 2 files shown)


YaLTeR:
cc @zimward

zimward (approved):
LGTM, less inputs are generally good

YaLTeR:
Thanks

PR #4210, opened 2026-06-18
mru: selected window is *always* in the center of the screen

Axlefublr:
Think of it as `scrolloff = 999` you might be used to in text editors.

> [!WARNING]
> This pr is a draft because I don't intend it to get into upstream. This is more so a easily findable, publically available patch that people with niri forks can use.


Axlefublr:
here is how it looks

https://github.com/user-attachments/assets/e30cc291-244f-4eae-9f9d-ab306330798e

PR #4218, opened 2026-06-20
Add suspend action to IPC

### LINKED ISSUE #3260: Suspend action missing from cli
not sure if this is intended but I noticed that the suspend action is missing from the cli.

coleleavitt:
## Summary
- Expose `suspend` in `niri_ipc::Action` so `niri msg action suspend` parses.
- Map the IPC action to the existing compositor `Action::Suspend` path.

## Issues
- Fixes #3260

## Testing
- `cargo test -q parses_suspend_action`
- pre-push hook: `cargo check`


bluelinden:
not seeing how this is needed... shouldn't this be handled by an idle daemon or systemd?

PR #4234, opened 2026-06-23
Feature: Allow keyboard input to pass to window while in Overview Mode (If user sets flag in config)

r0ckHopper:
### Issue Summary
---
When in Overview mode there is no way for user to push keyboard input into the focused window
If a user needs to input into a window they will need to exit Overview mode, then go back into Overview mode to continue rearranging.

### Proposed change
---
Allow user to set a flag allow_overview_throughput_keyboard in niri's config which tells niri *"don't block keyboard input from reaching focus window in overview mode"*

### Rationale for change
---
I find myself reflexively trying to type into newly spawned applications when in Overview mode. 
This happens when I am setting up multiple terminals for projects and arranging them, I keep going back and forth between overview mode and normal.
This change enables people like me to reap the benefits of overview mode without having to juggle overview state.
 
### Implementation
---
- Updated  ```Overview``` struct, ```OverviewPart```, and ```MergeWith<OverviewPart> for Overview``` with ```allow_overview_throughput_keyboard```, which is false by default **(used  ```disable_power_key_handling``` in niri-config/src/input.rs for an example)**
- Check Flag in ```update_keyboard_focus()``` and skip Overview mode specific logic so inputs can go to defaults which is the focused window
- Check Flag in ```on_keyboard()``` from src/input/mod.rs and skip Overview logic again, so no hardcoded overview commands intercept our our keyboard inputs (escape, arrow keys, enter)

### Testing
---
- Go into Overview mode -> spawn a new terminal -> type a command -> press enter 
- for change on line 1228, I have not found a way to test because I can't find any apps that live on Bottom or Background layer to request focus. But this should behave the exact same way. If you know of an app that could help test this behavior please let me know and I will test and update this bullet
- All cargo tests pass

###Demo

https://github.com/user-attachments/assets/475afcb5-6737-4be9-a594-2a3129379543

### DIFF
--- a/niri-config/src/misc.rs
+++ b/niri-config/src/misc.rs
@@ -123,6 +123,7 @@ pub struct Overview {
     pub zoom: f64,
     pub backdrop_color: Color,
     pub workspace_shadow: WorkspaceShadow,
+    pub allow_overview_throughput_keyboard: bool,
 }
 
 impl Default for Overview {
@@ -131,6 +132,7 @@ impl Default for Overview {
             zoom: 0.5,
             backdrop_color: DEFAULT_BACKDROP_COLOR,
             workspace_shadow: WorkspaceShadow::default(),
+            allow_overview_throughput_keyboard: false,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


arbunian:
I just tested it out and it works. I don't know any bottom or background layers so I didn't get to use it for that but I did get to use it in floorp, kitty, and obsidian and it works without any problems on my end. 

I'd want it to be added since it's one of the functionalities I wanted from [scroll](https://github.com/dawsers/scroll) along with the varying zoom functionality #3194.

PR #4247, opened 2026-06-25
Specify gnome portal as the choice for ScreenCast xdg-desktop-portal endpoint

itsjunetime:
I was having difficulty getting screencasting to work, and this change eventually made it work.

Because gnome is already required for this to work, I think this change makes sense and just allows niri to work better with systems where multiple portals are already set up.

### DIFF
--- a/resources/niri-portals.conf
+++ b/resources/niri-portals.conf
@@ -3,3 +3,4 @@ default=gnome;gtk;
 org.freedesktop.impl.portal.Access=gtk;
 org.freedesktop.impl.portal.Notification=gtk;
 org.freedesktop.impl.portal.Secret=gnome-keyring;
+org.freedesktop.impl.portal.ScreenCast=gnome;


HigherOrderLogic:
GNOME portal is already set in the default portal list.

The reason why it doesnt work on your system might be because you overwrote it with some other config file somewhere.

itsjunetime:
I did consider that, but I can't find anywhere where that might be the case. I've run `sudo rg portal.ScreenCast /` and the only portal files that come up that mention it at all are the niri one (since I've added this) and the gnome one. As far as I can tell (based on the [`portals.conf` spec](https://flatpak.github.io/xdg-desktop-portal/docs/portals.conf.html)), the `default=` should be sufficient, but it simply wasn't.

I guess I could close this in favor of opening an issue (and I'm happy to do that if that's how maintainers would prefer this sort of thing to be dealt with); but as someone who has been bit by this portals configuration stuff many times in the past as niri changes, I think this would probably help others.

And I doubt this will screw up anyone else's setup, as the default location of `/usr/share/xdg-desktop-portal/niri-portals.conf` is almost at the very bottom of the list of places for this configuration to exist at, so if anyone else has custom changes in any location other than `/usr/local/share/xdg-desktop-portal`, they'll still take precedence over this one.

YaLTeR:
It already works fine on Fedora, Arch, Aeryn, probably others. If it doesn't then your setup is likely broken.

PR #4291, opened 2026-07-09
​[SUPERSEDED] feat(backend/tty): dynamically suspend idle secondary GPUs

gimgiwer:
**This PR was automatically closed due to Git history mismatch during a force push.**  
Please track the updated and rebased Pull Request here: **[PR #4346](https://github.com/YaLTeR/niri/pull/4346)**

### DIFF
--- a/docs/mkdocs.yaml
+++ b/docs/mkdocs.yaml
@@ -81,7 +81,7 @@ nav:
     - Layer‐Shell Components: Layer‐Shell-Components.md
     - IPC, niri msg: IPC.md
     - Application-Specific Issues: Application-Issues.md
-    - Nvidia: Nvidia.md
+    - GPUs: GPUs.md
     - Xwayland: Xwayland.md
     - Gestures: Gestures.md
     - Fullscreen and Maximize: Fullscreen-and-Maximize.md
--- a/docs/wiki/Configuration:-Debug-Options.md
+++ b/docs/wiki/Configuration:-Debug-Options.md
@@ -33,6 +33,7 @@ debug {
     honor-xdg-activation-with-invalid-serial
     skip-cursor-only-updates-during-vrr
     deactivate-unfocused-windows
+    disable-idle-drm-device-suspend
 }
 
 binds {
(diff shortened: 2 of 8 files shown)


emekoi:
Does niri not already do this? I have an AMD laptop with a RTX 3070 and my laptop's discrete GPU enters D3cold without issue, even after hotplugging displays to the DisplayPort output that's connected directly to the discrete GPU. I think I did have to boot with the `nvidia-drm.fbdev=0` kernel parameter to get this to work in the past (c.f. https://github.com/NVIDIA/open-gpu-kernel-modules/issues/759) otherwise my GPU would never enter D3cold after niri used it, even after I killed niri. But now I boot with `nvidia-drm.fbdev=1` and it still works fine. I'm using Nvidia's open drivers, version 610.43.02.

gimgiwer:
> Does niri not already do this? I have an AMD laptop with a RTX 3070 and my laptop's discrete GPU enters D3cold without issue, even after hotplugging displays to the DisplayPort output that's connected directly to the discrete GPU. I think I did have to boot with the `nvidia-drm.fbdev=0` kernel parameter to get this to work in the past (c.f. [NVIDIA/open-gpu-kernel-modules#759](https://github.com/NVIDIA/open-gpu-kernel-modules/issues/759)) otherwise my GPU would never enter D3cold after niri used it, even after I killed niri. But now I boot with `nvidia-drm.fbdev=1` and it still works fine. I'm using Nvidia's open drivers, version 610.43.02.

It does, but only for the display path — and only until the first PRIME offload app.
The difference is what actually gets opened on the dGPU. A monitor plugged into the dGPU only needs the primary node (`/dev/dri/cardX`) for modesetting. Rendering still happens on the iGPU, so niri never touches EGL on the dGPU's render node (`/dev/dri/renderDX`). Unplug the monitor and the KMS state goes idle — and since there's no EGL context, the driver is free to drop back into D3cold. The kernel doesn't mind the primary fd staying open as long as there are no active outputs.
PRIME offload is a different story. As soon as an app offloads (or we import a dmabuf from the dGPU), niri has to bring up EGL on the render node. The EGL driver then holds that fd open and keeps the context cached basically forever. That fd pins the runtime PM refcount, so the card never suspends again — even after the offloaded app is long gone.
What this PR does: when the render node goes idle, we drop it from the GPU manager, which tears down the dGPU EGL context and closes the fd. The primary fd stays open so the session doesn't break, but the render node fd was the thing keeping the refcount up — once it's closed, the kernel can finally kick the card into D3cold.

PR #4307, opened 2026-07-13
layout: refill the view from the left after closing a column

SHORiN-KiWATA:
## Description

Niri's scrolling layout commonly shows two half-width columns side by side. Consider three such columns. The dotted lines below mark the view.

When columns 1 and 2 are visible, closing the column 2 keeps column 1 in place while column 3 enters from the right:

```text
before:

..............
[  1  ][  2  ] [  3  ]
..............


after closing column 2:

..............
[  1  ][  3  ]
..............
```

Column 1 remains in the left half of the screen.
In the mirror case, columns 2 and 3 are visible and column 1 is outside the view on the left:

```
before:

        ..............
[  1  ] [  2  ][  3  ]
        ..............
```

Currently, closing column 2 moves column 3 from the right half of the screen to the left, while column 1 remains outside the view:

```
current:

       ..............
[  1  ][  3  ]
       ..............
```

This PR changes the mirror case by moving the view to the left and activating column 1:

```
with this PR:

..............
[  1  ][  3  ]
..............
```

Column 1 enters from the left, while column 3 remains in the right half of the screen.

This introduces additional view movement, but that movement compensates for the layout change caused by removing column 2. As a result, the neighboring column that was already visible remains stable in screen space, making the two mirror cases behave consistently.

## Before

https://github.com/user-attachments/assets/925b8c73-63d0-41b5-8a9b-d2af5852e161

## After

https://github.com/user-attachments/assets/e4ddb6f7-0de0-4322-b288-e2673ce7d60b

## Testing

- `cargo test --all`
- `RUN_SLOW_TESTS=1 cargo test --all`
- Manually tested both left-side and right-side refills in a nested niri session

### DIFF
--- a/src/layout/closing_window.rs
+++ b/src/layout/closing_window.rs
@@ -47,6 +47,9 @@ pub struct ClosingWindow {
     /// Position in the workspace.
     pos: Point<f64, Logical>,
 
+    /// Horizontal movement synchronized with a layout change.
+    move_x: Option<Animation>,
+
     /// How much the texture should be offset.
     buffer_offset: Point<f64, Logical>,
 
@@ -149,6 +152,7 @@ impl ClosingWindow {
             block_out_from: snapshot.block_out_from,
             geo_size,
             pos,
+            move_x: None,
             buffer_offset,
             buffer_with_blocked_out_bg_offset,
             blocked_out_buffer_offset,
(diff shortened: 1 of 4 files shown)


zxc-rv:
I think it would be better if the refill would work on closing the right column too. At least as an option.

PR #4315, opened 2026-07-15
input: ignore power key presses immediately after resume

Hermit-Avon:
When niri handles the power key itself, the key press used to wake a suspended
system can be delivered after resume and interpreted as a fresh suspend request.
This can make the system immediately suspend again after waking up.

The observed journal sequence is:

```text
Lid opened
Power key pressed short
System returned from sleep operation 'suspend'
suspend requested from client ... ('niri')
The system will suspend now
```

This change listens for login1's `PrepareForSleep` signal. After
`PrepareForSleep(false)`, niri opens a short post-resume window and swallows
`XF86_PowerOff` presses during that window.

This preserves the normal behavior where pressing the power key while the
session is already awake suspends the system, while avoiding a suspend loop
caused by a wake-up key event.

Tested:

```text
cargo check --features dbus
```

I could not run `cargo +nightly fmt --all` on this machine because the installed
Cargo is not rustup-managed. I ran `cargo fmt --all`; it completed, but ignored
nightly-only rustfmt options.

### DIFF
--- a/src/dbus/freedesktop_login1.rs
+++ b/src/dbus/freedesktop_login1.rs
@@ -4,6 +4,17 @@ use zbus::names::InterfaceName;
 
 pub enum Login1ToNiri {
     LidClosedChanged(bool),
+    PrepareForSleep(bool),
+}
+
+#[zbus::proxy(
+    interface = "org.freedesktop.login1.Manager",
+    default_service = "org.freedesktop.login1",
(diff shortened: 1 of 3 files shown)


HigherOrderLogic:
Maybe see #3773.

Hermit-Avon:
Closing in favor of the existing related PR #3773. Thanks for pointing it out.

PR #4318, opened 2026-07-16
Relative Output Placement

blackd:
Implements #3793

I know this suggestion didn't get a lot of traction, but I implemented it for my own use and decided to put it here for posterity. Anyway.

#### Configuration

```kdl
output "eDP-1" {}

output "HDMI-A-1" {
    position left-of="eDP-1"
}

output "HDMI-A-2" {
    position above="HDMI-A-1" align="center"
}
```

Supported relations are `above=`, `below=`, `left-of=` and `right-of=`. These are self explanatory I hope

Supported alignments are `beginning` which is the default, `center` and `end`. The code has more about these. And in general this might need some work because the results may not be what the user expects. I would really want to discuss this if someone is interested.

The change keeps the current behaviour and just add the relative resolution on top of it. Outputs that have a missing anchor fall back to the missing position resolving. And are resolved to the right position when the anchor output is plugged in. 

I hope the commits are at the proper granularity.... I had to recreate them to follow the guidelines.

### DIFF
--- a/docs/wiki/Configuration:-Outputs.md
+++ b/docs/wiki/Configuration:-Outputs.md
@@ -173,6 +173,9 @@ Set the position of the output in the global coordinate space.
 This affects directional monitor actions like `focus-monitor-left`, and cursor movement.
 The cursor can only move between directly adjacent outputs.
 
+
+#### Absolute Positioning
+
 > [!NOTE]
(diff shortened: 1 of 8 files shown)


glitchifyed:
This would be great for my workflow!

PR #4320, opened 2026-07-17
fix: enforce pointer constraints per-axis

### LINKED ISSUE #3128: Confined cursors stick to region borders
When a client attempts to confine the cursor to a given region, and the mouse approaches the borders of the region, it never actually reaches the borders of the region and instead gets stuck right before it. This happens in floating and tiled windows, but not in fullscreen windows as long as there is not a another monitor sharing region edge. This issue occurs with Wine, SDL3, and libwayland applications (I think Wine and SDL3 both use libwayland). The program in the video is in [this](https://gist.github.com/emekoi/68fc5a70356baf09c33e618b82d3bd7d) gist. I don't have a good way of recording two monitors at once, so that isn't included in the video but the behavior on that shared edge is the same as in the floating case shown.

https://github.com/user-attachments/assets/79ac1fc5-050d-4bb4-b32f-0744045cdaa3

The only thing in my config is keybinds. 

### System Information

* niri version: niri 25.11 (v25.11-44-gb5640d52)

* Distro: arch linux (6.18.1)

* GPU: 
  * AMD Ryzen 9 5900HS with Radeon Graphics
  * NVIDIA GeForce RTX 3070[^1]

* CPU: AMD Ryzen 9 5900HS with Radeon Graphics

[^1]: This GPU is not used by Niri.

emekoi:
This fixes #3128 by enforcing pointer constraints on a per-axis basis instead of globally. This mean that if pointer leaves the x confines of a region, but does not leave the y confines, we simply discard any the change in x, without effecting any change in y. This turned out to be a "bug" in both anvil and niri, but unrelated to smithay. Just how pointer constraints were enforced.

This is based off of the work I did in https://github.com/Smithay/smithay/pull/2079 and I can confirm that both the smithay pointer constraints test client and the program I showed in the original issue work a lot better now.

The same caveats w.r.t correctness apply here as well since this is essentially discrete collision detection+resolution instead of continuous.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -2427,9 +2427,6 @@ impl State {
 
         let pos = pointer.current_location();
 
-        // We have an output, so we can compute the new location and focus.
-        let mut new_pos = pos + event.delta();
-
         // We received an event for the regular pointer, so show it now.
         self.niri.pointer_visibility = PointerVisibility::Visible;
         self.niri.tablet_cursor_location = None;
@@ -2504,6 +2501,10 @@ impl State {
             }
             None
         });
+
+        // We have an output, so we can compute the new location and focus.
+        let mut new_pos = pos + event.delta();
+
         if let Some((output, horizontal)) = spatial_grab.flatten() {
             if let Some(geo) = self.niri.global_space.output_geometry(&output) {
                 let geo = geo.to_f64();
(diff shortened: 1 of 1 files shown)


emekoi:
Discovered some jank: when the cursor is moving really quickly and might escape a confinement region, it stops dead in tracks which looks pretty jarring: 

https://github.com/user-attachments/assets/75a850ef-66c7-4c6c-b004-1b6db23baa81

I think I might try to find the closest confinement region and clamp the pointer to that rectangle instead of simply discarding all movement in the X or Y directions, or at the very least clamp to the surface.

PR #4329, opened 2026-07-18
Normalize full-height windows after manual resizing

### LINKED ISSUE #1811: A fixed height window can take up the whole column, and behaves unexpectedly compared to a visually-identical auto-height window.
# Background

[Spotted in Brodie's new video](https://www.youtube.com/watch?v=DeYx2exm04M&t=6m45s), this is actually an issue i've experienced a lot but was never a big enough annoyance for me to investigate. I was reminded of it just now, and realized i knew what was going on to cause this.

[As this code comment so helpfully describes](https://github.com/YaLTeR/niri/blob/f3f6e79eeca8924ff9cfea4b30006e5b782bc93e/src/layout/scrolling.rs#L221-L233), niri allows at most one window in a column to be non-auto size. By default, when a new column is created, its lone window has a size of `WindowHeight::Auto(1.0)`; which means to fill all available space.

# Explanation

That video shows a window on the left which has been manually resized in the vertical axis. But, it was put *back* to the max height possible. It now looks just like a newly created column. It is indistuinguishable from `WindowHeight::Auto(1.0)`, but it is actually `WindowHeight::Fixed(_)` with a maximal height.

So, when another window is consumed into its column, *that* window is autosized beneath a window taking up the whole column, and the newly consumed window only gets its minimum height. That's certainly never what i intend to do, or expect, when i consume a window into a column that *looks* like it is auto-sized to take up the full height.

# Steps to Reproduce

1. Spawn two terminals.
2. Focus the left one.
3. Resize it vertically in a way that cancels out to an identity transform:
    - `set-window-height -10%`
    - `set-window-height +10%`
4. `consume-window-into-column` the right terminal into the left one.
5. Observe that the right terminal is now only taking up its minimum height, rather than being split fairly.

If you take the same steps, but skip resizing the left terminal, then they will be split evenly, as expected.

Alternatively, after step 3, you can `reset-window-height` the left terminal. This has absolutely zero visual effect, as it does not affect the height of the window, but it will reset the internal state of the column to be `WindowHeight::Auto(1.0)`, and then consuming the right terminal will split them evenly, as expected.

# How i think it should work

I reckon, if we have a column that contains exactly one tile, and it is resized to take up the max height of the column, then it should be normalized into being auto-sized. But this check should only happen when it is manually resized.

If it was *not* manually resized vertically in order to become full-height, it should *not* be normalized like this. Currently, you can move a single-window column to a smaller display; it will visually update to have a smaller height; and then you can move it back to the larger display; and it will remember its height. In this case, the window has height `WindowHeight::Fixed(_)` with a height *larger* than the screen height. I think this makes a lot of sense, and the height should not be normalized to `WindowHeight::Auto(1.0)` when moving across displays.

lindestad:
Fixes #1811.

When the only window in a column is manually resized back up to the maximum height, store it as auto-height rather than fixed-height. This makes a later `consume-window-into-column` split the available space (which is intuitive) instead of leaving the consumed window at its minimum height (confusing from a user perspective).

The conversion only happens while resizing upward in a single-window column, preserving fixed heights that are merely clamped by output changes. The comparison uses rounded logical-pixel heights to avoid fractional-scale precision issues.

## Manual validation

Both captures use identical nested niri sessions and the following steps:

1. Spawn two terminals in separate columns.
2. Focus the left terminal.
3. `set-window-height -10%`
4. `set-window-height +10%`
5. `consume-window-into-column`

### Before
The image shows the resulting state, meaning there are actually two terminals here:
before
The consumed window collapses to one pixel.

### After
Now the terminal is auto resized, so it is actually visible:
after
The windows split the column evenly.

A regression test covers the same sequence at output scales 1.0 and 1.5.

Tested with:

- `cargo test --all --exclude niri-visual-tests`
- Matched nested niri reproduction before and after the fix using real Wayland Alacritty clients

### DIFF
--- a/src/layout/tests.rs
+++ b/src/layout/tests.rs
@@ -2467,6 +2467,46 @@ fn set_window_height_recomputes_to_auto() {
 }
 
 #[test]
+fn set_window_height_to_max_recomputes_to_auto() {
+    for scale in [1., 1.5] {
+        let ops = [
+            Op::AddScaledOutput {
+                id: 1,
+                scale,
+                layout_config: None,
+            },
+            Op::AddWindow {
+                params: TestWindowParams::new(0),
+            },
+            Op::Communicate(0),
+            Op::AddWindow {
+                params: TestWindowParams::new(1),
+            },
+            Op::Communicate(1),
+            Op::FocusColumnLeft,
+            Op::SetWindowHeight {
+                id: None,
+                change: SizeChange::AdjustProportion(-10.),
+            },
+            Op::Communicate(0),
+            Op::SetWindowHeight {
+                id: None,
+                change: SizeChange::AdjustProportion(10.),
+            },
+            Op::Communicate(0),
+            Op::ConsumeWindowIntoColumn,
+        ];
+
+        let layout = check_ops(ops);
+        let mut windows = layout.windows().map(|(_, win)| win);
+        let first = windows.next().unwrap().requested_size().unwrap();
+        let second = windows.next().unwrap().requested_size().unwrap();
+
+        assert_eq!(first.h, second.h, "scale: {scale}");
+    }
+}
+
+#[test]
 fn one_window_in_column_becomes_weight_1() {
     let ops = [
         Op::AddOutput(1),
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -4957,6 +4957,7 @@ impl<W: LayoutElement> Column<W> {
                 tile.window_height_for_tile_height(tile_height)
             }
         };
+        let is_resizing_up = window_height > current_window_px;
 
         // Clamp the height according to other windows' min sizes, or simply to working area height.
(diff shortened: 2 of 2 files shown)


src/layout/tests.rs
```diff
@@ -2466,6 +2466,46 @@ fn set_window_height_recomputes_to_auto() {
...
+            Op::Communicate(1),
+            Op::FocusColumnLeft,
+            Op::SetWindowHeight {
+                id: None,
+                change: SizeChange::AdjustProportion(-10.),
+            },
+            Op::Communicate(0),
+            Op::SetWindowHeight {
+                id: None,
+                change: SizeChange::AdjustProportion(10.),
```
megabyte6, line 2494: I'm personally not a huge fan of leaving the trailing zero off (10. instead of 10.0). It's more of a preference, though. I know some other places in the codebase use that pattern.

megabyte6 (approved):
Looks good to me 👍 
I tested it locally and it seems to work well.

PR #4330, opened 2026-07-19
only refresh windows after a surface commit

bdjnk:
We add a new top level surface_committed flag. When a commit occurs in compositor we set it to true, and on refresh_layout if it is true we refresh all windows and reset it. This helps prevent unnecessary window refreshes, reducing CPU use modestly in certain scenarios. It also centralizes the window refresh mechanism which was previously spread across three files.

### DIFF
--- a/src/handlers/compositor.rs
+++ b/src/handlers/compositor.rs
@@ -61,6 +61,8 @@ impl CompositorHandler for State {
         on_commit_buffer_handler::<Self>(surface);
         self.backend.early_import(surface);
 
+        self.niri.surface_committed = true;
+
         let mut root_surface = surface.clone();
         while let Some(parent) = get_parent(&root_surface) {
             root_surface = parent;
--- a/src/layout/floating.rs
+++ b/src/layout/floating.rs
@@ -1218,8 +1218,6 @@ impl<W: LayoutElement> FloatingSpace<W> {
             ) {
                 win.send_pending_configure();
             }
-
-            win.refresh();
         }
     }
 
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -4875,7 +4875,6 @@ impl<W: LayoutElement> Layout<W> {
             win.set_bounds(output_size(&move_.output).to_i32_round());
 
             win.send_pending_configure();
-            win.refresh();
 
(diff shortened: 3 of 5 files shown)


bdjnk:
**I did a performance analysis, with the help of AI, and here is its summary:**

On every input dispatch, `refresh()` calls `window.refresh()` for **every** window,
which walks that window's full surface tree (`output_update`, plus `Output` enter/leave).
While the pointer moves this runs at the input-event rate — hundreds of times a second —
even though nothing about the windows has changed. This PR sets a `surface_committed`
flag on commit and only runs the per-window refresh when it's set, and folds the refresh
(previously duplicated across three call sites) into one place.

No behavioral change: `output_update` recomputes output overlap and `wl_surface`
enter/leave, which can only change on a commit; niri already drives enter/leave
explicitly on map and output moves.

**Redundant work removed** — Tracy call counts over 10 s of continuous mouse motion:

| `window.refresh()` → `output_update` calls | 1 window | 50 windows |
|---|---|---|
| main            | 1,748 | 42,330 |
| this PR         | **0** | **51** |

After the change the walk fires only on real surface commits (~once per window at map
time), not once per window on every input event.

**CPU self-time removed** at 50 windows (the four surface-tree-walk zones):

| Zone | main | this PR |
|---|---|---|
| `output_update`      | 48.3 ms | 0.02 ms |
| `Window::refresh`    | 38.3 ms | 0.02 ms |
| `Output::enter`      |  8.5 ms | 0.00 ms |
| `Output::leave`      |  8.1 ms | 0.00 ms |
| **total**            | **103.1 ms** | **~0** |

**Scaling** — the removed cost is O(open windows), paid on every dispatch while the
pointer moves, so it's negligible at typical window counts and grows with them:

| Open windows | self-time removed / 10 s | ≈ CPU (one core) |
|---|---|---|
| 1  |   6.3 ms | 0.06% |
| 50 | 103.1 ms | ~1.0% |

Measured on a 12th-gen i5-1240P, single output @ 2256×1504, build with
`--features=profile-with-tracy-ondemand`; 10 s of scripted pointer motion per window
count, cursor parked clear of any hot corner.

mossbraxtin (approved):

cornservant:
I've tried this for over a week without issues

PR #4332, opened 2026-07-19
fix 32bit build: use portable-atomic

peppergrayxyz:
building on 32bit fails with error:
```no `AtomicU64` in `sync::atomic```

add portable-atomic to use native types when available and emulate them when not.

Bugs: #4331

Please also consider adding a 32bit CI to prevent regressions!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ pango = { version = "0.21.5", features = ["v1_44"] }
 pangocairo = "0.21.5"
 pipewire = { version = "0.10.0", optional = true, features = ["v0_3_33"] }
 png = "0.18.1"
+portable-atomic = "1"
 profiling = "1.0.17"
 sd-notify = "0.5.0"
 serde.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2267,6 +2267,7 @@ dependencies = [
  "pipewire",
  "pkg-config",
  "png",
+ "portable-atomic",
  "profiling",
  "proptest",
  "proptest-derive",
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by YaLTeR** [comment]
Oh so this is what it was for. Sorry I removed it recently.

**Comment by YaLTeR** [comment]
> Sorry I removed it recently.

nvm, this code didn't use it in the first place

**Comment by YaLTeR** [comment]
I'll merge this fix (just fix the formatting please), but I'm not sure I want to add a 32-bit CI. Like, even my ancient Eee PC is 64-bit. I'm not sure 32-bit systems are relevant enough to warrant actively testing them upstream.

**Comment by peppergrayxyz** [comment]
Sure, thanks! Formatting should be fixed!

**Comment by YaLTeR** [comment]
Note that you need the nightly rustfmt

**Comment by peppergrayxyz** [comment]
I hope `+nightly` did the trick now!

File: Cargo.toml [hunk 1]
```diff
@@ -82,6 +82,7 @@ pango = { version = "0.21.5", features = ["v1_44"] }
 pangocairo = "0.21.5"
 pipewire = { version = "0.10.0", optional = true, features = ["v0_3_33"] }
 png = "0.18.1"
+portable-atomic = "1"
```
HigherOrderLogic, line 85: Small nit, but shouldn't we specify the whole version to match other parts of the manifest, and preventing unwanted version bump?

```suggestion
portable-atomic = "1.14.0"
```

peppergrayxyz: I did it like this, because this is what the [portable-atomic#usage](https://github.com/taiki-e/portable-atomic#usage) says to do 👀
```toml
[dependencies]
portable-atomic = "1"
```

YaLTeR: It's fine, I wanted to eventually start switching to less constrained versions. It just needs to be done carefully with checking minimal versions, ideally on CI like smithay already does.

YaLTeR:
Thanks

PR #4340, opened 2026-07-21
screencast: report window stream dimensions

pikyusufaslan:
## Problem

The Mutter-compatible `Parameters` property reports a fixed `1x1` size for every window stream. Consumers that use this property to initialize window capture get the wrong geometry, including when the selected window is on an inactive workspace.

## Change

Query the compositor state asynchronously during `RecordWindow` and store the window's actual physical stream size. Window stream position remains `(0, 0)` because the stream is isolated from the global output.

The existing `1x1` value is kept for Niri's dynamic target pseudo-window, which has no fixed geometry until a target is assigned.

## Testing

- `cargo check --offline`
- `cargo test --offline` (195 passed)
- Nested Niri with a private D-Bus session:
  - moved an Alacritty window to an inactive workspace without following focus
  - `RecordWindow` returned a stream with `size: (525, 1038)`
  - the dynamic target still returned `size: (1, 1)`

### DIFF
--- a/src/dbus/mutter_screen_cast.rs
+++ b/src/dbus/mutter_screen_cast.rs
@@ -73,7 +73,10 @@ pub struct Stream {
 enum StreamTarget {
     // FIXME: update on scale changes and whatnot.
     Output(niri_ipc::Output),
-    Window { id: u64 },
+    Window {
+        id: u64,
+        parameters: StreamParameters,
+    },
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


YaLTeR:
What's the reason for using that parameter instead of the actual size of the video frames / pipewire stream? Especially considering window streams will resize at runtime

pikyusufaslan:
You are right. I used `Parameters` because xdg-desktop-portal-gnome forwards it as the initial stream size, but that makes it only a snapshot. Niri can renegotiate the actual PipeWire video size when a window resizes, so storing the initial window size there would become stale and could disagree with the frames.

I changed the consumer to treat an invalid or placeholder portal size as a signal to inspect the negotiated PipeWire format directly, so this is no longer needed upstream. I will close this PR. Thanks for pointing that out.

PR #4344, opened 2026-07-21
built-in screen sharing

pantarune:
serve `org.freedesktop.impl.portal.ScreenCast` directly so users don't need
`xdg-desktop-portal-gnome` for screen sharing. niri already handles screenshots
in-house; this extends that to screencasting.

the portal daemon discovers the backend through the existing `.portal` file
mechanism. only config needed is one line in `portals.conf`.

## changes

- `src/dbus/portal_screen_cast.rs` — portal backend. uses the same
  `ScreenCastToNiri` channel the mutter interface uses. adds `node_tx` to
  `StartCast` so the pipewire node id can be returned directly instead of
  going through a dbus signal roundtrip.
- `src/dbus/mutter_screen_cast.rs` — add `node_tx` field to
  `ScreenCastToNiri::StartCast`.
- `src/screencasting/mod.rs` — carry `node_tx` through `PendingCast`.
- `src/screencasting/pw_utils.rs` — send node id through `node_tx` when set.
- `src/dbus/mod.rs` — register the module behind `xdp-gnome-screencast`.

no new dependencies.

## testing

start niri, add this to `portals.conf`:

```
[preferred]
default=gtk
org.freedesktop.impl.portal.ScreenCast=niri
```

restart the portal daemon, open obs → add screen capture (pipewire) source.
should show the monitor and capture frames. also works with discord and
firefox webrtc screen sharing.

test that the mutter screencast interface still works alongside the new
portal interface (they share the same backend infrastructure). test that
stopping/cancelling a screen share cleans up properly — no orphaned
pipewire nodes or sessions.

### DIFF
--- a/src/dbus/mod.rs
+++ b/src/dbus/mod.rs
@@ -15,7 +15,11 @@ pub mod mutter_service_channel;
 #[cfg(feature = "xdp-gnome-screencast")]
 pub mod mutter_screen_cast;
 #[cfg(feature = "xdp-gnome-screencast")]
+pub mod portal_screen_cast;
+#[cfg(feature = "xdp-gnome-screencast")]
 use mutter_screen_cast::ScreenCast;
+#[cfg(feature = "xdp-gnome-screencast")]
+use portal_screen_cast::PortalScreenCast;
 
 use self::freedesktop_screensaver::ScreenSaver;
 use self::gnome_shell_introspect::Introspect;
@@ -35,6 +39,8 @@ pub struct DBusServers {
     pub conn_introspect: Option<Connection>,
     #[cfg(feature = "xdp-gnome-screencast")]
     pub conn_screen_cast: Option<Connection>,
+    #[cfg(feature = "xdp-gnome-screencast")]
+    pub conn_portal_screen_cast: Option<Connection>,
     pub conn_login1: Option<Connection>,
     pub conn_locale1: Option<Connection>,
     pub conn_a11y_manager: Option<Connection>,
@@ -126,8 +132,12 @@ impl DBusServers {
                         }
                     })
                     .unwrap();
+                let portal_to_niri = to_niri.clone();
                 let screen_cast = ScreenCast::new(backend.ipc_outputs(), to_niri);
                 dbus.conn_screen_cast = try_start(screen_cast);
+
+                let portal = PortalScreenCast::new(backend.ipc_outputs(), portal_to_niri);
+                dbus.conn_portal_screen_cast = try_start(portal);
             }
 
             let (to_niri, from_a11y) = calloop::channel::channel();
--- a/src/dbus/mutter_screen_cast.rs
+++ b/src/dbus/mutter_screen_cast.rs
@@ -4,6 +4,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
 use serde::Deserialize;
+use std::sync::mpsc;
 use zbus::fdo::RequestNameFlags;
 use zbus::object_server::{InterfaceRef, SignalEmitter};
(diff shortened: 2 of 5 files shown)


HigherOrderLogic:
> serve `org.freedesktop.impl.portal.ScreenCast` directly so users don't need
> `xdg-desktop-portal-gnome` for screen sharing

What GUI are you using to select screencast stream with this?

pantarune:
there isnt a gui, it automatically selects the entire screen like in windows or macOs. less friction and majortiy of apps are built for Windows screenshare which has no selector so i think this is better cause the picker is rarely used

HigherOrderLogic:
> there isnt a gui, it automatically selects the entire screen like in windows or macOs

So you just dont respect the users' choice?



> less friction and majortiy of apps are built for Windows screenshare which has no selector

This is simply false. What's on true Windows doesnt always apply to Linux desktop. Many apps dont have built-in selector and relies on portal to select the stream. OBS is one of them.



> cause the picker is rarely used

I find myself use it extensively, and I'm pretty sure I'm not alone here.

pantarune:
how about i make the dynamic cast target the default? it does what the picker gui does but with keybinds

HigherOrderLogic:
> how about i make the dynamic cast target the default

It's a bad default, and there's already [a solution](https://codeberg.org/debugloop/niri-autoselect-portal) without making Niri worse if that's what you want.

Sempyos:
> how about i make the dynamic cast target the default? it does what the picker gui does but with keybinds

The issue is that is not accessible at all for disabled users, along with the fact that many people do not read the wiki entirely

pantarune:
> > how about i make the dynamic cast target the default? it does what the picker gui does but with keybinds
> 
> The issue is that is not accessible at all for disabled users, along with the fact that many people do not read the wiki entirely

i agree, but adding a picker will use zenity which might go past the scope of niri which is keybind based. though i will implement that later on.


> > how about i make the dynamic cast target the default
> 
> It's a bad default, and there's already [a solution](https://codeberg.org/debugloop/niri-autoselect-portal) without making Niri worse if that's what you want.

i made [my own solution](https://github.com/pantarune/niri-screenshare) and got [good support ](https://www.reddit.com/r/niri/comments/1v296jv/i_made_a_simple_niri_screenshare_portal_15_mb_no/), screenshare working right out of the box is suitable for niri which already does the same for screenshots, this is why i think this should be builtin. if you want i can implement a picker as well but the current solution for screenshare is using a 33 MB w/ depends gnome package which depend on nautilius forcing you to have it installed if you want to use screenshare, thats the main reason i made this in the first place.

HigherOrderLogic:
> i made [my own solution](https://github.com/pantarune/niri-screenshare) and got [good support ](https://www.reddit.com/r/niri/comments/1v296jv/i_made_a_simple_niri_screenshare_portal_15_mb_no/)

Having a few dozen upvotes on Reddit doesnt makes your ideas a good default. And if it's already working separately, it's not necessary to integrate into Niri, especially when it's very lackluster and doesnt cover half of the features in the portal spec.





> screenshare working right out of the box is suitable for niri which already does the same for screenshots

If Niri is packaged and installed correctly, the Gnome portal should be installed as an *optional* dependency and works ootb with [the default portal config](https://github.com/niri-wm/niri/blob/main/resources/niri-portals.conf).



> if you want i can implement a picker

No I dont want that. Pre-existing solutions have been working perfectly fine for me.




> a 33 MB w/ depends gnome package which depend on nautilius forcing you to have it installed if you want to use screenshare

Sounds like a packaging issue, the portal shouldn't require Nautilus to function at all. And this also sounds like from Arch, so maybe make an issue there.

YaLTeR:
yeah we're not looking for integrated portal impl at this time

justpav05:
I have a question for the author regarding how the backend is selected and how it works

PR #4348, opened 2026-07-22
ipc: expose atomic cursor state

pikyusufaslan:
## Summary

Add `niri msg cursor-state`, with an equivalent JSON response, to return the current global pointer position and compositor-visible cursor image as one snapshot.

## Why

IPC clients that draw or react to the cursor currently need to combine separate observations. Between those observations the pointer can move or the active cursor can change, so the pair can describe two different compositor moments. A single snapshot is useful for screen-recording overlays, input visualizers, accessibility tools, and other integrations that need to mirror the cursor state faithfully.

## Scope

- The API exposes only the current state; it does not add cursor pixel data, cursor-surface contents, subscriptions, or a new event stream.
- Cursor images are intentionally classified as a named standard shape, an application-owned surface, or hidden.
- The position and image are sampled together from the compositor event loop.

## Validation

- Added IPC JSON round-trip coverage for the request and response.
- Ran `cargo test --offline -p niri-ipc` and `cargo check --offline`.

I would appreciate feedback on the API shape and naming, especially whether this belongs alongside the existing one-shot IPC queries.

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -119,6 +119,8 @@ pub enum Request {
     OverviewState,
     /// Request information about screencasts.
     Casts,
+    /// Request the current global cursor position and image state.
+    CursorState,
 }
 
 /// Reply from niri to client.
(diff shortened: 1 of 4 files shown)


miku4k:
#1924

Issue #1402, opened 2025-04-07
Two 50% windows don't always exactly fit on screen with fractional scale


kosz85:
When switching focus between windows (with option `on-overflow`) there is no way to have both windows on screen in 2 columns. There is always only one centered. If I resize window to have few less pixels, then I can have back both columns on single screen.

I assume it's result of fractional rest of width. I have standard scale 1.25 set on this display. Don't know if it's important but I have 2 monitors and the main one is scaled 1.0 and it behaves without any problems there.

```
➜ niri msg outputs
Output "Dell Inc. DELL*****" (DP-1)
  Current mode: 3840x2160 @ 59.997 Hz (preferred)
  Variable refresh rate: not supported
  Physical size: 700x390 mm
  Logical position: 1920, -1300
  Logical size: 3072x1728
  Scale: 1.25
  Transform: normal
  Available modes:
    3840x2160@59.997 (current, preferred)
...
```

```
layout {
    // Set gaps around windows in logical pixels.
    gaps 3
    center-focused-column "on-overflow"
```

Switching  scale to 1 or 1.2 it works properly, switching back to 1.25 it has overflow and always center window.

### System Information

* niri version: niri 25.02 (v25.02-98-g7cfecf4b)

* Distro: Ubuntu 24.04.2 LTS

* GPU: Radeon RX Vega 6 (Ryzen 4000/5000 Mobile Series)

* CPU: AMD Ryzen 7 PRO 4750U with Radeon Graphics

YaLTeR:
I have noticed this but haven't looked into it yet

nekopsykose:
> Switching scale to 1 or 1.2 it works properly, switching back to 1.25 it has overflow

with a 2880x1800 display it seems to work with 1.25 but overflows at 1.2 scale instead :)  

the gap value also matters; with no gaps or 1 it seems to be fine, anything bigger and it starts overflowing for me. so there's probably some calculation including the gap size that overflows by a pixel and makes it not fit

and of course this applies to any combined proportion scale of the two windows (0.5 + 0.5, 0.8 + 0.2, etc) having the same issue

YaLTeR:
There's rounding (to physical pixels) in several places with these setups. The obvious window size to logical integers rounding is a `floor()` so it's not to blame. I need to look what exactly overflows where, then think what's the best way to deal with it.

all-yall:
[...]

PR #4354, opened 2026-07-24
fix: ignore subpixel scroll when activating columns

MrVampy:
At fractional output scales, fully visible adjacent columns can yield a tiny
nonzero result from `scroll_amount_to_activate`. With
`focus-follows-mouse max-scroll-amount="0%"`, that rounding residue prevents
hover focus.

Treat view changes smaller than one physical pixel as zero, matching
`animate_view_offset_with_config`, and add a regression test using fractional
scale, gaps, and struts.

This addresses the focus-follows-mouse case reported in #3790, which was
marked as a duplicate of #1402.

## Tests

- `cargo fmt --all -- --check`
- `cargo test --all --exclude niri-visual-tests -- --nocapture --skip=::egl`

## Manual validation

- Compared unpatched commit `0777769e` with this PR's `6fde0c6` build in
  otherwise identical nested compositor sessions: scale `2.3`, gaps `8`,
  struts `2`, `max-scroll-amount="0%"`, and two fully visible Kitty columns.
- Unpatched: pointer crossings did not change focus and emitted no
  `WindowFocusChanged` events.
- Patched: pointer crossings changed focus on every crossing and emitted the
  expected alternating `WindowFocusChanged` events.
- Deployed the patched build on a daily-driver NixOS session and repeated the
  test with Thunar and Kitty in adjacent columns on the same scale `2.3`
  output. Focus followed the pointer in both directions at `0%`, with the live
  event stream confirming alternating focus events.

### DIFF
--- a/src/layout/scrolling.rs
+++ b/src/layout/scrolling.rs
@@ -1429,7 +1429,15 @@ impl<W: LayoutElement> ScrollingSpace<W> {
         let new_col_x = self.column_x(column_idx);
         let from_view_offset = target_x - new_col_x;
 
-        (from_view_offset - new_view_offset).abs() / self.working_area.size.w
+        let amount = (from_view_offset - new_view_offset).abs();
+
+        // Column activation treats changes smaller than one physical pixel as no movement.
+        // Mirror that tolerance here so rounding noise does not prevent focus-follows-mouse.
+        if amount < 1. / self.scale {
+            return 0.;
+        }
+
+        amount / self.working_area.size.w
     }
 
     pub fn activate_window(&mut self, window: &W::Id) -> bool {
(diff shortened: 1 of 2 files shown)


src/layout/scrolling.rs
```diff
@@ -1429,7 +1429,15 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
         let from_view_offset = target_x - new_col_x;
 
-        (from_view_offset - new_view_offset).abs() / self.working_area.size.w
+        let amount = (from_view_offset - new_view_offset).abs();
+
+        // Column activation treats changes smaller than one physical pixel as no movement.
+        // Mirror that tolerance here so rounding noise does not prevent focus-follows-mouse.
+        if amount < 1. / self.scale {
+            return 0.;
+        }
```
megabyte6, line 1438: I think that `1.0` and `0.0` should be used instead of `1.` and `0.` for readability. Not major, though.

megabyte6 (approved):
I haven't tested it, but the code looks good to me 👍

PR #4384, opened 2026-08-02
Add option to not expel window from column when fullscreening

cornservant:
This PR addresses issue #426.

It adds an option `layout::maximized-window-placement` with default value `"expel-from-column"`, which is the current behaviour of expelling a tile from its column before fullscreening/maximizing. By setting the value to `"keep-in-column"` the fullscreened/maximized window will not be expelled and stays in the column. Animations are adjusted to make the transition seamless.

## TODO
- [x] Add tests
- [x] ~~Not sure if events stream reports the correct layout~~ Not an issue for non-floating tiles, as the view-/workspace-relative location is not being reported https://github.com/niri-wm/niri/issues/2381
- [x] ~~Animation issue under the presence of firefox, maybe something related to this comment https://github.com/cornservant/niri/blob/46a76eee2fb0dc72570c2616a35d65d85ae6ce3c/src/layout/scrolling.rs#L1338-L1407 ~~ Seems to be a bug (in vanilla niri) https://github.com/niri-wm/niri/issues/4456
- [x] ~~maximizing/unmaximizing fractal is delayed and at times freezes the window, check if this is due to this PR~~ https://github.com/niri-wm/niri/pull/4384#issuecomment-5181770773
- [x] ~~vram usage can jump up for a period of time (for example: put firefox and ghostty in a column, maximize ghostty, vram usage increases massively until you switch focus to firefox); doesn't occur with animations off~~ https://github.com/niri-wm/niri/pull/4384#issuecomment-5181529210

### DIFF
--- a/docs/wiki/Configuration:-Layout.md
+++ b/docs/wiki/Configuration:-Layout.md
@@ -12,6 +12,7 @@ layout {
     empty-workspace-above-first
     default-column-display "tabbed"
     background-color "#003300"
+    maximized-window-placement "expel-from-column"
 
     preset-column-widths {
         proportion 0.33333
@@ -561,3 +562,16 @@ layout {
 ```
 
 You can also set the color per-output [in the output config](./Configuration:-Outputs.md#layout-config-overrides).
+
+### maximized-window-placement
+
+<sup>Since: next release</sup>
+
+Determines whether windows are expelled from their column on maximize/fullscreen.
+Can be `expel-from-column` or `keep-in-column`.
+
+```kdl
+layout {
+    maximized-window-placement "expel-from-column"
+}
+```
--- a/niri-config/src/layout.rs
+++ b/niri-config/src/layout.rs
@@ -24,6 +24,7 @@ pub struct Layout {
     pub gaps: f64,
     pub struts: Struts,
     pub background_color: Color,
+    pub maximized_window_placement: MaximizedWindowPlacement,
 }
 
 impl Default for Layout {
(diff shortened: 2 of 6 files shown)


Sempyos:
Out of curiosity, could you provide a video showing the problem with the animations while Firefox is open..?

cornservant:
Sure

https://github.com/user-attachments/assets/1ddd490f-8af8-44e9-b28c-190d3023f111

cornservant:
Regarding the VRAM usage spiking, this is probably caused by the creation of a lot of offscreen buffers during animation. It happens with niri 26.04 too when maximizing in tabbed mode, so at least it doesn't appear to be a regression introduced in this PR.

cornservant:
Fractal handling maximization poorly happens on the latest Niri release too. Didn't happen with Cosmic and Hyprland, but there's a bug report showing it happens with Mutter: https://gitlab.gnome.org/World/fractal/-/boards?search=resize&show=eyJpaWQiOiIxMjg3IiwiZnVsbF9wYXRoIjoiV29ybGQvZnJhY3RhbCIsImlkIjoxODkwNTN9#

https://github.com/user-attachments/assets/c0d22164-08a0-4123-8b80-3677bb7b2f0b

I'm not going to open a bug report on this (unless someone deems it worthy), considering other gnome apps seem perfectly fine.

cornservant:
I'm not too happy with the option names, but I think this PR is ready to leave the draft stage.

PR #4388, opened 2026-08-03
Adding `--now` flag for `niri msg pick-window` for immediate window picking under the cursor.

tsukiiru:
originally from pr https://github.com/niri-wm/niri/pull/4189.  
it seems like the pr was stale (inactive for a few months) and the original author (@KapDecy) wasn't applying the suggestions in the comments.  

### Changes
- added `now` as a flag in `cli.rs`
- added `now` boolean as a field in the `PickWindow` Request and Msg
- added a conditional check for the `now` boolean in `Request::PickWindow` enum `event_loop` closure using `window_under_cursor()` function

### Command
`niri msg pick-window --now` - immediate pick  
`niri msg pick-window`       - pick by choosing (as usual)

## Tests
all tests from `cargo test --all` with the recommended env went through  
also tested by running niri on a tty

### DIFF
--- a/niri-ipc/src/lib.rs
+++ b/niri-ipc/src/lib.rs
@@ -82,7 +82,10 @@ pub enum Request {
     /// Request information about the focused window.
     FocusedWindow,
     /// Request picking a window and get its information.
-    PickWindow,
+    PickWindow {
+        /// Immediately picking the window under the cursor.
+        now: bool,
+    },
     /// Request picking a color from the screen.
     PickColor,
     /// Perform an action.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -76,7 +76,11 @@ pub enum Msg {
     /// Print information about the focused window.
     FocusedWindow,
     /// Pick a window with the mouse and print information about it.
-    PickWindow,
+    PickWindow {
+        /// Immediately picking the window under the cursor.
+        #[arg(long, short)]
+        now: bool,
+    },
     /// Pick a color from the screen with the mouse.
     PickColor,
     /// Perform an action.
--- a/src/ipc/client.rs
+++ b/src/ipc/client.rs
@@ -34,7 +34,7 @@ pub fn handle_msg(mut msg: Msg, json: bool) -> anyhow::Result<()> {
         Msg::Outputs => Request::Outputs,
         Msg::FocusedWindow => Request::FocusedWindow,
         Msg::FocusedOutput => Request::FocusedOutput,
-        Msg::PickWindow => Request::PickWindow,
+        Msg::PickWindow { now } => Request::PickWindow { now: *now },
         Msg::PickColor => Request::PickColor,
         Msg::Action { action } => Request::Action(action.clone()),
         Msg::Output { output, action } => Request::Output {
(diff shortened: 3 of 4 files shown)


megabyte6 (approved):
Looks good to me :+1: 
I did a quick test, and everything seems to work as expected.

cornservant:
Without `focus-follows-mouse` enabled the --now flag stops being self-explanatory. Furthermore there is no `picking` involved when using --now, somewhat contradicting the command name. 

I'd rather add a field `is_under_cursor` to the items in `niri msg -j windows` that indicates whether a window is the one under the cursor. Just like you can already query the focused window via `niri msg -j windows | jq '.[] | select(.is_focused)'` you'd be able to query for the window under the cursor via `niri msg -j windows | jq '.[] | select(.is_under_cursor)'`

tsukiiru:
> Without `focus-follows-mouse` enabled the --now flag stops being self-explanatory. Furthermore there is no `picking` involved when using --now, somewhat contradicting the command name.

right. now that you pointed it out, the flag is quite out of place.

> I'd rather add a field `is_under_cursor` to the items in `niri msg -j windows` that indicates whether a window is the one under the cursor. Just like you can already query the focused window via `niri msg -j windows | jq '.[] | select(.is_focused)'` you'd be able to query for the window under the cursor via `niri msg -j windows | jq '.[] | select(.is_under_cursor)'`

isn't there already `msg focused-window` to get info of the focused window? the `is_focused` field exists because it can be (and is being) used in other places. however the same can't really be said for `is_under_cursor`.

cornservant:
> isn't there already `msg focused-window` to get info of the focused window? the `is_focused` field exists because it can be (and is being) used in other places. however the same can't really be said for `is_under_cursor`.

Ye, I agree. Maybe it should just be a straight forward `msg window-under-cursor`?

tsukiiru:
> Ye, I agree. Maybe it should just be a straight forward `msg window-under-cursor`?

i think an extra msg would clutter the command list  
image  

and as you mentioned above, if `focus-follows-mouse` was set, picking under cursor would pretty much does the same thing as `msg focused-window`.  
i'm thinking about putting `--focused` and `--under-cursor` (basically `--now`) flags for `pick-window` and remove `focused-window` since they are all picking a window?
although i'm not a really huge fan of changing an existing cli command due to possible breakages...

Apollo314:
> Without `focus-follows-mouse` enabled the --now flag stops being self-explanatory. Furthermore there is no `picking` involved when using --now, somewhat contradicting the command name.
> 
> I'd rather add a field `is_under_cursor` to the items in `niri msg -j windows` that indicates whether a window is the one under the cursor. Just like you can already query the focused window via `niri msg -j windows | jq '.[] | select(.is_focused)'` you'd be able to query for the window under the cursor via `niri msg -j windows | jq '.[] | select(.is_under_cursor)'`

`msg pick-window` gives you a special picker that allows you to pick with your mouse when you click. `msg pick-window --now` assumes you already clicked and picked it now. what is so hard to grasp?

you can want is_under_cursor flag, that's okay. nobody's stopping you from opening another pr. that feature would not be mutually exclusive to this one though. after all you can get focused window that way or with `niri msg focused-window`.

PR #4392, opened 2026-08-04
feat: force render (v2)

MithicSpirit:
Renders a window when it is being screencast or as configured by the force-render window rule.  Maximum FPS of windows being rendered this way is controlled by the force-render-fps window rule (unlimited by default, or when explicitly set to 0).

This is inspired by #2609, but is not a direct update of it. In particular, this avoids storing the force render state globally, and instead puts it in the data map, which should prevent the memory leak (https://github.com/niri-wm/niri/pull/2609/changes#r3588566980).

Closes #2609.

TODO: bikeshedding, docs

### DIFF
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -571,7 +571,7 @@ struct SurfaceFrameThrottlingState {
 
 struct SurfaceForceRenderState {
     /// Whether a forced render is scheduled in the event loop.
-    scheduled: RefCell<bool>,
+    scheduled: Cell<bool>,
 }
 
 pub enum CenterCoords {
@@ -668,7 +668,35 @@ impl Default for SurfaceFrameThrottlingState {
 impl Default for SurfaceForceRenderState {
     fn default() -> Self {
         Self {
-            scheduled: RefCell::new(false),
+            scheduled: Cell::new(false),
+        }
+    }
+}
+
+pub fn setup_force_render(event_loop: &mut LoopHandle<'static, State>, mapped: &Mapped) -> bool {
+    if mapped.force_render().is_none() {
+        return false;
+    }
+
+    let already_scheduled = with_states(mapped.toplevel().wl_surface(), |states| {
+        let force_render_state = states
+            .data_map
+            .get_or_insert(SurfaceForceRenderState::default);
+        force_render_state.scheduled.replace(true)
+    });
+    if already_scheduled {
+        return false;
+    }
+
+    let mapped_id = mapped.id();
+    let res = event_loop.insert_source(Timer::immediate(), move |_, _, state| {
+        force_render_callback(state, mapped_id)
+    });
+    match res {
+        Ok(_) => true,
+        Err(e) => {
+            warn!("error scheduling forced render: {e}");
+            false
         }
     }
 }
@@ -696,9 +724,7 @@ fn force_render_callback(state: &mut State, mapped_id: MappedId) -> TimeoutActio
             let force_render_state = states
                 .data_map
                 .get_or_insert(SurfaceForceRenderState::default);
-            let mut scheduled = force_render_state.scheduled.borrow_mut();
-
-            *scheduled = false;
+            force_render_state.scheduled.set(false);
             return None;
         };
 
@@ -5133,12 +5159,8 @@ impl Niri {
 
         let frame_callback_time = get_monotonic_time();
 
-        // Borrow checker decided to be annoying here.
-        let mut to_force_render = Vec::new();
         for mapped in self.layout.windows_for_output_mut(output) {
-            if mapped.force_render().is_some() {
-                to_force_render.push(mapped.id());
-            }
+            setup_force_render(&mut self.event_loop, mapped);
             mapped.send_frame(
                 output,
                 frame_callback_time,
@@ -5146,9 +5168,6 @@ impl Niri {
                 should_send,
             );
         }
-        to_force_render
-            .drain(..)
-            .for_each(|x| self.setup_force_render(x));
 
         for surface in layer_map_for_output(output).layers() {
             surface.send_frame(
@@ -5257,36 +5276,6 @@ impl Niri {
                 |_, _| None,
             );
         }
-    }
-
-    fn setup_force_render(&mut self, mapped_id: MappedId) {
-        let Some(mapped) = self
-            .layout
-            .windows()
-            .find(|(_, m)| m.id() == mapped_id)
-            .map(|(_, m)| m)
-        else {
-            return;
-        };
-        with_states(mapped.toplevel().wl_surface(), |states| {
-            let force_render_state = states
-                .data_map
-                .get_or_insert(SurfaceForceRenderState::default);
-            let mut scheduled = force_render_state.scheduled.borrow_mut();
-            if *scheduled {
-                return;
-            }
-            *scheduled = true;
-
-            let res = self
-                .event_loop
-                .insert_source(Timer::immediate(), move |_, _, state| {
-                    force_render_callback(state, mapped_id)
-                });
-            if let Err(e) = res {
-                warn!("error scheduling forced render: {e}")
-            };
-        })
     }
 
     pub fn take_presentation_feedbacks(
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -178,9 +178,9 @@ use crate::utils::vblank_throttle::VBlankThrottle;
 use crate::utils::watcher::Watcher;
 use crate::utils::xwayland::satellite::Satellite;
 use crate::utils::{
-    center, center_f64, expand_home, get_monotonic_time, ipc_transform_to_smithay, is_mapped,
-    logical_output, make_screenshot_path, output_matches_name, output_size, panel_orientation,
-    send_scale_transform, write_png_rgba8, xwayland,
+    center, center_f64, expand_home, frequency_to_period, get_monotonic_time,
+    ipc_transform_to_smithay, is_mapped, logical_output, make_screenshot_path, output_matches_name,
+    output_size, panel_orientation, send_scale_transform, write_png_rgba8, xwayland,
 };
 use crate::window::mapped::MappedId;
 use crate::window::{InitialConfigureState, Mapped, ResolvedWindowRules, Unmapped, WindowRef};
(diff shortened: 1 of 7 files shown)


YaLTeR:
I haven't looked at this but my intuitive feeling is that force render shouldn't have an fps setting, rather it should render at the same fps (and (possibly estimated) redraw cycle) as the output that the window is on, or in case there are no outputs, as the output the window was last on (or if that's too annoying then default to some value like 30).

phisch:
> I haven't looked at this but my intuitive feeling is that force render shouldn't have an fps setting, rather it should render at the same fps (and (possibly estimated) redraw cycle) as the output that the window is on, or in case there are no outputs, as the output the window was last on (or if that's too annoying then default to some value like 30).

You could make a window rule matching windows that are being screencast, and force them to render at a maximum of 60fps. I think this would only be useful when this max_fps counter applies while the window is fully off-screen, because you would want it to render at the output refresh rate while visible to yourself.

There are also a few games that need a minimum of frames to do their game logic, or else they "fall behind", and catch up once the window is focused again. So setting them to the minimum required FPS to not fall behind would be a nice performance optimization.

I wouldn't just assume a value and use that as the default for all. 

My only critique would be `force_render` and `force_render_fps` sound like they apply always, but my understanding was that this is only applied if the window is fully off-screen. So something like `force_offscreen_render` and `force_offscreen_render_fps` or something along those lines to make it clearer what the config does.

YaLTeR:
Ah, when writing that comment I was only thinking about screencasts and not about just force render on its own. I think for that some max FPS cap makes sense indeed, probably on the lower side, and perhaps overridable with a window rule. Maybe the current 1 FPS frame callback throttle could just be treated as the "default" max FPS value then?

MithicSpirit:
I think it would make sense to decouple screencasting force-render FPS from window-rule force-render FPS, and make the former unconfigurable. Then we would be free to screencast at last-output FPS while establishing a more strict default FPS limit for the window-rule.

I suspect almost all users will not want the screencasting FPS to suddenly drop when they switch to a different workspace, and the few who do want that can patch it in themselves (should be pretty easy).

From my experience, 1 FPS is too low for games, and I think something like 15 FPS would be better (it's what I use in my config).

Sabin-DC:
I'm testing this and so far so good (i was using your older PR).  I just don't know if there is a systematic way to test for memory leaks if that was the issue I recall from the prev PR? 

Is there a monitoring tool I could use to run something overtime and report back?

MithicSpirit:
Did some further testing with this, and it broke on multi-monitor setups (the window would get frames slower/inconsistently when on an inactive monitor). I tried a lot of things to avoid going back to an event loop, but in the end I couldn't find it (perhaps someone who is better versed in the smithay-niri event loop can point it out). Now it should be working, but does use the event loop (one event per force-rendered window).

I'm not sure why, but this implementation does not have the same frame rate consistency issues on my system as the previous one. This was reimplemented from scratch, so I guess I just didn't inherit whatever caused that bug.

As part of this most recent rewrite, the force-render frame rate is now also clamped to the refresh rate of the window's most recent output. In my opinion, the code is also easier to follow than my previous iteration.

MithicSpirit:
> I'm testing this and so far so good (i was using your older PR). I just don't know if there is a systematic way to test for memory leaks if that was the issue I recall from the prev PR?
> 
> Is there a monitoring tool I could use to run something overtime and report back?

Not that I know of. This isn't a true memory leak, in that it was still accessible via the map, and hence would get cleaned up at program exit (so things like valgrind couldn't catch it).

I think the best option would be to just look at memory usage, although the memory footprint per window for force-render is quite small. If you really want to try, you can compare the memory usage for opening-and-closing thousands to millions of windows when force-render is on or off.

YaLTeR:
> I think it would make sense to decouple screencasting force-render FPS from window-rule force-render FPS, and make the former unconfigurable. Then we would be free to screencast at last-output FPS while establishing a more strict default FPS limit for the window-rule.

Yes, of course. Screencast should continue without changes when going off screen.

MithicSpirit:
I'm thinking having two window rules for this is a bit excessive. Technically it's a bit less flexible, but maybe there should just be one window rule `force-render` with an optional parameter for choosing the frame rate (`force-render fps=15`).

src/utils/mod.rs
```diff
@@ -164,6 +164,16 @@ pub fn get_monotonic_time() -> Duration {
     Duration::new(ts.tv_sec as u64, ts.tv_nsec as u32)
 }
 
+/// Converts a frequency in millihertz to the corresponding period, clamping to
+/// `Duration::ZERO` for nonpositive rates and `Duration::MAX` on overflow.
+pub fn frequency_to_period(rate: u32) -> Duration {
+    if rate > 0 {
+        Duration::try_from_secs_f64(1000.0 / rate as f64).unwrap_or(Duration::MAX)
```
MithicSpirit, line 171: In hindsight, I guess overflow is impossible here (bounded above by 1000 since the divisor is ≥1).

urayde:
> I'm thinking having two window rules for this is a bit excessive. Technically it's a bit less flexible, but maybe there should just be one window rule `force-render` with an optional parameter for choosing the frame rate (`force-render fps=15`).

Yeah, I think it makes more sense, unless there's any reason to add another option. In that case I would opt in for: 
```kdl
force-render {
  fps 15
  whatever true
}
```

Though `force-render` is quite misleading in my opinion, something like `offscreen-render` would be better.

YaLTeR:
Why do you need two window rules or one with a separate parameter? We already send frame callbacks once a second to all windows. It's already "always enabled", just need one rule to control this frequency

src/niri.rs
```diff
@@ -5066,14 +5133,22 @@ impl Niri {
...
             mapped.send_frame(
                 output,
                 frame_callback_time,
                 FRAME_CALLBACK_THROTTLE,
                 should_send,
             );
         }
+        to_force_render
+            .drain(..)
+            .for_each(|x| self.setup_force_render(x));
```
MithicSpirit: There ought to be a better time to set this up than while the window is being rendered. I suppose for screencasting it could be done when the screencasting is initializing (although I don't think screencasting has a reference to the event loop, so I'll need to handle that somehow), but what about for the window rule?

src/niri.rs
```diff
@@ -5066,14 +5133,22 @@ impl Niri {
 
         let frame_callback_time = get_monotonic_time();
 
+        // Borrow checker decided to be annoying here.
+        let mut to_force_render = Vec::new();
```
cornservant: wouldn't it be better to iterate twice over `windows_for_output_mut` rather than allocate on each redraw?

e.g.
```patch
 src/niri.rs | 30 +++++++++---------------------
 1 file changed, 9 insertions(+), 21 deletions(-)

index 535f42d8..ec291c73 100644
--- a/src/niri.rs
+++ b/src/niri.rs
@@ -5133,12 +5133,12 @@ impl Niri {
 
         let frame_callback_time = get_monotonic_time();
 
-        // Borrow checker decided to be annoying here.
-        let mut to_force_render = Vec::new();
-        for mapped in self.layout.windows_for_output_mut(output) {
+        for mapped in self.layout.windows_for_output(output) {
             if mapped.force_render().is_some() {
-                to_force_render.push(mapped.id());
+                Self::setup_force_render(&mut self.event_loop, &mapped);
             }
+        }
+        for mapped in self.layout.windows_for_output_mut(output) {
             mapped.send_frame(
                 output,
                 frame_callback_time,
@@ -5146,9 +5146,6 @@ impl Niri {
                 should_send,
             );
         }
-        to_force_render
-            .drain(..)
-            .for_each(|x| self.setup_force_render(x));
 
         for surface in layer_map_for_output(output).layers() {
             surface.send_frame(
@@ -5259,16 +5256,9 @@ impl Niri {
         }
     }
 
-    fn setup_force_render(&mut self, mapped_id: MappedId) {
-        let Some(mapped) = self
-            .layout
-            .windows()
-            .find(|(_, m)| m.id() == mapped_id)
-            .map(|(_, m)| m)
-        else {
-            return;
-        };
+    fn setup_force_render(event_loop: &mut LoopHandle<'static, State>, mapped: &Mapped) {
         with_states(mapped.toplevel().wl_surface(), |states| {
+            let mapped_id = mapped.id();
             let force_render_state = states
                 .data_map
                 .get_or_insert(SurfaceForceRenderState::default);
@@ -5278,11 +5268,9 @@ impl Niri {
             }
             *scheduled = true;
 
-            let res = self
-                .event_loop
-                .insert_source(Timer::immediate(), move |_, _, state| {
-                    force_render_callback(state, mapped_id)
-                });
+            let res = event_loop.insert_source(Timer::immediate(), move |_, _, state| {
+                force_render_callback(state, mapped_id)
+            });
             if let Err(e) = res {
                 warn!("error scheduling forced render: {e}")
             };
```

MithicSpirit: Yeah, that's true. I'm kinda hoping there's a better time to do this stuff though, such that I don't even need to repeatedly check it.

MithicSpirit: Actually, I think your changing the signature of `setup_force_render` to not even take a `self` makes it possible to do everything in the same loop.

thuiop:
I tried this; it indeed works for running some games in the background which would run at slower speeds otherwise.

Ecamika:
how is the progress about this pr? have anything new?

MithicSpirit:
Still works. From discussion on the discord, I think this is probably not good enough to be merged (too hacky) and will probably require some refactoring of niri internals (likely in a v3). You can pull it into your own tree for the time being though.

Brisingr05:
FWIW this also fixes an issue where Steam's Settings window would become very laggy even when I'm actively using it (unless it's floating).

edralzar:
I do livestreaming and I have created a dynamic overlay element using Godot that I can capture in OBS as a specific Pipewire Window capture. The thing is, I don't want it using up space on my own screen all the time but unfortunately if I put it in another workspace it runs at 1 FPS :disappointed:
This is another use case for this PR :+1:

Yurihaia:
For the record, there is a bug with this that makes the force render callback get cleared whenever a window is moved in the overlay via mouse. The fix is fairly trivial (just check the moving window), but does require an extra method `moving_window_mut` on `Layout`.

MithicSpirit:
Does this bug have any concrete effects? Or is the problem just that the event gets skipped while the window is being dragged?

Yurihaia:
It means that the window permanently stops being forcibly rendered after being dragged. Specifically what happens is the render callback gets cleared without the "active" state flag being unset.

https://forge.fiftysix.dev/rime/niri/commit/37e6b8d63d0dace57937912bd67bec8e48cbb29a an example of the fix on my personal fork

MithicSpirit:
Oh right, it can't unset the flag because it didn't even find the window. Yeah, I'll fix this.

PR #4395, opened 2026-08-04
Fix non-xray blur capture on render-only framebuffers

### LINKED ISSUE #4378: Blur is not applied on iGPU-connected output when render-drm-device is set to the dGPU (niri: 26.04)
On a hybrid Intel + NVIDIA laptop, config-defined blur (window rules) is lost on the output connected to the secondary GPU (integrated), but only when the NVIDIA (dedicated) card is used as the render device.

The behaviour is asymmetric:
- With render-drm-device unset (default), niri uses renderD128 (Intel) as the render node, and blur works on both eDP-1 and HDMI-A-1.
- With render-drm-device set to the NVIDIA render node, niri uses renderD129 (NVIDIA), and blur works on HDMI-A-1 but is missing on eDP-1.

**Client-requested blur is unaffected: a kitty window continues to render its own blur on eDP-1 in both configurations. Only blur that niri applies itself (via window rules) is lost.

**--Steps to reproduce**
1. Hybrid laptop, internal panel wired to the iGPU, HDMI wired to the dGPU.
2. Add a window rule with blur enabled, and confirm it works on both outputs with no render-drm-device set.
3. Set the dGPU as the render device:
4. Restart niri.
5. Open a window matched by the blur rule on HDMI-A-1 — blur renders correctly.
6. Drag that same window to eDP-1 — blur disappears as it crosses. Dragging it back restores blur.

**--System information**
niri: 26.04 (from repo)
Distro: CachyOS (Arch-based)
Kernel: 7.1.5-1-cachyos
Machine: Acer Nitro AN515-57
CPU: Intel Core i7 (Tiger Lake-H)
GPU 1: Intel TigerLake-H GT1 [UHD Graphics] — 0000:00:02.0 → card1, renderD128
GPU 2: NVIDIA GA107M [GeForce RTX 3050 Mobile] — 0000:01:00.0 → card0, renderD129
NVIDIA driver: 610.43.03-1 (nvidia-utils)
Mesa: 3:26.1.6-1
nvidia_drm.modeset: Y

**--Relevant config**
debug {
    render-drm-device "/dev/dri/by-path/pci-0000:01:00.0-render"
}

environment {
    NVD_BACKEND "direct"
    LIBVA_DRIVER_NAME "nvidia"
    __GLX_VENDOR_LIBRARY_NAME "nvidia"
}

Acters:
Fixes #4378.

The multi-GPU renderer uses an offscreen framebuffer whose read buffer may be `GL_NONE`. Non-xray blur calls `glBlitFramebuffer()` to capture the contents behind a window, but it previously changed only the draw framebuffer. On this target, the capture was all zeroes.

If the read buffer is `GL_NONE`, select `GL_BACK` for the default framebuffer or `GL_COLOR_ATTACHMENT0` for other framebuffers while blitting, then restore the previous state.

I reproduced this on an Intel + NVIDIA system using DRM CRTC CRC, so screencopy would not trigger a separate render. I tested both Smithay's CPU-copy path and a separate Vulkan copy path. In both cases, non-xray blur updated the Intel-connected output immediately after this change.

`cargo test --workspace` and `cargo check -p niri` pass.

### DIFF
--- a/src/render_helpers/framebuffer_effect.rs
+++ b/src/render_helpers/framebuffer_effect.rs
@@ -254,7 +254,20 @@ impl RenderElement<GlesRenderer> for FramebufferEffectElement {
                 while gl.GetError() != ffi::NO_ERROR {}
 
                 let mut current_fbo = 0i32;
+                let mut current_read_buffer = 0i32;
                 gl.GetIntegerv(ffi::DRAW_FRAMEBUFFER_BINDING, &mut current_fbo as *mut _);
+                gl.GetIntegerv(ffi::READ_BUFFER, &mut current_read_buffer as *mut _);
+
+                // Render targets may be configured as render-only.
+                let temporary_read_buffer =
+                    (current_read_buffer == ffi::NONE as i32).then_some(if current_fbo == 0 {
+                        ffi::BACK
+                    } else {
+                        ffi::COLOR_ATTACHMENT0
+                    });
+                if let Some(read_buffer) = temporary_read_buffer {
+                    gl.ReadBuffer(read_buffer);
+                }
 
                 // BlitFramebuffer is affected by the scissor test, we don't want that.
                 gl.Disable(ffi::SCISSOR_TEST);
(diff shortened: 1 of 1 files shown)


Acters:
while my PR would have fixed it, Smithay now exposes a DMABUF texture target. 

My bisect lead me to the following Smithay commit:  
fa509c7b renderer/gles: Bind Dmabufs as textures instead of renderbuffers

Of course this required Niri to move the targeted Smithay version past this commit.

It had replaced the old renderbuffer-backed path with a texture-backed DMABUF targets.  
Merging the extra GL state handling in my PR for a path that the current Niri will no longer use in smithay is probably unnecessary. I think I can close my PR now because the issue is fixed upstream now by simply just doing things differently. 

It's a weird issue. I wouldn't say it was on Smithay to fix but because they changed the way it worked, it just naturally fixed issues that had occured from exposing a renderbuffer, which Niri also didn't handle correctly.

I should have noticed this commit fixed it on my first pass. My bad

PR #4407, opened 2026-08-08
config: default quit keybinds lock when session locks

### LINKED ISSUE #4406: [Config] Default niri configuration enables quitting the session while it is locked
With the default niri configuration, it is possible to lock the session (using swaylock for example) and quit the session with any of the default "quit" keybinds without need for unlocking first

This is possible because the modifier "allow-when-locked" is unused (and unavailable) for any of the default "quit" commands

With physical access to the machine, this can possibly be exploited to gain access to an user's shell session or just to destroy their session which is also undesirable

<details><summary>Config</summary>

```kdl
...
// Suggested binds for running programs: terminal, app launcher, screen locker.
    Mod+T hotkey-overlay-title="Open a Terminal: alacritty" { spawn "alacritty"; }
    Mod+D hotkey-overlay-title="Run an Application: fuzzel" { spawn "fuzzel"; }
    Super+Alt+L hotkey-overlay-title="Lock the Screen: swaylock" { spawn "swaylock"; }
...
    // The quit action will show a confirmation dialog to avoid accidental exits.
    Mod+Shift+E { quit; }
    Ctrl+Alt+Delete { quit; }
...
```

</details> 

### System Information

* niri version: niri 26.04 (8ed0da4)

* Distro: Artix Linux

* GPU: Intel Atom Processor D2xxx/N2xxx Integrated Graphics

* CPU: Intel Atom N2600 (4) @ 1.60 GHz

Coki91:
change the default configuration's "quit" keybinds to spawn-sh niri ipc calls for the quit action; and disable such keybinds when the session is locked

closes #4406

### DIFF
--- a/resources/default-config.kdl
+++ b/resources/default-config.kdl
@@ -623,9 +623,10 @@ binds {
     // which ensures niri always processes them, even when an inhibitor is active.
     Mod+Escape allow-inhibiting=false { toggle-keyboard-shortcuts-inhibit; }
 
-    // The quit action will show a confirmation dialog to avoid accidental exits.
-    Mod+Shift+E { quit; }
-    Ctrl+Alt+Delete { quit; }
+    // The quit action will be unavailable when the session is locked,
+    // and show a confirmation dialog to avoid accidental exits.
+    Mod+Shift+E allow-when-locked=false { spawn-sh "niri msg action quit"; }
+    Ctrl+Alt+Delete allow-when-locked=false { spawn-sh "niri msg action quit"; }
 
     // Powers off the monitors. To turn them back on, do any input like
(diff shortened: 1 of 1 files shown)


HigherOrderLogic:
Iirc Yalter said it's set that way to let user recover from crashed lockscreen, but Im not sure.

Coki91:
I have not faced a screenlocker crashing thus far so I'm unable to 100% understand the logic, but I would think it's sensible to let the user handle that from a different session (for example ttys2 to 6 are usually enabled on most distros out of the box) than to leave them exposed by default

YaLTeR:
This is intentional: https://niri-wm.github.io/niri/Security-Model.html#lock-screen

PR #4411, opened 2026-08-09
Keep pointer constraints in sync with keyboard focus

### LINKED ISSUE #3729: Cursor locked state lost if the new 'cursor position hint' position is on a floating window
For some reason proton (with waylanddrv enabled via `PROTON_ENABLE_WAYLAND=1`) will send a cursor position hint on mouse button clicks (no matter left or right or middle) when you move the camera. If the hint position is within a floating window, the cursor locked state will be lost. That is, one will lost the camera control until moving the cursor within the game window again. It's kinda annoying in the case, say, one is watching a video or something as a floating window and meanwhile playing a game as a windowed window.

Demonstration video:

https://github.com/user-attachments/assets/215e5867-acf1-43d8-864d-4072cc7d9619

I patched it (in `src/handlers/mod.rs`, method `cursor_position_hint`) to print the log rightside to demonstrate the hint events from proton.

<details><summary>Config</summary>

```kdl
// This config is in the KDL format: https://kdl.dev
// "/-" comments out the following node.
// Check the wiki for a full description of the configuration:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Introduction
config-notification {
    disable-failed
}
// Outputs from existing configuration
output "eDP-1" {
    off
}
output "DP-2" {
    // Uncomment this line to disable this output.
    // off

    // Resolution and, optionally, refresh rate of the output.
    // The format is "<width>x<height>" or "<width>x<height>@<refresh rate>".
    // If the refresh rate is omitted, niri will pick the highest refresh rate
    // for the resolution.
    // If the mode is omitted altogether or is invalid, niri will pick one automatically.
    // Run `niri msg outputs` while inside a niri instance to list all outputs and their modes.
    mode "2560x1440@165.001"

    // You can use integer or fractional scale, for example use 1.5 for 150% scale.
    scale 1.33333

    // Transform allows to rotate the output counter-clockwise, valid values are:
    // normal, 90, 180, 270, flipped, flipped-90, flipped-180 and flipped-270.
    transform "normal"

    // Position of the output in the global coordinate space.
    // This affects directional monitor actions like "focus-monitor-left", and cursor movement.
    // The cursor can only move between directly adjacent outputs.
    // Output scale and rotation has to be taken into account for positioning:
    // outputs are sized in logical, or scaled, pixels.
    // For example, a 3840×2160 output with scale 2.0 will have a logical size of 1920×1080,
    // so to put another output directly adjacent to it on the right, set its x to 1920.
    // If the position is unset or results in an overlap, the output is instead placed
    // automatically.
    position x=0 y=0
}

spawn-at-startup "wl-clip-persist" "--clipboard" "regular" "--all-mime-type-regex" r#"(?i)^(?!image/x-inkscape-svg).+"#
spawn-at-startup "wl-paste" "--type" "text" "--watch" "cliphist" "store"
spawn-at-startup "wl-paste" "--type" "image" "--watch" "cliphist" "store"
spawn-at-startup "fcitx5" "-d"
spawn-at-startup "bash" "-c" r##"echo $NIRI_SOCKET > /tmp/niri-socket.txt"##

gestures {
    // Enable this to open niri overview on mouse moved to the top-left corner.
    hot-corners {
        // off
    }
}

// Input device configuration.
// Find the full list of options on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Input
input {
    keyboard {
        xkb {
            // You can set rules, model, layout, variant and options.
            // For more information, see xkeyboard-config(7).

            // For example:
            // layout "us,ru"

            // If this section is empty, niri will fetch xkb settings
            // from org.freedesktop.locale1. You can control these using
            // localectl set-x11-keymap.
        }

        // Enable numlock on startup, omitting this setting disables it.
        numlock
    }

    // Next sections include libinput settings.
    // Omitting settings disables them, or leaves them at their default values.
    // All commented-out settings here are examples, not defaults.
    touchpad {
        // off
        tap
        // dwt
        // dwtp
        // drag false
        // drag-lock
        natural-scroll
        // accel-speed 0.2
        // accel-profile "flat"
        // scroll-method "two-finger"
        // disabled-on-external-mouse
    }

    mouse {
        // off
        // natural-scroll
        // accel-speed 0.2
        // accel-profile "flat"
        // scroll-method "no-scroll"
        accel-profile "adaptive"
        accel-speed 0.2
    }

    trackpoint {
        // off
        // natural-scroll
        // accel-speed 0.2
        // accel-profile "flat"
        // scroll-method "on-button-down"
        // scroll-button 273
        // scroll-button-lock
        // middle-emulation
    }

    // Uncomment this to make the mouse warp to the center of newly focused windows.
    // warp-mouse-to-focus

    // Focus windows and outputs automatically when moving the mouse into them.
    // Setting max-scroll-amount="0%" makes it work only on windows already fully on screen.
    // focus-follows-mouse max-scroll-amount="0%"
}
// You can configure outputs by their name, which you can find
// by running `niri msg outputs` while inside a niri instance.
// The built-in laptop monitor is usually called "eDP-1".
// Find more information on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Outputs
// Remember to uncomment the node by removing "/-"!

// Settings that influence how windows are positioned and sized.
// Find more information on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Layout
layout {
    // Set gaps around windows in logical pixels.
    background-color "transparent"
    // - "never", default behavior, focusing an off-screen column will keep at the left
    //   or right edge of the screen.
    // - "always", the focused column will always be centered.
    // - "on-overflow", focusing a column will center it if it doesn't fit
    //   together with the previously focused column.
    center-focused-column "never"
    // You can customize the widths that "switch-preset-column-width" (Mod+R) toggles between.
    preset-column-widths {
        // Proportion sets the width as a fraction of the output width, taking gaps into account.
        // For example, you can perfectly fit four windows sized "proportion 0.25" on an output.
        // The default preset widths are 1/3, 1/2 and 2/3 of the output.
        proportion 0.33333
        proportion 0.5
        proportion 0.66667
        // Fixed sets the width in logical pixels exactly.
        // fixed 1920
    }
    // You can also customize the heights that "switch-preset-window-height" (Mod+Shift+R) toggles between.
    // preset-window-heights { }
    // You can change the default width of the new windows.
    // default-column-width {}
    // By default focus ring and border are rendered as a solid background rectangle
    // behind windows. That is, they will show up through semitransparent windows.
    // This is because windows using client-side decorations can have an arbitrary shape.
    //
    // If you don't like that, you should uncomment `prefer-no-csd` below.
    // Niri will draw focus ring and border *around* windows that agree to omit their
    // client-side decorations.
    //
    // Alternatively, you can override it with a window rule called
    // `draw-border-with-background`.
    border {
        off
        width 4
        active-color   "#707070"      // Neutral gray
        inactive-color "#d0d0d0"      // Light gray
        urgent-color   "#cc4444"      // Softer red
    }
    shadow {
        softness 30
        spread 5
        offset x=0 y=5
        color "#0007"
    }
    struts {
    }
}
layer-rule {
    match namespace="^quickshell$"
    place-within-backdrop true
}
overview {
    workspace-shadow {
        off
    }
}
// Add lines like this to spawn processes at startup.
// Note that running niri as a session supports xdg-desktop-autostart,
// which may be more convenient to use.
// See the binds section below for more spawn examples.
environment {
    // IME
    XMODIFIERS "@im=fcitx"

    // XDG
    XDG_CURRENT_DESKTOP "niri"
    XDG_CONFIG_HOME "/home/bczhc/.config"
    XDG_DATA_DIRS "/home/bczhc/.local/xdg-override:/home/bczhc/.local/share:/usr/local:/usr/share:/usr/local/share"
    XDG_STATE_HOME "/home/bczhc/.local/state"
    // For screencasting
    XDG_CURRENT_DESKTOP "niri:GNOME"

    QUIRK_APP_DIR "/home/bczhc/bin/scripts/exec-override"

    // NVIDIA driver
    __GL_SHADER_DISK_CACHE_SIZE "100000000000"
    __GL_SHADER_DISK_CACHE_SKIP_CLEANUP "1"

    // Electron Wayland hint
    ELECTRON_OZONE_PLATFORM_HINT "auto"
}
hotkey-overlay {
    skip-at-startup
}
prefer-no-csd
animations {
    workspace-switch {
        spring damping-ratio=0.80 stiffness=523 epsilon=0.0001
    }
    window-open {
        duration-ms 150
        curve "ease-out-expo"
    }
    window-close {
        duration-ms 150
        curve "ease-out-quad"
    }
    horizontal-view-movement {
        spring damping-ratio=0.85 stiffness=423 epsilon=0.0001
    }
    window-movement {
        spring damping-ratio=0.75 stiffness=323 epsilon=0.0001
    }
    window-resize {
        spring damping-ratio=0.85 stiffness=423 epsilon=0.0001
    }
    config-notification-open-close {
        spring damping-ratio=0.65 stiffness=923 epsilon=0.001
    }
    screenshot-ui-open {
        duration-ms 200
        curve "ease-out-quad"
    }
    overview-open-close {
        spring damping-ratio=0.85 stiffness=800 epsilon=0.0001
    }
}
// Window rules let you adjust behavior for individual windows.
// Find more information on the wiki:
// https://github.com/YaLTeR/niri/wiki/Configuration:-Window-Rules
// Work around WezTerm's initial configure bug
// by setting an empty default-column-width.
window-rule {
    // This regular expression is intentionally made as specific as possible,
    // since this is the default config, and we want no false positives.
    // You can get away with just app-id="wezterm" if you want.
    match app-id=r#"^org\.wezfurlong\.wezterm$"#
    default-column-width {}
}
window-rule {
    match app-id=r#"^org\.gnome\."#
    draw-border-with-background false
    geometry-corner-radius 12
    clip-to-geometry true
}
window-rule {
    match app-id=r#"^gnome-control-center$"#
    match app-id=r#"^pavucontrol$"#
    match app-id=r#"^nm-connection-editor$"#
    open-floating false
}
window-rule {
    match app-id=r#"^gnome-calculator$"#
    match app-id=r#"^galculator$"#
    match app-id=r#"^blueman-manager$"#
    match app-id=r#"^org\.gnome\.Nautilus$"#
    match app-id=r#"^xdg-desktop-portal$"#
    open-floating true
}
window-rule {
    match app-id=r#"^steam$"# title=r#"^notificationtoasts_\d+_desktop$"#
    default-floating-position x=10 y=10 relative-to="bottom-right"
    open-focused false
}
window-rule {
    match app-id=r#"^org\.wezfurlong\.wezterm$"#
    match app-id="Alacritty"
    match app-id="zen"
    match app-id="com.mitchellh.ghostty"
    match app-id="kitty"
    draw-border-with-background false
}
window-rule {
    match app-id=r#"firefox$"# title="^Picture-in-Picture$"
    match app-id="zoom"
    open-floating true
}
// Open dms windows as floating by default
window-rule {
    match app-id=r#"org.quickshell$"#
    match app-id=r#"com.danklinux.dms$"#
    open-floating true
}

window-rule {
    match app-id="firefox$"

    open-maximized-if-alone true
}

debug {
    render-drm-device "/dev/dri/card1"
    honor-xdg-activation-with-invalid-serial
    //disable-direct-scanout
}

// Override to disable super+tab
recent-windows {
    binds {
        Alt+Tab         { next-window scope="output"; }
        Alt+Shift+Tab   { previous-window scope="output"; }
        Alt+grave       { next-window filter="app-id"; }
        Alt+Shift+grave { previous-window filter="app-id"; }
    }
}

// Include dms files
include "dms/colors.kdl"
include "dms/layout.kdl"
include "dms/alttab.kdl"
include "dms/binds.kdl"
include "dms/outputs.kdl"
include "dms/cursor.kdl"
include "dms/windowrules.kdl"

```

</details> 

### System Information

* niri version: niri 25.11 (v25.11-137-g5a39c513-modified)

* Distro: Arch rolling

* GPU: NVIDIA GeForce RTX 3060 Mobile

* CPU: Intel i7-11800H

<del>Though I don't know if this is what proton should not do i.e. sending the position hint even when controlling a 3D camera will make no sense.</del>
Okay it needs to send the hint. 2D UI that captures the mouse + uses raw input is possible. If completely ignoring the hint update, i got mouse position issues with them.

### LINKED ISSUE #4121: Cursor constraints get activated when it's not keyboard focused

### LINKED ISSUE #4430: Pointer lock should be released on keyboard focus loss
After pointer constraints behavior change of 0dce5f125c1445992b5d800d01e228ad21108163, it fixes #3729 there's still a new bug.

Demonstration video:

https://github.com/user-attachments/assets/90861000-72ca-4236-a621-0f1925d0378d

Expected behavior: Pointer lock should be released once the window loses its keyboard focus (e.g. workspace change operation, Mod+HJKL, or Alt+Tab), and should be restored automatically on keyboard focus regain. Otherwise the user won't be able to escape from the lock state unless the app itself actively unlocks the pointer.

Tested sway/hyprland/kwin_wayland/mutter, all does the "expected behavior".

Vantrongs:
Fixes #3729.
Fixes #4121.
Fixes #4430.

## Behavior

Pointer constraints now follow both pointer and keyboard focus:

- activation requires pointer and keyboard focus to resolve to the same root shell surface;
- an active lock or confinement is released when an explicit focus action moves keyboard focus away;
- a persistent constraint is restored when keyboard focus returns, while a one-shot constraint is not;
- a committed cursor-position hint remains deferred while locked and is applied at most once after deactivation/removal;
- overlapping surfaces, ordinary clicks, absolute input, and pointer refreshes do not steal focus from an active lock;
- layer-shell surfaces follow their keyboard-interactivity policy: `exclusive` can activate, `on-demand` activates after receiving keyboard focus, and `none` cannot activate.

The lifecycle also accounts for current surface geometry, input/constraint regions, workspace/output changes, popups and subsurfaces, layer surfaces, overview animation, hidden cursors, and multiple persistent constraints.

## Smithay prerequisite

This branch pins [`1b0e75ab`](https://github.com/Vantrongs/smithay/commit/1b0e75abb473847c08ed83b1d18519c06fed9a47), which is exactly one commit on top of Niri's current Smithay revision `7609ab82`. Therefore it includes the already merged Smithay [#2107](https://github.com/Smithay/smithay/pull/2107) and does not drop any of Niri's existing Smithay changes.

The additional commit separates temporary constraint deactivation from final protocol-object removal, carries the concrete surface/pointer/constraint identity through both callbacks, exposes the lifetime/object identity needed for cleanup, and documents that pointer operations must be deferred outside the pointer-motion callback. Terminal removal is matched against the exact Wayland object identity, so destroying an old defunct one-shot object cannot remove a newer constraint created for the same surface and pointer. Deferring pointer operations avoids re-entering `PointerHandle`'s mutex when focus loss deactivates a constraint during motion dispatch.

This history matters because the earlier #2107 head (`46a875`) was older than the Smithay revision Niri's update work already used (`5fb12b`); pinning that old head directly would have dropped four unrelated fixes. The present fork is based on the newer upstream `7609ab82`, not that older branch. Its single lifecycle commit still needs separate Smithay upstream review before this Niri PR can merge without a fork.

## Validation

On exact PR head [`5532ed92`](https://github.com/Vantrongs/niri/commit/5532ed925b03165d4cbefd19afb721594c0c66f4), using Smithay `1b0e75ab`:

- all 28 pointer-constraint regression tests pass, including the defunct one-shot/replacement-constraint lifecycle regression;
- all 223 Niri tests pass;
- `cargo fmt --all -- --check`;
- `git diff --check`.

The preceding implementation head also passed `cargo clippy --locked --all-targets --all-features -- -D warnings`. The final head changes only the Smithay pin/lockfile and adds the lifecycle regression test described above.

The core lock/focus lifecycle was also tested successfully with the native Linux War Thunder client on NixOS 26.05, NVIDIA, and one physical display in generation 449. That manual generation used the pre-final-rebase implementation (`b902a70b`); the later changes cover layer-shell, overview animation, hidden-cursor, and lifecycle edge cases with automated regressions on the exact PR head. Proton titles and multi-display behavior still benefit from independent testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3232,7 +3232,7 @@ checksum = "e8e2fb0f499abb4d162f2bedad68f5ef91a1682b5a03596ddb67efd37768d100"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/Vantrongs/smithay.git?rev=fe0320bc4586383dd774fd1768775e7d941e1e47#fe0320bc4586383dd774fd1768775e7d941e1e47"
+source = "git+https://github.com/Vantrongs/smithay.git?rev=1b0e75abb473847c08ed83b1d18519c06fed9a47#1b0e75abb473847c08ed83b1d18519c06fed9a47"
 dependencies = [
  "aliasable",
  "appendlist",
@@ -3311,7 +3311,7 @@ dependencies = [
 [[package]]
 name = "smithay-drm-extras"
 version = "0.1.0"
-source = "git+https://github.com/Vantrongs/smithay.git?rev=fe0320bc4586383dd774fd1768775e7d941e1e47#fe0320bc4586383dd774fd1768775e7d941e1e47"
+source = "git+https://github.com/Vantrongs/smithay.git?rev=1b0e75abb473847c08ed83b1d18519c06fed9a47#1b0e75abb473847c08ed83b1d18519c06fed9a47"
 dependencies = [
  "drm",
  "libdisplay-info",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3232,7 +3232,7 @@ checksum = "e8e2fb0f499abb4d162f2bedad68f5ef91a1682b5a03596ddb67efd37768d100"
 [[package]]
 name = "smithay"
 version = "0.7.0"
-source = "git+https://github.com/Smithay/smithay.git?rev=7609ab82f002436188ebd4df1d8f7e1905f90c4a#7609ab82f002436188ebd4df1d8f7e1905f90c4a"
+source = "git+https://github.com/Vantrongs/smithay.git?rev=fe0320bc4586383dd774fd1768775e7d941e1e47#fe0320bc4586383dd774fd1768775e7d941e1e47"
 dependencies = [
  "aliasable",
  "appendlist",
@@ -3311,7 +3311,7 @@ dependencies = [
 [[package]]
 name = "smithay-drm-extras"
 version = "0.1.0"
-source = "git+https://github.com/Smithay/smithay.git?rev=7609ab82f002436188ebd4df1d8f7e1905f90c4a#7609ab82f002436188ebd4df1d8f7e1905f90c4a"
+source = "git+https://github.com/Vantrongs/smithay.git?rev=fe0320bc4586383dd774fd1768775e7d941e1e47#fe0320bc4586383dd774fd1768775e7d941e1e47"
 dependencies = [
  "drm",
  "libdisplay-info",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,15 +27,14 @@ tracy-client = { version = "0.18.4", default-features = false }
 
 [workspace.dependencies.smithay]
 # version = "0.4.1"
-git = "https://github.com/Smithay/smithay.git"
-rev = "7609ab82f002436188ebd4df1d8f7e1905f90c4a"
+git = "https://github.com/Vantrongs/smithay.git"
+rev = "fe0320bc4586383dd774fd1768775e7d941e1e47"
 default-features = false
 
 [workspace.dependencies.smithay-drm-extras]
 # version = "0.1.0"
-git = "https://github.com/Smithay/smithay.git"
-rev = "7609ab82f002436188ebd4df1d8f7e1905f90c4a"
-# path = "../smithay/smithay-drm-extras"
+git = "https://github.com/Vantrongs/smithay.git"
+rev = "fe0320bc4586383dd774fd1768775e7d941e1e47"
 
 [package]
 name = "niri"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,13 +28,13 @@ tracy-client = { version = "0.18.4", default-features = false }
 [workspace.dependencies.smithay]
 # version = "0.4.1"
 git = "https://github.com/Vantrongs/smithay.git"
-rev = "fe0320bc4586383dd774fd1768775e7d941e1e47"
+rev = "1b0e75abb473847c08ed83b1d18519c06fed9a47"
 default-features = false
 
 [workspace.dependencies.smithay-drm-extras]
 # version = "0.1.0"
 git = "https://github.com/Vantrongs/smithay.git"
-rev = "fe0320bc4586383dd774fd1768775e7d941e1e47"
+rev = "1b0e75abb473847c08ed83b1d18519c06fed9a47"
 
 [package]
 name = "niri"
--- a/docs/wiki/Configuration:-Debug-Options.md
+++ b/docs/wiki/Configuration:-Debug-Options.md
@@ -283,6 +283,8 @@ Most of the time, these fresh tokens will have invalid serials, because the app 
 By default, niri ignores xdg-activation tokens with invalid serials, to prevent windows from randomly stealing focus.
 This debug flag makes niri honor such tokens, making the aforementioned widely-used apps get focus when clicking on their tray icon or notification.
 
+Use the [`on-xdg-activate` window rule](./Configuration:-Window-Rules.md#on-xdg-activate) to control what niri does for individual windows when it accepts an xdg-activation request.
+
 Amusingly, clicking on a notification sends the app a perfectly valid activation token from the notification daemon, but these apps seem to simply ignore it.
 Maybe in the future these apps/toolkits (Electron, Qt) are fixed, making this debug flag unnecessary.
 
--- a/docs/wiki/Configuration:-Window-Rules.md
+++ b/docs/wiki/Configuration:-Window-Rules.md
@@ -58,6 +58,7 @@ window-rule {
     default-column-display "tabbed"
     default-floating-position x=100 y=200 relative-to="bottom-left"
     scroll-factor 0.75
+    on-xdg-activate "focus"
 
     focus-ring {
         // off
@@ -593,6 +594,35 @@ window-rule {
 >
 > This is because window title (and app ID) are not double-buffered in the Wayland protocol, so they are not tied to specific window contents.
 > There's no robust way for Firefox to synchronize visibly showing a different tab and changing the window title.
+
+#### `on-xdg-activate`
+
+<sup>Since: next release</sup>
+
+Set what niri does when a window requests activation through [xdg-activation](https://wayland.app/protocols/xdg-activation-v1).
+
+Values:
+
+- `"ignore"`: do nothing: neither focus the window nor mark it urgent.
+- `"set-urgent"`: always mark the window urgent instead of focusing it.
+- `"focus"`: always focus the window, even for activation requests without a serial.
+
+The default behavior, when this rule is unset, picks between focusing and marking the window urgent based on the serial that the application provides when creating the activation token.
+Requests with a valid serial focus the target window, and requests without a serial mark it urgent.
+
+Requests with a set, but invalid, serial, are always ignored, which you can change with the [`honor-xdg-activation-with-invalid-serial`](https://niri-wm.github.io/niri/Configuration%3A-Debug-Options.html#honor-xdg-activation-with-invalid-serial) debug flag.
+
+This is useful for apps that steal focus on incoming messages or when opening popup windows like Picture-in-Picture.
+
+```kdl
+// Don't let Firefox PiP steal focus.
+window-rule {
+    match title="^Picture-in-Picture$"
+
+    open-floating true
+    on-xdg-activate "set-urgent"
+}
+```
 
 #### `opacity`
 
--- a/docs/wiki/Window-Geometry-IPC.md
+++ b/docs/wiki/Window-Geometry-IPC.md
@@ -1,0 +1,19 @@
+# On-demand window geometry (local fork)
+
+`niri msg --json window-geometry --id <window-id>` returns the exact focused
+window's current global logical visual rectangle and the logical output layout
+from the same compositor callback. It uses rendered placement, including the
+client geometry offset; it does not derive tiled positions from cached IPC
+layout data and does not emit per-frame window-layout events.
+
+The request fails while locked, in overview/screenshot/MRU/exit UI, during output
+layout animations or screen transitions, for a dragged window, or if actual
+keyboard focus belongs to another window or a layer-shell surface.
+Consumers must handle these errors instead of guessing coordinates. Geometry
+is a snapshot, not a lock: validate it around capture and again after input
+backend preparation, immediately before dispatch. Never replay an input operation
+after failure. A compositor restart is required after installing
+this fork update; replacing only the `niri msg` CLI is insufficient.
+
+Regression: `cargo test -p niri live_window_geometry` exercises tiled and moved
+floating windows against the real headless compositor's pointer hit-testing.
--- a/flake.nix
+++ b/flake.nix
@@ -49,6 +49,7 @@
               ./src
               ./Cargo.toml
               ./Cargo.lock
+              ./build.rs
             ];
           };
 
@@ -164,6 +165,11 @@
     in
     {
       checks = forAllSystems (system: {
+        niri-source = nixpkgsFor.${system}.runCommand "niri-source-contract" { } ''
+          # Cargo must discover the libinput capability probe in the filtered source.
+          cmp ${./build.rs} ${self.packages.${system}.niri.src}/build.rs
+          touch "$out"
+        '';
         # We use the debug build here to save a bit of time
         inherit (self.packages.${system}) niri-debug;
       });
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -60,7 +60,7 @@ pub use crate::recent_windows::{MruDirection, MruFilter, MruPreviews, MruScope, 
 pub use crate::utils::FloatOrInt;
 use crate::utils::{Flag, MergeWith as _};
 pub use crate::window_rule::{
-    FloatingPosition, PopupsRule, RelativeTo, ResolvedPopupsRules, WindowRule,
+    FloatingPosition, OnXdgActivate, PopupsRule, RelativeTo, ResolvedPopupsRules, WindowRule,
 };
 pub use crate::workspace::{Workspace, WorkspaceLayoutPart};
 
@@ -656,6 +656,30 @@ mod tests {
     }
 
     #[test]
+    fn parse_on_xdg_activate() {
+        let parsed = do_parse(
+            r#"
+            window-rule { on-xdg-activate "ignore"; }
+            window-rule { on-xdg-activate "set-urgent"; }
+            window-rule { on-xdg-activate "focus"; }
+            "#,
+        );
+
+        assert_eq!(
+            parsed
+                .window_rules
+                .iter()
+                .map(|rule| rule.on_xdg_activate)
+                .collect::<Vec<_>>(),
+            vec![
+                Some(OnXdgActivate::Ignore),
+                Some(OnXdgActivate::SetUrgent),
+                Some(OnXdgActivate::Focus),
+            ]
+        );
+    }
+
+    #[test]
     fn parse() {
         let parsed = do_parse(
             r##"
@@ -891,6 +915,7 @@ mod tests {
                 default-window-height { fixed 500; }
                 default-column-display "tabbed"
                 default-floating-position x=100 y=-200 relative-to="bottom-left"
+                on-xdg-activate "ignore"
 
                 focus-ring {
                     off
@@ -1799,6 +1824,9 @@ mod tests {
                     ),
                     open_focused: Some(
                         true,
+                    ),
+                    on_xdg_activate: Some(
+                        Ignore,
                     ),
                     min_width: None,
                     min_height: None,
--- a/niri-config/src/window_rule.rs
+++ b/niri-config/src/window_rule.rs
@@ -34,6 +34,8 @@ pub struct WindowRule {
     pub open_floating: Option<bool>,
     #[knuffel(child, unwrap(argument))]
     pub open_focused: Option<bool>,
+    #[knuffel(child, unwrap(argument))]
+    pub on_xdg_activate: Option<OnXdgActivate>,
 
     // Rules applied dynamically.
     #[knuffel(child, unwrap(argument))]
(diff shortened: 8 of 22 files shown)


bczhc:
Behavior with this patch:

https://github.com/user-attachments/assets/d193039d-a48a-47ca-86af-dad40261a0c8

Note the red point below the semitransparant alacrity window. The upper window still gains the keyboard focus (on click) but with the lower window keeping locking the pointer.

This fixes for some games but not all (namely okay for 'Escape The Backrooms' but still fails for 'Palworld'. Both are UE games, run with etaash-proton 11.0). I guess for those failed cases they have a "stop receiving any mouse events on keyboard focus loss" or "explicitly unlock the pointer on keyboard focus loss" logic.

update: even for the "okay" case because the upper window gets the keyboard focus, the lower game loses its character control via wasd keys.

Vantrongs:
Thanks for testing this. Pointer lock and keyboard focus are independent here: this patch only prevents a committed cursor hint/content refresh from causing a synthetic surface leave and deactivating the pointer constraint. It intentionally does not stop another window from gaining keyboard focus. If Palworld destroys its lock or stops consuming relative motion after keyboard-focus loss, that is a separate client/focus interaction rather than the premature unlock fixed here. I will keep this PR scoped to #3729 and the behavior covered by its regression test.

bczhc:
> Thanks for testing this. Pointer lock and keyboard focus are independent here: this patch only prevents a committed cursor hint/content refresh from causing a synthetic surface leave and deactivating the pointer constraint. It intentionally does not stop another window from gaining keyboard focus. If Palworld destroys its lock or stops consuming relative motion after keyboard-focus loss, that is a separate client/focus interaction rather than the premature unlock fixed here. I will keep this PR scoped to #3729 and the behavior covered by its regression test.

There's no such thing as "keep scoped". Your patch does not fix issue #3729. Please re-read the issue and try testing yourself to see what mutter/hyprland/kde_wayland/sway behaves - The upper window should not gain keyboard focus on click when the lower window is locking the pointer. Otherwise, the lower window will lost its character move due to "wasd" keys get passed to the uppwer window. See the "Hyprland" video in comment https://github.com/niri-wm/niri/issues/3729#issuecomment-4287089677.

Vantrongs:
You were right, and my previous reply drew the boundary too narrowly. Keeping
the pointer locked while allowing the click to move keyboard focus still makes
the game lose WASD, so it does not fully solve #3729.

I have rebased and updated the patch. It now keeps ordinary click focus on the
locked surface as well as preventing content refreshes, overlapping surfaces,
and absolute motion from breaking pointer focus. The latest cursor hint is
still applied normally after the lock is removed.

There was also a Smithay complication behind the delay. When I first worked on
this, Smithay #2107's head (`46a875`) was based on an older revision than the
one Niri's `update-smithay` work already used (`5fb12b`). Pointing Cargo at that
PR head would have silently dropped four unrelated Smithay fixes. I therefore
used a local prototype for investigation instead of putting that dependency in
this PR. Smithay #2107 has since been merged, and current Niri `main` now pins a
newer revision (`7609ab82`) that includes it, so this branch no longer needs a
Smithay fork or override.

Could you please retry the latest commit with this sequence?

1. Put the same semi-transparent floating Alacritty window over the game.
2. Let the game lock the pointer, move the hidden cursor/hint under Alacritty,
   and click there.
3. Check that Alacritty never gains keyboard focus, WASD still controls the
   game, and relative camera motion continues without a jump.
4. Check that the game does not receive `unlocked` until it intentionally
   releases or destroys the lock.
5. Open the game menu or otherwise release the lock and confirm that the cursor
   appears at the latest hint and normal window focusing works again.
6. If possible, repeat this in both Palworld and Escape The Backrooms. A second
   display or an absolute/virtual pointer input would also be useful extra
   coverage.

If it still fails, the most useful detail would be which happens first: the
upper window gains keyboard focus, the client receives `unlocked`, or relative
motion stops while both focuses still look correct.

Locally, the original aim jump no longer reproduced in the native Linux War
Thunder client on NixOS/NVIDIA with one display. The exact PR head also passes
four focused regression tests and the full Niri test suite, but your Proton
cases cover behavior that this local setup cannot, so your feedback is still
important.

bczhc:
Well thanks for the update. Tested it. The patch literally merely changed the keyboard behavior which _does_ fix my game issue but also raises new bugs: The cursor is not able to be unlocked on keyboard focus leave (`wl_keyboad::leave`) (e.g. via alt+tab or mod+HJKL keys).

Your current approach may be going the wrong way and only solving the symptom rather than the underlying pointer constraint behavior. Again, read the original issue full context please. See Hyprland vs niri demonstration video in https://github.com/niri-wm/niri/issues/3729#issuecomment-4287089677. So I'm using a simple winit+wgpu app in the comment to demonstrate the issue. The expected behaviors - from my tests against other compositors - I'll summarize for you here:

1. Once the pointer is locked, the visual pointer position (if it's not set hidden, you'll be able to observe) should not change on `set_cursor_position_hint` received (that is, so why it's called a "hint", is that the pointer only updates the "actual position" on unlock. It's a deferred behavior). niri now always updates the visual pointer position on that event received.
2. Once the pointer is locked, when `set_cursor_position_hint` hints a position outside the original window (e.g. go outside the border or inside another overlapped upper window), the cursor should not be unlocked and clicks should not change keyboard focuses.
3. So, according to above, the _only way_ to unlock the pointer while a window is locking your pointer, is 1) closing the original window, 2) using some keybinds to switch the keyboard focus (something like alt+tab), or 3) the app itself actively unlocks it. (ofc. in way "2)" when the keyboard focus is switched back to the original window which locks the pointer, the pointer lock should be restored, automatically)

The above describes what Hyprland/sway/kwin_wayland/mutter _all_ does. It's better regarding this niri should also behave the same as them I think. Your patch doesn't fix 1 and breaks 3.

Vantrongs:
Thanks — your three-point summary identified the missing lifecycle, and the latest head now implements all three rather than only preventing focus theft.

**Update:** the exact head to test is now [`5532ed92`](https://github.com/Vantrongs/niri/commit/5532ed925b03165d4cbefd19afb721594c0c66f4). It supersedes the previously mentioned `2a97e5ee`; the Niri behavior is unchanged by the final commit, which updates the Smithay prerequisite and adds a regression for the corrected object lifecycle.

1. `set_cursor_position_hint` remains deferred while locked. The compositor pointer location does not change until deactivation/removal, and the latest hint is then applied at most once.
2. While active, overlap, pointer refresh, absolute input, and an ordinary click cannot move pointer or keyboard focus away from the constrained root surface.
3. An explicit keyboard-focus action such as Alt+Tab or Mod+HJKL sends `unlocked`/`unconfined`; returning keyboard focus restores the same persistent constraint. A one-shot constraint is not restored.

This also follows YaLTeR's suggestion in #4430: activation checks that pointer and keyboard focus resolve to the same root shell surface, and `update_keyboard_focus()` explicitly suspends the old root's constraint. Both Locked and Confined constraints are covered. Layer-shell behavior is covered as well: `exclusive` can activate, `on-demand` activates after click focus, and `none` cannot activate.

Could you please retry your demo against that exact head, including the unpublished mouse-toggle layer variant if convenient? The most useful matrix is:

- Locked and Confined;
- Alt+Tab and Mod+HJKL: exactly one unlock/unconfine, then one re-lock/re-confine when returning for a persistent constraint;
- moving the pointer back over the surface without keyboard focus must not reactivate it (#4121);
- a committed hint or click under an overlapping window must not move the compositor cursor or keyboard focus while active;
- layer `exclusive`, `on-demand`, and `none` as above;
- if possible, overview close while hide-when-typing has hidden the cursor.

There is one important Smithay prerequisite. The branch now pins [`1b0e75ab`](https://github.com/Vantrongs/smithay/commit/1b0e75abb473847c08ed83b1d18519c06fed9a47), exactly one lifecycle commit on top of Niri's current `7609ab82` (which already includes Smithay #2107). It is not the old #2107 head `46a875`; that older head predated Niri's then-used `5fb12b` and would have dropped four unrelated Smithay fixes.

The final Smithay commit distinguishes temporary deactivation from terminal protocol-object removal, carries the exact constraint identity through both callbacks, and matches terminal destruction against the exact Wayland object. In particular, destroying an old defunct one-shot object can no longer remove a newer constraint created for the same surface and pointer. Niri includes a regression for that sequence. This Smithay commit still needs separate upstream review before the Niri PR can merge without depending on the fork.

On the exact Niri/Smithay pair, all 28 pointer-constraint tests and all 223 Niri tests pass. The native Linux War Thunder scenario also worked in the tested NixOS generation; that manual run covered the core lifecycle, while the final lifecycle edge case is covered by the automated regression.

PR #4422, opened 2026-08-10
 input: honor keyboard-shortcuts-inhibit for mouse input

LilFishyChan:
Currently, niri consults `is_inhibiting_shortcuts()` only in keyboard bind handling. Mouse bind handling and Mod+mouse gestures don't, so apps like Moonlight that request inhibit to forward input to a remote machine still see their `Mod+Wheel*` binds consumed locally instead of reaching the app.

The protocol covers keyboard keys, and `Mod` (Super/Alt) is one. Two changes here, with the same Mod-is-keyboard argument:

1. Mouse binds (commit 1): user-bound binds are the unit of meaning, not the input device. `Mod+WheelDown { focus-workspace-down }` and `Mod+U { focus-workspace-down }` are the same shortcut in user semantics — only the trigger differs.

2. Mod+mouse gestures (commit 2): the gesture is conditional on `Mod` held; since `Mod` is a keyboard key in protocol scope, the gesture that depends on it shouldn't fire when the focused surface has requested inhibit. This covers `Mod+LMB` (interactive move), `Mod+RMB` (interactive resize), `Mod+MMB` (spatial workspace switch)

Concrete failure mode that motivates this: same niri config on two machines, bridged via Moonlight. On machine A I press `Mod+WheelDown` expecting the remote (machine B) to switch workspaces — but niri consumes the event locally, so A's workspaces change while B's don't. Same with `Mod+LMB` drag: the drag motion never reaches machine B.

Tested locally on Arch (niri 26.04 base); binary verifies against moonlight and remmina.

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -3131,6 +3131,11 @@ impl State {
                 || is_mru_open
                 || self.niri.mods_with_wheel_binds.contains(&modifiers);
             if should_handle {
+                // Honor keyboard-shortcuts-inhibit for wheel binds, mirroring the
+                // FilterResult::Forward check in on_keyboard_key_event.
+                let inhibited = self.is_inhibiting_shortcuts();
+                let mut any_bind_ran = false;
+
                 let horizontal = horizontal_amount_v120.unwrap_or(0.);
                 let ticks = self.niri.horizontal_wheel_tracker.accumulate(horizontal);
                 if ticks != 0 {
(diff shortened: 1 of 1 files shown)


YaLTeR:
well, it *is* called **keyboard**-shortcuts-inhibit?

LilFishyChan:
> well, it is called **keyboard**-shortcuts-inhibit?

The protocol covers keyboard keys. `Mod` is a keyboard key. That puts keyboard binds squarely inside the protocol's scope

`Mod + non-keyboard-input` — `Mod+WheelDown`, `Mod+LMB` drag, etc. Two angles:
1. Binds: user-bound shortcuts are the unit of meaning. `Mod+WheelDown { focus-workspace-down }` and `Mod+U { focus-workspace-down }` are the same shortcut, only the trigger differs.
2. Gestures: the gesture is conditional on `Mod` held. `Mod` is in protocol scope. So a gesture whose precondition (`Mod` held by the user) is being undone by the inhibit shouldn't fire.

Concrete case: same niri config on two machines, bridged via Moonlight. On machine A I press `Mod+WheelDown` expecting the remote (machine B) to switch workspaces — but niri consumes the event locally, so A's workspaces change while B's don't.

If needed, I can rework the PR description, commit messages, and scope.

YaLTeR:
I guess that makes sense. Do other compositors also do this btw?

LilFishyChan:
> I guess that makes sense. Do other compositors also do this btw?

Yes — Hyprland does the same thing: mouse binds go through the same `processEvent` that checks `shortcutsInhibit->isInhibited()` ([`Manager.cpp:576`](https://github.com/hyprwm/Hyprland/blob/main/src/keybinds/Manager.cpp#L576)).

PR #4435, opened 2026-08-12
fix: use `active_window` for geometry calculation

NSPC911:
Niri currently uses the topmost floating window for `warp-mouse-to-focus`. However, this conflicts with `focus-follows-mouse` since it implicitly(?) focuses on a window; it doesn't set it as topmost until you click on it, resulting in the video below

https://github.com/user-attachments/assets/da48cec5-0a04-4120-a0ae-6ab90df3475a

This PR fixes the problem by using the active window to resolve geometry

https://github.com/user-attachments/assets/5278ea47-0b11-4085-b0e6-92ad4daf54db

No tests were added; I'm not sure how to do those yet.

### DIFF
--- a/src/layout/floating.rs
+++ b/src/layout/floating.rs
@@ -360,7 +360,10 @@ impl<W: LayoutElement> FloatingSpace<W> {
     ///
     /// During animations, assumes the final tile position.
     pub fn active_window_visual_rectangle(&self) -> Option<Rectangle<f64, Logical>> {
-        let (tile, offset) = self.tiles_with_offsets().next()?;
+        let active_id = self.active_window_id.as_ref()?;
+        let (tile, offset) = self
+            .tiles_with_offsets()
+            .find(|(tile, _)| tile.window().id() == active_id)?;
 
         let window_pos = offset + tile.window_loc();
         let window_size = tile.window_size();


YaLTeR:
That's correct in principle, however I have one concern: how does it interact if some window overlaps from above, and warp mouse to focus places the mouse over the window above? Does it work fine or does it cause unwanted focus?

Also, the second place this is used is for tablet `map-to-focused-window` but I think that's fine because it doesn't do focus-follows-mouse.

NSPC911:
> That's correct in principle, however I have one concern: how does it interact if some window overlaps from above, and warp mouse to focus places the mouse over the window above? Does it work fine or does it cause unwanted focus?

It does not cause unwanted focus


https://github.com/user-attachments/assets/fac43a3b-6780-4e64-9b97-af386c34a282


https://github.com/user-attachments/assets/41d41b6e-f59b-46bb-9369-62cfc022f636

^ This seems quite reasonable

> Also, the second place this is used is for tablet `map-to-focused-window` but I think that's fine because it doesn't do focus-follows-mouse.


Yes, just double-checked; tablet doesn't do `focus-follows-mouse`


https://github.com/user-attachments/assets/f04a5f96-e602-4ba0-8117-6463a5d3772c

YaLTeR:
Thanks, that answers all of my questions. Come to think of it, this situation also arises when an active scrolling window is covered by a floating window, so it's not new.

PR #4444, opened 2026-08-14
niri-session: fix shellcheck warnings

Ordoviz:
[ANSI-C strings](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html) like `$''` are Bash-only. The one in niri-session can be replaced with an ordinary single-quoted string because it contains no backslash escapes.

### DIFF
--- a/resources/niri-session
+++ b/resources/niri-session
@@ -77,15 +77,15 @@ elif hash dinitctl >/dev/null 2>&1; then
     fi
 
     # Create the directory for the logfile, if doesn't exist
-    mkdir --parents $HOME/.local/share/niri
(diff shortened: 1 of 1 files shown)


neunato:
Actually this is now part of POSIX too

https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_02_04
https://pubs.opengroup.org/onlinepubs/9799919799/xrat/V4_xcu_chap01.html#tag_23_02_02_04

YaLTeR:
Thanks for checking. I guess this change is fine either way

PR #4459, opened 2026-08-17
input: per-keyboard-device xkb and repeat configuration

ArthurViniciusBA1:
## Summary

Adds the ability to configure keyboard settings (xkb layout, repeat rate/delay, numlock) per physical keyboard device, e.g. a different layout on an external keyboard than on a laptop's built-in one. Closes part of #371 (the keyboard piece specifically; touchpad/mouse/etc. per-device config is still not covered by this PR).

```kdl
input {
    keyboard {
        xkb {
            layout "br"
        }
    }

    keyboard "My External Keyboard" {
        xkb {
            layout "us"
        }
    }
}
```

Run `niri msg keyboard-layouts` or check `libinput list-devices` / `/proc/bus/input/devices` to find a device's exact name.

Named keyboard blocks inherit any setting they don't specify from the unnamed block in the same `input` section (so you don't need to repeat `repeat-rate`/`numlock`/etc. in every block), then fall back to the regular defaults.

## Background

This picks up where #417 left off — same underlying idea (there's no way to have more than one active Wayland keymap per seat, so per-device config means re-resolving and re-applying the config for whichever device most recently sent a key event), but rebased on current `main` and adjusted where the codebase has moved on since 2024:

- `Input::keyboards: Vec<Keyboard>` replaces the resolve-independently approach from #417 with cascading resolution: named blocks inherit from the unnamed block in the same section rather than always falling back to bare defaults for anything unset.
- Config reload now clears the per-device resolved-config cache and resets the tracked active keyboard to the fallback when `input.keyboards` changes, so edits take effect on the next keystroke instead of being masked by stale cached per-device configs (this was left as a known gap in #417's review).
- `xkb.file` (raw keymap file override) is honored per-device too, going through the existing `set_xkb_file` path instead of always falling through to the rules-based `XkbConfig` path.
- Bind key repeat and the a11y repeat-delay heuristic now use the currently tracked keyboard's repeat settings instead of always the fallback keyboard's.
- Device name matching is case-insensitive, per review feedback on #417.

## Known limitation

`track-layout "window"` combined with multiple configured keyboards has the same open question raised on #417: per-surface layout tracking currently stores one `KeyboardLayout` per surface, not one per (keyboard, surface) pair, so switching between two `Window`-tracked keyboards and expecting independent per-keyboard-per-window layout memory isn't fully correct. This doesn't affect the common case of keyboards each having a single fixed layout (`track-layout` defaults to `Global`, which is unaffected), or of only the fallback keyboard using `Window` tracking. Flagging this here rather than trying to redesign per-surface layout tracking in this PR — happy to follow up if that's wanted.

XWayland clients generally see one process-wide X11 keymap, so per-keyboard layouts may not be reflected correctly in X11-only clients (noted in the wiki docs).

## Testing

- `cargo test --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --check` all pass.
- Manually tested on a real tty/udev session with two physical keyboards (a laptop's built-in keyboard on `br` and an external keyboard named via a `keyboard "..."` block on `us`): typing on each keyboard produces the correct layout's characters, confirming the device is resolved correctly and the keymap switches on keystroke as expected.

### DIFF
--- a/docs/wiki/Configuration:-Input.md
+++ b/docs/wiki/Configuration:-Input.md
@@ -4,7 +4,7 @@ In this section you can configure input devices like keyboard and mouse, and som
 
 There's a section for each device type: `keyboard`, `touchpad`, `mouse`, `trackpoint`, `trackball`, `tablet`, `touch`.
 Settings in those sections will apply to every device of that type.
-Currently, there's no way to configure specific devices individually (but that is planned).
+Currently, there's no way to configure specific touchpad/mouse/etc. devices individually (but that is planned), except for `keyboard`, which can be given a device name to configure just that one keyboard — see [Per-Keyboard Configuration](#per-keyboard-configuration) below.
 
 All settings at a glance:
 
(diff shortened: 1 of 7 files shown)


Severin-Nitsche:
I tested this today on my setup and found that omitting part of the `xkb`-Block does not automagically fall back to what is defined in the unnamed Block, so I was hit with a sudden `us` layout when I just wanted to change some `xkb->options` for my external keyboard. Otherwise it seems to be working fine.

PR #4472, opened 2026-08-24
Fix client triggerable overflow panics

WitteShadovv:
Follow-up to the report I emailed you.

Three places where niri does raw integer arithmetic on client-controlled coordinates and overflows. With `overflow-checks = true` in the release profile each one is a panic, which aborts the compositor and takes every client in the session with it.

Every fix is a `saturating_*` call, matching what smithay's `Point`/`Size` operators already do everywhere else in the coordinate math.

### 1. Blur region rect extremities (`src/utils/region.rs`)

This is the one that matters. `region_to_non_overlapping_rects()` computes `r.loc.y + r.size.h` and `r.loc.x + r.size.w` on raw `i32`s that came straight from `wl_region.add()` and were never clamped, so a region containing `add(0, i32::MAX, 1, 1)` overflows.

It's reachable from a **sandboxed** client. `ext-background-effect` is one of the few globals created without the `client_is_unrestricted` filter, so a client on a security-context socket can bind it, attach that region as a blur region, and niri panics on the next repaint. Tile rendering calls `render_background_effect()` unconditionally, so it doesn't depend on blur being enabled in the config, and `get_cached_blur_region()` runs before any visibility check.

Introduced in 931123f3 ("Implement ext-background-effect protocol"), so this affects 26.04 and current main.

### 2. `axis_discrete` v120 value (`src/protocols/virtual_pointer.rs`)

`discrete * 120` overflows for `discrete > 17_895_697`.

### 3. CVT timing sums (`src/backend/tty.rs`)

`zwlr_output_configuration_head_v1::set_custom_mode` accepts width and height up to `u16::MAX`, and the CVT timing sums are unchecked `u16` addition. `1920x65535` panics on the horizontal sums, `65000x65000` on the vertical ones.

2 and 3 are both behind `client_is_unrestricted`, so they're only reachable by clients your security model already trusts. That makes them robustness fixes rather than boundary crossings, but they're the same one-line class, so they're in here as well.

### Testing

`cargo test` passes and the two regression tests both fail without the corresponding arithmetic fix:

```
---- utils::region::tests::test_extremity_overflow stdout ----
panicked at src/utils/region.rs:107:42: attempt to add with overflow

---- backend::tty::tests::test_calc_cvt_extreme_size stdout ----
panicked at src/backend/tty.rs:3115:23: attempt to add with overflow
```

I left the virtual pointer one untested. There's no virtual-pointer harness in `src/tests/` yet, and building one for a `saturating_mul` felt out of proportion for this PR. Happy to add it if you'd rather have it.

To reproduce the first one by hand: a client that binds `ext_background_effect_manager_v1`, maps a toplevel with a buffer, then does

```
region = wl_compositor.create_region()
region.add(0, i32::MAX, 1, 1)
effect = ext_background_effect_manager_v1.get_background_effect(surface)
effect.set_blur_region(region)
surface.commit()
```

On main niri aborts on the next repaint. On this branch it keeps running and the region is simply clamped away, which is correct since the blur region gets clipped to the surface during rendering anyway.

### DIFF
--- a/src/backend/tty.rs
+++ b/src/backend/tty.rs
@@ -3112,13 +3112,13 @@ pub fn calculate_mode_cvt(width: u16, height: u16, refresh: f64) -> DrmMode {
     };
     let cvt_timing = libdisplay_info::cvt::Timing::compute(options);
 
-    let hsync_start = width + cvt_timing.h_front_porch as u16;
+    let hsync_start = width.saturating_add(cvt_timing.h_front_porch as u16);
     let vsync_start = (cvt_timing.v_lines_rnd + cvt_timing.v_front_porch) as u16;
-    let hsync_end = hsync_start + cvt_timing.h_sync as u16;
-    let vsync_end = vsync_start + cvt_timing.v_sync as u16;
+    let hsync_end = hsync_start.saturating_add(cvt_timing.h_sync as u16);
+    let vsync_end = vsync_start.saturating_add(cvt_timing.v_sync as u16);
 
-    let htotal = hsync_end + cvt_timing.h_back_porch as u16;
-    let vtotal = vsync_end + cvt_timing.v_back_porch as u16;
+    let htotal = hsync_end.saturating_add(cvt_timing.h_back_porch as u16);
+    let vtotal = vsync_end.saturating_add(cvt_timing.v_back_porch as u16);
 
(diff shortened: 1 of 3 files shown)


YaLTeR:
Thanks

YaLTeR:
Also notified the upcoming Smithay virtual-pointer impl here https://github.com/Smithay/smithay/pull/2159/changes#r4005047070

PR #4495, opened 2026-08-28
feat: add layout option to maximize columns by default

MithicSpirit:
With `layout { default-column-width maximize=true }`, windows in the layout will spawn maximized, as if from the `open-maximized` window rule.  This behavior is overriden by said window rule. Note that `maximize=true` can be combined with the existing `default-column-width` behavior to also set the default width when un-maximizing.


Sempyos:
Could you add the configuration options you add to the description of the PR..?

PR #4497, opened 2026-08-28
feat: add ignore-client-size window rule

MithicSpirit:
Allows ignoring the client-provided minimum and maximum sizes via `window-rule { ignore-client-size true }`.

### DIFF
--- a/niri-config/src/lib.rs
+++ b/niri-config/src/lib.rs
@@ -1804,6 +1804,7 @@ mod tests {
                     min_height: None,
                     max_width: None,
                     max_height: None,
+                    ignore_client_size: None,
                     focus_ring: BorderRule {
                         off: true,
                         on: false,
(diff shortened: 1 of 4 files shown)


Sempyos:
Could you add an example of the configuration options you add to the description of the PR..?

YaLTeR:
Why would we want to do that? In any case, I'm not a fan of adding features just to work around buggy clients in niri. At most it can be a debug flag, but even then it has to be well justified (although, we don't have debug window rules.. idk). Also, you can probably do it outside niri through wl-proxy?

MithicSpirit:
Yeah, I figured this one wasn't very likely to be upstreamed.  This was added because I personally find that clients often have overly restrictive minimum sizes, and they work just fine when resized beyond that.  However, this is not very robust, so I think it makes sense to stay as an "opt-in patch".

I wasn't aware of wl-proxy as a possibility for doing this (I suppose I could just filter out the part where the client sets their size), but I guess that makes sense.

PR #4521, opened 2026-09-03
Fix: handle axis updates in tablet tip events

### LINKED ISSUE #2527: Glitched drawing tablet input adds fish hooks to stylus strokes
Drawing tablet stylus input has these weird fish hooks that come out of the strokes that I never actually draw. I can notice this issue both in Wayland native apps (Xournal++, Blender) and in XWayland apps (Krita, WINE apps). I checked in GNOME and KDE's Wayland sessions, and the hook problem wasn't there. Hyprland also exhibits this problem somewhat.

[Niri config, via niri-flake](https://bpa.st/IG4Q)

Image

###### Example with Xournal++, running natively on Wayland in Niri. Hooks are most noticable on the top center strokes.

Image

### System Information

* niri version: `niri stable v25.08 (commit 01be0e65f4eb91a9cd624ac0b76aaeab765c7294)`
* Distro: NixOS unstable (25.11 Xantusia)
* GPU: NVIDIA GeForce MX130 (via proprietary driver) / Intel UHD Graphics 620
* CPU: Intel Core i7-8550U
* Tablet: Wacom Intuos Art

DeSpecTDr:
Fixes #2527

LLM usage: llm helped to find the bug and review the fix, human (me) wrote the code, beep-boop

Basically niri ignores axis information provided by libinput along with tip up/down event, that's why there is a visible jump

For reference, sway (wlroots) handles this correctly
https://github.com/external-mirrors/wlroots/blob/2cec06a4617cd9c712c6b35831cd1d75c41fe2af/backend/libinput/tablet_tool.c#L274-L298

Fixed, tested with Wacom Intuos Pro M in Xournal++, cargo {fmt, clippy, test} pass

image


DeSpecTDr:
Okay, so i decided to compare behavior by replaying hid inputs in windows vm and on niri with this patch. 
Windows 11 with official wacom driver:
image
Niri (with libinput AttrPressureRange=409:270):
image

(without AttrPressureRange thin lines at the end (end is at the bottom usually) are longer, because by default libinput emits up event at 1% pressure, while on windows it's somewhere around 3%, so i changed it for comparison)

Also these thin end lines look ugly on purpose, xournal++ uses min pressure 0.05 by default, while here it was changed to 0.01

Everything looks rather close to me (i couldn't set normal resolution for a vm, so the aspect ratio is different). It turns out wlroots is wrong on UpEvents. Gnome has the correct implementation: https://github.com/GNOME/mutter/blob/0da9a352a8a3069f9f5738478ed636bafba6beee/src/backends/native/meta-seat-impl.c#L2815-L2837

Also moved tool.axis(self, frame) after tool.motion(...), because otherwise it doesn't have the correct focused window when changing axis and switching windows at the same time.

That "if TabletToolTipState::Down" is outside the match on purpose, so everything has the correct time.

I'm almost sure everything is correct now. I could add some comments to the code explaining everything if needed, but i tried to follow the code style around it (aka "an endless blob of code")

YaLTeR:
Thanks, looks reasonable.

> Also moved tool.axis(self, frame) after tool.motion(...), because otherwise it doesn't have the correct focused window when changing axis and switching windows at the same time.

Seems like a deficiency in Smithay's tablet code that should ideally be fixed there. For now the niri change is fine.

> I could add some comments to the code explaining everything if needed, but i tried to follow the code style around it (aka "an endless blob of code")

I mean, we do have comments for non-obvious things. I added two

bigbitbox:
Tested the merged commit `d18b201` on an ONEXPLAYER X1Air with a Dell PN7522W pen / built-in NVTK0603 digitizer (NixOS, Linux 6.18.44, native Wayland Xournal++ 1.3.6).

The repeatable down-right hooks on pen-up that I had with Niri 26.04 are now gone after restarting into this build:
```text
niri 26.04 (d18b201)
```
This fixes the issue on my hardware. Thanks!

PR #4522, opened 2026-09-03
updated dependencies

ghost:
updated project dependencies


ghost:
i didn't change because is a project  configuration but why do you guys have on release builds overflows check  to panic instead of wrap?

YaLTeR:
So that we can find and fix overflow bugs. Also thanks for the PR but I prefer to update dependencies myself at specific points in time

ghost:
> So that we can find and fix overflow bugs. Also thanks for the PR but I prefer to update dependencies myself at specific points in time

but in release? that could cause problems to users and reduce optimizations a bit. its more usual to see in debug

YaLTeR:
If it caused problems for you then I want to know about it, this is the entire point

PR #4523, opened 2026-09-03
animation: optimize clamped spring duration

ghost:
Replace repeated transcendental evaluations in the underdamped spring duration sampler with the equivalent fixed-step recurrence. Critical and overdamped springs retain direct evaluation to avoid numerical instability.

Add coverage comparing the optimized path with direct millisecond sampling across damping ratios, stiffnesses, velocities, and both animation directions.

Tests:
- `cargo test animation::spring::tests --lib`
- `cargo clippy --lib --tests -- -D warnings`

### DIFF
--- a/src/animation/spring.rs
+++ b/src/animation/spring.rs
@@ -124,10 +124,11 @@ impl Spring {
         // The first frame is not that important and we avoid finding the trivial 0 for in-place
         // animations.
         let mut i = 1u16;
-        let mut y = self.oscillate(SAMPLE_INTERVAL);
-
         let omega0 = (self.params.stiffness / self.params.mass).sqrt();
-        if beta < omega0 && (beta - omega0).abs() > f64::from(f32::EPSILON) {
+        let x0 = self.from - self.to;
+        let critical = (beta - omega0).abs() <= f64::from(f32::EPSILON);
+
+        if beta < omega0 && !critical {
             // An underdamped spring sampled at a fixed interval satisfies the recurrence
             //
             // x[n] = 2e^(-beta h)cos(omega h)x[n-1] - e^(-2 beta h)x[n-2].
@@ -135,16 +136,21 @@ impl Spring {
             // Use it here to avoid evaluating exp(), sin(), and cos() for every millisecond.
             let omega = ((omega0 * omega0) - (beta * beta)).sqrt();
             let decay = (-beta * SAMPLE_INTERVAL).exp();
-            let current_coefficient = 2. * decay * (omega * SAMPLE_INTERVAL).cos();
+            let angle = omega * SAMPLE_INTERVAL;
+            let cos = angle.cos();
+            let sin = angle.sin();
+            let current_coefficient = 2. * decay * cos;
             let previous_coefficient = -(decay * decay);
-            let mut previous_displacement = self.from - self.to;
+            let mut previous_displacement = x0;
+            let y =
+                self.to + decay * (x0 * cos + ((beta * x0 + self.initial_velocity) / omega) * sin);
             let mut displacement = y - self.to;
 
             while (distance > f64::EPSILON && -displacement > self.params.epsilon)
                 || (-distance > f64::EPSILON && displacement > self.params.epsilon)
             {
                 if i > 3000 {
-                    return None;
+                    break;
                 }
 
                 i += 1;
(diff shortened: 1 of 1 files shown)


YaLTeR:
> optimize

Do you mind showing benchmarks?

SomeFlyingThing:
i benchmarkerd it with hyprfine
  This workload cycles through the default underdamped configurations (damping-ratio=0.6, stiffness 500/1000) with different positions and velocities:
  
  Benchmark 1: before (500000 calls)
     Time (mean ± σ):     587.0 ms ± 0.5 ms
     Range (min … max):   586.2 ms … 587.7 ms    8 runs
  
   Benchmark 2: after (500000 calls)
     Time (mean ± σ):      72.2 ms ± 0.1 ms
     Range (min … max):    72.1 ms … 72.4 ms     8 runs
  
   Summary
     after ran 8.13 ± 0.02 times faster than before

HigherOrderLogic:
Seems like you forgot to delete a part of the clanker's answer.

ghost:
Comparison:

- Original: `upstream/main`
- Commit 1: `0ddfaa37`
- Commit 2: `0486fc3c`

All three implementations were benchmarked in the same binary and measurement session. Return-value checksums matched for these workloads.

### Underdamped defaults — 500,000 calls

```text
Original
  Time (mean ± σ):     573.6 ms ± 0.6 ms
  Range (min … max):   572.8 ms … 574.7 ms    8 runs

Commit 1
  Time (mean ± σ):      72.2 ms ± 0.1 ms
  Range (min … max):    72.1 ms … 72.3 ms     8 runs

Commit 2
  Time (mean ± σ):      85.5 ms ± 0.1 ms
  Range (min … max):    85.4 ms … 85.7 ms     8 runs
```

- Commit 1: **7.95× faster** than original
- Commit 2: **6.71× faster** than original
- Commit 1 is 1.18× faster than Commit 2 here

### Critical defaults — 500,000 calls

```text
Original
  Time (mean ± σ):     805.6 ms ± 2.4 ms
  Range (min … max):   803.1 ms … 808.6 ms    8 runs

Commit 1
  Time (mean ± σ):     786.9 ms ± 4.9 ms
  Range (min … max):   782.5 ms … 794.7 ms    8 runs

Commit 2
  Time (mean ± σ):     224.6 ms ± 0.7 ms
  Range (min … max):   223.9 ms … 225.5 ms    8 runs
```

- Commit 1: **1.02× faster** than original
- Commit 2: **3.59× faster** than original
- Commit 2 is 3.50× faster than Commit 1 here

### Mixed defaults — 300,000 calls

This contains six critical and two underdamped configurations.

```text
Original
  Time (mean ± σ):     459.9 ms ± 2.4 ms
  Range (min … max):   457.3 ms … 466.2 ms    10 runs

Commit 1
  Time (mean ± σ):     374.2 ms ± 0.9 ms
  Range (min … max):   372.6 ms … 375.4 ms    10 runs

Commit 2
  Time (mean ± σ):     116.2 ms ± 0.4 ms
  Range (min … max):   115.7 ms … 117.0 ms    10 runs
```

- Commit 1: **1.23× faster** than original
- Commit 2: **3.96× faster** than original
- Commit 2 is 3.22× faster than Commit 1

Commit 1 is fastest for underdamped springs, but accumulated recurrence rounding produced a one-millisecond difference in a generated edge case. Commit 2 adds direct boundary refinement and optimizes critical damping. It matched the original result across one million generated underdamped and critical inputs, while providing the best overall performance.

cornservant:
Setup up a benchmark with criterion. It generates random parameters per benchmark (but uses the same parameters to compare each algorithm ofc) and you can fiddle with the random distribution of those parameters.

https://github.com/cornservant/optimized-clamped-duration-bench

I got 16.0ms / 9.40ms / 9.43ms for baseline / commit1/ commit2. Do mind that I have no idea what the realistic values are for these params 😅

cornservant:
Can you also describe parameters that causes the greatest speedup compared to baseline?

ghost:
> Can you also describe parameters that causes the greatest speedup compared to baseline?

The largest speedup should be for underdamped springs that require a relatively large number of 1 ms samples before reaching the target.

In particular, that tends to mean a damping ratio below but fairly close to 1, lower stiffness, a small epsilon relative to the displacement, and/or an initial velocity pointing away from the target.

Those cases favor the new implementation because the baseline evaluates exp/sin/cos through oscillate() on every sample, while the optimized underdamped path pays for those once and then advances with the recurrence. The boundary refinement only adds a few direct evaluations at the end.

So the speedup generally grows with the number of samples, and underdamped cases benefit more than critical ones because the baseline’s per-sample calculation is more expensive. Overdamped springs should see little or no improvement since they still use the direct-evaluation path.

For the Niri-like defaults I benchmarked, the underdamped set was 6.71× faster, compared with 3.59× for the critical set.

cornservant:
Right, with exponentially distributed variables and expected values

- E(damping ratio) = 0.1
- E(stiffness) = 1.0
- E(epsilon) = 0.0001
- E(initial velocity) = 1.0
- from = 0.0
- E(to) = 1.0

I get 10.8ms / 1.23ms / 1.33ms (baseline / commit1 / commit2) evaluating 1000 samples on my machine.

ghost:
> Right, with exponentially distributed variables and expected values
> 
> 
> 
> - E(damping ratio) = 0.1
> 
> - E(stiffness) = 1.0
> 
> - E(epsilon) = 0.0001
> 
> - E(initial velocity) = 1.0
> 
> - from = 0.0
> 
> - E(to) = 1.0
> 
> 
> 
> I get 10.8ms / 1.23ms / 1.33ms (baseline / commit1 / commit2) evaluating 1000 samples on my machine.

In commit 1 recurrence calculations accumulate tiny floating-point rounding errors.Thats fixed in commit 2

PR #4544, opened 2026-09-07
feat: have concurrent holds timed separately

NSPC911:
fixes the `FIXME` comment in `src/input/mod.rs`:

original problem i faced was if im holding mod+shift+down (to move the window down) and mod+shift+right (to move the window right), it moves it down once, then moves it to the right while held down, even though i wanted it to move diagonally.

possibly unrelated but unavoidable change in `src/tests/mod.rs`: couldnt run tests without it

also video

https://github.com/user-attachments/assets/4480b017-6f40-4f2f-a981-aa113b0edc26

### DIFF
--- a/src/input/mod.rs
+++ b/src/input/mod.rs
@@ -420,15 +420,8 @@ impl State {
         let time = Event::time_msec(&event);
         let pressed = event.state() == KeyState::Pressed;
 
-        // Stop bind key repeat on any release. This won't work 100% correctly in cases like:
-        // 1. Press Mod
-        // 2. Press Left (repeat starts)
-        // 3. Press PgDown (new repeat starts)
-        // 4. Release Left (PgDown repeat stops)
-        // But it's good enough for now.
-        // FIXME: handle this properly.
         if !pressed {
-            if let Some(token) = self.niri.bind_repeat_timer.take() {
+            if let Some(token) = self.niri.bind_repeat_timers.remove(&event.key_code()) {
                 self.niri.event_loop.remove(token);
             }
         }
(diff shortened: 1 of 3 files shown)


src/input/mod.rs
```diff
@@ -476,6 +469,25 @@ impl State {
                 let raw = keysym.raw_latin_sym_or_raw_current_sym();
                 let modifiers = modifiers_from_state(*mods);
 
+                // Existing repeats no longer match when the modifier state changes.
+                if matches!(
```
Kruemmelspalter: why not use `modified.is_modifier_key()` ([`Keysym::is_modifier_key`](https://docs.rs/smithay/latest/smithay/input/keyboard/struct.Keysym.html#method.is_modifier_key)) instead?

NSPC911: What the heck, I didn't know that was a thing, thanks!

src/input/mod.rs
```diff
@@ -5275,6 +5274,33 @@ mod tests {
 
     use super::*;
     use crate::animation::Clock;
+    use crate::tests::Fixture;
+
+    #[test]
+    fn bind_repeats_are_tracked_per_key() {
+        let mut fixture = Fixture::new();
```
YaLTeR: Fixture is for client-server tests. If you want to use it put the test there. But I'm not sure it's needed for this test, maybe can be done in a lighter weight way?

src/input/mod.rs
```diff
@@ -5275,6 +5274,33 @@ mod tests {
...
+            cooldown: None,
+            allow_when_locked: false,
+            allow_inhibiting: false,
+            hotkey_overlay_title: None,
+        };
+        let down = Keycode::from(Keysym::Down.raw() + 8);
+        let right = Keycode::from(Keysym::Right.raw() + 8);
+
+        state.start_key_repeat(down, bind(Keysym::Down));
+        state.start_key_repeat(right, bind(Keysym::Right));
```
YaLTeR: I want to see tests that do two repeats at once and verify not only that they start, but also that they stop when expected. And also that releasing modifiers stops the repeats as expected.

src/input/mod.rs
NSPC911: Can't find a lighter way, I'll move it to `src/tests/input.rs`, unless you have a suggestion

PR #4550, opened 2026-09-08
Clean up renderer caches when no frames are rendered

RiversJin:
Clients can import and destroy linux-dmabuf buffers without rendering a frame. With monitors powered off, dead imports remain in the renderer cache, retaining their backing memory until rendering resumes. This reproduces the accumulation-and-release pattern reported in #3295.

Run `Renderer::cleanup_texture_cache()` from the existing one-second fallback timer. The TTY backend visits all GPU renderers, continuing if one fails; winit and headless use their primary renderer. Cleanup retains live resources and does not change frame callback policy.

Two EGL regression tests exercise the real timer with no outputs and with monitors off. They check deferred texture deletion, preservation of live textures, and cleanup on later ticks. Both fail without the fix and pass with it. A standalone linux-dmabuf client and reproduction instructions cover the cached EGLImage path separately.

Validation:

- `cargo test --locked --all --exclude niri-visual-tests`: passed (197 compositor tests, plus config, IPC, and doc tests).
- Nightly rustfmt check, Clippy with `-D warnings` (excluding niri-visual-tests), and `cargo check --locked --no-default-features`: passed.
- RX 7900 XTX, nested winit backend, upstream `dd75865f` versus this patch: 128 destroyed 4 MiB imports left GPU allocation at 581 MiB from a 69 MiB baseline without the fix. With the fix it returned to 69 MiB while outputs remained off. A live 128 MiB batch survived three cleanup ticks and was reclaimed after destruction.
- The same fix backported to the installed version was also tested on the DRM backend, including a 4 GiB import run and physical hotplug events. Exact-upstream runtime A/B above used winit; multi-GPU hardware was not tested.

Prepared with AI assistance; commands and runtime checks were executed locally.

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -5,6 +5,7 @@ use std::time::Duration;
 use niri_config::{Config, ModKey};
 use smithay::backend::allocator::dmabuf::Dmabuf;
 use smithay::backend::renderer::gles::GlesRenderer;
+use smithay::backend::renderer::Renderer;
 use smithay::output::Output;
 use smithay::reexports::wayland_server::protocol::wl_surface::WlSurface;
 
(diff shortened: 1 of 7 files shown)


AlexAllocated:
Thanks for working on this! I hit what looks like the same issue today on NixOS with an RTX 5080 (NVIDIA 595.71.05), using a Niri 26.04-based build with virtual-output support.

After leaving my LG monitor off overnight, turning it back on gave me a black screen. Niri was holding about 10.5 GB of VRAM, and the logs showed NVKMS buffer allocation failures. Temporarily switching to 1080p got it rendering again and dropped Niri to about 135 MB; I could then restore 3440x1440 at 160 Hz with VRR without restarting the session.

I have not tested this patch yet, so this is another matching report rather than confirmation of the fix. Very glad to see someone tackling it!

trevarj:
I'm also on NixOS and just started seeing this after bumping flake versions. Niri didn't change, but some deps did for it so I'm guessing it comes from something there.

Hope this works and gets merged cause it causes my docked laptop to just die when left on and idling

YaLTeR:
If I'm reading correctly, neither of you actually tested this PR? Then why are you asking for it to be merged?..

> I have not tested this patch yet, so this is another matching report rather than confirmation of the fix.

> The proposed renderer-cache cleanup in https://github.com/niri-wm/niri/pull/4550 looks relevant but is not tested here.

Also, the patch likely needs to look something different, but it would be good to get some confirmation that it works to know that it has the right idea.

AlexAllocated:
Please show me where I asked for it to be merged. All I did was provide additional information.

However, your sass aside, I have since pulled this patch into my install but I won't be able to report back until I leave the monitor off all night again.

PR #4548, opened 2026-09-08
feat: use new virtual keyboard/pointer implementations

pgaskin:
Fixes a bunch of virtual input bugs, most notably keybind handling and virtual keyboard keymaps.

See Smithay/smithay#2159 for more details.

needs niri-wm/niri#4552
needs Smithay/smithay#2159 (DO NOT MERGE UNTIL AFTER THIS)
closes #3309
fixes #3394
fixes #403
closes #2246
fixes #3595
fixes #3274

PR #4552, opened 2026-09-08
Update Smithay

pgaskin:
Mostly minor bugfixes and refactors.

Split from #4548 so it's easier to squash later.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,7 +66,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
 dependencies = [
  "cfg-if",
- "getrandom 0.3.4",
  "once_cell",
  "version_check",
  "zerocopy",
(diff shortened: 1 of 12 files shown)


HigherOrderLogic:
Seems like Smithay's toplevel protocol impl now can set custom identifier (Smithay/smithay#2143), can you try refactoring Niri's custom impl to use it too?

pgaskin:
@HigherOrderLogic done

YaLTeR:
Thanks. I went through this and merged it with some changes in 6a9ed5bc7c7afc53ae50773b49e73e7ffa6855d5.

> Seems like Smithay's toplevel protocol impl now can set custom identifier ([Smithay/smithay#2143](https://github.com/Smithay/smithay/pull/2143)), can you try refactoring Niri's custom impl to use it too?

But why? The whole point was that we already have the wlr-foreign-toplevel-management code, and ext-foreign-toplevel-list mirrors it closely, so it's easy to keep the two in sync.

I reverted this change from the merged commit. If needed can be reconsidered in the future.

HigherOrderLogic:
> But why?

This would help with implementing the ext-image-copy protocol for toplevel, which uses the `ForeignToplevelHandle` struct from Smithay.

YaLTeR:
Oh, huh. Idk, is there another way maybe? If not then I guess we could do it

HigherOrderLogic:
Well yes, basically you need to copy the whole top level manager block over and adapt to the code. Seems to work but kinda unnecessary if we only want to keep the custom impl imo.

PR #4553, opened 2026-09-08
render_to_shm: follow the buffer's stride and pool offset

pgaskin:
Don't assume we have a single tightly packed buffer of `stride=width*4` at the beginning of the pool.

Clients can pad rows, and have multiple buffers in a pool.

This PR fixes this by properly accounting for the offset and copying one row at a time.

Without this, screencopy clients (and ext-image-copy-capture in the near future) may break with "invalid buffer format or size".

I've been using this since mid-June as part of #4554.

closes #4519 (which is an incomplete version of this)

### DIFF
--- a/src/render_helpers/mod.rs
+++ b/src/render_helpers/mod.rs
@@ -305,7 +305,9 @@ pub fn render_to_shm(
     states: RenderElementStates,
 ) -> anyhow::Result<()> {
     let _span = tracy_client::span!();
-    shm::with_buffer_contents_mut(buffer, |shm_buffer, shm_len, buffer_data| {
+    // The pointer and length are the client's entire pool, which may hold other
+    // buffers besides this one... buffer_data is the one we want.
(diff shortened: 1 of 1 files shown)


pgaskin:
@Sempyos this is technically a bugfix, I think.

Sempyos:
Alright

YaLTeR:
Thanks

iciclejj:
This is probably a more appropriate fix than mine for the protocol-agnostic `render_to_shm()` helper, but this MR doesn't include the same `with_buffer_contents()` fix in the wlr_screencopy's buffer-size verification at https://github.com/niri-wm/niri/blob/a787723fdb6c3a175fef23c572d44293bf08179e/src/protocols/screencopy.rs#L553-L558, so the screencopies (the only current user of this helper) will still fail when clients "have multiple buffers in a pool" or the buffer doesn't occupy the entire pool.

YaLTeR:
Do you mind opening a PR?

iciclejj:
> Do you mind opening a PR?

Sure!

PR #4554, opened 2026-09-08
feat: implement ext-image-copy-capture for outputs and cursors

### LINKED ISSUE #1558: Implement ext-image-copy-capture

pgaskin:
Implement ext-image-copy-capture with output/cursor capture sources, and shm/dma buffers.

I've been using this (along with #4548) daily since mid-June, with RealVNC via [`xwlrvnc`](https://github.com/pgaskin/xwlrvnc), and wayvnc.

I've also tested it with grim, lamco-rdp-server, and briefly with wl-screencopy.

I've tested various output layouts and DPI settings, on both real TTYs and winit.

Compared to #3942, this one doesn't implement window capture, but it implements cursor capture, correctly handles sandboxing (i.e., `client_is_unrestricted`), properly handles BufferConstraints (the other one breaks with many actual clients), and properly cleans up stale sessions.

In addition, I've optimized it to not re-render more than necessary, even with multiple capture sessions.

includes #4553 (merge that one first)
partially replaces #3942
closes #1558

### DIFF
--- a/src/handlers/image_copy_capture.rs
+++ b/src/handlers/image_copy_capture.rs
@@ -1,0 +1,439 @@
+use std::time::Duration;
+
+use smithay::backend::allocator::dmabuf::Dmabuf;
+use smithay::backend::allocator::{Buffer as _, Fourcc, Modifier};
+use smithay::backend::drm::DrmNode;
+use smithay::backend::renderer::damage::OutputDamageTracker;
+use smithay::backend::renderer::gles::GlesRenderer;
+use smithay::backend::renderer::sync::SyncPoint;
+use smithay::output::{Output, WeakOutput};
+use smithay::reexports::calloop::generic::Generic;
+use smithay::reexports::calloop::{Interest, LoopHandle, Mode, PostAction};
+use smithay::reexports::wayland_server::protocol::wl_buffer::WlBuffer;
+use smithay::reexports::wayland_server::protocol::wl_pointer::WlPointer;
+use smithay::reexports::wayland_server::protocol::wl_shm;
+use smithay::utils::{
+    Buffer as BufferCoords, Logical, Physical, Point, Rectangle, Scale, Size, Transform,
+};
+use smithay::wayland::dmabuf::get_dmabuf;
+use smithay::wayland::image_capture_source::{
+    ImageCaptureSource, ImageCaptureSourceHandler, OutputCaptureSourceHandler,
+    OutputCaptureSourceState,
+};
+use smithay::wayland::image_copy_capture::{
+    BufferConstraints, CaptureFailureReason, CursorSession, CursorSessionRef, DmabufConstraints,
+    Frame, FrameRef, ImageCopyCaptureHandler, ImageCopyCaptureState, Session, SessionRef,
+};
+use smithay::wayland::shm;
+
+use crate::cursor::{RenderCursor, XCursor};
+use crate::niri::{Niri, State};
+use crate::utils::{CastSessionId, CastStreamId};
+
+/// Output capture session.
+pub struct ImageCopySession {
+    pub session: Session,
+    pub damage_tracker: OutputDamageTracker,
+    /// Frame waiting for output damage.
+    pub pending_frame: Option<Frame>,
+    /// Cast session id, shared with the cursor session of the same source.
+    pub session_id: CastSessionId,
+    /// Cast stream id, unique to this session.
+    pub stream_id: CastStreamId,
+}
+
+/// Cursor capture session of an output.
+pub struct ImageCopyCursorSession {
+    pub session: CursorSession,
+    /// Damage to the cursor image (i.e., not movement).
+    pub damage_tracker: OutputDamageTracker,
+    /// Frame waiting for cursor image change.
+    pub pending_frame: Option<Frame>,
+    /// Cast session id, shared with the output session of the same source.
+    pub session_id: CastSessionId,
+    /// Cast stream id, unique to this session.
+    pub stream_id: CastStreamId,
+}
+
+/// Cast session id of an image capture source.
+///
+/// Stored on the source so that the output session and the corresponding cursor
+/// session are reported as two streams of a single cast session.
+fn source_session_id(source: &ImageCaptureSource) -> CastSessionId {
+    source.user_data().insert_if_missing(CastSessionId::next);
+    *source.user_data().get::<CastSessionId>().unwrap()
+}
+
+/// Output captured by a session, or `None` if it's gone (sessions can outlive
+/// their output).
+pub fn source_output(source: &ImageCaptureSource) -> Option<Output> {
+    source.user_data().get::<WeakOutput>()?.upgrade()
+}
+
+/// Buffer constraints for capturing an output.
+///
+/// `render_node` is the primary renderer's DRM render node (see
+/// `Backend::primary_render_node()`), and only shm will be supported without
+/// one. It is required since querying it from the EGL context fails for the TTY
+/// backend: since Mesa 23.3 and until at least 26.1.8, `_eglGetGbmDisplay()`
+/// clears the display's EGLDevice once a second EGLDisplay is created for the
+/// same GBM device, which the TTY backend does during initialization. See
+/// https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/44351.
+pub fn output_capture_constraints(
+    renderer: &GlesRenderer,
+    render_node: Option<DrmNode>,
+    output: &Output,
+) -> Option<BufferConstraints> {
+    let mode = output.current_mode()?;
+    let size = Size::<i32, BufferCoords>::from((mode.size.w, mode.size.h));
+
+    let dma = (|| {
+        let node = render_node?;
+        let egl = renderer.egl_context();
+
+        // Offer all formats the renderer can draw into to avoid unnecessary
+        // conversions, preserving the original order (many clients depend on
+        // the order being stable to select the same format when renegotiating).
+        let mut formats: Vec<(Fourcc, Vec<Modifier>)> = Vec::new();
+        for format in egl.dmabuf_render_formats().iter() {
+            match formats.iter_mut().find(|(code, _)| *code == format.code) {
+                Some((_, modifiers)) => modifiers.push(format.modifier),
+                None => formats.push((format.code, vec![format.modifier])),
+            }
+        }
+        if formats.is_empty() {
+            return None;
+        }
+
+        // Put Xrgb8888 and Argb8888 first since some clients always take the
+        // first advertised format (e.g. wl-mirror, grim).
+        formats.sort_by_key(|(code, _)| match code {
+            Fourcc::Xrgb8888 => 0,
+            Fourcc::Argb8888 => 1,
+            _ => 2,
+        });
+
+        Some(DmabufConstraints { node, formats })
+    })();
+
+    Some(BufferConstraints {
+        size,
+        shm: vec![wl_shm::Format::Xrgb8888],
+        dma,
+    })
+}
+
+/// Buffer constraints for capturing the cursor of an output. Argb8888 since it has alpha.
+pub fn cursor_capture_constraints(niri: &Niri, output: &Output) -> BufferConstraints {
+    BufferConstraints {
+        size: cursor_capture_size(niri, output),
+        shm: vec![wl_shm::Format::Argb8888],
+        dma: None,
+    }
+}
+
+/// Size the cursor renders at on this output.
+fn cursor_capture_size(niri: &Niri, output: &Output) -> Size<i32, BufferCoords> {
+    let int_scale = output.current_scale().integer_scale();
+    let scale = Scale::from(output.current_scale().fractional_scale());
+
+    let size: Size<i32, Physical> = match niri.cursor_manager.get_render_cursor(int_scale) {
+        RenderCursor::Hidden => Size::from((0, 0)),
+        RenderCursor::Surface { surface, .. } => {
+            let bbox = smithay::desktop::utils::bbox_from_surface_tree(&surface, (0, 0));
+            bbox.to_f64().to_physical_precise_up(scale).size
+        }
+        RenderCursor::Named {
+            scale: buffer_scale,
+            cursor,
+            ..
+        } => {
+            // All frames are the same size since CursorManager::load_xcursor()
+            // picks one size and rejects frames which differ.
+            let (_idx, frame) = cursor.frame(niri.start_time.elapsed().as_millis() as u32);
+            // The image is loaded at the integer scale but drawn at the fractional one, so it
+            // ends up smaller than its own buffer whenever the two differ.
+            let logical = Size::<f64, Logical>::from((
+                f64::from(frame.width) / f64::from(buffer_scale),
+                f64::from(frame.height) / f64::from(buffer_scale),
+            ));
+            logical.to_physical_precise_ceil(scale)
+        }
+    };
+
+    // Fall back to the nominal cursor size when the cursor is currently hidden or has no
+    // buffer, so that the session always has valid constraints.
+    if size.is_empty() {
+        let fallback = i32::from(niri.config.borrow().cursor.xcursor_size) * int_scale;
+        return Size::from((fallback, fallback));
+    }
+
+    Size::from((size.w, size.h))
+}
+
+/// Cursor hotspot in capture buffer coordinates.
+pub fn cursor_capture_hotspot(niri: &Niri, output: &Output) -> Point<i32, BufferCoords> {
+    let int_scale = output.current_scale().integer_scale();
+    let scale = Scale::from(output.current_scale().fractional_scale());
+
+    let hotspot: Point<i32, Physical> = match niri.cursor_manager.get_render_cursor(int_scale) {
+        RenderCursor::Hidden => Point::from((0, 0)),
+        RenderCursor::Surface { surface, hotspot } => {
+            // The tree is shifted to put its bounding box at the origin in
+            // render_cursor_for_capture(), so shift the hotspot too.
+            let bbox = smithay::desktop::utils::bbox_from_surface_tree(&surface, (0, 0));
+            (hotspot - bbox.loc)
+                .to_f64()
+                .to_physical_precise_round(scale)
+        }
+        RenderCursor::Named {
+            scale: buffer_scale,
+            cursor,
+            ..
+        } => {
+            let (_idx, frame) = cursor.frame(niri.start_time.elapsed().as_millis() as u32);
+            // Same rescaling as in cursor_capture_size().
+            XCursor::hotspot(frame)
+                .to_logical(buffer_scale)
+                .to_f64()
+                .to_physical_precise_round(scale)
+        }
+    };
+
+    Point::from((hotspot.x, hotspot.y))
+}
+
+pub enum CaptureBuffer {
+    Dma(Dmabuf),
+    Shm,
+}
+
+/// Checks a frame's buffer for compatibility with the render helpers.
+///
+/// Smithay only validates buffers against the protocol constraints, which allow
+/// a client to attach a larger buffer than asked for, but the render helpers
+/// need an exact match, and checking them here allows mismatches to be reported
+/// to the client.
+pub fn capture_buffer(
+    buffer: &WlBuffer,
+    size: Size<i32, BufferCoords>,
+    format: wl_shm::Format,
+) -> Option<CaptureBuffer> {
+    if let Ok(dmabuf) = get_dmabuf(buffer) {
+        let size_matches = dmabuf.width() == size.w as u32 && dmabuf.height() == size.h as u32;
+        return size_matches.then(|| CaptureBuffer::Dma(dmabuf.clone()));
+    }
+
+    // Stride and pool placement are defined by the client, but the format and
+    // dimensions need to match.
+    let matches = shm::with_buffer_contents(buffer, |_ptr, _len, data| {
+        data.format == format && data.width == size.w && data.height == size.h
+    })
+    .unwrap_or(false);
+    matches.then_some(CaptureBuffer::Shm)
+}
+
+/// Finishes a frame, waiting for `sync_point` first if the render is still in flight.
+///
+/// Like `Screencopy::submit_after_sync()`.
+pub fn frame_success_after_sync<T>(
+    frame: Frame,
+    transform: Transform,
+    damage: Vec<Rectangle<i32, BufferCoords>>,
+    presented: Duration,
+    sync_point: Option<SyncPoint>,
+    event_loop: &LoopHandle<'static, T>,
+) {
+    let Some(sync_fd) = sync_point.and_then(|sync| sync.export()) else {
+        frame.success(transform, damage, presented);
+        return;
+    };
+
+    let mut pending = Some((frame, damage));
+    let source = Generic::new(sync_fd, Interest::READ, Mode::OneShot);
+    let res = event_loop.insert_source(source, move |_, _, _| {
+        let (frame, damage) = pending.take().unwrap();
+        frame.success(transform, damage, presented);
+        Ok(PostAction::Remove)
+    });
+    if let Err(err) = res {
+        // The client will retry (smithay sends an Unknown error when Drop is
+        // called on the Frame).
+        warn!("error waiting for image copy capture sync point: {err}");
+    }
+}
+
+impl ImageCaptureSourceHandler for State {}
+
+impl OutputCaptureSourceHandler for State {
+    fn output_capture_source_state(&mut self) -> &mut OutputCaptureSourceState {
+        &mut self.niri.output_capture_source_state
+    }
+
+    fn output_source_created(&mut self, source: ImageCaptureSource, output: &Output) {
+        source.user_data().insert_if_missing(|| output.downgrade());
+    }
+}
+
+impl ImageCopyCaptureHandler for State {
+    fn image_copy_capture_state(&mut self) -> &mut ImageCopyCaptureState {
+        &mut self.niri.image_copy_capture_state
+    }
+
+    fn capture_constraints(&mut self, source: &ImageCaptureSource) -> Option<BufferConstraints> {
+        let output = source_output(source)?;
+        if !self.niri.output_state.contains_key(&output) {
+            return None;
+        }
+
+        let render_node = self.backend.primary_render_node();
+        self.backend
+            .with_primary_renderer(|renderer| {
+                output_capture_constraints(renderer, render_node, &output)
+            })
+            .flatten()
+    }
+
+    fn cursor_capture_constraints(
+        &mut self,
+        source: &ImageCaptureSource,
+        _pointer: &WlPointer,
+    ) -> Option<BufferConstraints> {
+        let output = source_output(source)?;
+        if !self.niri.output_state.contains_key(&output) {
+            return None;
+        }
+
+        Some(cursor_capture_constraints(&self.niri, &output))
+    }
+
+    fn new_session(&mut self, session: Session) {
+        // Will be updated with the output properties before capture.
+        let damage_tracker = OutputDamageTracker::new((0, 0), 1.0, Transform::Normal);
+        let session_id = source_session_id(&session.source());
+        self.niri.image_copy_sessions.push(ImageCopySession {
+            session,
+            damage_tracker,
+            pending_frame: None,
+            session_id,
+            stream_id: CastStreamId::next(),
+        });
+    }
+
+    fn new_cursor_session(&mut self, session: CursorSession) {
+        // Will be updated with the output properties before capture.
+        let damage_tracker = OutputDamageTracker::new((0, 0), 1.0, Transform::Normal);
+        let session_id = source_session_id(&session.source());
+        self.niri
+            .image_copy_cursor_sessions
+            .push(ImageCopyCursorSession {
+                session,
+                damage_tracker,
+                pending_frame: None,
+                session_id,
+                stream_id: CastStreamId::next(),
+            });
+        // Send the initial cursor position and hotspot.
+        self.niri.refresh_image_copy_cursor_sessions();
+    }
+
+    fn frame(&mut self, session: &SessionRef, frame: Frame) {
+        let Some(s) = self
+            .niri
+            .image_copy_sessions
+            .iter_mut()
+            .find(|s| s.session == *session)
+        else {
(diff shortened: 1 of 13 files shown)


src/niri.rs
```diff
@@ -841,6 +860,64 @@ impl State {
...
+                .backend
+                .with_primary_renderer(|renderer| {
+                    image_copy_capture_impl::output_capture_constraints(renderer, &output)
+                })
+                .flatten();
+            let Some(constraints) = constraints else {
+                return false;
+            };
+
+            if s.session.current_constraints().map(|c| c.size) != Some(constraints.size) {
```
pgaskin: I need to see if there's a more efficient way to do this.

src/niri.rs
```diff
@@ -5415,6 +5508,369 @@ impl Niri {
...
+                            // Wait for the rendering to finish before sending the frame.
+                            let mut frame = Some((frame, damage));
+                            let source = Generic::new(sync_fd, Interest::READ, Mode::OneShot);
+                            self.event_loop
+                                .insert_source(source, move |_, _, _| {
+                                    let (frame, damage) = frame.take().unwrap();
+                                    frame.success(transform, damage, timestamp);
+                                    Ok(PostAction::Remove)
+                                })
+                                .unwrap();
```
pgaskin: Might see if I can get rid of this unwrap.

src/niri.rs
```diff
@@ -841,6 +860,64 @@ impl State {
...
+                return false;
+            };
+
+            let constraints = self
+                .backend
+                .with_primary_renderer(|renderer| {
+                    image_copy_capture_impl::output_capture_constraints(renderer, &output)
+                })
+                .flatten();
+            let Some(constraints) = constraints else {
```
pgaskin, line 910: Should look into if there's any dmabuf-specific things I need to check here.

src/niri.rs
```diff
@@ -5415,6 +5508,369 @@ impl Niri {
...
+    pub fn refresh_image_copy_cursor_sessions(&mut self) {
+        if self.image_copy_cursor_sessions.is_empty() {
+            return;
+        }
+
+        let _span = tracy_client::span!("Niri::refresh_image_copy_cursor_sessions");
+
+        let pointer_pos = self
+            .tablet_cursor_location
+            .unwrap_or_else(|| self.seat.get_pointer().unwrap().current_location());
```
HigherOrderLogic, line 5839: Wont this use a different pointer compared to the original `wl_pointer` if a touch device is presented?

src/niri.rs
```diff
@@ -5415,6 +5508,369 @@ impl Niri {
...
+        let output_scale = Scale::from(output.current_scale().fractional_scale());
+
+        let mut elements = Vec::new();
+        match self.cursor_manager.get_render_cursor(int_scale) {
+            RenderCursor::Hidden => (),
+            RenderCursor::Surface { surface, .. } => {
+                push_elements_from_surface_tree(
+                    renderer,
+                    &surface,
+                    Point::from((0, 0)),
```
HigherOrderLogic: I think you need to account for `bbox.loc` here.

HigherOrderLogic (commented):
Some comments from a quick skim, havent tried it out yet.

src/niri.rs
pgaskin: Done.

src/niri.rs
pgaskin: Done.

src/niri.rs
pgaskin: Fixed, I think, but haven't tested it yet.

src/niri.rs
pgaskin: I've never tested niri with a touch device, but if I understand it correctly, niri only has one seat/pointer/cursor, so no?

HigherOrderLogic: I dont have a touch device either, but I imagine people can possibly use both the touchscreen and a mouse at the same time?

pgaskin: I don't think so, but I'm really not sure.

src/niri.rs
pgaskin: I think the size check I did is enough:

- The dmabuf constraints don't depend on the output, only the renderer.
- Render node is never changed after startup.
- Multi-GPU setups still render through the main render node.
- Mode changes are caught by the size check, and transforms/scaling doesn't change the buffer size.

pgaskin:
I think I'm happy with this now, and it should be ready to fully review and merge (after #4553).

I've also been using this (along with niri-wm/niri#4548) for multiple hours today with RealVNC on [xwlrvnc](https://github.com/pgaskin/xwlrvnc) (also with x11vnc on xwlrvnc), and I've also tested it with wayvnc, grim, wl-mirror, and wl-screenrec.

I've optimized the per-frame constraint check, which reduces the CPU usage a bit, and verified that it works properly (also see https://github.com/niri-wm/niri/pull/4554#discussion_r3975005838).

I also fixed some edge-cases with cursor handling near output boundaries (where the cursor image spans outputs).

I also tweaked the output capture format list to always sort `Xrgb8888` and `Argb8888` first since quite a few clients always pick (or at least prefer) the first one on the list (e.g., grim, wl-mirror).

I've tested most of the edge cases too, especially scaled/rotated outputs, hidden cursors, and nested niri compositors.

I also looked at how it works with touch/tablet input (see https://github.com/niri-wm/niri/pull/4554#discussion_r3975121516).

src/niri.rs
pgaskin: After looking at it more I'm pretty sure it doesn't. I also did a bit of testing with uhid.

The cursor session captures the actual drawn cursor, not the logical position of any specific wl_pointer (and there's only one).

If a touch device is in use, the cursor will be hidden correctly. When switching, the cursor is warped to the last position.

This only needs to be handled differently if the FIXME in `tablet_tool_image` is changed.

https://github.com/niri-wm/niri/blob/dd75865f547f0eac0e9b6c4d86d2cd00c0744252/src/handlers/mod.rs#L140-L145

HigherOrderLogic: Oh yeah then it's probably fine I think.

pgaskin:
Also tested this with xdg-desktop-portal-wlr (via PipeWire, with GStreamer as the client).

pgaskin:
Also, with the work I did to optimize this and xwlrvnc, RealVNC+xwlrvnc+niri is now nearly as CPU-efficient and responsive as RealVNC+X11+i3. It's also similar (though slightly less efficient) with x11vnc+xwlrvnc+niri.

pgaskin:
Added a workaround for a bug in Mesa causing dmabuf to not work for the TTY backend.

Mesa clears the EGLDevice in `_eglGetGbmDisplay` when a second EGLDisplay is created for the same GBM device, which is done in the TTY backend.

https://github.com/intel-lgci-fdo-gitlab-mirror/mesa.mesa/blob/0ba4b08edc65075e9346d20d5310261939aaaf48/src/egl/main/egldisplay.c#L568-L601

I now keep track of the render node for each backend instead of querying it from the EGL display after the fact.

Also reported it upstream in [mesa!44351](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/44351).

pgaskin:
Rebased.

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+/// Buffer constraints for capturing an output.
+///
+/// `render_node` is the primary renderer's DRM render node (see
+/// `Backend::primary_render_node()`), and only shm will be supported without
+/// one. It is required since querying it from the EGL context fails for the TTY
+/// backend since as of Mesa 26.1.8, _eglGetGbmDisplay clears the display's
+/// EGLDevice once a second EGLDisplay is created for the same GBM device, which
+/// the TTY backend does during initialization.
```
YaLTeR, line 57: Link your Mesa MR here please

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,367 @@
...
+                .or_default()
+                .push(format.modifier);
+        }
+        if by_code.is_empty() {
+            return None;
+        }
+
+        // Xrgb8888 and Argb8888 go first since some clients just take the first
+        // advertised format. The rest is sorted so the order is stable across
+        // constraint updates for clients that depend on it.
```
YaLTeR, line 84: Which clients depend on it and why? From my reading of the protocol it doesn't seem like clients are supposed to depend on the order

YaLTeR: Also, why would the order change in the first place? Does `egl.dmabuf_render_formats()` shuffle the same formats around between invocations?

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,367 @@
...
+        RenderCursor::Surface { surface, .. } => {
+            let bbox = smithay::desktop::utils::bbox_from_surface_tree(&surface, (0, 0));
+            bbox.to_f64().to_physical_precise_up(scale).size
+        }
+        RenderCursor::Named {
+            scale: buffer_scale,
+            cursor,
+            ..
+        } => {
+            let (_idx, frame) = cursor.frame(niri.start_time.elapsed().as_millis() as u32);
```
YaLTeR, line 133: Maybe this should max across all frames?

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,367 @@
...
+            frame.fail(CaptureFailureReason::Unknown);
+            return;
+        };
+
+        // A session may only have one frame object in flight at a time; a
+        // second one is a duplicate_frame protocol error. Smithay raises that
+        // itself, but still check anyways since to avoid leaking frames if it
+        // doesn't.
+        if s.pending_frame.is_some() {
+            frame.fail(CaptureFailureReason::Unknown);
```
YaLTeR, line 300: Also add an `error!()` here since it's a programming error if this happens

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,367 @@
...
+            // The output may be gone already (the global lingwrs).
+            if self.niri.output_exists(&output) {
+                self.niri.queue_redraw(&output);
+            }
+        }
+    }
+
+    fn frame_aborted(&mut self, frame: FrameRef) {
+        for s in &mut self.niri.image_copy_sessions {
+            if s.pending_frame.as_ref().is_some_and(|f| *f == frame) {
```
YaLTeR, line 343: I usually do something like `s.pending_frame == Some(frame)`, if this works here

src/render_helpers/mod.rs
```diff
@@ -303,17 +303,39 @@ pub fn render_to_shm(
...
+    )
+}
+
+pub fn render_to_shm_with_format(
+    renderer: &mut GlesRenderer,
+    damage_tracker: &mut OutputDamageTracker,
+    buffer: &WlBuffer,
+    format: wl_shm::Format,
+    elements: &[impl RenderElement<GlesRenderer>],
+    states: RenderElementStates,
```
YaLTeR, line 323: I'm kinda thinking since there's only one other user of `render_to_shm()`, to just add the format argument to it and pass it explicitly there, without an extra function.

src/niri.rs
```diff
@@ -841,6 +860,79 @@ impl State {
...
+        let mut sessions = mem::take(&mut self.niri.image_copy_sessions);
+        sessions.retain_mut(|s| {
+            let Some(output) = live_output(&self.niri, &s.session.source()) else {
+                return false;
+            };
+
+            // This runs once per event loop iteration, and building the full
+            // constraints is expensive. We only compare the size, so a change
+            // in the dmabuf formats or device alone doesn't matter to us here
+            // (and if it does, polling is not the way to check for it).
```
YaLTeR, line 890: We have `output_resized()` btw, maybe can use that to avoid checking here?

src/niri.rs
```diff
@@ -841,6 +860,79 @@ impl State {
...
+            let Some(mode) = output.current_mode() else {
+                return false;
+            };
+            let size = Size::<i32, BufferCoords>::from((mode.size.w, mode.size.h));
+            if s.session.current_constraints().map(|c| c.size) == Some(size) {
+                return true;
+            }
+
```
YaLTeR, line 898: Add comment here saying that now we send the new constraints, bc it was slightly unobvious what's about to happen

src/niri.rs
```diff
@@ -5410,6 +5518,385 @@ impl Niri {
...
+    ) {
+        if self
+            .image_copy_sessions
+            .iter()
+            .all(|s| s.pending_frame.is_none())
+        {
+            return;
+        }
+
+        let _span = tracy_client::span!("Niri::render_for_image_copy_capture");
```
YaLTeR, line 5535: Probably can go below the mode check too

src/niri.rs
```diff
@@ -5410,6 +5518,385 @@ impl Niri {
...
+        output: &Output,
+        target_presentation_time: Duration,
+    ) {
+        if self
+            .image_copy_sessions
+            .iter()
+            .all(|s| s.pending_frame.is_none())
+        {
+            return;
+        }
```
YaLTeR, line 5533: Something tells me that this loop here isn't really saving any performance

src/niri.rs
```diff
@@ -5410,6 +5518,385 @@ impl Niri {
...
+                                Ok(PostAction::Remove)
+                            });
+                            if let Err(err) = res {
+                                // The client will retry (smithay sends an
+                                // Unknown error when Drop is called on the
+                                // Frame).
+                                warn!("error waiting for image copy capture sync point: {err}");
+                            }
+                        }
+                    }
```
YaLTeR, line 5653: Would be good to extract this (and possibly some other stuff) into functions, like there's `submit_after_sync` in wlr-screencopy

src/niri.rs
```diff
@@ -5410,6 +5518,385 @@ impl Niri {
...
+                Err(err) => {
+                    warn!("error rendering for image copy capture: {err:?}");
+                    frame.fail(CaptureFailureReason::Unknown);
+                    // Report full damage next time.
+                    s.damage_tracker = OutputDamageTracker::new(size, scale, transform);
+                }
+            }
+        }
+
+        sessions.append(&mut self.image_copy_sessions);
```
YaLTeR, line 5664: What's this for?

src/niri.rs
```diff
@@ -5410,6 +5518,385 @@ impl Niri {
...
+                renderer,
+                &mut s.damage_tracker,
+                &buffer,
+                wl_shm::Format::Argb8888,
+                elements,
+                states,
+            );
+            match res {
+                Ok(()) => {
+                    let full_damage = vec![Rectangle::from_size(constraints.size)];
```
YaLTeR, line 5756: Always full?

src/niri.rs
```diff
@@ -5410,6 +5518,385 @@ impl Niri {
...
+            let pos: Point<i32, Physical> =
+                (pointer_pos - geo.loc.to_f64()).to_physical_precise_round(scale);
+
+            // The cursor has entered while its image intersects the output, not only while the
+            // hotspot is inside it, so the position can be negative or past the edge.
+            let hotspot = Point::<i32, Physical>::from((hotspot.x, hotspot.y));
+            let image = Rectangle::new(pos - hotspot, Size::from((cursor_size.w, cursor_size.h)));
+            let output_rect =
+                Rectangle::from_size(output.current_transform().transform_size(mode.size));
+            if self.pointer_visibility.is_visible() && image.overlaps(output_rect) {
```
YaLTeR, line 5890: For PW I went with if pointer location is inside the output rect. Let's do the same here for consistency. It means no partial cursor but I think that's fine. It matters more for window casts where you don't want to leak the pointer image unnecessarily.

YaLTeR (commented):
Looks fairly good.

Please add this to the cast IPC. Search `src/ipc/server.rs` for `niri_ipc::Cast`.

Some comments also apply to the second instance of similar code to where I left the comment.

src/handlers/image_copy_capture.rs
pgaskin: `grim` and `wl-mirror` are two examples which directly depend on the order. In addition, ones which capture through xdg-desktop-portal-wlr pipewire usually depend at least partially on the order (the first acceptable one is picked) and will select it inconsistently every time they renegotiate if it changes.

src/niri.rs
pgaskin: The protocol does explicitly say:

> cursor entered captured area
>
> Sent when a cursor enters the captured area. It shall be generated before the "position" and "hotspot" events when and only when a cursor enters the area.
>
> The cursor enters the captured area when the cursor image intersects with the captured area. Note, this is different from e.g. wl_pointer.enter.

src/handlers/image_copy_capture.rs
pgaskin: And yeah, the protocol doesn't guarantee any order, but it doesn't say anything about stability or how clients should choose, just that they may choose any.

> The client may choose any of the modifiers advertised in the array of 64-bit unsigned integers.
>
> This event may be emitted multiple times, in which case the client may choose any given format.

pgaskin: And regarding `dmabuf_render_formats`, it doesn't shuffle (it returns in insertion order), but HashMap does when grouping by FourCC.

pgaskin: I'll change it to use a Vec to preserve the original order (apart from the Xrgb/Argb thing).

src/handlers/image_copy_capture.rs
pgaskin: Cursor size is already guaranteed in https://github.com/niri-wm/niri/blob/9e72e4917ca31baf4010496bf7f4aaf78d34d236/src/cursor.rs#L179.

src/handlers/image_copy_capture.rs
pgaskin: Will do.

src/niri.rs
pgaskin: Will do.

src/niri.rs
pgaskin: Which also explicitly is the opposite of what PW requires.

I'm going to leave it like this (also note that it's how wlroots does it).

src/handlers/image_copy_capture.rs
pgaskin: Done.

src/handlers/image_copy_capture.rs
pgaskin: Done, but I did it as a `warn!()` to match how other places log client bugs.

src/handlers/image_copy_capture.rs
pgaskin: Doesn't work (`E0369: binary operation == cannot be applied to type Option<Frame>`) and would move `frame`, which is needed again in the second loop.

I did simplify it to `s.pending_frame.as_deref() == Some(&frame)` though.

src/render_helpers/mod.rs
pgaskin: Done.

src/niri.rs
pgaskin: It doesn't work the same way (I need the `self.backend` to build constraints, which is on `State`, not `Niri`), and it won't reduce code duplication that much (especially since `refresh_image_copy_capture` is still needed for to do the cleanup), so I'd rather keep it this way.

A significant advantage of keeping it inline here is it makes what we're checking for explicit and eliminates the possibility of accidentally refactoring part of it away later.

src/niri.rs
pgaskin: Done.

src/niri.rs
pgaskin: I figured it's wasteful to send partial damage for a cursor small cursor image (where it's usually changing entirely).

Also, for animated cursors, a new frame results in a different `MemoryBufferRenderElement`, which means it'll get reported as full damage by the damage tracker anyways.

pgaskin:
Pushed some minor changes (mostly refactoring and some additional comments) based on your review.

I need to think about the cast IPC a bit more since it doesn't match up perfectly with how ext-image-copy-capture works (e.g., is_active has no direct equivalent). Edit: But it doesn't for screencopy either (it's only for PW), so I guess it's fine.

https://github.com/niri-wm/niri/compare/aa31f5b32bec687d3ae9deff5af463aa19f47510..b217f3f4801ad7f3e2941f21d403374c5ee5fc10

src/niri.rs
pgaskin: I don't *think* it's possible to hit a case where `image_copy_sessions` isn't empty at that point since handlers aren't re-entrant, but I added it just in case.

I'll add a comment and log message.

src/niri.rs
pgaskin: Yep, that's an old check I added, and it's effectively redundant if I move the `pending_frame.is_none()` check up in `render_image_for_copy_capture`, which I've done now.

pgaskin:
Added (and tested) support for cast IPC.

https://github.com/niri-wm/niri/compare/b217f3f4801ad7f3e2941f21d403374c5ee5fc10..aca0b6be31af0d7080e3f92d52864979df12457a

src/niri.rs
YaLTeR: Alright. For outputs it's fine, but for window capture I'd prefer the same logic as in PW because it's done that way to avoid exposing the cursor image over windows when it's not desired.

> Which also explicitly is the opposite of what PW requires.

Does PW *require* it? Back before I added the check, my PW sinks like OBS happily rendered the cursor way outside the screencast contents if niri signaled it this way. So it was just a natural way to restrict it. This way also avoids two cursors being drawn when it's crossing between two adjacent output streams.

src/handlers/image_copy_capture.rs
YaLTeR: Ah, that's good. Maybe worth to add a comment and pass 0 time so it doesn't look like measuring specifically the current frame.

src/handlers/image_copy_capture.rs
YaLTeR: If you're saying that Smithay is supposed to check this, then wouldn't it indicate a programming error?

src/niri.rs
YaLTeR: Fair enough

src/handlers/image_copy_capture.rs
YaLTeR: Ah, I see you changed it to "Smithay doesn't check itself it (create_frame pushes onto active_frames unconditionally)". Then a warn!() is correct yeah.

pgaskin: No, what I mean is that this would indicate a misbehaving client. Ideally, this would be checked in Smithay and handled there, but it isn't, so I do it here.

Edit: we posted at the same time lol.

src/handlers/image_copy_capture.rs
pgaskin: Yeah, I added a comment.

YaLTeR (approved):

src/niri.rs
pgaskin: Actually, it's not pipewire, but the downstream client (webrtc) I was primarily looking at for reference which depends on it: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/desktop_capture/linux/wayland/shared_screencast_stream.cc;l=875-885;drc=5c10f93376c1e9f0ff339bd1facf5b6d7d978f25

In pipewire itself, it's stored as a spa_point (which uses signed values), and there's no validation on it.

However, in pipewire's gstreamer plugin, it's stored as a uint32, which implies it  was intended to be within bounds.

OBS does look like it just accepts the coordinates.

But yeah, for toplevel capture, we could probably clamp it.

src/niri.rs
```diff
@@ -5410,6 +5533,393 @@ impl Niri {
...
+                        .map(Some)
+                }
+                CaptureBuffer::Shm => render_to_shm(
+                    renderer,
+                    &mut s.damage_tracker,
+                    &buffer,
+                    wl_shm::Format::Xrgb8888,
+                    elements,
+                    states,
+                )
```
YaLTeR, line 5642: The protocol says:

> During capture, the compositor will update the buffer with at least the union of the region [passed by the client](https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_frame_v1:request:damage_buffer) and the region advertised by ext_image_copy_capture_frame_v1.damage.

However, these render functions will use the damage tracker and render only over the damage that it reported (and not over the damage that the client indicated).

I'm not sure what's the best way to deal with this. The damage tracker can store up to 4 ages of damage, so maybe we could just damage_output() with some transparent rectangles first? Alternatively if it's too involved, could just do a full render if there's client damage outside ours.

Would be nice if there was some test client for this..

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,426 @@
...
+            .retain(|s| s.session != session);
+        self.niri.image_copy_capture_state.cleanup();
+    }
+
+    fn cursor_session_destroyed(&mut self, session: CursorSessionRef) {
+        self.niri
+            .image_copy_cursor_sessions
+            .retain(|s| s.session != session);
+        self.niri.image_copy_capture_state.cleanup();
+    }
```
YaLTeR, line 438: > This request doesn't affect ext_image_copy_capture_frame_v1 objects created by this object.

Seems like if there's a pending frame then it should be completed in this case.

pgaskin: It depends how literally you interpret "affect" (affect meaning destroy, or affect meaning anything at all). I'll see what other compositors do.

src/niri.rs
pgaskin: It's actually fine as is. It always re-renders everything since the `render_output_with_states` is ultimately called with an age of 0, which causes smithay to treat the whole buffer as damaged (like with screencopy).

The damage tracker is only used to send the `damage` event.

If we wanted to optimize the blitting to take damage into account, I'd rather do that in a separate PR.

src/handlers/image_copy_capture.rs
pgaskin: In wlroots, the pending frame is failed with `stopped` before `stopped` is sent for the session.

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/8697752bbebcf3b7e0c3af9f15296653d838cdef/types/wlr_ext_image_copy_capture_v1.c#L380-397

In hyprland, it does the same for cursor capture, but I don't completely understand what happens for regular output capture.

https://github.com/hyprwm/Hyprland/blob/f05d73f35795ded80d7e1264a37e41b461625f9f/src/protocols/ImageCopyCapture.cpp#L124-L125

https://github.com/hyprwm/Hyprland/blob/f05d73f35795ded80d7e1264a37e41b461625f9f/src/protocols/ImageCopyCapture.cpp#L161-L171

In cosmic, frames survive the session being destroyed (they are with the output rather than the session list), and get rendered on the next redraw. This seems wrong IMO.

https://github.com/pop-os/cosmic-comp/blob/a7349b72e39fd91d7ed25eddb2fa24b900a615d5/src/wayland/handlers/image_copy_capture/mod.rs#L336-L388

https://github.com/pop-os/cosmic-comp/blob/a7349b72e39fd91d7ed25eddb2fa24b900a615d5/src/backend/render/mod.rs#L1428

Also: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/124 (though it doesn't really say one way or the other).

src/niri.rs
pgaskin: And in wlroots, it does kinda what you said about combining the client damage when copying each frame.

In hyprland, it always copies the whole buffer.

In cosmic, there's a damage tracker per session, to which the client damage is added.

src/handlers/image_copy_capture.rs
pgaskin: I'm going to implement it like wlroots (`fail` the frame with `stopped` on destroy).

For the cursor session, it'll need a change in smithay, which currently doesn't call cursor_session_destroyed when the session is destroyed. I'll do that one in another PR (like #4571).

pgaskin: Done.

pgaskin:
Added an explicit `frame.fail(CaptureFailureReason::Stopped)` for pending frames when the session is destroyed. Smithay will need a fix (Smithay/smithay#2170) to make it also work in all cases for cursor sessions, but it isn't really a blocker for this PR.

https://github.com/niri-wm/niri/compare/aca0b6be31af0d7080e3f92d52864979df12457a..bc82a7ee517bbb197ab4cb8dee8ee5ce277b36c1

The buffer copy thing is fine since we always copy the whole frame, and only use the damage for `damage` events.

src/niri.rs
YaLTeR: > It always re-renders everything since the `render_output_with_states` is ultimately called with an age of 0, which causes smithay to treat the whole buffer as damaged (like with screencopy).

Hmm, right. Indeed, I forgot that this was how it worked. Especially in PipeWire where we get some random previous buffer and don't track its damage, so it needs to be fully rerendered. This is fine then.

src/handlers/image_copy_capture.rs
YaLTeR: Thanks for checking other compositors, sounds reasonable.

src/handlers/image_copy_capture.rs
```diff
@@ -0,0 +1,439 @@
...
+                frame.fail(CaptureFailureReason::Stopped);
+            }
+        }
+        self.niri.image_copy_capture_state.cleanup();
+    }
+
+    fn cursor_session_destroyed(&mut self, session: CursorSessionRef) {
+        // FIXME: this doesn't get called by smithay when the underlying
+        // ext_image_copy_capture_session_v1 is destroyed, only when the
+        // ext_image_copy_capture_cursor_session_v1 is destroyed
```
YaLTeR, line 428: Guess this comment is unnecessary bc it'll get fixed by a smithay update automatically?

pgaskin: #4574

YaLTeR:
I tried xdp-wlr and it seems to work with ext-copy-capture. Apparently it also doesn't support metadata cursor despite ext-copy-capture providing all the necessary infrastructure...

YaLTeR:
Thanks!

ids1024:
> I tried xdp-wlr and it seems to work with ext-copy-capture. Apparently it also doesn't support metadata cursor despite ext-copy-capture providing all the necessary infrastructure...

Not sure if it needs any changes to run on niri, but xdg-desktop-portal-cosmic does now have metadata cursor support. If you need something to test it, anyway.

PR #4566, opened 2026-09-11
Fix screencopy buffers when buffer-size != shm-pool-size

iciclejj:
See #4553 and https://github.com/niri-wm/niri/pull/4553#issuecomment-5632024235.

#4553 fixed the incorrect use of `with_buffer_contents_mut`, but missed the same issue in screencopy.rs.

Also added a helper to help ensure the compositor-decided/advertised buffer stride and later verification of it stays in sync.


YaLTeR:
Thanks

PR #4567, opened 2026-09-11
feat: add configurable pinch sensitivity

beixiyo:
Adds configurable sensitivity for pinch gestures forwarded to clients.

- `input.touchpad.pinch-sensitivity` sets the input-level sensitivity
- `window-rule.pinch-sensitivity` sets the sensitivity for matching windows
- the input and window-rule values are multiplied together
- both default to `1.0`, preserving the existing behavior

The client-facing scale is transformed with `scale.powf(sensitivity)`.

This is useful on touchpads where client pinch zoom is otherwise too slow. I have tested it with Figma and Firefox on my laptop touchpad, including configuration reload and a full compositor restart.

---

## Example configuration

  ```kdl
  input {
      touchpad {
          // Applied to all client-forwarded pinch gestures.
          pinch-sensitivity 1.8
      }
  }

  window-rule {
      match app-id="firefox$"

      // Multiplied with the input-level value for matching windows.
      pinch-sensitivity 1.2
  }
  ```

  In this example, the effective pinch sensitivity in Firefox is
  `1.8 × 1.2 = 2.16`. Other applications use the input-level value of `1.8`.

---

Discussed in https://github.com/niri-wm/niri/discussions/4474

## Testing

- `cargo test -p niri-config`
- `cargo check --workspace --all-targets`
- manual testing with Figma and Firefox


Sempyos:
Could you put an example configuration of the options you added in the description..?

beixiyo:
> Could you put an example configuration of the options you added in the description..?

Added an example showing both the input-level and window-rule options, including how their values combine. Thanks!

Sempyos:
Thanky

YaLTeR:
Thanks, all good

PR #4572, opened 2026-09-12
feat: Add a live cursor-following screen magnifier

ensismoebius:
## Summary

Adds a toggleable live magnifier, similar to accessibility magnifiers in other desktop environments:

- Crops a small area of the already-composited framebuffer around the cursor and blits it back scaled up (à la `framebuffer_effect.rs`'s capture trick).
- Toggle with the `toggle-magnifier` bind action or the equivalent IPC action. Scrolling with the configured `zoom-modifier` held (`Mod` by default) also turns the magnifier on and adjusts its zoom level directly, so a separate toggle press isn't required to start zooming.
- Zoom level changes ease in/out with a new configurable `magnifier-zoom` animation instead of snapping instantly.
- The cursor sprite tracks the same crop/zoom warp the magnifier applies to the framebuffer, so it stays visually aligned with the magnified content instead of drifting to its real screen position.
- New `magnifier` config options: `max-zoom` (cap for scroll-zoom), `zoom-speed` (scroll step size), `zoom-modifier` (`mod`/`mod-ctrl`/`none`, which modifier if any is needed to zoom by scrolling), and `hide-mouse` (hides the real cursor while active).

Split into three self-contained commits: the base magnifier feature, the zoom animation, and the zoom-modifier/cursor-accuracy polish.

## Test plan

- [x] `cargo check --workspace` and `cargo test -p niri-config` pass on each commit individually.
- [x] `niri validate` on a config exercising every new option (`zoom`, `max-zoom`, `zoom-speed`, `zoom-modifier`, `hide-mouse`, and the `magnifier-zoom` animation).
- [x] Wiki docs added for the animation and the `magnifier` config section, verified against the `wiki-parses` test.
- [x] Manual testing on a running compositor (toggle, fullscreen zoom in/out via scroll with each `zoom-modifier` setting, hide-mouse, cursor accuracy while zoomed), including in a nested niri instance.

I do needed this feature, so i implemented with help of Claude Code, and using now. Hope it helps someone else.


ensismoebius:
Addressed the Copilot review findings by amending the relevant commits (force-pushed):

- Magnifier now damages every frame while active, not just on toggle/zoom-change, so cursor movement and scene changes behind it are always re-captured.
- The intermediate capture texture is now sized via `transform.transform_size(dst.size)`, fixing a width/height mismatch on 90°/270° rotated outputs.
- The rendered zoom value is now clamped to `max-zoom` in addition to its target, since a bouncy `magnifier-zoom` spring could otherwise briefly overshoot the configured cap.

All three commits still build and pass `cargo test -p niri-config` individually.

ensismoebius:
Fixed a corner-dimming bug in the circle spotlight lens: dimming previously stopped at the lens's whole bounding square, but a circle lens only ever displays the area inside its inscribed circle (its own square corners are masked out). That left a square-ish patch of undimmed background visible at each of the four corners around the circle.

For circle lenses, the undimmed "hole" now shrinks down to the actual crop rectangle (mirroring the crop size computed in `magnifier.rs`) instead of the whole lens box — safe, since nothing between the crop and the lens box ever actually becomes visible through the circular mask. Rectangle lenses are unaffected.

Tested live in a nested niri instance (winit backend) against both a plain background and a terminal with visible text — corners now dim uniformly with the rest of the screen, no residual undimmed patches.

PR #1791, opened 2025-06-13
Support shm sharing

wrvsrx:
Fix #455

PR #4588, opened 2026-09-15
pw_utils: retain SHM buffer mappings

PurpleCloudX:
SHM capture currently maps and unmaps the same memfd on every frame and clear. This keeps the mapping for the buffer's lifetime instead.

Two commits:

1. Borrow `Shmbuf` during rendering and clearing, and pass only its `Copy` layout to completion metadata. This removes the SHM clones and lets the buffer own its file descriptor directly. The borrow ends before queueing to PipeWire.
2. Map the sealed memfd when allocating the buffer and unmap it when `Shmbuf` is dropped. The mapping constructor is `unsafe`, with explicit requirements for file size, lifetime, and synchronized access. Writes use short-lived slices with `copy_from_slice()` and `fill()`. Clearing no longer returns a `Result`, since it only writes to an existing mapping.

Each buffer retains a virtual-memory mapping until removal; no additional pixel storage is allocated. Mapping errors are reported during buffer allocation rather than during rendering.

Mutter uses the same lifetime for its MemFd mappings: [mapping in `on_stream_add_buffer`](https://github.com/GNOME/mutter/blob/3af3738795f6e3a37b77c3e0b64b1e671709a3aa/src/backends/meta-screen-cast-stream-src.c#L1787-L1825), and [unmapping in `on_stream_remove_buffer`](https://github.com/GNOME/mutter/blob/3af3738795f6e3a37b77c3e0b64b1e671709a3aa/src/backends/meta-screen-cast-stream-src.c#L1883-L1891).

I started this work while #1791 was open, with Codex assistance for implementation and testing.

### Validation

Passed on the reduced revision:

```sh
cargo check --locked -p niri
cargo fmt --all -- --check # nightly rustfmt from the repository's Nix dev shell
```

Live screencast tests and benchmarks have not been rerun for this revision.


YaLTeR:
I'm not against optimizing PW SHM, but this is just a huge patchset you're dropping all at once which makes me very reluctant to review it. The +4,084 -196 stat is also concerning: PW SHM is a fallback mode, I'm not prepared to have +4k lines to improve its performance.

Maybe you could split out a few of the changes that have the smallest diffs but provide the most benefit?

PurpleCloudX:
I've reduced this to five commits in five files (+449/-93): mapping and texture reuse, two unnecessary buffer clones removed, and fixes for stalled captures and accumulated frame-timing delays.

The description includes separate benchmarks for mapping and texture reuse, along with the design and testing details. Per-buffer damage tracking is deferred; DMA-BUF recovery is in #4590.

src/screencasting/pw_utils.rs
YaLTeR, line 1: How much is this really a problem? I understand this is more correct but this change as implemented here is kind of unwieldy and awkward. While you can move the mouse or otherwise cause a redraw if you run out of buffers.

src/screencasting/pw_utils/frame_pacing.rs
YaLTeR, line 1: Again, when is this actually a problem in practice? When and where can you experience this being problematic? This amount of code needs justification.

src/screencasting/pw_utils/shm_mapping.rs
```diff
@@ -0,0 +1,54 @@
...
+    len: usize,
+}
+
+impl ShmMapping {
+    pub(super) fn new(fd: BorrowedFd<'_>, len: usize) -> anyhow::Result<Self> {
+        ensure!(
+            len > 0 && len <= isize::MAX as usize,
+            "invalid mapping length"
+        );
+        // The caller seals the file against shrinking before creating this mapping.
```
YaLTeR, line 20: If a function relies on the caller to do something right for safety then the function must be marked as unsafe

src/screencasting/pw_utils/shm_mapping.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+        .context("error mapping SHM buffer")?;
+        Ok(Self { address, len })
+    }
+
+    /// Only call while the producer owns the dequeued PipeWire buffer.
+    pub(super) fn copy_frame(&self, bytes: &[u8]) -> anyhow::Result<()> {
+        ensure!(bytes.len() == self.len, "invalid SHM frame length");
+        // The source cannot alias this private mapping; no Rust references to it exist.
+        unsafe { ptr::copy_nonoverlapping(bytes.as_ptr(), self.address.cast(), self.len) };
```
YaLTeR, line 39: I prefer `slice::from_raw_parts_mut()` followed by `copy_from_slice()`, I believe it should be as fast. Same for `clear()`.

src/screencasting/pw_utils/shm_mapping.rs
YaLTeR, line 1: Does any other PW screencast producer, like Mutter, preserve the mapping? Can you link reference to the code?

src/screencasting/pw_utils.rs
```diff
@@ -2079,4 +2050,34 @@ mod tests {
         assert!(ShmLayout::new(Size::from((536_870_912, 1))).is_err());
         assert!(ShmLayout::new(Size::from((500_000_000, 3))).is_err());
     }
+
+    #[test]
+    fn shm_mapping_survives_buffer_clones_and_rejects_invalid_copies() {
```
YaLTeR, line 2055: I don't think this test is useful, it repeats the 2 line trivial logic from the impl

src/screencasting/pw_utils.rs
```diff
@@ -1342,16 +1342,16 @@ impl Cast {
                         .map(|x| (x, SharingBuf::Dma))
                 }
                 x if x == DataType::MemFd.as_raw() => {
-                    let shmbuf = inner_.shmbufs[&fd].clone();
+                    let shmbuf = &inner_.shmbufs[&fd];
```
YaLTeR, line 1345: This commit before all others please

src/render_helpers/mod.rs
YaLTeR, line 1: Since this barely improves performance let's skip it for now

YaLTeR (changes requested):
Thanks, please leave only Shmbuf reference commit and persistent mmap commit and address my comments. Everything else can go into a separate PR since usefulness is not very clear for now.

PurpleCloudX:
Thank you for taking the time to review this and for helping me narrow the scope.

I've kept only the Shmbuf reference commit, followed by the persistent mmap commit, and addressed your comments. The mapping constructor is now explicitly marked `unsafe` with its safety requirements documented, copying and clearing use slice APIs, and I've removed the trivial test.

I also checked Mutter: it maps MemFd buffers when they are added and unmaps them when they are removed. I've linked those code paths in the updated PR description.

The existing workspace tests, Clippy, and formatting checks all pass. Thanks again for the guidance.

YaLTeR:
I measured in winit and I can confirm the performance improvement, my test came out 50% faster:

image

YaLTeR:
Thanks

Issue #4587, opened 2026-09-14
Zoom window share is blank: static window casts send a single frame (plus 1x1 stream Parameters)


slynchDev:
Sharing a single window from Zoom (6.7.5, PipeWire mode, xdg-desktop-portal-gnome) on niri shows a blank picture, while sharing a monitor works. Traced it on niri 25.11 (`v25.11-131-gd74a4356`, NVIDIA, SHM streams); the relevant code is unchanged on main. Two separate things on niri's side, the second is the actual cause.

### 1. Window streams report a 1x1 `Parameters` size

`src/dbus/mutter_screen_cast.rs`, `Stream::parameters()`, returns `size: (1, 1)` for `StreamTarget::Window` with the comment "Does any consumer need this?". xdg-desktop-portal-gnome (`gnomescreencast.c`, `gnome_screen_cast_session_add_stream_properties`) copies `position` and `size` from that property into the stream properties it hands to the app, so every portal client is told a niri window stream is 1x1. Mutter reports the window's logical size there. Fix: look the window's logical bbox up on the niri thread at `RecordWindow` time and report it — slynchDev/niri@6d6875b2. (Verified via `busctl`; on its own it did not unblank Zoom.)

### 2. Static windows send exactly one frame; Zoom drops it

Casts are damage-driven. A window that isn't changing (a Chrome "New Tab" page, an idle terminal) produces one frame when the consumer connects and nothing afterwards. Zoom evidently discards the first frame and then waits forever, so the share stays blank until the window repaints. Sharing a window with a running `btop` in it works fine, and monitor shares work because cursor and bar clock damage keep frames flowing.

Measured with a libpipewire client on the Mutter ScreenCast D-Bus API (`RecordWindow` / `RecordMonitor`, BGRA/BGRx SHM):

| cast | frames delivered |
|---|---|
| Chrome window, static, 8 s | 1 |
| monitor (3440x1440), 8 s | 51 |
| kitty window printing `date` at 20 Hz, 6 s | 122 |

What fixed it: a 500 ms per-cast timer that, if an active window cast has produced no frame in that time, drops its damage trackers and redraws it (`Cast::force_full_redraw()` + `State::redraw_cast()`), so a static window streams at ~2 fps — slynchDev/niri@d76dbbb0. With that, Zoom window shares work on r16 (static Chrome window verified, monitor shares unaffected). Harness on the fixed build: 11 frames in 6 s for the same static window.

[...]

PR #4589, opened 2026-09-15
dbus: report logical window screencast size

PurpleCloudX:
## Summary

Window screencast streams currently report `size = (1, 1)` in the Mutter ScreenCast `Parameters` property. xdg-desktop-portal-gnome forwards this value to clients, although it does not describe the captured region.

This addresses the metadata part of #4587, reported by @slynchDev.

- Query the initial window bounds on the Niri thread when handling `RecordWindow`, using the same `bbox_with_popups()` as capture, in logical coordinates.
- Store only the resulting size. A bounded reply channel follows the existing D-Bus request pattern; replying never blocks the compositor thread.
- Omit `size` for the dynamic cast target, which has no source selected yet. Reject missing or empty ordinary windows instead of fabricating dimensions.
- Omit `position` for window streams. Monitor parameters remain unchanged.

These are initial parameters, not live resize notifications. PipeWire continues to negotiate the actual pixel dimensions. The [Mutter interface](https://github.com/GNOME/mutter/blob/gnome-48/data/dbus-interfaces/org.gnome.Mutter.ScreenCast.xml) defines optional source dimensions in compositor coordinates; the [portal interface](https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.portal.ScreenCast.xml) also makes size optional and restricts position to monitor streams.

## Testing

Two tests use the existing Wayland client/server fixture: logical capture bounds at scales 1, 1.5 and 2, client-side window geometry, resize, unmap, dynamic targets, missing IDs, and a dropped reply receiver. No new test dependencies.

Workspace tests passed (201 Niri tests, 22 config/IPC/wiki tests and one doctest), as did strict Clippy, no-default-features and D-Bus-only checks, and nightly rustfmt.

Also tested a nested Niri under headless Weston with a private D-Bus session at scale 1.5:

| Source | Initial logical size after this change | PipeWire pixels, unchanged |
| --- | --- | --- |
| 640x480 window | 640x480 | 960x720 |
| Same window with an extending popup | 840x540 | 1260x810 |

Before the size fix, both reported 1x1. GStreamer received frames in both cases. Monitor parameters remained `(0, 0)` / `1280x720`; dynamic-target parameters omitted size. Missing and closed window IDs returned errors without invalidating the new session.

To check manually, create a Mutter ScreenCast session, call `RecordWindow` with a window ID from `org.gnome.Shell.Introspect.GetWindows`, and read `org.gnome.Mutter.ScreenCast.Stream.Parameters`. Repeat on a scaled output and with a popup extending outside the parent. Logical dimensions should describe the capture bounds, not the scaled PipeWire buffer. Recording the dynamic target should still succeed without a size property.

I used Codex to assist with implementation and testing.


HigherOrderLogic:
#4340.

PR #4590, opened 2026-09-15
pw_utils: validate DMA-BUF planes and recover failed streams

PurpleCloudX:
## Summary

Validate DMA-BUF allocations before publishing them and renegotiate SHM when DMA-BUF capture fails. Developed with Codex assistance.

## Problem

An allocation that works for one advertised DMA-BUF modifier can fail for another, and successful negotiation does not guarantee that later allocation, rendering or fence handling will succeed. The current paths can stop capture after the first allocation/export failure, assert on a plane-count mismatch, or lose the notification needed to deliver an unfinished buffer.

Plane metadata also needs to describe the allocation rather than an assumed packed image. Padding and modifier-specific auxiliary storage cannot be inferred from width and height, and a multi-plane buffer must not leave its later planes marked valid after a failed frame.

## Design

- Try the driver's preferred allocation first, then individual offered modifiers if allocation or export fails.
- Validate the allocated modifier, plane count, backing extents and SPA-representable strides before publishing buffers. Update metadata for every plane, including when returning a corrupted buffer.
- Attempt SHM renegotiation once after DMA-BUF allocation, rendering or fence-export/registration failure. Keep unfinished buffers owned until the old pool is removed; do not block the compositor or deliberately publish unfinished contents.
- Ignore stale DMA negotiation during recovery, preserve the current size/rate, and stop the session if SHM allocation fails or recovery has not completed after ten seconds.

The modifier probe keeps the driver's choice as the fast path. Only a failed probe tries individual alternatives, so a working preferred allocation is not repeated. Candidate lists are borrowed slices; plane extents use an `ArrayVec` bounded by Smithay's maximum plane count.

Allocation is checked before publishing the plane descriptors. Backing-file lengths are queried rather than guessed from `stride * height`, which would be wrong for some auxiliary planes. Both corruption and successful reuse update every plane while preserving its offset and stride.

Recovery runs on the next event-loop dispatch: rendering temporarily removes casts from `State`, and renegotiating inside a PipeWire callback would also complicate re-entrancy. A small state enum allows one transition to SHM and prevents late DMA callbacks from restarting the failed path. Unfinished buffers remain owned while the old pool is being replaced. Blocking the compositor on a failed fence or repeatedly retrying DMA would make the failure affect more than the capture session.

## Trade-offs

After recovery, that stream stays on SHM rather than repeatedly attempting DMA-BUF again. This favors a usable capture over continued retries, at the cost of SHM's capture overhead. The ten-second limit is a recovery policy, not a protocol requirement: a consumer that does not complete SHM negotiation causes the session to stop instead of remaining suspended indefinitely.

An allocation whose size cannot be queried or represented in SPA takes the fallback path, even if a particular DMA importer could have used it.

## Validation

Workspace tests, strict Clippy and nightly formatting passed. Focused tests cover modifier retry order/bounds, padded and multi-plane storage, invalid offsets and restoration of all plane metadata.

A nested Niri under headless Weston also completed a live SHM capture with a slow GStreamer consumer (`identity sleep-time=100000`), consumer reconnects and explicit session teardown. This checks that ordinary SHM operation still works; it does not exercise DMA recovery.

The layout tests use memfds to exercise metadata handling; they do not simulate a GPU driver. GBM negotiation, fence failure and recovery timeout paths have not been validated on a DRM-backed session. Intel/NVIDIA hardware was not available for testing.

```sh
```


src/screencasting/pw_utils/dmabuf_layout.rs
YaLTeR, line 1: What is the actual problem that this commit is solving and when does it occur in practice?

src/screencasting/pw_utils/recovery.rs
YaLTeR, line 1: When can this occur in practice?

YaLTeR (commented):
My main concern here is adding a lot of code and logic for things that cannot obviously even happen in practice. For DMA-BUF validation for instance, can it even happen? I hope that PipeWire doesn't let the client give us modifiers that we didn't give it in the first place for example.

PurpleCloudX:
Thanks for reviewing this. After checking the PipeWire documentation and existing reports, I do not have a concrete reproducer that justifies the additional validation and recovery logic in this PR.

In particular, the [PipeWire DMA-BUF documentation](https://github.com/PipeWire/pipewire/blob/master/doc/dox/internals/dma-buf.dox) explicitly allows `maxsize` and `chunk.size` to be zero, so filling in backing extents is not a protocol correctness fix. The synthetic tests exercise the checks, but do not establish that these invalid layouts occur with buffers allocated by niri.

Allocation and framebuffer import failures can happen, as the historical NVIDIA issue #511 shows, but that issue was resolved by modifier-aware allocation. I have not reproduced a remaining failure on current niri or verified that this PR successfully recovers from it on real hardware. The SHM capture tests do not validate that recovery path.

I am closing this PR rather than asking you to maintain this complexity without a demonstrated benefit. A future fix should start with a reproducible failure and target that specific path. Thanks for pointing out the gap in the justification.

PR #4591, opened 2026-09-15
screencasting: recover consumer progress and preserve capture cadence

PurpleCloudX:
## Summary

Resume capture when consumer buffers become available and prevent redraw jitter from accumulating into a lower capture rate. Developed with Codex assistance.

## Problem

Damage tracking records what changed in the scene, not whether a consumer received those pixels. If capture updates the tracker but finds no available PipeWire buffer, a later redraw of the same static scene can appear to have nothing to send. A resumed consumer or replacement buffer pool has a similar problem: it needs an initial frame even when the scene has not changed.

Separately, computing every deadline from the last actual frame time adds each small scheduling delay to the next frame's deadline. That accumulated lateness reduced the effective capture rate in the nested SHM tests.

## Design

- Invalidate scene/cursor damage when a consumer resumes or a new buffer pool becomes usable, so a static scene can produce its first frame again.
- Remember buffer starvation and retry from the PipeWire process callback when buffers return. Only request a redraw while active and waiting; this does not add polling or periodic idle frames.
- Release the `RefCell` borrow before queueing buffers, since queueing can invoke callbacks. Use `VecDeque` to preserve delivery order without repeatedly shifting a `Vec`.
- Keep a bounded frame deadline rather than accumulating redraw lateness. Use the same non-stale presentation time for admission and recording, retaining future predictions and the negotiated rate limit.

Invalidating damage asks the existing rendering path to reconstruct the frame; it does not renegotiate the format or duplicate pixel storage. The `waiting_for_buffer` flag distinguishes a skipped delivery from ordinary inactivity. PipeWire's process event then provides the retry opportunity, instead of adding a timer that repeatedly wakes an idle capture.

Queueing a completed buffer can call back into PipeWire synchronously. Removing it from the queue and releasing the mutable borrow first avoids re-entering a borrowed `CastInner`. A FIFO `VecDeque` fits that operation without copying the remaining entries or changing their delivery order.

The pacing helper retains the existing cadence through small delays, but resets after a long pause, clock reversal or rate change. It does not accumulate a backlog of missed frames to send in a burst. Presentation predictions are kept when they are in the future; stale predictions are clamped once and the same time is used for both admission and recording.

## Trade-offs

Recovery can require a full redraw of an otherwise static scene. That is intentional: after a missed delivery, the damage tracker alone cannot prove that the consumer has the current contents. Pacing follows a sustained cadence rather than demanding an exact minimum gap after every late frame; the next frame can be closer to a late frame while remaining on the negotiated schedule. No periodic frames are added to a healthy idle stream.

These problems became visible while testing cheaper SHM readback. #4588 uses this branch on both sides of its performance comparison so a change in capture cadence is not mistaken for an optimization.

## Validation

Workspace tests, strict Clippy and nightly formatting passed. Focused tests cover static-scene invalidation, jitter, idle/clock resets, rate changes and faster input against a negotiated cadence.

Release builds were run under GPU-rendered headless Weston with private D-Bus sessions and real GStreamer/PipeWire consumers. Three 300-frame, 60 Hz runs completed in 5.074 / 5.073 / 5.074 seconds with this branch. Consumer reconnects and explicit session teardown succeeded. The mapping/readback integration builds also survived a 100ms-per-frame consumer, with hundreds of actual buffer-starvation events in the producer logs.

A 30 FPS consumer with a 60 Hz source received 300 frames in 10.109 seconds. DRM/VRR timing has not been tested; the nested runs used an AMD Radeon 680M.

```sh
```


PurpleCloudX:
Included in #4588 as the first two commits. Keeping these delivery fixes with the resource-reuse changes makes that PR independently mergeable, so this separate PR is no longer needed.

PR #4592, opened 2026-09-15
pw_utils: read back only changed regions of SHM frames

PurpleCloudX:
## Summary

Read back only the damage accumulated since each SHM consumer buffer was last filled, using Smithay's synchronous readback and `DamageBag`. Developed with Codex assistance.

Deferred while the resource-reuse changes in #4588 are reviewed. The branch and measurements are preserved; per-buffer damage tracking can be evaluated against `main` after that work lands. The original texture/readback commits are visible in this [fixed comparison](https://github.com/PurpleCloudX/niri/compare/06a45a81e4941d6296f80a56cac539f86bbe7f6d...a39dd3d53349d1472c4b4d274f4c3fb904eb4fb3): +329/-33 across three files.

## Problem

SHM capture currently downloads and copies the whole frame even if only a small region changed. At 1920x1080 with four bytes per pixel, that is about 8.3 MB per frame. Reusing the mapping avoids mapping overhead, but it does not reduce that pixel traffic.

Using only the newest frame's changed region would be incorrect. PipeWire rotates consumer buffers: buffer A might still contain frame 10 when buffer B has already received frame 11. If A is used for frame 12, it needs the changes from both frames 11 and 12. There is no guarantee that the next returned buffer contains the previously delivered frame.

## Design

- Reuse texture storage while the renderer context, dimensions and format match. Still redraw the texture completely, avoiding a second GPU-content history.
- Lazily keep bounded damage history for SHM streams. Each destination buffer records the commit of its last successful contents, so returned buffers need not arrive in order.
- Read back the bounding damaged region since that buffer's commit, then copy its rows directly into the retained mapping. Unknown/expired history and non-normal transforms use a full readback. Clearing invalidates the buffer's commit; failed rendering resets history.

This keeps the state tied to the actual consumer buffer, rather than assuming the previous delivered frame is the contents of the next available buffer.

Each successful copy stores a Smithay commit counter on the shared mapping. Clones of a buffer therefore refer to the same contents and counter. Damage is recorded even when no destination buffer is available, and the counter advances only after the copy succeeds. The history is allocated only for SHM capture and retains 16 entries; a buffer older than that gets a full refresh, so correctness does not depend on the chosen history length.

The cached texture belongs to the cast, and is replaced when its renderer context, size or pixel format changes. Reusing storage avoids allocating a render target for every frame. It is still fully redrawn with the existing damage-tracker rendering API: preserving partial GPU contents would require a second history separate from the consumer buffers.

## Trade-offs

One bounding rectangle keeps the readback and row-copy path small. It can include unchanged pixels between distant updates, but avoids issuing a separate GPU readback for every tiny rectangle. Full-screen changes still take the full-copy path and are not expected to gain much from damage tracking.

Non-normal output transforms use full readback rather than introducing another coordinate-conversion path. Expired history, cleared buffers and render errors also force a full refresh. This gives up some optimization in those cases to keep the fallback behavior straightforward.

## Measurements

The measured builds were `b232ca78` (mapping reuse and delivery fixes) and `7e5bd0eb` (also texture reuse and incremental readback). 1920x1080 capture, 60 Hz input, 300 BGRx frames per run; medians of three alternating runs:

| Workload | Mapping only CPU (ms) | Incremental readback CPU (ms) | Elapsed before / after (s) |
| --- | ---: | ---: | ---: |
| 64x64 updates | 670 | 300 | 5.073 / 5.074 |
| Full 1920x1080 updates | 820 | 790 | 5.124 / 5.074 |

The useful gain is on small damage, about 55% less producer CPU time beyond mapping reuse. Full-frame cost is similar; the small difference should not be treated as a general speedup.

The [method in #4588](https://github.com/niri-wm/niri/pull/4588) uses release builds, a synthetic Wayland SHM client, headless GPU-rendered Weston, private D-Bus sessions and a real PipeWire/GStreamer consumer. Full-screen runs additionally verify captured pixels across the 1920x1080 image.

## Validation

Workspace tests, strict Clippy, the no-default-features build and nightly formatting passed. Manually invoked EGL tests compare actual pixels from cached/fresh textures and rotating SHM destinations against full readback, including moving content, skipped delivery, expired history, clears, size changes and a rotated output. A focused row-copy test checks bounds and preservation of untouched pixels.

Live integration runs cover consumer reconnects, explicit Stop and a slow consumer that causes real buffer starvation. Tests used an AMD Radeon 680M in nested Niri; other GPU drivers have not been tested.

```sh
cargo test --locked --lib -- --ignored --nocapture
cargo check --locked --no-default-features
```


PurpleCloudX:
Deferring per-buffer damage tracking to keep #4588 focused on resource reuse. The branch and measurements are preserved; this can be evaluated against main after the resource-reuse changes land, without maintaining a dependent PR in the meantime.

PR #4593, opened 2026-09-15
feat(input): Signal scroll stop events for continuous sources

eriedaberrie:
Niri currently only sends stop events for finger axis sources. This is fine and entirely in-spec but one of my apps (Vesktop specifically) seems to misbehave when continuous scroll sources don't get their terminating stop events, failing to release scroll latching until it sees one. Wayland makes sending stop events optional for non-finger sources, but sending one here would make Vesktop work better, and is also the default behavior in both hyprland and wlroots.

I should also mention that Niri currently uses a deprecated input API that doesn't technically guarantee that continuous sources are actually terminated with a zero-delta event but it does to work with my trackpoint.


YaLTeR:
Could you link some example from other compositors?

eriedaberrie:
https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/f5528eab3ec2bdf2a241865df445c229cd87b387/types/seat/wlr_seat_pointer.c#L390
https://github.com/hyprwm/Hyprland/blob/c26dbf9337998ba9afcfe84050e36969bb787598/src/managers/SeatManager.cpp#L497

None of these gate the checks behind the type being finger. Hyprland does skip it if the type is wheel while wlroots doesn't but AFAIK this doesn't matter because wheel sources don't send events with zero value.

YaLTeR:
Alright, thanks

PR #4596, opened 2026-09-15
feat: lock-cursor-when-focused window rule to force a cursor grab (when in focus)

lynthium:
Adds a "lock-cursor-when-focused" to window rule options that effectively binds a cursor to that window when in focus.
Use case for this is mostly gaming, where not all applications/games have a solidly working mouse grab.

Related discussion: https://github.com/niri-wm/niri/discussions/3355

Seems to work in testing so far with single and dual monitor setup with regular and floating windows.

Configuration example:
```
window-rule {
     match app-id="org.freedesktop.Xwayland" title=r#"^Xwayland on :2\d"#
     open-floating true
     lock-cursor-when-focused true
}
```


Sempyos:
Hiya, could you please add a example config of the option you added to the description of the PR..?

lynthium:
> Hiya, could you please add a example config of the option you added to the description of the PR..?

Added an example, but formatting is being annoying, will take another try at it after work.

Update: fixed indentation

Sempyos:
Thanky

src/input/mod.rs
```diff
@@ -2431,6 +2431,33 @@ impl State {
...
+            .window_rect(crate::layout::LayoutElement::id(win))
+        else {
+            return pos;
+        };
+        let output_geo = self.niri.global_space.output_geometry(output).unwrap();
+        rect.loc += output_geo.loc.to_f64();
+
+        Point::from((
+            pos.x.clamp(rect.loc.x, rect.loc.x + rect.size.w),
+            pos.y.clamp(rect.loc.y, rect.loc.y + rect.size.h),
```
HigherOrderLogic: I saw other place use `rect.loc.x + rect.size.w - 1`. Is there any reason for the difference?

PR #4597, opened 2026-09-15
portals: add oo7 as secret portal

hecknt:
Mirroring [this commit](https://gitlab.gnome.org/GNOME/gnome-session/-/commit/3a77ebd28b07dac36efadc93df7ebb3dcb3155c9) on gnome-session. oo7 is being considered as a replacement for gnome-keyring in Fedora 45, and is already used in GNOME OS.

Fedora 45 Change Set: https://fedoraproject.org/wiki/Releases/45/ChangeSet#oo7_Secrets_Service_Provider
GNOME OS MR: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/merge_requests/4788


HigherOrderLogic:
You probably also need to change the docs.

hecknt:
> You probably also need to change the docs.

Done.

YaLTeR:
Makes sense. Can't test myself yet bc it's not packaged in aeryn

Sempyos:
I'll probably poke around with it today for a bit

PR #4599, opened 2026-09-16
handlers: implement ext-image-copy for toplevel

### LINKED ISSUE #1558: Implement ext-image-copy-capture

HigherOrderLogic:
Closes #3942, actually closes #1558.

Implement the other half of the ext-image-copy protocol., including cursor capture. I have only tested with `wayshot` as that's the only application I could find that do something meaningful with the protocol.


src/protocols/foreign_toplevel.rs
```diff
@@ -480,6 +493,76 @@ where
...
+                        source: capture_source.clone(),
+                    },
+                );
+                capture_source.add_instance(&source);
+            }
+            ext_foreign_toplevel_image_capture_source_manager_v1::Request::Destroy => {}
+            _ => unreachable!(),
+        }
+    }
+}
```
HigherOrderLogic, line 564: As I said in #4552, this is because we use a custom ext-foreign-toplevel implementation, so we cannot interop with Smithay's impl of this protocol. 

Migrating to Smithay's ext-foreign-toplevel impl would help to get rid of this block, and also help us in the future as more protocols depend on ext-foreign-toplevel.

src/niri.rs
```diff
@@ -5612,55 +5677,95 @@ impl Niri {
...
+                    let buf_pos = win_pos + bbox.loc.to_f64().to_logical(scale);
+                    let pos = buf_pos.to_physical_precise_round(scale).upscale(-1);
+                    self.render_pointer(renderer, output, &mut |elem| {
+                        elements.push(WindowCaptureRenderElement::Pointer(
+                            RelocateRenderElement::from_element(elem, pos, Relocate::Relative),
+                        ));
+                    });
                 }
             }
+            mapped.render_for_screen_cast(renderer, scale, false, &mut |elem| {
```
YaLTeR: Why do you think window casts should exclude popups? The ones in PipeWire intentionally include popups, I think that's how window casts should work

src/niri.rs
```diff
@@ -7011,6 +7108,38 @@ impl Niri {
     }
 }
 
+fn reset_damage_tracker_if_changed(
```
YaLTeR: These extractions into separate commits pls

src/niri.rs
```diff
@@ -5954,8 +5947,47 @@ impl Niri {
...
+                    .layout
+                    .window_render_location(mapped.toplevel().wl_surface())
+                    .and_then(|win_pos| {
+                        let ipos = (pointer_pos
+                            - geo.loc.to_f64()
+                            - (win_pos + bbox.loc.to_f64().to_logical(scale)))
+                        .to_physical_precise_round(scale);
+                        let hotspot = Point::<i32, Physical>::from((hotspot.x, hotspot.y));
+                        Rectangle::new(ipos - hotspot, Size::from((cursor_size.w, cursor_size.h)))
+                            .overlaps(Rectangle::from_size(bbox.size))
```
YaLTeR: Imo for window casts the pointer check should be like in PipeWire, for a single point, not rectangle overlap, so that pointer is not leaked when you don't want it

src/niri.rs
HigherOrderLogic: Last time I checked other implementations like Mutter and Hyprland, they all excludes popups.

src/niri.rs
HigherOrderLogic: I originally did that, but I changed to checking rectangle overlap to be consistent with the output one. Will revert later.

src/niri.rs
HigherOrderLogic: I just checked again, Mutter doesnt implement it, but for those that do, [Hyprland does include popup](https://github.com/hyprwm/Hyprland/blob/main/src/managers/screenshare/ScreenshareFrame.cpp#L336), and [Cosmic doesnt](https://github.com/pop-os/cosmic-comp/blob/master/src/wayland/handlers/image_copy_capture/render.rs#L681).

I based this implementation on the Cosmic one, so it doesnt include popups.

YaLTeR: Well I think they're wrong about not including popups, which is why niri includes popups

HigherOrderLogic: I will add it back then.

pgaskin:
I'll take a look at this next week.

src/niri.rs
pgaskin: FWIW, I wouldn't go with cosmic blindly since they've already definitely [deviated](https://github.com/niri-wm/niri/pull/4554#discussion_r3995427661) from what seems to be the correct implementation before.

src/niri.rs
pgaskin: Yeah, we did overlap for the output one since the spec explicitly states that in a few ways:

> The cursor enters the captured area when the cursor image intersects with the captured area. Note, this is different from e.g. wl_pointer.enter.

> The given position is the position of the cursor's hotspot and it is relative to the main buffer's top left corner in transformed buffer pixel coordinates. The coordinates may be negative or greater than the main buffer size.

Arguably, it should be that for toplevels too, but idk.

src/niri.rs
HigherOrderLogic: > I wouldn't go with cosmic blindly

I mean, it's a big compositor. Also I was using its codebase to reference its Smithay API usage so it's just coincidence.

src/niri.rs
```diff
@@ -5708,17 +5851,7 @@ impl Niri {
...
-                scale: last_scale,
-                ..
-            } = s.damage_tracker.mode().clone()
-            else {
-                unreachable!("damage tracker must have static mode");
-            };
-            if size != last_size || scale != last_scale {
-                s.damage_tracker = OutputDamageTracker::new(size, scale, Transform::Normal);
-            }
+            reset_damage_tracker_if_changed(&mut s.damage_tracker, size, scale, Transform::Normal);
```
cornservant: This additionally resets if `last_transform != Transform::Normal` which didn't happen before. Is this intentional?

src/handlers/image_copy_capture.rs
```diff
@@ -85,6 +88,25 @@ pub fn source_output(source: &ImageCaptureSource) -> Option<Output> {
     source.user_data().get::<WeakOutput>()?.upgrade()
 }
 
+/// Window captured by a session, and the output it is currently on.
+pub fn source_window<'a>(
+    niri: &'a Niri,
+    source: &ImageCaptureSource,
+) -> Option<(&'a Mapped, Output)> {
```
cornservant: Not being a member function of niri, you might as well make it only depend the layout:
```rust
pub fn source_window<'a, W: LayoutElement>(layout: &'a Layout<W>, source: &ImageCapture Source) -> Option<(&'a W, Output)>
```

src/handlers/image_copy_capture.rs
```diff
@@ -85,6 +88,25 @@ pub fn source_output(source: &ImageCaptureSource) -> Option<Output> {
...
+        .user_data()
+        .get::<Weak<WlSurface>>()?
+        .upgrade()
+        .ok()?;
+    let (mapped, output) = niri.layout.find_window_and_output(&surface)?;
+    Some((mapped, output.cloned()?))
+}
+
+/// Output of a session's source.
+pub fn source_session_output(niri: &Niri, source: &ImageCaptureSource) -> Option<Output> {
```
cornservant:
```rust
pub fn source_session_output<W: LayoutElement>(
    layout: &Layout<W>,
    source: &ImageCaptureSource,
) -> Option<Output> {
    source_output(source).or_else(|| source_window(layout, source).map(|(_, o)| o.clone()))
}
```

src/niri.rs
```diff
@@ -5713,17 +5856,7 @@ impl Niri {
...
```
cornservant: This additionally resets if `last_transform != Transform::Normal` which didn't happen before. Was the previous behaviour buggy?

cornservant: This additionally resets if `last_transform != Transform::Normal` which didn't happen before. Was the previous behaviour buggy?

HigherOrderLogic: The cursor is always created with `Transform::Normal`, so comparing the mode will always return false, basically keeping the same behavior compared to the previous implementation.

PR #4608, opened 2026-09-17
layout: respect input region holes

midischwarz12:
Niri currently treats points outside a client's Wayland input as compositor activation hits when they remain inside the tile area. This prevents transparent input-region holes or "click through" from passing pointer events through to windows underneath.

The change distinguishes three areas:
      - Client input regions receive normal input.
      - Compositor decorations outside the client window geometry remain activatable.
      - Holes inside the client window geometry pass through to underlying surfaces.

This follows the semantics of `wl_surface.set_input_region` without making transparent pixels implicitly click-through.

The behavior applies to both floating and tiled windows. Overview mode retains its existing behavior through a separate activation-only hit-testing path, so transformed thumbnails remain selectable.

This change has been tested, and I am currently daily driving this change.


YaLTeR:
How does it compare to https://github.com/niri-wm/niri/pull/4381? That one is much simpler while applying my suggestion that a visible border behind window should remain clickable

midischwarz12:
This implementation preserves activation-only hit (click through) in overview and interactive move (InteractiveMoveState) where transformed windows cannot use normal input-region hit. #4381 seems to not account for that which could inccorectly select a window underneath or no window at all.

Also, this implementation accounts for floating or tiling, fractional scaling, and empty-background. 

I'm adding your desired border behavior now.

midischwarz12:
@YaLTeR Okay, I implemented the desired behavior for a clickable border when visible. This should be ready now.

See the above comment for edge cases that the other PR did not consider.

src/input/mod.rs
YaLTeR, line 1: Commit seems unrelated to this pull request?

src/layout/tile.rs
```diff
@@ -914,19 +914,49 @@ impl<W: LayoutElement> Tile<W> {
...
+    fn draws_background_behind_window(&self) -> bool {
+        if self.border.is_off() && self.focus_ring.is_off() {
+            return false;
+        }
+
+        self.window
+            .rules()
+            .draw_border_with_background
+            .unwrap_or_else(|| !self.window.has_ssd())
+    }
```
YaLTeR, line 946: If border is off and focus ring is on, but the window is not currently active, then there's no focus ring, so the click can go through

src/layout/tests.rs
```diff
@@ -2180,6 +2190,374 @@ fn large_max_size() {
     check_ops_with_options(options, ops);
 }
 
+#[test]
```
YaLTeR, line 2193: Are all of these tests needed? As in, do they all test different code paths? Can their amount of code be reduced?

src/layout/scrolling.rs
```diff
@@ -3013,6 +3013,18 @@ impl<W: LayoutElement> ScrollingSpace<W> {
...
+
+    pub fn window_under_for_activation(&self, pos: Point<f64, Logical>) -> Option<(&W, HitType)> {
+        self.window_under_impl(pos, true)
+    }
+
+    fn window_under_impl(
+        &self,
+        pos: Point<f64, Logical>,
+        for_activation: bool,
+    ) -> Option<(&W, HitType)> {
```
YaLTeR, line 3027: This "for activation" is a bit of a weird name and not very descriptive. Maybe instead have an `enum InputRegion { Honor, Ignore }` and pass that instead of `for_activation: bool`? And also remove the two helper functions and pass it always explicitly. Same in `HitType::hit_tile()`.

YaLTeR (changes requested):
See comments + please clean up the commits [like described in CONTRIBUTING.md](https://github.com/niri-wm/niri/blob/main/CONTRIBUTING.md#writing-pull-requests). Also: what clients can I use to test these changes?

src/input/mod.rs
midischwarz12: Yea, that was left over from testing. Woops.

src/layout/tests.rs
midischwarz12: Not fully. But I thought it was good coverage. I'll reduce it though.

src/layout/tile.rs
midischwarz12: Ah, good catch. I'll fix this

midischwarz12:
@YaLTeR When working on this, I could never find a client that produced reliable enough results for input regions. So I just ended up writing my own. I created a [public repo](https://github.com/midischwarz12/wayland-input-region-test) for what I was using to test this as I went. Feel free to use it as well.

midischwarz12:
@YaLTeR Okay, this should be ready for review again.

PR #4609, opened 2026-09-17
nix: retain the libinput capability build script

Vantrongs:
# nix: retain the libinput capability build script in the package source

The flake source filter omits `build.rs`. Cargo therefore skips the libinput >= 1.30 capability probe and builds without `have_libinput_plugin_system`, even when the installed libinput supports it.

Include `build.rs` and add a small check comparing it with the file in the filtered package source. No compositor/input behavior or dependency pins change.

Validation on upstream `818678e8`:
- Before: `packages.x86_64-linux.niri.src` contains no `build.rs`.
- After: `nix build --no-link .#checks.x86_64-linux.niri-source` passes.
- `git diff --check` passes.
- The current flake formatter reports two existing formatting differences outside this patch; they were left unchanged.

A full build of this exact upstream candidate and a new live compositor session were not run. Prepared with OpenAI Codex assistance.


flake.nix
```diff
@@ -164,6 +165,11 @@
     in
     {
       checks = forAllSystems (system: {
+        niri-source = nixpkgsFor.${system}.runCommand "niri-source-contract" { } ''
```
LuckShiba, line 168: I don't think this check should exist, it's pretty much nonsense

LuckShiba (changes requested):

cornservant:
Why close the PR? just remove the check and keep the addition of ./build.rs to the fileset.

PR #4614, opened 2026-09-18
tty: allow opting into software EGL

johnrichardrinehart:
QEMU guests without GPU acceleration may expose only a software EGL device. Niri currently rejects that device, leaving the compositor without a renderer or outputs.

This adds an explicit `NIRI_ALLOW_SOFTWARE_EGL` opt-in while preserving the existing hardware-safe default. When opted in, it also keeps the configured primary node for software EGL so Smithay registration and renderer lookup use the same key.

Validation:
- `nix develop -c cargo fmt --all -- --check`
- `nix build .#niri-debug --no-link`
- both commits apply cleanly to v26.04, the downstream version that exposed the issue


YaLTeR:
I don't much like having an env variable for this. Can it work automatically?

bitbloxhub:
Also see: https://github.com/niri-wm/niri/pull/3959
Related (potentially overlapping) software rendering PR.

PR #4616, opened 2026-09-19
feat(ipc): mark ipc-types as non-exhaustive

titaniumtraveler:
This might be useful to reduce breaking changes for users of the `niri-ipc` crate.

All `non_exhaustive` attributes are behind a `#[cfg_attr(not(feature = "niri-internals"))]` so that `niri` *itself* doesn't have to pretend it doesn't *exactly* know which fields/attributes there are.

(The details of all this are of cause easily changeable via vim macro lol)


niri-ipc/src/lib.rs
```diff
@@ -191,6 +195,9 @@ pub struct PickedColor {
 #[cfg_attr(feature = "clap", command(subcommand_value_name = "ACTION"))]
 #[cfg_attr(feature = "clap", command(subcommand_help_heading = "Actions"))]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
+#[cfg_attr(not(feature = "niri-internals"), non_exhaustive)]
+// TODO: how to minimize breaking on the action variants?
+// `#[non_exhaustive]` doesn't really work on those...
```
YaLTeR: I think it's fine here and in OutputAction, because these are never sent from the compositor to the client?

titaniumtraveler: The main problem is when the client tries to construct them and there is a field added to them.
(As that would break compatibility if an application upgrades the `niri-ipc`-crate)

I *guess* a possible fix would be to put them into their own struct and have them be structured as `EnumVariant(EnumVariant)`, (and the struct would then impl `Default`), but that would be kinda annoying tbh

But it is probably fine anyways.

titaniumtraveler:
If it's fine like this, I would remove all the `TODO:`s and undraft the PR.

niri-ipc/src/lib.rs
YaLTeR: > (As that would break compatibility if an application upgrades the `niri-ipc`-crate)

That's fine. The bigger problem is that currently if an application uses niri-ipc and then niri upgrades, the application has a high chance of failing at runtime due to new JSON variants being added in data sent from compositor.

titaniumtraveler: The other question I have would be whether `niri_ipc::state` should be considered "stable"/not internal.
As in *should* an application potentially make use of it to maintain a local cache of all the relevant information.

(Like what @Axlefublr discussed yesterday in `#development` on discord; Ignoring the usefulness of the discussion itself of cause lol)

niri-ipc/Cargo.toml
```diff
@@ -20,3 +20,4 @@ serde_json.workspace = true
 [features]
 clap = ["dep:clap"]
 json-schema = ["dep:schemars"]
+niri-internals = []
```
HigherOrderLogic: Nit: it might be a good idea to prefix this with `_` or `__` to better signal that it's an internal feature.

```suggestion
__niri-internals = []
```

Axlefublr: ah yes: “niri internals (they are internal btw)”

titaniumtraveler: Fair

YaLTeR: or maybe smth descriptive like "exhaustive"

titaniumtraveler: > or maybe smth descriptive like "exhaustive"

Alternatively maybe just `non_exhaustive` (instead of `not(feature = "exhaustive")`) and explicitly tell the user that they *should* enable the feature (and make it be default) to protect them from a changing `niri` version?

YaLTeR: I guess non_exhaustive + default makes sense

titaniumtraveler: In that case we also only need the `non_exhaustive` attribute for enums.
(Per default) serde just ignores unmentioned fields *anyway*.

YaLTeR:
Could you also test this change with, say, some tool built with the previous niri-ipc release, and current niri sending new stuff?

titaniumtraveler:
> Could you also test this change with, say, some tool built with the previous niri-ipc release, and current niri sending new stuff?

Don't think I have anything at hand that would fit that description.

PR #4618, opened 2026-09-19
Support multiple open-on-output fallbacks for named workspaces

atimofeev:
Implements https://github.com/niri-wm/niri/discussions/3371.

`open-on-output` in a named workspace now takes several outputs:

```kdl
workspace "browser" {
    open-on-output "DP-9" "DP-4"
}
```

Behavior:

- The workspace goes to the first listed output that is connected; if none of them is, it stays on the default monitor, as before.
- The list is reevaluated on output connect/disconnect, so plugging in a higher-priority output moves the workspace there, and unplugging it falls back to the next match. Connection order doesn't matter.
- Moving a workspace to another monitor by hand replaces the configured affinity for the running session.

`Workspace::open_on_output` changes from `Option<String>` to `Vec<String>`, and the layout keeps ordered `original_outputs` instead of a single original output. A single `open-on-output` argument behaves exactly as before, so existing configs are unaffected.

Testing: added layout tests for fallback selection, the preferred output appearing later, the preferred output disappearing, no match at all, and an explicit workspace move overriding the configured list. Each of them keeps a third, non-candidate output connected as the primary monitor, so the assertions depend on the candidate order rather than on there being a single monitor. I've also been running this branch as my daily driver since mid-August on a laptop with two docking stations that expose the external monitor under different connector names — named workspaces land on the external monitor in both setups and follow hotplug correctly.

The named workspaces wiki page is updated in the same commit.


docs/wiki/Configuration:-Named-Workspaces.md
```diff
@@ -12,6 +12,20 @@ workspace "chat" {
...
 ```
 
+Repeat `open-on-output` to define ordered fallbacks:
+
+```kdl
+workspace "browser" {
+    open-on-output "DP-9"
+    open-on-output "DP-4"
+}
+```
```
HigherOrderLogic: I think this syntax is a bit obscure. Imo something like this would be better:

````suggestion
```kdl
workspace "browser" {
    open-on-output "DP-9" "DP-4"
}
```
````

Niri then will try to place the window on monitor from left to right.

Even then this still doesnt look really good, but I fear something better would require too much custom parsing.

File: src/layout/mod.rs [hunk 2]
```diff
@@ -907,8 +916,28 @@ impl<W: LayoutElement> Layout<W> {
...
                     }
 
-                    let primary = &mut monitors[primary_idx];
-                    primary.append_workspaces(workspaces);
+                    let connected_outputs = monitors
+                        .iter()
+                        .map(|monitor| monitor.output.clone())
+                        .collect::<Vec<_>>();
+                    let mut workspaces_by_monitor =
+                        (0..monitors.len()).map(|_| Vec::new()).collect::<Vec<_>>();
```
HigherOrderLogic:
```suggestion
                    let mut workspaces_by_monitor = vec![Vec::new(); monitors.len()];
```

src/layout/workspace.rs
```diff
@@ -474,6 +476,16 @@ impl<W: LayoutElement> Workspace<W> {
         self.output.as_ref()
     }
 
+    pub(super) fn first_connected_output<'a>(&self, outputs: &'a [Output]) -> Option<&'a Output> {
```
HigherOrderLogic: Why does this have to take a list of owned `Output`.

src/layout/tests.rs
```diff
@@ -1644,6 +1647,86 @@ fn check_ops(ops: impl IntoIterator<Item = Op>) -> Layout<TestWindow> {
...
+        Op::MoveWorkspaceToMonitor {
+            ws_name: Some(1),
+            output_id: 3,
+        },
+        Op::AddOutput(1),
+    ]);
+
+    assert_eq!(workspace_output_name(&layout, 1), "output3");
+}
+
```
HigherOrderLogic, line 1765: I dont think these even test what you change.

src/layout/workspace.rs
```diff
@@ -474,6 +476,16 @@ impl<W: LayoutElement> Workspace<W> {
...
+        self.original_outputs
+            .iter()
+            .find_map(|id| outputs.iter().find(|output| id.matches(output)))
+    }
+
+    pub(super) fn set_original_output(&mut self, output: &Output) {
+        self.original_outputs = vec![OutputId::new(output)];
+    }
```
HigherOrderLogic: Is this complex enough to warrant its own function?

src/layout/workspace.rs
```diff
@@ -508,9 +520,10 @@ impl<W: LayoutElement> Workspace<W> {
         self.output = output;
 
         if let Some(output) = &self.output {
-            // Normalize original output: possibly replace connector with make/model/serial.
```
HigherOrderLogic, line 511: I believe the comment should still be here.

docs/wiki/Configuration:-Named-Workspaces.md
atimofeev: Agreed, switched to arguments on one node:

```kdl
workspace "browser" {
    open-on-output "DP-9" "DP-4"
}
```

Config side is now `#[knuffel(child, unwrap(arguments), default)]`, same as `calibration_matrix` in `input.rs`. A single argument parses exactly as before, so existing configs are untouched. Wiki example updated too.

src/layout/mod.rs
atimofeev: That does not compile, since `Workspace<W>` is not `Clone`:

```
error[E0277]: the trait bound `Vec<_>: Clone` is not satisfied
   --> src/layout/mod.rs:946:58
    |
946 |                     let mut workspaces_by_monitor = vec![Vec::new(); monitors.len()];
    |                                                     -----^^^^^^^^^^-----------------
    = note: required for `Vec<layout::workspace::Workspace<W>>` to implement `Clone`
```

Used `iter::repeat_with(Vec::new).take(monitors.len()).collect()` instead, which reads better than the `(0..n).map(|_| ...)` version.

src/layout/workspace.rs
atimofeev: It does not — it was owned only because the old `add_output` loop mutated `monitors` while needing the output list. Signature is now `fn first_connected_output<'a>(&self, outputs: &[&'a Output]) -> Option<&'a Output>`, and `add_output` is split into two phases: first collect the `(monitor, workspace)` indices that prefer the newly connected output while only immutably borrowing, then do the removal pass. No `Output` clones left in any of the three call sites.

src/layout/workspace.rs
atimofeev: No, removed it — the field is `pub(super)` anyway, so the call sites just assign `workspace.original_outputs = vec![OutputId::new(&self.output)]`, like the old code did with `original_output`. Kept `normalize_original_outputs`, which has two callers and is an actual loop.

src/layout/workspace.rs
atimofeev: Restored (pluralized for the new field).

src/layout/tests.rs
atimofeev: You are right about two of them — with only one or two outputs connected the expected output was also the only/primary monitor, so they passed regardless of the candidate logic.

Rewrote all four so a third, non-candidate output stays connected and is the primary monitor, which makes every assertion depend on the ordered candidates:

- `workspace_output_fallback_promotes_preferred_in_reverse_connection_order`: starts on output3, moves to output2 when the second candidate connects, then to output1.
- `workspace_output_preferred_removal_uses_fallback`: output1 -> output2 on disconnect, then output3 (primary) when no candidate is left.
- `workspace_output_without_match_stays_on_default_output`: stays on output3, and connecting the unrelated output4 does not steal it.
- `explicit_workspace_move_overrides_output_candidates`: after an explicit move to output3, connecting output1 does not take it back.

Sanity check: with `first_connected_output` limited to the first candidate, the first, second and fourth fail; with the candidate order reversed, the first and second fail. The third one is order-independent by nature — it guards the no-match path.

atimofeev:
Thanks for the review! Addressed everything and force-pushed (single commit, `b4153733`):

- `open-on-output` now takes multiple arguments on one node instead of repeated nodes.
- `first_connected_output` borrows the outputs; `add_output` split into an immutable scan pass and a mutating pass, so no `Output` clones are needed.
- Dropped the `set_original_output` one-liner, restored the normalization comment.
- Rewrote the four new layout tests so a non-candidate output stays primary and the assertions actually depend on the candidate order.

`cargo +nightly fmt --all`, `cargo clippy --all-targets` and `cargo test --all` are clean.

docs/wiki/Configuration:-Named-Workspaces.md
YaLTeR: Hm, you think so? I was actually going to suggest separate nodes. It's sort of consistent with how you can have multiple of `window-rule`s and such. Also multiple arguments to one `open-on-output` kinda reads like open on multiple outputs at once, instead of fallback. Is it just me?

HigherOrderLogic: > Hm, you think so?

Tbh I think they're nearly as obscure as each other, but on one line it's somewhat better.



> Also multiple arguments to one `open-on-output` kinda reads like open on multiple outputs at once, instead of fallback.

This is the exactly same thought I had when I first saw the multiline output config format. And as I said, I dont think either of these are better than each other.

src/layout/workspace.rs
```diff
@@ -474,6 +476,12 @@ impl<W: LayoutElement> Workspace<W> {
         self.output.as_ref()
     }
 
+    pub(super) fn first_connected_output<'a>(&self, outputs: &[&'a Output]) -> Option<&'a Output> {
```
YaLTeR:
```suggestion
    pub(super) fn find_preferred_output<'a>(&self, outputs: &[&'a Output]) -> Option<&'a Output> {
```
I think this is clearer

src/layout/workspace.rs
```diff
@@ -521,6 +531,14 @@ impl<W: LayoutElement> Workspace<W> {
         }
     }
 
+    fn normalize_original_outputs(outputs: &mut [OutputId], output: &Output) {
+        for id in outputs {
+            if id.matches(output) {
+                *id = OutputId::new(output);
+            }
+        }
+    }
```
YaLTeR: This can make several ids match, e.g. if you have `open-on-output "DP-1" "SomeMonitor"` where SomeMonitor is connected to DP-1. I guess this isn't a problem, but perhaps would be cleaner to dedup the list (though it may be slightly annoying).

src/layout/mod.rs
```diff
@@ -2538,34 +2586,18 @@ impl<W: LayoutElement> Layout<W> {
...
-                assert!(
-                    monitor
-                        .workspaces
-                        .iter()
-                        .any(|workspace| workspace.original_output.matches(&monitor.output)),
-                    "secondary monitor must not have any non-own workspaces"
+                assert_eq!(
+                    preferred,
+                    Some(&monitor.output),
+                    "workspace must be on its first connected output candidate"
```
YaLTeR: If `preferred.is_none()` and `idx != primary_idx` then this assert message is confusing (it was specially handled before for this reason).

docs/wiki/Configuration:-Named-Workspaces.md
YaLTeR: Separate open-on-output lines are also easier to comment out individually (though this is a bit of a fringe case).

Guess I'll ask in the chatroom what ppl think

src/layout/workspace.rs
```diff
@@ -474,6 +476,12 @@ impl<W: LayoutElement> Workspace<W> {
         self.output.as_ref()
     }
 
+    pub(super) fn first_connected_output<'a>(&self, outputs: &[&'a Output]) -> Option<&'a Output> {
+        self.original_outputs
+            .iter()
+            .find_map(|id| outputs.iter().copied().find(|output| id.matches(output)))
+    }
```
YaLTeR, line 486: I think this function can take an `outputs: impl Iterator<Item = &'a Output> + Clone` and then you can avoid collecting outputs into a `Vec` in the callers. I have a suspicion that this will result in less overhead, especially considering that most cases will have `self.original_outputs` with a single output.

src/layout/mod.rs
```diff
@@ -907,8 +928,31 @@ impl<W: LayoutElement> Layout<W> {
...
+                        .iter()
+                        .map(|monitor| &monitor.output)
+                        .collect::<Vec<_>>();
+                    let target_indices = workspaces
+                        .iter()
+                        .map(|ws| {
+                            ws.first_connected_output(&connected_outputs)
+                                .and_then(|target| {
+                                    connected_outputs.iter().position(|o| *o == target)
+                                })
```
YaLTeR: This (and a similar loop in the tests) looks like the function should return an `Option<usize>` index itself instead of the `&'a Output`

src/layout/mod.rs
```diff
@@ -907,8 +928,31 @@ impl<W: LayoutElement> Layout<W> {
...
+                    let mut workspaces_by_monitor = iter::repeat_with(Vec::new)
+                        .take(monitors.len())
+                        .collect::<Vec<_>>();
+                    for (ws, target_idx) in workspaces.into_iter().zip(target_indices) {
+                        workspaces_by_monitor[target_idx].push(ws);
+                    }
+
+                    for (monitor, workspaces) in monitors.iter_mut().zip(workspaces_by_monitor) {
+                        monitor.append_workspaces(workspaces);
+                    }
```
YaLTeR: Can this somehow avoid these intermediate Vecs?

src/layout/mod.rs
```diff
@@ -764,31 +764,52 @@ impl<W: LayoutElement> Layout<W> {
...
+                    .iter()
+                    .map(|monitor| &monitor.output)
+                    .chain(iter::once(&output))
+                    .collect::<Vec<_>>();
 
-                let mut stopped_primary_ws_switch = false;
+                // Find the workspaces that prefer the new output, before mutating the monitors.
+                let to_move = monitors
+                    .iter()
+                    .map(|monitor| {
```
YaLTeR: Is this outer loop/vec collection needed here or can it be done inside the monitors loop below, removing this Vec?

YaLTeR (commented):
Thanks, the tests look fairly comprehensive. Some comments.

docs/wiki/Configuration:-Named-Workspaces.md
YaLTeR: Seems like ppl in chat also prefer the multiple arguments on a single line approach

src/layout/workspace.rs
atimofeev: Renamed to `find_preferred_output`. It now returns `Option<usize>` as suggested below, so the lifetime is no longer part of the return type.

src/layout/workspace.rs
atimofeev: Deduplicated during normalization. Connector and make/model/serial aliases that resolve to the same output now become one canonical `OutputId`; reverse-index removal preserves the first configured candidate and its priority.

src/layout/mod.rs
atimofeev: Restored the distinction: a workspace with no connected candidate now separately asserts that it is on the primary monitor (`workspace without a connected output candidate must be on the primary monitor`). The other assertion only covers a connected candidate on the wrong monitor.

src/layout/workspace.rs
atimofeev: `find_preferred_output` now takes `impl Iterator<Item = &Output> + Clone`. All callers pass `monitors.iter().map(...)` directly; no output Vec allocations remain.

src/layout/mod.rs
atimofeev: Done. `find_preferred_output` returns `Option<usize>` directly, removing both follow-up position scans.

src/layout/mod.rs
atimofeev: Removed them. Redistribution now walks target monitor indices and uses `workspaces.extract_if(..)` for each target, appending each collected group immediately. There is no target-index Vec or per-monitor bucket Vec.

src/layout/mod.rs
atimofeev: Done. The selection check now runs inside the source monitor/workspace loop. It iterates workspaces in reverse, computes the preferred index against the cloneable output iterator, and removes immediately, so the outer `to_move` Vec and its inner Vecs are gone.

docs/wiki/Configuration:-Named-Workspaces.md
atimofeev: Thanks for checking. Kept the current multiple-arguments-on-one-node syntax, then.

src/layout/mod.rs
```diff
@@ -799,28 +810,30 @@ impl<W: LayoutElement> Layout<W> {
...
+                    if stopped_ws_switch
+                        || (source.options.layout.empty_workspace_above_first
+                            && source.workspaces.len() == 2)
+                    {
+                        source.clean_up_workspaces();
+                    }
 
-                workspaces.reverse();
+                    source_workspaces.reverse();
+                    workspaces.extend(source_workspaces);
```
YaLTeR: Just push everything to `workspaces` and then reverse it once at the end I think. Then there's no need for `source_workspaces`.

YaLTeR (commented):
One minor thing

src/layout/mod.rs
atimofeev: Done. Removed `source_workspaces`, push retained workspaces directly into `workspaces`, and reverse `workspaces` once after processing all source monitors.

YaLTeR:
Upon final review, I'm afraid I would like to override the syntax decision back to multiple `open-on-output` lines because:
1. It would be really annoying to make the multi-argument version fail parsing for `open-on-output` with no arguments.
2. I realized that we do have a precedent for OR on separate lines: window rule `match` and `exclude`.

If you're fine with this, it's good to merge.

atimofeev:
I'm perfectly fine with both approaches.
This is done, then.

Thank you guys!

YaLTeR:
Alright, thanks!

PR #4620, opened 2026-09-19
~/.local/share is deprecated for MIME types association

msmafra:
~/.local/share is usually the value of $XDG_DATA_HOME, or it's "fallback" if not set as stated in the XDG Base Directory specs, but in the case of Association between MIME types and applications the usage of $XDG_DATA_HOME is deprecated. And $XDG_DATA_DIRS, $XDG_CONFIG_DIRS and $XDG_CONFIG_HOME are the ones valid.
In the case of setting specifically for Niri, $XDG_CONFIG_HOME seems the most appropriate:
>  ln -s /usr/share/applications/gnome-mimeapps.list $XDG_CONFIG_HOME/niri-mimeapps.list
or 
>  ln -s /usr/share/applications/gnome-mimeapps.list $XDG_CONFIG_HOME/$XDG_CURRENT_DESKTOP-mimeapps.list

https://specifications.freedesktop.org/basedir/latest/
https://specifications.freedesktop.org/mime-apps/latest/file.html


YaLTeR:
If you try it it doesn't work correctly. It takes precedence over user overrides then.

PR #4623, opened 2026-09-20
pw_utils: Preserve the screencast frame cadence

PurpleCloudX:
## Problem

Capturing a fullscreen 1080p60 video in nested niri can deliver only about 45 FPS even when rendering the captured frame takes much less than one refresh interval. I reproduced this with GStreamer's `videotestsrc` playing into `waylandsink`, as well as a small Wayland test client.

`compute_extra_delay()` limits capture using `last_frame_time + min_time_between_frames`. After a slightly late frame, recording its time as the new baseline shifts every following deadline. A subsequent frame can then fall outside the 100 us allowance and be skipped. In the nested backend, output submission also happens between predicting the presentation time and checking the capture deadline, so that deadline may already have passed by the time we schedule the redraw.

In two instrumented 300-frame fullscreen runs, this produced 96 and 112 timing rejections. All of those deadlines had already passed when checked. The runs took 6.60 and 6.87 seconds rather than roughly 5 seconds: the difference closely matches an extra refresh interval per rejection. Capture rendering/readback had a p95 of 5.1-6.3 ms, and neither run ran out of PipeWire buffers. Delivery intervals had a p95 of 33.7-34.5 ms.

## Change

Keep `last_frame_time` on the capture cadence rather than moving it to every late frame's time. Advance it by one negotiated interval after a successful frame, provided the delay is less than one interval. Restart the cadence after a longer gap, a rate change, or a backward time prediction instead of accumulating overdue frames.

The state stays in `Cast`: one additional `Duration` and a short method used by the four existing successful-frame paths. The existing damage checks, rate limiter, and one-shot redraw timer remain in place. Nothing advances on an unsuccessful capture, and there is no periodic wakeup or idle-frame generation.

I separately tested correcting stale presentation timestamps, preserving the cadence, and combining both. Timestamp correction alone reached about 57 FPS in the fullscreen case; cadence preservation reached about 60 FPS without it. Combining them did not improve the result, so this change only preserves the cadence.

The timing principle has precedents, although these are not identical implementations:

- [KWin's `VsyncSource::arm()`](https://github.com/KDE/kwin/blob/b8ef79a70431550832671daa28f1a3201aeca575/src/utils/vsyncsource.cpp#L38-L56) aligns the next wakeup to a steady-clock interval boundary. [Screencasting uses that source](https://github.com/KDE/kwin/blob/b8ef79a70431550832671daa28f1a3201aeca575/src/plugins/screencast/screencaststream.cpp#L509-L530).
- [Mutter's `calculate_next_update_time_us()`](https://github.com/GNOME/mutter/blob/888a7b7dac0c58612007c46bbad75de9f9437f48/clutter/clutter/clutter-frame-clock.c#L1093-L1134) subtracts dispatch lateness when deriving the next update from a previous dispatch, rather than carrying that lateness into each following interval.

## Measurements

Release builds of `62c230a6` and this change, with the same lockfile and build settings. AMD Radeon 680M, Mesa 26.1.4, Weston 15.0.1 headless GL backend, nested niri, GStreamer 1.28.4, and PipeWire server/plugin 1.6.7. Capture used MemFd/BGRx.

Medians, with the order alternating between baseline/fix and fix/baseline:

| Workload | Runs per version | Frames per run | Main FPS | This change FPS | Niri CPU time per run, main / change |
| --- | ---: | ---: | ---: | ---: | ---: |
| Fullscreen changing test client, monitor capture | 6 | 600 | 44.96 | 59.88 | 3315 / 3035 ms |
| 64x64 changing region, monitor capture | 6 | 600 | 59.36 | 59.96 | 2625 / 2625 ms |
| Changing 640x480 window, window capture | 3 | 300 | 46.04 | 58.81 | 490 / 440 ms |
| Fullscreen GStreamer video, monitor capture | 3 | 300 | 45.16 | 59.91 | 1360 / 1340 ms |

The primary benefit is more regular delivery, not a faster pixel-copy implementation. CPU time is measured for an equal number of delivered frames, not as CPU utilization over a fixed wall-clock period. The small-region case has little improvement. In the six fullscreen test-client runs, baseline FPS ranged from 44.49 to 52.75, while the fix ranged from 59.78 to 59.95.

## Test Method

Each run starts a fresh nested niri under a private D-Bus session in a 1920x1080 headless Weston instance. Capture is started through the Mutter ScreenCast D-Bus API using `RecordMonitor` or `RecordWindow`. The changing test client commits at 60 Hz using a periodic timerfd. The GStreamer playback case instead uses this standard client on niri's Wayland socket:

```sh
gst-launch-1.0 videotestsrc is-live=true pattern=blink \
  ! video/x-raw,format=BGRx,width=1920,height=1080,framerate=60/1 \
  ! waylandsink fullscreen=true
```

After warming up the client, consume the node returned by `PipeWireStreamAdded`:

```sh
gst-launch-1.0 -v pipewiresrc path="$NODE" num-buffers=300 \
  ! video/x-raw,format=BGRx ! fakesink sync=false silent=false
```

Count actual `fakesink` chain messages and require the requested number of frames and EOS. Divide that count by GStreamer's reported pipeline execution time. Niri CPU time is the delta of its process user/system counters across the consumer lifetime, including consumer setup and teardown. No builds from this task ran during measurement. The final comparison binaries contain no diagnostic instrumentation. Fullscreen runs additionally check a captured frame's dimensions and pixels at the left, center, and right of every row.

I also exercised consumer reconnection; 30, 15, and 60 FPS renegotiation within a session; a 120 Hz producer request rate with a 30 FPS consumer cap; a 20 Hz source; and a consumer sleeping 50 ms per frame. Monitor and window captures both stopped delivering while their content was paused for two seconds and resumed afterwards, with no Niri CPU counter increase during that idle interval.

This reproduces a useful improvement in the tested nested capture paths. It is not a measurement of native DRM/VRR behavior or other GPUs.

Validation: release build; `cargo test --locked --workspace --exclude niri-visual-tests`; `cargo clippy --locked --workspace --all-targets -- -D warnings`; `cargo check --locked --no-default-features --features xdp-gnome-screencast`; and nightly `cargo fmt --all -- --check` all pass.

I used Codex to assist with the investigation, implementation, and testing.


src/screencasting/pw_utils.rs
```diff
@@ -106,7 +106,9 @@ pub struct Cast {
     formats: FormatSet,
     offer_alpha: bool,
     cursor_mode: CursorMode,
-    pub last_frame_time: Duration,
+    // The last slot in the capture cadence, rather than the time a late frame was rendered.
```
YaLTeR:
```suggestion
```

src/screencasting/pw_utils.rs
```diff
@@ -1019,6 +1022,23 @@ impl Cast {
...
+        self.last_frame_time = if self.last_frame_interval == interval
+            && time >= self.last_frame_time
+            && time < next + interval
+        {
+            next
+        } else {
+            time
+        };
+        self.last_frame_interval = interval;
+    }
```
YaLTeR: For some reason despite "Maintainers are allowed to edit this pull request." showing in the GitHub UI, it does not allow me to push changes.

```suggestion
    pub fn record_frame_time(&mut self, recorded: Duration) {
        let interval = self.inner.borrow().min_time_between_frames;
        let ideal = self.last_frame_time + interval;

        // Absorb small (< 1 frame) differences in output refresh interval vs. screencast framerate
        // to keep the screencast time base consistent instead of shifting forward every frame.
        //
        // After a missed interval or a rate change, restart instead of catching up in a burst.
        self.last_frame_time = if self.last_frame_interval == interval
            && recorded >= self.last_frame_time
            && recorded < ideal + interval
        {
            ideal
        } else {
            recorded
        };
        self.last_frame_interval = interval;
    }
```

YaLTeR (commented):
Thanks for making the patch 10× smaller, it considerably helps.

I thought over this change and I think it is correct. But the comment motivation was not entirely accurate.

In most cases, `record_frame_time()` will be called with `target_presentation_time` which is computed *exactly* on the TTY, as in, last presentation time plus refresh interval. (This is only not the case on winit and on some NVIDIA drivers.)

So in the common case, there are no "small scheduling delays" to absorb -- the times are computed exactly. Rather, what is absorbed is the difference between monitor refresh rate and screencast negotiated frame rate.

The suggested change updates the comment accordingly. Please confirm that my intuition is correct.

src/screencasting/pw_utils.rs
PurpleCloudX: Thanks for clarifying and suggesting the wording. Your explanation makes sense: on the usual TTY path, `target_presentation_time` comes from the presentation timing, not when the code happens to run. I focused too much on the nested backend when writing the original comment. Describing it as keeping the capture cadence steady when the output refresh interval and negotiated capture interval differ is more accurate.

YaLTeR:
Thanks

PR #4629, opened 2026-09-21
use CTM hardware acceleration when GAMMA_LUT is not available

MrFantOlas:
## Description

This change allows devices that not dot have a GAMMA_LUT but have a CTM to use the later for gamma operations.
Although more driver support GAMMA_LUT, two at least only support CTM (apple and msm). This allows user using devices for these drives to benefit from hardware accelerated colour transformation even if less accurate.

I have a device running the apple driver and can confirm that the change works here.

## Related Issue

https://github.com/niri-wm/niri/issues/3672

## Checklist

* [x] I ran tests and `cargo +nightly fmt`.

* [x] I did not use an LLM to write this pull request code or description.


YaLTeR:
Would be good if someone else with this hardware could test

PR #4632, opened 2026-09-22
workspace focus/move on specified output

cornservant:
## Description

Addresses https://github.com/niri-wm/niri/issues/2778

Changes:
- extend the commands `focus-workspace-down`, `focus-workspace-up`, `move-workspace-down`, `move-workspace-up` by an optional string parameter `output` which can take on the values
    - `"active"` (current/default behaviour),
    - `"under-cursor"` (applies change to workspace/monitor under the cursor),
    - `"{output name}"` (applies change to specified output/workspace on specified output)
- changes the default mouse wheel binds to use `output="under-cursor"`

## Checklist

* [x] I ran tests and `cargo +nightly fmt`.

* [x] I did not use an LLM to write this pull request code or description.


niri-config/src/binds.rs
```diff
@@ -1048,6 +1061,66 @@ impl FromStr for Key {
...
+/// Reference to an output
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum OutputReference {
+    /// refers to the active output
+    Active,
+    /// refers to the output under the cursor
+    UnderCursor,
+    /// refers to the output by name
+    OutputName(String),
+}
```
HigherOrderLogic, line 1074: Somewhat an extreme edge case, but what if a monitor is named "active"?

niri-ipc/src/lib.rs
```diff
@@ -471,9 +471,21 @@ pub enum Action {
     /// Center all fully visible columns on the screen.
     CenterVisibleColumns {},
     /// Focus the workspace below.
-    FocusWorkspaceDown {},
+    FocusWorkspaceDown {
+        /// Output on which to focus the workspace below.
+        ///
+        /// Possible values: "active", "under-cursor", name of the output
```
HigherOrderLogic:
```suggestion
        /// Possible values: "active", "under-cursor", name of the output.
```

niri-ipc/src/lib.rs
```diff
@@ -471,9 +471,21 @@ pub enum Action {
...
+        #[cfg_attr(feature = "clap", arg(long, default_value_t = OutputReferenceArg::Active))]
+        output: OutputReferenceArg,
+    },
     /// Focus the workspace above.
-    FocusWorkspaceUp {},
+    FocusWorkspaceUp {
+        /// Output on which to focus the workspace above.
+        ///
+        /// Possible values: "active", "under-cursor", name of the output
```
HigherOrderLogic:
```suggestion
        /// Possible values: "active", "under-cursor", name of the output.
```

Also should be applicable to other similar spots.

niri-config/src/binds.rs
```diff
@@ -1048,6 +1061,66 @@ impl FromStr for Key {
...
+            ));
+        }
+    }
+
+    fn raw_decode(
+        value: &knuffel::span::Spanned<knuffel::ast::Literal, S>,
+        ctx: &mut knuffel::decode::Context<S>,
+    ) -> Result<Self, DecodeError<S>> {
+        match &**value {
+            knuffel::ast::Literal::String(ref s) => Ok(s.parse::<OutputReference>().unwrap()),
```
HigherOrderLogic:
```suggestion
            knuffel::ast::Literal::String(ref s) => Ok(s.parse().unwrap()),
```

niri-config/src/binds.rs
```diff
@@ -1048,6 +1061,66 @@ impl FromStr for Key {
...
+            _ => {
+                ctx.emit_error(DecodeError::unsupported(
+                    value,
+                    "Unsupported value, only strings are recognized",
```
HigherOrderLogic: Nit: I think Knuffel error message usually starts with lowercase letter, and is something like `"output must be a string"`.

Ref: https://github.com/niri-wm/niri/blob/5f4469b6a992492cf7221b269e9379f42e737649/niri-config/src/workspace.rs#L91

niri-config/src/binds.rs
cornservant: I'd agree that a more correct approach would be to use a prefix like "name:{output name}", but I haven't seen this approach taken elsewhere in the code, in particular for workspaces where it would also make sense to distinguish `id:13` from `idx:13` or `name:13`. On the other hand, should someone run into this problem they could write `"Active"` instead because the output matcher is case-insensitive.

niri-config/src/binds.rs
cornservant: I'll change it to "output reference must be a string" and hope it's not too jargony

